PR #18, opened 2017-10-06
Added class for loading common page metadata. Added page template take-action

koyan:
Added class for loading common page metadata. Added page template take-action

### DIFF
--- a/page_take-action.php
+++ b/page_take-action.php
@@ -1,0 +1,8 @@
+<?php /* Template Name: Take Action */
+
+$page = new P4_Page();
+
+//If there is a background image, use it instead of the default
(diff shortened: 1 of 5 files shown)


page_take-action.php
```diff
@@ -0,0 +1,8 @@
+<?php /* Template Name: Take Action */
+
+$page = new P4_Page();
```
angtheod: Here is where you 'd create the child class:

`$page = new P4_Take_Action_Page();

class P4_Take_Action_Page extends P4_Page {
    public __construct() {
        parent::__construct();
        $page->context['background_image']  = $page->page_meta_data['p4_background_image'][0];
    }
}`

PR #24, opened 2017-10-11
Feature/1312 act explore menu

kirdia:
Navigation bar menu for act & explore pages


templates/navigation-bar.twig
```diff
@@ -80,12 +80,11 @@
...
                                     </li>
-                                    <li class="nav-item active hidden-md-down hidden-lg-down">
-                                        <a href="#">{{ __( 'ACT', domain ) }}</a>
-                                    </li>
-                                    <li class="nav-item hidden-md-down hidden-lg-down">
-                                        <a href="#">{{ __( 'EXPLORE', domain ) }}</a>
-                                    </li>
+                                    {% for item in navbar_menu.get_items %}
+                                        <li class="nav-item active {{ item.classes | join(' ') }}">
+                                            <a href="{{ item.get_link }}">{{ __( item.title, domain ) }}</a>
```
angtheod: @kirdia These don't need translations. They will be able to change the name from the WordPress panel.
Besides, the WP translation functions can not have variables as an argument (see https://codex.wordpress.org/I18n_for_WordPress_Developers).

PR #29, opened 2017-10-17
Feature/1359 page background image

sagarsdeshmukh:


functions.php
```diff
@@ -215,6 +215,26 @@ public function register_header_metabox() {
 				'type' => 'text_medium',
 			)
 		);
+
+		$p4_header->add_field(
+			array(
+				'name'    => 'Background overide',
```
koyan: Please use localised messages (translatable texts).

PR #30, opened 2017-10-18
Feature/1361 unify wrapper all pages

sagarsdeshmukh:
Unify wrapper for all pages (home, act, take action etc) to a single markup


koyan:
Looks good.

PR #34, opened 2017-10-23
PLANET-1365 Applied header block markup

sagarsdeshmukh:


templates/blocks/header.twig
```diff
@@ -1,21 +1,25 @@
...
\ No newline at end of file
+	<div class="top-tags-wrap">
+		{% if ( page_tags ) %}
+			{% for page_tag in page_tags %}
+				<span>#{{ page_tag.name }}</span>
+			{% endfor %}
+		{% endif %}
+	</div>
+	{% if ( header_title ) %}
+		<h1>{{ header_title|raw }}</h1>
```
johnbillion: What's the reason for removing the `wp_kses_post` filtering?

sagarsdeshmukh: We are using autoescape so we won't need `wp_kses_post ` here. Also I need to remove `raw` as well.
- If we used `raw` then we need `wp_kses_post `

koyan:
Looks good 👍

PR #36, opened 2017-10-23
Implemented Tag <> Image functionality so that Editors can associate …

angtheod:
…an Image with a Tag. WP Media window opens in order to select an Image. Saved Image appears also in the Image column on Tags page. Added code to separate class. Separated a bit better the code in P4_Master_Site class cause it was getting big and messy. Added auto-loading for the classes under the classes directory. Added jquery admin code to separate admin.js file which is requested while on WP panel pages.

### DIFF
--- a/assets/js/admin.js
+++ b/assets/js/admin.js
@@ -1,0 +1,40 @@
+$ = jQuery;
+
+$(document).ready(function() {
+
+	/**
+	 * Taxonomy_Image
+	 */
+	$("#insert_image_tag_button").off("click").on("click", function () {
+		var $media_modal = wp.media({
+			title: 'Insert Image',
+			library: {
+				type: 'image'
+			},
+			button: {
+				text: 'Insert'
+			},
+			multiple: false
+		});
+		$media_modal.open();
+
+		$media_modal.on('select', function () {
+			var $selected_image = $media_modal.state().get('selection').first().toJSON();
+			$("#tag_attachment_id").val($selected_image.id);
+			$(".tag_attachment .attachment-thumbnail").attr("src", $selected_image.sizes.thumbnail.url);
+		});
+
+		return false;
+	});
+
+	$(".tag_attachment .attachment-thumbnail").off("click").on("click", function () {
+		$("#insert_image_tag_button").click();
+	});
+
+	$("#addtag #submit").off("click").on("click", function () {
+		setTimeout(function () {
+			jQuery("#tag_attachment_id").val("");
+			jQuery(".tag_attachment .attachment-thumbnail").attr("src", "");
+		}, 300);
+	});
+} );
--- a/assets/js/admin.js
+++ b/assets/js/admin.js
@@ -6,23 +6,26 @@ $(document).ready(function() {
 	 * Taxonomy_Image
 	 */
 	$("#insert_image_tag_button").off("click").on("click", function () {
-		var $media_modal = wp.media({
-			title: 'Insert Image',
-			library: {
-				type: 'image'
-			},
-			button: {
-				text: 'Insert'
-			},
-			multiple: false
-		});
-		$media_modal.open();
+		if ( typeof wp !== "undefined" ) {
+			var media_modal = wp.media({
+				title: "Insert Image",
+				library: {
+					type: "image"
+				},
+				button: {
+					text: "Insert"
+				},
+				multiple: false
+			});
 
-		$media_modal.on('select', function () {
-			var $selected_image = $media_modal.state().get('selection').first().toJSON();
-			$("#tag_attachment_id").val($selected_image.id);
-			$(".tag_attachment .attachment-thumbnail").attr("src", $selected_image.sizes.thumbnail.url);
-		});
+			media_modal
+				.on("select", function () {
+					var $selected_image = media_modal.state().get("selection").first().toJSON();
+					$("#tag_attachment_id").val($selected_image.id);
+					$(".tag_attachment .attachment-thumbnail").attr("src", $selected_image.sizes.thumbnail.url);
+				})
+				.open();
+		}
 
 		return false;
 	});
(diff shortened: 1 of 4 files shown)


classes/class-p4-taxonomy-image.php
```diff
@@ -0,0 +1,148 @@
...
+
+		/**
+		 * Save user custom input field.
+		 *
+		 * @param int $term_id The ID of the WP_Term object that is added or edited.
+		 */
+		public function save_tag_meta( $term_id ) {
+			$attachment_id = filter_input( INPUT_POST, 'tag_attachment_id', FILTER_VALIDATE_INT );
+			if ( $attachment_id ) {
+				$attachment_id = $this->validate( $attachment_id );
```
dashaluna: Not sure we need `$this->validate` here. The `filter_input` above validates if it's int and converts to int on success ref: http://php.net/manual/en/filter.filters.validate.php

Also if ( $attachment_id ) above will fail on int=0.

So, I'd remove this line and corresponding `validate()` function as it's not used anywhere else. Also remove the `if` statement below.

classes/class-p4-taxonomy-image.php
```diff
@@ -0,0 +1,148 @@
...
+				}
+			}
+		}
+
+		/**
+		 * Add custom column.
+		 *
+		 * @param array $columns Associative array with the columns of the page.
+		 *
+		 * @return array
```
dashaluna: Docs: description of what's returned.

classes/class-p4-taxonomy-image.php
```diff
@@ -0,0 +1,148 @@
...
+		}
+
+		/**
+		 * Apply custom output to a custom column.
+		 *
+		 * @param string $output The html to be applied to each row of the $column.
+		 * @param string $column The name of the column to be managed.
+		 * @param int    $term_id The ID of the WP_Term object that is added or edited.
+		 *
+		 * @return string
```
dashaluna: Docs: description of what's returned.

classes/class-p4-taxonomy-image.php
```diff
@@ -0,0 +1,148 @@
...
+
+			return $output;
+		}
+
+		/**
+		 * Make column sortable.
+		 *
+		 * @param array $columns Associative array with the columns of the page.
+		 *
+		 * @return array
```
dashaluna: Docs: description of what's returned.

classes/class-p4-taxonomy-image.php
```diff
@@ -0,0 +1,148 @@
...
+		}
+
+		/**
+		 * Validates the input.
+		 *
+		 * @param string $input The user input to be validated.
+		 *
+		 * @return int|bool Int if validation is ok, false if validation fails.
+		 */
+		public function validate( $input ) {
```
dashaluna: I'd remove this redundant function as explained in an earlier comment.

functions.php
```diff
@@ -38,10 +40,52 @@ class P4_Master_Site extends TimberSite {
...
+	}
+
+	/**
+	 * Define settings for the Planet4 Master Theme.
+	 */
+	protected function settings() {
+		Timber::$autoescape = true;
+		Timber::$dirname = array( 'templates', 'views' );
+		$this->theme_dir = get_template_directory_uri();
+		$this->theme_images_dir = $this->theme_dir . '/images/';
```
dashaluna: Minor: align `=` signs.

assets/js/admin.js
```diff
@@ -0,0 +1,43 @@
...
+		$("#insert_image_tag_button").click();
+	});
+
+	$("#submit", $("#addtag")).off("click").on("click", function () {
+		setTimeout(function () {
+			jQuery("#tag_attachment_id").val("");
+			jQuery(".tag_attachment .attachment-thumbnail").attr("src", "");
+		}, 300);
+	});
+} );
```
dashaluna: Needs new line at the end of the file.

Also, a thought for the future - does this need to be named more granular and only loaded on appropriate admin page rather than on all admin pages as it is done at the moment?

assets/js/admin.js
```diff
@@ -0,0 +1,43 @@
...
+	 */
+	$("#insert_image_tag_button").off("click").on("click", function () {
+		if ( typeof wp !== "undefined" ) {
+			var media_modal = wp.media({
+				title: "Insert Image",
+				library: {
+					type: "image"
+				},
+				button: {
+					text: "Insert"
```
dashaluna: @mcshanea @tcrsavage Are these need to be registered for translations, or we're not bothering with admin side for translations?

classes/class-p4-taxonomy-image.php
angtheod: The validate function does not validate if it's int. It validates if its positive int.
Failing on 0 was ok, that is why I did not care about it.
But now I changed the function to return only bool.

dashaluna:
@angtheod just a thought, it looks like you're using cmb2, can that handle image-tag association?

classes/class-p4-taxonomy-image.php
dashaluna: @angtheod Hm, I'm still a bit confused here. The `validate()` will return `true` if it's `int and it's >0`. 

So `filter_input()` will convert value to int, so in `if` statement we can just check if it's `>0`:
`if ( $attachment_id > 0 ) {`

I don't see the value of separating this logic into separate method if it's used just ones, and it's clear to see the check closer to the code where it's performed. If that makes sense?

dashaluna: Actually here I wonder if a check to see if term actually exist would be a good one, as anything can be passed in `$_POST`

dashaluna:
@angtheod looks good, but added some comments re validation function https://github.com/greenpeace/planet4-master-theme/pull/36#discussion_r146876423 and following comment re checking term actually exists. Let me know what you think.

classes/class-p4-taxonomy-image.php
angtheod: You mean the 'tag_attachment_id' key in $_POST? If this is not set then filter_input will return NULL or if the filter fails for whatever reason it will return false.
In both cases the if statement inside validate() results to True (since NULL == 0 and false == 0 are both True), so validation will not pass.

As for checking this in a separate function, I guess it is just a matter of preference... I know it is a trivial function, but since I put this all inside a class I prefer to have a separate method for validation logic just to know at first glance that this class handles also validation.

angtheod:
> just a thought, it looks like you're using cmb2, can that handle image-tag association?

No idea... I was asked by design team to do it this way, so I proceeded this way. @dashaluna

PR #38, opened 2017-10-25
Feature/1389 post page template

kirdia:
Initial version of post page template


functions.php
```diff
@@ -289,6 +289,34 @@ public function register_header_metabox() {
...
+
+		$p4_post->add_field( array(
+			'name' => __( 'Articles Title', 'planet4-master-theme' ),
+			'desc' => __( 'Articles title comes here', 'planet4-master-theme' ),
+			'id'   => $prefix . 'articles_title',
+			'type' => 'text_medium',
+		) );
+
+		$p4_post->add_field(
+			array(
```
dashaluna: `array(` should appear on the line above, so:

`$p4_post->add_field( array(`

single.php
```diff
@@ -9,8 +9,17 @@
...
+
 $context = Timber::get_context();
-$post = Timber::query_post();
+$post    = Timber::query_post();
+
+$page_meta_data = get_post_meta( $post->ID );
+$articles_title = $page_meta_data['p4_articles_title'][0];
+$articles_count = $page_meta_data['p4_articles_count'][0];
+
+if ( ! empty( $articles_title ) and $articles_count and is_int( intval( $articles_count ) ) ) {
```
dashaluna: Use logical `&&` instead of `and`.

dashaluna: Worth moving `intval` up to variable declaration to simply the `if` conditions.

templates/single.twig
```diff
@@ -4,27 +4,66 @@
...
 				</div>
-				<div class="responses">
-					{% for cmt in post.get_comments() %}
-						{% include "comment.twig" with {comment:cmt} %}
-					{% endfor %}
+				<div class="article-author">
+					<div class="row">
+						<div class="col-md-6">
+							<div>
+								{{ __('ABOUT THE AUTHOR') }}
```
dashaluna: Needs domain specified, why is it in all caps? I'd thought that kind of styling is done in CSS?

templates/single.twig
```diff
@@ -4,27 +4,66 @@
 	<div class="content-wrapper">
 		<article class="post-type-{{ post.post_type }}" id="post-{{ post.ID }}">
 			<section class="article-content">
+				<div><span>{{ __('POST') }}</span>
```
dashaluna: Needs domain specified, why is it in all caps? I'd thought that kind of styling is done in CSS?

functions.php
```diff
@@ -289,6 +289,34 @@ public function register_header_metabox() {
...
+				'name'            => __( 'Articles Count', 'planet4-master-theme' ),
+				'desc'            => __( 'Articles count comes here', 'planet4-master-theme' ),
```
dashaluna: Perhaps more descriptive `desc` is needed for editors? I know I don't know GP data, but that desc doesn't really clarify anything.

koyan:
Looks good 👍

PR #40, opened 2017-10-27
Feature/937: Search sort

angtheod:
Initial version of the P4 Search Results page. It is using SearchWP plugin's search under the hood. (Sorting by most recent is not completed). Added simple markup to resemble the final design. Replaced Timber's get_preview() deprecated function calls with preview(). Truncated preview down to 25 words.

### DIFF
--- a/search.php
+++ b/search.php
@@ -22,10 +22,9 @@ $context['sort_options'] = [
 	'recent'   => __( 'Most recent', 'planet4-master-theme' ),
 ];
 
-$search        = get_search_query();
 $default_sort  = 'relevant';
-$field         = 'search_results_sort';
-$selected_sort = filter_input( INPUT_POST, $field, FILTER_SANITIZE_STRING );
+$selected_sort = filter_input( INPUT_GET, 'order', FILTER_SANITIZE_STRING );
+$search        = get_search_query();
 
 if ( ! in_array( $selected_sort, array_keys( $context['sort_options'] ), true ) ) {
 	$context['selected_sort'] = $default_sort;
@@ -35,27 +34,26 @@ if ( ! in_array( $selected_sort, array_keys( $context['sort_options'] ), true ) 
 
 switch ( $context['selected_sort'] ) {
 	case 'recent':
-		$context['posts']  = Timber::get_posts( [
+		$context['posts'] = Timber::get_posts( [
+			// TODO - Find solution for Timber bug (see https://github.com/timber/timber/issues/935).
+			// which does not include attachments to get_posts() results if we supply an array as an argument.
 			's'       => $search,
 			'orderby' => 'post_date',
 			'order'   => 'DESC',
 		] );
 		break;
 	default:
-		$context['posts'] = Timber::get_posts( [
-			's' => $search,
-		] );
+		$context['posts'] = Timber::get_posts();
 }
-
 $found_posts = count( $context['posts'] );
-$context['title']  = $found_posts . ' results for \'' . get_search_query() . '\'';
 
+$context['title']  = "$found_posts results for '$search'";
+$context['domain'] = 'planet4-master-theme';
 $context['issues'] = get_categories( [
 	'child_of' => get_category_by_slug( 'issues' )->term_id,
 	'orderby'  => 'name',
 	'order'    => 'ASC',
 ] );
-
 $context['campaigns'] = [
 	[
 		'name'    => '#CampaignName1',
@@ -70,7 +68,6 @@ $context['campaigns'] = [
 		'results' => 0,
 	],
 ];
-
 $context['categories'] = [
 	[
 		'name'    => '#1',
@@ -85,7 +82,6 @@ $context['categories'] = [
 		'results' => 0,
 	],
 ];
-
 $context['content_types'] = [
 	[
 		'name'    => '#1',
@@ -100,7 +96,5 @@ $context['content_types'] = [
 		'results' => 0,
 	],
 ];
-
-$context['domain'] = 'planet4-master-theme';
 
 Timber::render( $templates, $context );
--- a/search.php
+++ b/search.php
@@ -16,16 +16,14 @@
 global $wp_query;
 
 $templates = array( 'search.twig', 'archive.twig', 'index.twig' );
-$context = Timber::get_context();
-
-$default_sort = 'relevant';
-$found_posts = $wp_query->found_posts;
-$context['title']  = $found_posts . ' results for \'' . get_search_query() . '\'';
+$context   = Timber::get_context();
 $context['sort_options'] = [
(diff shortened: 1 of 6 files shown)


angtheod:
I have created my own markup to resemble the final design, provided [here](https://jira.greenpeace.org/secure/attachment/24594/search_results_2.jpg) , since we don't have markup for the search page ready yet.

search.php
```diff
@@ -9,10 +9,92 @@
...
+			// TODO - Find solution for Timber bug (see https://github.com/timber/timber/issues/935).
+			// which does not include attachments to get_posts() results if we supply an array as an argument.
+			's'       => $search,
+			'orderby' => 'post_date',
+			'order'   => 'DESC',
+		] );
+		break;
+	default:
+		$context['posts'] = Timber::get_posts();
+}
```
kirdia: You can build the filter options array outside switch block and add only the order options in the switch block.
Also search string is missing for default option.

search.php
```diff
@@ -9,10 +9,92 @@
...
+		$context['posts'] = Timber::get_posts( [
+			// TODO - Find solution for Timber bug (see https://github.com/timber/timber/issues/935).
+			// which does not include attachments to get_posts() results if we supply an array as an argument.
+			's'       => $search,
+			'orderby' => 'post_date',
+			'order'   => 'DESC',
+		] );
+		break;
+	default:
+		$context['posts'] = Timber::get_posts();
```
kirdia: You can call Timber get post outside switch block, after you have populated the filters array

search.php
```diff
@@ -9,10 +9,92 @@
...
+} else {
+	$context['selected_sort'] = $selected_sort;
+}
+
+switch ( $context['selected_sort'] ) {
+	case 'recent':
+		$context['posts'] = Timber::get_posts( [
+			// TODO - Find solution for Timber bug (see https://github.com/timber/timber/issues/935).
+			// which does not include attachments to get_posts() results if we supply an array as an argument.
+			's'       => $search,
```
tcrsavage: Looks like this is unrelated to Timber assuming you're trying to get attachments returning as top level items in the result set. 
https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-query.php#L2323

Attachments have a special 'inherit' post status which means they take the status from their parent post. I hadn't realised this before but this looks like this status only gets handled transparently by core if the is_tax is set to true on the query object. Unless there needs to be a concept of unpublished attachments on site, we can just add the 'inherit' status to our post_status arg 

```
'post_status' => ['publish', 'inherit'],
'post_type' => ['post', 'attachment']
```

Otherwise need to set is_tax to true (needs testing) or filtering `posts_where` and `posts_join` might be the best option.
https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-query.php#L2376

templates/search.twig
```diff
@@ -0,0 +1,143 @@
...
+			vertical-align: top;
+		}
+		#results_settings {
+			float: right;
+		}
+		.result-post-date {
+			font-style: italic;
+			color: grey;
+		}
+	</style>
```
tcrsavage: Is there an issue associated with moving this into a stylesheet?

templates/tease-search.twig
```diff
@@ -0,0 +1,16 @@
...
+	<td>
+		<article class="tease tease-{{ post.post_type }}" id="tease-{{ post.ID }}">
+			{% block content %}
+				<h2 class="h2"><a href="{{ post.link }}">{{ post.title|raw }}</a></h2>
+				<span class="result-post-date">{{ post.post_date|date('M d, Y') }}</span>
+				<p>{{ post.preview()|truncate( 25, true )|raw }}</p>
+			{% endblock %}
+		</article>
+	</td>
+</tr>
```
tcrsavage: Newline

tcrsavage:
@angtheod couple of other quick questions:

- I understand that the plan is to integrate Elasticsearch into the site at some point. This feels like it undermines the use of SearchWP, I'd expect issues running those integrations together. Has this been discussed?
- The linked Jira issue 937 is a user story but this PR doesn't satisfy the story. Assuming 1 PR per issue, this PR should have an associated subtask of 937?

search.php
```diff
@@ -9,10 +9,92 @@
...
+	],
+	[
+		'name'    => '#2',
+		'results' => 0,
+	],
+	[
+		'name'    => '#3',
+		'results' => 0,
+	],
+];
```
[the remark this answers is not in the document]
angtheod: Hello, yes you are right, I did not have time to refine the code and separate logic but will do. That is one of the reasons I added the do not merge label :)

angtheod:
> I understand that the plan is to integrate Elasticsearch into the site at some point. This feels like it undermines the use of SearchWP, I'd expect issues running those integrations together. Has this been discussed?

@tcrsavage yes you are correct the plan is to switch to ES when we have the time to do so and we will stop using SearchWP when we get there

templates/tease-search.twig
```diff
@@ -0,0 +1,16 @@
+<tr id="result-row-{{ post.ID }}">
+	<th>
+		{% if ( post.get_thumbnail ) %}
+			<img src="{{ post.thumbnail.src }}" height="150" width="150" alt="image" />
```
sagarsdeshmukh: Can we use dynamic `alt` text here ?

templates/search.twig
angtheod: @tcrsavage 
The styling I have added is temporary until our front end developer delivers the styling of this page, so this will be removed later on.

search.php
```diff
@@ -9,10 +9,92 @@
...
+$context['content_types'] = [
```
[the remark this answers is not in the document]
angtheod: @tcrsavage 
Not yet, but there will be one.

search.php
angtheod: @tcrsavage 
Hi Theo, I had already tried the 'inherit' post_status option, does not work. Also, tried with 'any'.

Can you explain more what you mean with the is_tax ? Will try to use the posts_where, posts_join filters, see what I can get.

angtheod: @tcrsavage 
Looks like this might have to do with Timber and SearchWP not collaborating very well.
If I don't pass a parameter to Timber::get_posts it works as expected with SearchWP...
but if I do pass a parameter then it looks like it skips SearchWP somehow.

angtheod: @tcrsavage @koyan 
Looks like this might have to do with Timber and SearchWP not collaborating very well.
If I don't pass a parameter to Timber::get_posts it works as expected with SearchWP...
but if I do pass a parameter then it looks like it skips SearchWP somehow.

tcrsavage: @angtheod will ping you to help with some debugging on this. I'm able to get attachments pulling in on my local with 'inherit' status but haven't dug into the SearchWP side of things yet.

PR #47, opened 2017-11-01
Feature/1428 Search attachments

angtheod:
Fix issue with Timber and SearchWP not collaborating well when we provided parameter to Timber's get_posts() function. Now Most Recent sorting includes attachments as well. Also, fixed issue with displaying max 10 results. Also, added pagination as a temp solution until we add lazy loading.

### DIFF
--- a/search.php
+++ b/search.php
@@ -19,12 +19,12 @@ global $wp_query;
 $templates = array( 'search.twig', 'archive.twig', 'index.twig' );
 $context   = Timber::get_context();
 $context['sort_options'] = [
-	'relevant' => __( 'Most relevant', 'planet4-master-theme' ),
-	'recent'   => __( 'Most recent', 'planet4-master-theme' ),
+	'relevant'  => __( 'Most relevant', 'planet4-master-theme' ),
+	'post_date' => __( 'Most recent', 'planet4-master-theme' ),
 ];
 
 $default_sort  = 'relevant';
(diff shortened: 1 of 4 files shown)


search.php
```diff
@@ -30,37 +33,24 @@
...
-// Cast to array for forward compatibility with php 7.2 which requires count parameter to be array or object that implements Countable.
-$found_posts = count( (array) $context['posts'] );
+/*
+  The issue seems like it is related with Timber and SearchWP not collaborating very well here. After investigating this we found that
+  If we do not pass an argument to Timber::get_posts() then it falls back to the main WP_Query which works with SearchWP.
+  If we pass an argument to Timber::get_posts() then it creates a subquery and SearchWP is not aware of that and therefore we do not get attachemnts included in search results.
+  A solution is to proceed without passing query options to get_posts at all
+  and instead use the `edit_searchwp_query_orderby` of searchwp to edit the order by part of the main WP_Query directly.
+*/
+$context['posts'] = Timber::get_posts();
```
tcrsavage: I think for this description we can just leave instructions not to insert query args directly into this call and that get_posts uses the main query if no params are passed in. The solution we came out with here is more technically correct because we don't introduce unnecessary queries on the page load and instead modify the main query.

```
/* With no args passed to this call, Timber uses the main query which we filter for customisations via 
 * P4_Master_Site class.
 *
 * When customising this query, use filters on the main query to avoid bypassing SearchWP's 
 * handling of the query
 */

tcrsavage:
👍

PR #53, opened 2017-11-06
Added custom field in Tag for allowing selection of HappyPoint backgr…

angtheod:
…ound image inside Campaigns.


koyan:
Looks good 👍

PR #56, opened 2017-11-07
Feature/1389 post page template

kirdia:
Added cmb2 take action page selector. Changed post page template markup


koyan:
Looks good 👍

PR #57, opened 2017-11-07
Get block names dynamically.

angtheod:
Now that we use a const for the BLOCK_NAME we can retrieve them easier dynamically.


koyan:
Looks good 👍

PR #61, opened 2017-11-08
Don't split content if the take action page for the take action boxou…

kirdia:
…t block is not defined in the post page template.


koyan:
Looks good 👍

PR #62, opened 2017-11-08
Fix Navigation bar disparities. Added new markup for the Select a web…

angtheod:
…site links. Now they are links as suggested by WCAG report. Added active link to the correct menu item. Updated front end Javascript according to Presentation Layer.


koyan:
Looks good 👍

PR #65, opened 2017-11-09
Add footer menu items.

angtheod:
It does not have as many changes as it looks. Looks like that because I added some of the code inside and  if statement


koyan:
Looks good 👍

PR #67, opened 2017-11-09
Add custom taxonomy for planet4 post types

kirdia:


functions.php
```diff
@@ -100,6 +100,7 @@ protected function hooks() {
 		add_filter( 'get_twig',               array( $this, 'add_to_twig' ) );
 		add_action( 'init',                   array( $this, 'register_post_types' ) );
 		add_action( 'init',                   array( $this, 'register_taxonomies' ) );
+		add_action( 'init',                   array( $this, 'register_p4_post_type_taxonomy' ) );
```
angtheod: @kirdia 
Could call your method from inside the register_taxonomies() method that we already have for this.

kirdia: Yes, good call. Will call it from inside

PR #68, opened 2017-11-09
Added Issue name and link in Header block Added Campaigns names and l…

angtheod:
…inks in Header block. These navigation links do not show up inside the Issue page itself.


koyan:
Looks good 👍

PR #69, opened 2017-11-09
Presentation templates disparities

angtheod:
Do not include the Header block's markup at all if the header_title is not set (e.g. in Front Page). Small possible bug fix. Markup fixes.


koyan:
Looks good 👍

PR #71, opened 2017-11-10
Changed post-type to page-type

mcshanea:
Just changing the wording here, as "Post Type" is an existing WordPress term and could cause confusion and 'P4' was an unnecessary qualification. We talked about Presentation Type and Content Type but settled on Page Type for least conflict across editorial terms.


koyan:
Looks good 👍

PR #72, opened 2017-11-10
Remove sticky Header block.

angtheod:


koyan:
Looks good 👍
As always, please make sure that you communicate changes to the .js file with @ashish173

PR #74, opened 2017-11-13
Feature/1503 search

angtheod:
Added markup for Search page (with or without results) according to presentation layer. Added missing fonts. Removed old markup. Removed unused image.


koyan:
Looks good 👍

PR #78, opened 2017-11-14
Title for post template should use whole width. Removed kses_post esc…

kirdia:
…aping from post content.

### DIFF
--- a/templates/single.twig
+++ b/templates/single.twig
@@ -18,7 +18,7 @@
 		<div class="post-block" id="post-{{ post.ID }}">
 
 			<div class="row">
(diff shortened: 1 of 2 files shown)


templates/single.twig
```diff
@@ -72,7 +72,7 @@
 			<div class="row">
 				<div class="col-lg-12">
 					<div class="post-details">
-						{{ post.content|e('wp_kses_post')|raw }}
```
koyan: ??? why are you removing the escaping? Since we pass it raw, we should not remove the escaping

koyan:
Looks good 👍

PR #79, opened 2017-11-14
Feature/1503 search

angtheod:
Update Search page markup. Provide all required data to the search and tease-search templates. Result items now display the correct content type, page type, campaign tags. Document items now link directly to the attachment instead of the post that contains it, so that the document can be opened with one click. Download button now works. Action items now have the set featured image as background. Added Take Action button links. Hide Suggested terms block if there are no suggestions.

### DIFF
--- a/search.php
+++ b/search.php
@@ -118,7 +118,7 @@ $context['campaigns'] = [
 		'results' => 0,
 	],
 ];
-$context['categories'] = [
+$context['page_types'] = [
 	[
 		'name'    => 'Organisation',
 		'results' => 0,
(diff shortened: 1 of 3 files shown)


search.php
```diff
@@ -76,32 +113,44 @@
...
+		'name'    => 'Organisation',
+		'results' => 0,
+	],
+	[
+		'name'    => 'Press Release',
 		'results' => 0,
 	],
 	[
-		'name'    => '#2',
+		'name'    => 'Publication',
```
koyan: Publication and story are not categories, but are controlled by the extra taxonomy.

koyan:
Looks good 👍

PR #85, opened 2017-11-15
Feature/1480 cookies

angtheod:
Implemented Cookies functionality. Added backend field with that allows for some basic customization of the text and links to be added. Input is sanitized and output is escaped properly. Added JS funcionality to store cookie for 1 day.


functions.php
```diff
@@ -135,6 +136,45 @@ private function services( $services = array() ) {
...
+	 * @param array $context The associative array with data to be passed to the main template.
+	 *
+	 * @return mixed
+	 */
+	public function add_to_context( $context ) {
+		$context['data_nav_bar'] = [
+			'websites'     => $this->websites,
+			'images'       => $this->theme_images_dir,
+			'home_url'     => home_url( '/' ),
+			'act_url'      => '/act',
```
koyan: Why do you hardcode this here?

angtheod: I didn't. It was already there (in the bottom of the file). I just moved the function up.
This file is getting big and we should clear up the code sometime soon. To be honest I do not remember if I put this there, but I will see how to make it dynamic.

koyan: Ok. There is already a jira issue about making this dynamic.

koyan:
Looks good 👍

PR #89, opened 2017-11-16
Feature/1440 add data layer pushes

sagarsdeshmukh:
The dataLayer pushes for all pages are done. Please merge this so I will test and take reviews from Anton.


koyan:
Looks good 👍

PR #90, opened 2017-11-16
1426 - Search page code refinement

angtheod:
Refined code for the P4 Search results page, so that it is reusable, customisable and extensible. Created class P4_Search which sets the context and renders the page. Separated logic into different methods based on their functionality. Pagination and suggestions are optional and are added only if they are added from the backend. Refined code for the Sort options so that it is quick to add/remove Sort options, just by adding/removing an array key. Added better protection for the Orderby Sql clause. Added better xss protection in the search template.


classes/class-p4-search.php
```diff
@@ -0,0 +1,244 @@
...
+		 * Adds the context for each Search result item.
+		 *
+		 * @param array $context Associative array with the data to be passed to the view.
+		 */
+		protected function add_items_context( &$context ) {
+
+			foreach ( $this->posts as $post ) {
+				switch ( $post->post_type ) {
+					case 'page':
+						if ( 'act' === basename( get_permalink( $post->post_parent ) ) ) {
```
koyan: Note: in the future this will have to change according to https://jira.greenpeace.org/browse/PLANET-1411

koyan:
Looks good 👍

PR #92, opened 2017-11-17
Feature/1502 search filtering

angtheod:
Implemented dynamic Filters and result counting for each one of them (Content type filters are hardcoded). Filters are sorted alphabetically and only those that have results appear. Filter boxes show/expand only if they have filters inside of them. Replaced pagination with Load more button functionality. The button shows the next 5 results and hides itself if there are no more to show.

Currently, we are not caching anything yet and the actual filtering is not yet completed.
We are retrieving all results in order to be able to have dynamic filters and counters.


koyan:
Looks good 👍

PR #97, opened 2017-11-22
Presentation templates disparities

angtheod:
Fix Header Block disparity so that description text does not spread across page width. Also, fixed small issue where the Separator bullet between Issue and Campaign links appeared even if there were no Campaign links. Adjusted Campaign page markup to be similar to presentation layer markup and remove unecessary classes and aligned its header to start at the same height as the other pages headers.

### DIFF
--- a/templates/tag.twig
+++ b/templates/tag.twig
@@ -8,26 +8,25 @@
 				background-size: cover;
 			}
 			.tag_template {
-				padding-top: 100px;
+				padding-top: 200px;
 			}
-			.tag_template .tag_header_container {
+			.tag_template > .container {
 				color: white;
 			}
-			.category_link {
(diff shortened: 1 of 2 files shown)


templates/tag.twig
```diff
@@ -8,26 +8,25 @@
...
 			.tag_template {
-				padding-top: 100px;
+				padding-top: 200px;
 			}
-			.tag_template .tag_header_container {
+			.tag_template > .container {
 				color: white;
 			}
-			.category_link {
+			.page-header > a {
```
angtheod: @ashish173 
This is the css that I had used for the code-generated Campaign page. I guess it should be in presentation layer as well.

koyan:
Looks good 👍

PR #100, opened 2017-11-23
Feature/1503 search

angtheod:
Change Load more button behavior to reveal more results just by scrolling down the first 2 times as requested by design initially. Apply wp js standards.

*There is a [PR](https://github.com/greenpeace/planet4-presentation-layer/pull/37) for this on presentation-layer.


koyan:
Looks good 👍

PR #101, opened 2017-11-24
PLANET-1411 Added settings for ACT, EXPLORE PAGE and ISSUE CATEGORY

sagarsdeshmukh:
Added admin settings option for ACT, EXPLORE PAGE and ISSUE CATEGORY

### DIFF
--- a/classes/class-p4-settings.php
+++ b/classes/class-p4-settings.php
@@ -10,7 +10,7 @@ if ( ! class_exists( 'P4_Settings' ) ) {
 		 * Option key, and option page slug
 		 * @var string
 		 */
-		private $key = 'cmb2_planet4_options';
+		private $key = 'planet4_options';
 
 		/**
 		 * Array of metaboxes/fields
@@ -37,39 +37,42 @@ if ( ! class_exists( 'P4_Settings' ) ) {
 		public function __construct() {
 
 			// Set our title
-			$this->title = __( 'Planet4', 'planet4-master-theme' );
-			$pages = get_pages();
+			$this->title   = __( 'Planet4', 'planet4-master-theme' );
+			$pages         = get_pages();
 
+			$pages_array[] = __( 'Select Page' ,'planet4-master-theme' ) ;
 			foreach ( $pages as $all_pages ) {
-				$new_pages[$all_pages->ID] = __( $all_pages->post_title ,'planet4-master-theme' ) ;
+				$pages_array[$all_pages->ID] = __( $all_pages->post_title ,'planet4-master-theme' ) ;
 			}
 
-			$categories = get_categories();
+			$categories         = get_categories();
+
+			$categories_array[] = __( 'Select Category' ,'planet4-master-theme' ) ;
 			foreach ( $categories as $category ) {
-				$category_names[$category->term_id] = __( $category->cat_name ,'planet4-master-theme' );
+				$categories_array[$category->term_id] = __( $category->cat_name ,'planet4-master-theme' );
 			}
 
 			// Set our CMB2 fields, wrap them in a filter so others can easily tap in and add their own as well.
-			$this->fields = apply_filters( 'cmb2_planet4_options', array(
+			$this->fields = apply_filters( 'planet4_options', array(
 					array(
 						'name'    => __( 'Select act page', 'planet4-master-theme' ),
 						'id'      => 'select_act_page',
 						'type'    => 'select',
-						'options' => $new_pages
+						'options' => $pages_array
 					),
 
 					array(
 						'name'    => __( 'Select explore page', 'planet4-master-theme' ),
 						'id'      => 'select_explore_page',
 						'type'    => 'select',
-						'options' => $new_pages
+						'options' => $pages_array
 					),
 
 					array(
 						'name'    => __( 'Select category', 'planet4-master-theme' ),
 						'id'      => 'select_category',
 						'type'    => 'select',
-						'options' => $category_names
+						'options' => $categories_array
 					)
 
 				) );
@@ -155,13 +158,13 @@ if ( ! class_exists( 'P4_Settings' ) ) {
  * @param  string  $key Options array key
  * @return mixed        Option value
  */
-function cmb2_planet4_get_option( $key = '' ) {
-	global $cmb2_planet4_options;
+function planet4_get_option( $key = '' ) {
+	global $planet4_options;
 
 	if( function_exists( 'cmb2_get_option' ) ) {
-		return cmb2_get_option( 'cmb2_planet4_options', $key );
+		return cmb2_get_option( 'planet4_options', $key );
 	} else {
-		$options = get_option( 'cmb2_planet4_options' );
+		$options = get_option( 'planet4_options' );
 		return isset( $options[ $key ] ) ? $options[ $key ] : false;
 	}
 }
--- a/classes/class-p4-settings.php
+++ b/classes/class-p4-settings.php
@@ -32,7 +32,6 @@ if ( ! class_exists( 'P4_Settings' ) ) {
 
 		/**
 		 * Constructor
-		 * @since 0.1.0
 		 */
 		public function __construct() {
 
@@ -52,7 +51,6 @@ if ( ! class_exists( 'P4_Settings' ) ) {
 				$categories_array[$category->term_id] = __( $category->cat_name ,'planet4-master-theme' );
 			}
 
-			// Set our CMB2 fields, wrap them in a filter so others can easily tap in and add their own as well.
 			$this->fields = apply_filters( 'planet4_options', array(
 					array(
 						'name'    => __( 'Select act page', 'planet4-master-theme' ),
@@ -81,7 +79,6 @@ if ( ! class_exists( 'P4_Settings' ) ) {
 
 		/**
 		 * Initiate our hooks
-		 * @since 0.1.0
 		 */
 		public function hooks() {
 			add_action( 'admin_init', array( $this, 'init' ) );
@@ -90,7 +87,6 @@ if ( ! class_exists( 'P4_Settings' ) ) {
 
 		/**
 		 * Register our setting to WP
-		 * @since  1.0
 		 */
 		public function init() {
 			register_setting( $this->key, $this->key );
@@ -98,7 +94,6 @@ if ( ! class_exists( 'P4_Settings' ) ) {
 
 		/**
 		 * Add menu options page
-		 * @since 0.1.0
 		 */
 		public function add_options_page() {
 			$this->options_page = add_options_page( $this->title, $this->title, 'manage_options', $this->key, array( $this, 'admin_page_display' ) );
@@ -106,7 +101,6 @@ if ( ! class_exists( 'P4_Settings' ) ) {
 
 		/**
 		 * Admin page markup. Mostly handled by CMB2
-		 * @since  1.0
 		 */
 		public function admin_page_display() {
 			?>
@@ -119,7 +113,6 @@ if ( ! class_exists( 'P4_Settings' ) ) {
 
 		/**
 		 * Defines the theme option metabox and field configuration
-		 * @since  1.0
 		 * @return array
 		 */
 		public function option_metabox() {
@@ -133,7 +126,6 @@ if ( ! class_exists( 'P4_Settings' ) ) {
 
 		/**
 		 * Public getter method for retrieving protected/private variables
-		 * @since  1.0
 		 * @param  string  $field Field to retrieve
 		 * @return mixed          Field value or exception is thrown
 		 */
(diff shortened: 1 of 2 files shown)


classes/class-p4-settings.php
```diff
@@ -0,0 +1,167 @@
...
+	/**
+	 * Class P4_Settings
+	 */
+	class P4_Settings {
+
+		/**
+		 * Option key, and option page slug
+		 * @var string
+		 */
+		private $key = 'cmb2_planet4_options';
```
angtheod: I wouldn't include 'cmb2' in this string.

classes/class-p4-settings.php
```diff
@@ -0,0 +1,170 @@
...
+
+		/**
+		 * Options Page hook
+		 * @var string
+		 */
+		protected $options_page = '';
+
+		/**
+		 * Constructor
+		 * @since 0.1.0
```
koyan: What are all those since 0.1.0 lines in the comments?

sagarsdeshmukh: Yes we do not need versions - i will update with the proper docblock

classes/class-p4-settings.php
```diff
@@ -0,0 +1,161 @@
...
+			foreach ( $pages as $all_pages ) {
+				$pages_array[$all_pages->ID] = __( $all_pages->post_title ,'planet4-master-theme' ) ;
+			}
+
+			$categories         = get_categories();
+
+			$categories_array[] = __( 'Select Category' ,'planet4-master-theme' ) ;
+			foreach ( $categories as $category ) {
+				$categories_array[$category->term_id] = __( $category->cat_name ,'planet4-master-theme' );
+			}
```
angtheod: @sagarsdeshmukh Don't we want to be able to choose the 'parent' category, so that we
can then get all the children categories dynamically like referenced [here](https://github.com/greenpeace/planet4-plugin-blocks/blob/5c8f29efe07e6bf490896013ab06006cc5bb37ee/classes/controller/blocks/class-p4bks-blocks-splittwocolumns-controller.php#L24)

classes/class-p4-settings.php
```diff
@@ -0,0 +1,161 @@
...
+						'type'    => 'select',
+						'options' => $pages_array
+					),
+
+					array(
+						'name'    => __( 'Select category', 'planet4-master-theme' ),
+						'id'      => 'select_category',
+						'type'    => 'select',
+						'options' => $categories_array
+					)
```
angtheod: @sagarsdeshmukh Please, use short arrays [] and add a comma even in the last key => value pair.

classes/class-p4-settings.php
```diff
@@ -0,0 +1,161 @@
...
+		 */
+		public function __construct() {
+
+			// Set our title
+			$this->title   = __( 'Planet4', 'planet4-master-theme' );
+			$pages         = get_pages();
+
+			$pages_array[] = __( 'Select Page' ,'planet4-master-theme' ) ;
+			foreach ( $pages as $all_pages ) {
+				$pages_array[$all_pages->ID] = __( $all_pages->post_title ,'planet4-master-theme' ) ;
```
angtheod: spaces need fixing in this line

angtheod: Also, WP's translation functions expect string literals and not variables so that the translator can actually translate the string. Here you do not need to pass the post title to the translation function. The title will already be in the language that the editor wrote it.

classes/class-p4-settings.php
```diff
@@ -0,0 +1,161 @@
...
+		}
+
+		/**
+		 * Defines the theme option metabox and field configuration
+		 * @return array
+		 */
+		public function option_metabox() {
+			return array(
+				'id'         => 'option_metabox',
+				'show_on'    => array( 'key' => 'options-page', 'value' => array( $this->key, ), ),
```
angtheod: Need short arrays and each pair of values in separate line.

functions.php
```diff
@@ -598,10 +598,19 @@ function edit_searchwp_query_orderby( $orderby ) {
 	public function populate_act_page_children_options() {
 
 		// Get the id of the ACT page. We need this to get the children posts/pages of the ACT Page.
-		$arguments = [
-			'post_type'     => 'page',
-			'post_name__in' => [ 'act', 'ACT', 'Act' ],
-		];
+		$parent_act_id = planet4_get_option( 'select_act_page' );
+		if( 0 != $parent_act_id ) {
```
angtheod: Better check type also here by using !==

classes/class-p4-settings.php
```diff
@@ -0,0 +1,161 @@
...
+	}
+}
+
+/**
+ * Wrapper function around cmb2_get_option
+ * @param  string  $key Options array key
+ * @return mixed        Option value
+ */
+function planet4_get_option( $key = '' ) {
+	global $planet4_options;
```
angtheod: Unused global var.

functions.php
```diff
@@ -598,10 +598,19 @@ function edit_searchwp_query_orderby( $orderby ) {
...
+			$act_page_object  = get_post( $parent_act_id );
+			$arguments = [
+				'post_type'     => 'page',
+				'post_name__in' => [ $act_page->post_name ],
```
angtheod: @sagarsdeshmukh did you mean to use $act_page_object here?

functions.php
```diff
@@ -598,10 +598,19 @@ function edit_searchwp_query_orderby( $orderby ) {
 	public function populate_act_page_children_options() {
 
 		// Get the id of the ACT page. We need this to get the children posts/pages of the ACT Page.
-		$arguments = [
-			'post_type'     => 'page',
-			'post_name__in' => [ 'act', 'ACT', 'Act' ],
-		];
+		$parent_act_id = planet4_get_option( 'select_act_page' );
```
angtheod: @sagarsdeshmukh Correct me if I am wrong but you do have the chosen Act Parent page id here and then you are doing things in lines 603-621 just to get that id that you already have?

classes/class-p4-settings.php
```diff
@@ -0,0 +1,161 @@
...
+		}
+	}
+}
+
+/**
+ * Wrapper function around cmb2_get_option
+ * @param  string  $key Options array key
+ * @return mixed        Option value
+ */
+function planet4_get_option( $key = '' ) {
```
angtheod: In my opinion I don't think you really need this global function here. Better to let in here only the class.
I would call the cmb2_get_option() directly.

classes/class-p4-settings.php
sagarsdeshmukh: @angtheod Yes, Angelos if here we don't choose parent category then it will work as per reference. [here](https://github.com/greenpeace/planet4-plugin-blocks/blob/feature/1411-make-dynamic-p4-settings/classes/controller/blocks/class-p4bks-blocks-splittwocolumns-controller.php#L24)

classes/class-p4-settings.php
sagarsdeshmukh: I will update the array syntax with short array [].

classes/class-p4-settings.php
sagarsdeshmukh: Fixed.

functions.php
sagarsdeshmukh: If we use !== with 0 it check boolean datatypes, because of that I use !=.

classes/class-p4-settings.php
sagarsdeshmukh: Fixed.

classes/class-p4-settings.php
sagarsdeshmukh: Yes, got it.

classes/class-p4-settings.php
sagarsdeshmukh: If we use wrapper function planet4_get_option(), we will get more control over output of cmb2_get_option() function.

functions.php
sagarsdeshmukh: Yes I mean the $act_page_object variable. I fixed it.

functions.php
sagarsdeshmukh: Yes agree, I optimize the populate_act_page_children_options() function.

classes/class-p4-settings.php
angtheod: @sagarsdeshmukh @koyan 
Hey Sagar, I think we need to give them the option to select the parent category (for example called 'Issues') and not the children categories.
Also, I think it would be better to name the drop down label 'Select Issues parent category' instead of just 'Select category'.

angtheod: @sagarsdeshmukh @koyan 
Hey Sagar, I think you are getting only the categories that have at least 1 post assigned to them or that have children categories. I believe we want all categories to be listed in there.

It is worth noting that we don't need to use CMB2 for these fields. WP has 2 built-in functions that do exactly what we want (and better actually, since they listed options inside the drop down can be displayed in hierarchical way).

Take a look at those 2 functions:

`wp_dropdown_pages( [
				'hide_empty'   => 0,
				'hierarchical' => true,
			] );
			wp_dropdown_categories( [
				'hide_empty'   => 0,
				'hierarchical' => true,
				'orderby'      => 'name',
			] );`

Also, I think it would be better to name the drop down label 'Select Issues parent category' instead of just 'Select category'.

classes/class-p4-settings.php
```diff
@@ -0,0 +1,164 @@
...
+			$pages_array[] = __( 'Select Page', 'planet4-master-theme' ) ;
+			foreach ( $pages as $all_pages ) {
+				$pages_array[$all_pages->ID] = $all_pages->post_title;
+			}
+
+			$categories         = get_categories();
+
+			$categories_array[] = __( 'Select Category', 'planet4-master-theme' ) ;
+			foreach ( $categories as $category ) {
+				$categories_array[$category->term_id] = __( $category->cat_name, 'planet4-master-theme' );
```
angtheod: @sagarsdeshmukh 
Same, for this one, object property doesn't need translation

classes/class-p4-settings.php
```diff
@@ -0,0 +1,193 @@
...
+			if ( in_array( $field, [ 'key', 'fields', 'title', 'options_page' ], true ) ) {
+				return $this->{$field};
+			}
+
+			if ( 'option_metabox' === $field ) {
+				return $this->option_metabox();
+			}
+
+			throw new Exception( 'Invalid property: ' . $field );
+		}
```
angtheod: @sagarsdeshmukh 
Do we really need to declare this magic method?
I do not see any checks prior to returning the protected/private properties, so I do not see why to allow that.
Also, option_metabox() method is already publicly accessible.

functions.php
angtheod: Yes, Sagar, that is the point. It is in the WordPress coding standards to use strict comparisons.

classes/class-p4-settings.php
```diff
@@ -0,0 +1,193 @@
...
+		 */
+		public function __construct() {
+
+			// Set our title
+			$this->title   = __( 'Planet4', 'planet4-master-theme' );
+
+			$this->fields = apply_filters( 'planet4_options', [
+				[
+					'name'    => __( 'Select act page', 'planet4-master-theme' ),
+					'id'      => 'select_act_page',
```
koyan: As noted in the blocks plugin, please change the name from "select_act_page" to "act_page"

classes/class-p4-settings.php
```diff
@@ -0,0 +1,193 @@
...
+					'type'    => 'act_page_dropdown',
+				],
+
+				[
+					'name'    => __( 'Select explore page', 'planet4-master-theme' ),
+					'id'      => 'select_explore_page',
```
koyan: change name of param

classes/class-p4-settings.php
```diff
@@ -0,0 +1,193 @@
...
+					'type'    => 'explore_page_dropdown',
+				],
+
+				[
+					'name'    => __( 'Select issues parent category', 'planet4-master-theme' ),
+					'id'      => 'select_category',
```
koyan: change name of param

classes/class-p4-settings.php
angtheod: @sagarsdeshmukh 
But why do you need cmb2_get_option? Does it do something that get_option() does not?
In my opinion wherever we have perfectly good and trusted WP functions that do the job, we should use those to avoid getting too tightly coupled with CMB2 and for future developers to be able to understand faster the code.

functions.php
```diff
@@ -760,5 +750,5 @@ public function register_header_metabox() {
 }
 
 new P4_Master_Site( [
-	'P4_Taxonomy_Image',
+	'P4_Taxonomy_Image','P4_Settings',
```
angtheod: Please, if that's ok with you, put each class name in new line :)

koyan:
Looks good 👍

PR #104, opened 2017-11-28
Fix php warnings for unset variables in page and post files.

kirdia:


Kluny:
Noting that this commit failed the travis check because it uses the null coalescing operator (??) which is new in PHP 7 and incompatible with 5.6. The client confirms that 5.6 compatibility is not required, so this is fine.

PR #105, opened 2017-11-28
Feature/1502 search filtering

angtheod:
Applied filtering. Set a limit (100) in returned results to avoid returning too many. Search for empty string now returns 0 results (instead of everything). Bug fixes in filtering and search page template. JS code standards.

### DIFF
--- a/classes/class-p4-search.php
+++ b/classes/class-p4-search.php
@@ -184,15 +184,19 @@ if ( ! class_exists( 'P4_Search' ) ) {
 
 			$ids          = [];
 			$timber_posts = [];
-			$swp_posts    = ( new SWP_Query( $args ) )->posts;
 
-			foreach ( $swp_posts as $post ) {
+			$posts  = ( new SWP_Query( $args ) )->posts;
+			foreach ( $posts as $post ) {
 				$ids[] = $post->ID;
 			}
-			$args['post__in'] = $ids;
-			$args['orderby'] = 'post__in';
 
-			$posts = ( new WP_Query( $args ) )->posts;
+			// WP_Query is not capable of searching within rich text documents.
+			if ( 'attachment' !== $args['post_type'] ) {
+				$args['post__in'] = $ids;
+				$args['orderby']  = 'post__in';
+				$posts            = ( new WP_Query( $args ) )->posts;
+			}
+
 			foreach ( $posts as $post ) {
 				// Use Timber's Post instead of WP_Post so that we can make use of Timber within the template.
 				$timber_posts[] = new TimberPost( $post->ID );
--- a/classes/class-p4-search.php
+++ b/classes/class-p4-search.php
@@ -21,6 +21,8 @@ if ( ! class_exists( 'P4_Search' ) ) {
 		protected $all_posts;
 		/** @var array|bool|null $posts */
 		protected $posts;
+		/** @var array $selected_sort */
+		protected $selected_sort;
 		/** @var array $filters */
 		protected $filters;
 		/** @var array $templates */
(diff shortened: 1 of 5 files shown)


classes/class-p4-search.php
```diff
@@ -43,45 +45,195 @@ public function __construct( $search_query, $templates = [ 'search.twig', 'archi
...
+			}
+		}
+
+		/**
+		 * Handle form submit.
+		 *
+		 * @param array $context Associative array with the data to be passed to the view.
+		 */
+		public function handle_submit( &$context ) {
+			if ( 'GET' === $_SERVER['REQUEST_METHOD'] ) {
```
kirdia: I think we should eliminate the use of superglobals in the class. This check (if request is GET) should be in an upper level (eg. search.php)

classes/class-p4-search.php
```diff
@@ -43,45 +45,195 @@ public function __construct( $search_query, $templates = [ 'search.twig', 'archi
...
+				// Handle submitted sort options.
+				$selected_sort = filter_input( INPUT_GET, 'orderby', FILTER_SANITIZE_STRING );
+				if ( ! in_array( $selected_sort, array_keys( $context['sort_options'] ), true ) ) {
+					$context['selected_sort'] = self::DEFAULT_SORT;
+				} else {
+					$context['selected_sort'] = $selected_sort;
+				}
+
+				// Handle submitted filter options.
+				if ( is_array( $_GET['f'] ) ) {
```
kirdia: Again I think we should pass the search query parameters in another way in the class constructor to eliminate superglobals use in the class. That way the class would be more easy to test the class and it's functions.

classes/class-p4-search.php
```diff
@@ -43,45 +45,195 @@ public function __construct( $search_query, $templates = [ 'search.twig', 'archi
...
+				$ids = [];
+				foreach ( $posts as $post ) {
+					$ids[] = $post->ID;
+				}
+				$args['post__in'] = $ids;
+				$args['orderby']  = 'post__in';
+				$posts            = ( new WP_Query( $args ) )->posts;
 			}
+
+			$timber_posts = [];
```
kirdia: Maybe it is better to move the conversion of WP_Post objects to TimberPost objects in a different function to keep the function from serving several purposes. Because search is a vital component of the app (and probably it will get more complex in the future) we should implement tests at some point. So I think is better to keep the functions as small as possible, to server a single concern.

classes/class-p4-search.php
```diff
@@ -43,45 +45,195 @@ public function __construct( $search_query, $templates = [ 'search.twig', 'archi
...
+		/**
+		 * Applies user selected filters to the search if there are any and gets the filtered posts.
+		 * If there are not then uses Timber's get_posts to retrieve all of them (up to the limit set).
+		 *
+		 * @return array The posts of the search.
+		 */
+		protected function get_posts() : array {
+			global $wp_query;
+
+			if ( ! $wp_query->is_main_query() || ! $wp_query->is_search() ) {
```
kirdia: I don't understand the purpose for these checks, but maybe it is better to move them in a an upper level (search.php).
I think it would be better for the separation of concerns.

classes/class-p4-search.php
```diff
@@ -137,26 +133,26 @@ protected function get_posts() : array {
...
+										'terms'    => $filter['id'],
 									];
 								}
 								break;
 							case 'ctype':
 								switch ( $filter['id'] ) {
 									case 0:
-										$args['post_type']   = 'page';
-										$args['post_parent'] = get_page_by_path( 'act', 'OBJECT', 'page' )->ID;
+										$args['post_type'] = 'page';
```
sagarsdeshmukh: Please fix indentations here

classes/class-p4-search.php
```diff
@@ -137,26 +133,26 @@ protected function get_posts() : array {
...
+										$options = get_option( 'planet4_options' );
+										$args['post_parent'] = $options['act_page'];
 										break;
 									case 1:
 										$args['post_type'] = 'attachment';
 										break;
 									case 2:
-										$args['post_type']             = 'page';
-										$args['post_parent__not_in'][] = get_page_by_path( 'act', 'OBJECT', 'page' )->ID;
+										$args['post_type'] = 'page';
```
sagarsdeshmukh: Please fix indentations here

kirdia:
Maybe we can break down the twig template further because it is heavy indented, but in a future iteration.
For example in `filter-sidebar-options-modal` we can make a twig template for filter types (`filter-item`) and pass the data that vary.
Overall it looks good 👍

angtheod:
Imagine it is already broken down. The markup for each result item is in tease-search.twig.
But yes, I think we could place into one template file the markup for the filter checkboxes, which is similar for the Modal and the Normal page version.

PR #108, opened 2017-11-29
Add dropdown for picking p4 custom taxo type

Kluny:
p4_metabox_markup() - callback method for adding a metabox with a dropdown containing three pr custom post types, Publication, Press Release, and Story.

p4_save_post_type() - action on the save hook for saving custom post type as submitted through dropdown.

Second attempt, correcting a git mistake.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -120,6 +120,7 @@ class P4_Master_Site extends TimberSite {
 		add_action( 'admin_enqueue_scripts',  array( $this, 'enqueue_admin_assets' ) );
 		add_action( 'wp_enqueue_scripts',     array( $this, 'enqueue_public_assets' ) );
 		add_filter( 'wp_kses_allowed_html',   array( $this, 'set_custom_allowed_attributes_filter' ) );
+		add_action( 'save_post', 			  array($this, 'p4_save_post_type'), 10, 2 );
 
 		remove_action( 'wp_head', 'print_emoji_detection_script', 7 );
 		remove_action( 'wp_head', 'wp_generator' );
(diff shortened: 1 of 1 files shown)


functions.php
```diff
@@ -278,7 +281,7 @@ public function engaging_network_id_show_settings( $args ) {
 	 */
 	public function cookies_show_settings( $args ) {
 		$cookies_text = get_option( 'cookies_field', '' );
-		$args = [
+		$args         = [
```
mikeselander: There's quite a bit of un-related spacing updates in this PR, I'd prefer to see just the changes that are required and any directly-related code.

functions.php
```diff
@@ -527,10 +532,75 @@ public function register_p4_page_type_taxonomy() {
 		}
 	}
 
+	/**
+	 * Save custom taxonomy for planet4 post types
+	 */
+	public function p4_save_post_type( $post_id ) {
+		// some of these checks might be redundant, but they're all nicely
```
mikeselander: Minor tick: Multi-line comments are generally in the `/** */` format

functions.php
```diff
@@ -527,10 +532,75 @@ public function register_p4_page_type_taxonomy() {
...
+		// separated and easy to delete so I'll leave them for now.
+		// ignore autosave
+		if ( defined( 'DOING_AUTOSAVE' ) && DOING_AUTOSAVE ) {
+			return;
+		}
+		// check nonce
```
mikeselander: Another minor tick, but inline comments should start with a capital letter and end with a stop character like a period.

functions.php
```diff
@@ -527,10 +532,75 @@ public function register_p4_page_type_taxonomy() {
...
+		if ( ! isset( $_POST['p4-page-type'] ) ) { // Input var okay.
+			return;
+		}
+		// make sure the term exists
+		if ( ! $selected = get_term_by( 'slug', sanitize_text_field( $_POST['p4-page-type'] ), 'p4-page-type' ) ) { // Input var okay.
+			return;
+		}
+		// make sure it's not an error
+		if ( is_wp_error( $selected ) ) {
+			return $post_id;
```
mikeselander: Why return `$post_id` here and null in the rest of the checks?

functions.php
```diff
@@ -527,10 +532,75 @@ public function register_p4_page_type_taxonomy() {
...
+        <div>
+        <select name="p4-page-type"><?
+			foreach ( $terms as $term ) {
+				$selected = ( $current === $term->slug ) ? 'selected="selected"' : '';
+				?>
+                <option <?= $selected ?> value="<?= $term->slug ?>"><?= $term->name ?></option>
+			<? }
+			$selected = ( - 1 === $current ) ? 'selected="selected"' : '';
+			?>
+            <option value="-1" <?= $selected ?> >none</option>
```
mikeselander: It's not good practice to use the `<?=` syntax as some servers can't handle it and it's also less expressive.

functions.php
```diff
@@ -527,10 +532,75 @@ public function register_p4_page_type_taxonomy() {
...
+		}
+		$terms = get_terms( 'p4-page-type', [ 'hide_empty' => false ] );
+		wp_nonce_field( basename( __FILE__ ), 'p4-page-type-nonce' );
+		?>
+        <div>
+        <select name="p4-page-type"><?
+			foreach ( $terms as $term ) {
+				$selected = ( $current === $term->slug ) ? 'selected="selected"' : '';
+				?>
+                <option <?= $selected ?> value="<?= $term->slug ?>"><?= $term->name ?></option>
```
mikeselander: `$term->name` and `$term->slug` need to be escaped and also should use core's `selected()` function instead of writing your own logic for this.

functions.php
```diff
@@ -761,4 +831,4 @@ public function register_header_metabox() {
 
 new P4_Master_Site( [
 	'P4_Taxonomy_Image',
-] );
+] );
```
mikeselander: Missing EOF line here.

functions.php
```diff
@@ -527,10 +532,75 @@ public function register_p4_page_type_taxonomy() {
...
+            <option value="-1" <?= $selected ?> >none</option>
+        </select>
+        </div><?
+	}
+
 	/**
 	 * Registers custom post types.
 	 */
-	public function register_post_types() {}
+	public function register_post_types() {
```
mikeselander: What's going on with this empty function here?

functions.php
```diff
@@ -527,10 +532,75 @@ public function register_p4_page_type_taxonomy() {
...
+			if ( ! is_wp_error( $current_term ) ) {
+				$current = $current_term[0]->slug;
+			}
+		}
+		$terms = get_terms( 'p4-page-type', [ 'hide_empty' => false ] );
+		wp_nonce_field( basename( __FILE__ ), 'p4-page-type-nonce' );
+		?>
+        <div>
+        <select name="p4-page-type"><?
+			foreach ( $terms as $term ) {
```
mikeselander: When breaking up a logic block (`if`, `foreach`, etc.) with HTML output we prefer to use the 

```

File: functions.php [hunk 10]
```diff
@@ -527,10 +532,75 @@ public function register_p4_page_type_taxonomy() {
...
+		// check user's capabilities
+		if ( ! current_user_can( 'edit_post', $post_id ) ) {
+			return;
+		}
+		// make sure there's input
+		if ( ! isset( $_POST['p4-page-type'] ) ) { // Input var okay.
+			return;
+		}
+		// make sure the term exists
+		if ( ! $selected = get_term_by( 'slug', sanitize_text_field( $_POST['p4-page-type'] ), 'p4-page-type' ) ) { // Input var okay.
```
mikeselander: `$selected` should be set outside of the if statement and the if statement then used to check both that it is not false and is not a `WP_Error`. This is mostly for readability.

functions.php
```diff
@@ -527,10 +532,75 @@ public function register_p4_page_type_taxonomy() {
 		}
 	}
 
+	/**
+	 * Save custom taxonomy for planet4 post types
```
mikeselander: Missing `@param` in the docblock.

functions.php
```diff
@@ -527,10 +532,75 @@ public function register_p4_page_type_taxonomy() {
...
+			return;
+		}
+		// make sure it's not an error
+		if ( is_wp_error( $selected ) ) {
+			return $post_id;
+		}
+		// save post type
+		wp_set_post_terms( $post_id, $selected->slug, 'p4-page-type', $append = false );
+
+		return;
```
mikeselander: No need to return at the end of a function.

functions.php
```diff
@@ -527,10 +532,75 @@ public function register_p4_page_type_taxonomy() {
...
+	}
+
+	/**
+	 * Add a dropdown to choose planet4 post type.
```
mikeselander: Missing `@param` in docblock.

functions.php
```diff
@@ -527,10 +532,75 @@ public function register_p4_page_type_taxonomy() {
...
+	 */
+	public function p4_metabox_markup( $object ) {
+		get_post_meta( $object->ID );
+		$current = - 1;
+		if ( $current_term = get_the_terms( $object, 'p4-page-type' ) ) {
+			if ( ! is_wp_error( $current_term ) ) {
```
mikeselander: You're using 2 levels of `if`s to do what one would achieve just as well. In my opinion it would be cleaner to do something along the lines of:

```
$current_term = get_the_terms( $object, 'p4-page-type' );
$current = ( $current_term && ! is_wp_error( $current_term ) ) ? $current_term[0]->slug : -1;
```

functions.php
```diff
@@ -527,10 +532,75 @@ public function register_p4_page_type_taxonomy() {
...
+		// save post type
+		wp_set_post_terms( $post_id, $selected->slug, 'p4-page-type', $append = false );
+
+		return;
+	}
+
+	/**
+	 * Add a dropdown to choose planet4 post type.
+	 */
+	public function p4_metabox_markup( $object ) {
```
mikeselander: It might be worth checking that we're dealing with a WP_Post upon entry into this function. Makes it easier for IDEs to run interpret what's going on and makes the errors pretty easy to read through if something does go wrong.

functions.php
Kluny: Added.

functions.php
Kluny: Spacing reverted.

functions.php
Kluny: Changed to /** */ format.

functions.php
Kluny: Added capitals and punctuation to single-line comments in this method.

functions.php
Kluny: Amended as suggested.

functions.php
Kluny: Totally didn't notice it! Removed.

functions.php
Kluny: Removed.

functions.php
Kluny: Added @param.

functions.php
Kluny: Added a check.

functions.php
Kluny: That solution works; changed it. I noticed while testing that choosing "none" doesn't do anything though. Not sure whether the right behaviour would be to remove the "none" option after one of the others has been selected, or add a method to remove the chosen page type if "none" is selected. Probably the latter.

functions.php
Kluny: Changed foreach to alternative syntax.

functions.php
Kluny: Replaced my logic with `selected()`.

functions.php
Kluny: Amended just in case.

functions.php
Kluny: I'm not sure. There's a reference to it at line 118 - maybe it's a stub that someone else left there. I'll remove it in a different PR, I guess.

functions.php
Kluny: Weird. Added?

functions.php
Kluny: Thinking about it, I'm not sure how the spacing got like that. I didn't go through the file correcting spacing or anything, and I don't recall running phpcbf. Maybe phpstorm did it automatically somehow? Then it might be evidence of the revision tree getting out of order. I'll watch it closely on my next commit.

Kluny: Looks like my attempt to fix it just made the next commit messier. I'll leave it and take note of that for the future.

functions.php
```diff
@@ -333,9 +344,87 @@ public function register_p4_page_type_taxonomy() {
...
+		// Make sure the term exists and it's not an error.
+        $selected = get_term_by( 'slug', sanitize_text_field( $_POST['p4-page-type'] ), 'p4-page-type' ); // Input var okay.
+
+		if ( false === $selected || is_wp_error( $selected ) ) {
+			return;
+		}
+		// Save post type.
+		wp_set_post_terms( $post_id, $selected->slug, 'p4-page-type', $append = false );
+    }
+	/**
```
mikeselander: Missing space here.

functions.php
```diff
@@ -333,9 +344,87 @@ public function register_p4_page_type_taxonomy() {
...
+	 * Removes primary category from a post.
+	 *
+	 * @param $post_id
```
mikeselander: Missing information on this param comment.

functions.php
```diff
@@ -333,9 +344,87 @@ public function register_p4_page_type_taxonomy() {
...
+	 */
+	public function p4_remove_page_type( $post_id ) {
+		if ( $term = get_the_terms( $post_id, 'p4-page-type' ) ) {
```
mikeselander: Can we change this to an early return instead of this nested conditional?

Nested conditionals are far harder to maintain and review.

functions.php
```diff
@@ -333,9 +344,87 @@ public function register_p4_page_type_taxonomy() {
...
+			if ( ! is_wp_error( $term ) ) {
+				wp_remove_object_terms( $post_id, $term[0]->term_id, 'p4-page-type' );
+			}
+		}
+	}
+	/**
```
mikeselander: Missing space.

functions.php
```diff
@@ -333,9 +344,87 @@ public function register_p4_page_type_taxonomy() {
...
+	 * Add a dropdown to choose planet4 post type.
+     *
+     * @param WP_Post $object
+	 */
+	public function p4_metabox_markup( $object ) {
```
mikeselander: Instead of the conditional below, you could do the following which would be cleaner:

`p4_metabox_markup( WP_Post  $object ) {`

This will enforce the correct type coming into this function for you.

functions.php
```diff
@@ -333,9 +344,87 @@ public function register_p4_page_type_taxonomy() {
...
+		$current = ( $current_term && ! is_wp_error( $current_term ) ) ? $current_term[0]->slug : -1;
+
+		$terms = get_terms( 'p4-page-type', [ 'hide_empty' => false ] );
+		wp_nonce_field( basename( __FILE__ ), 'p4-page-type-nonce' );
+		?>
+        <div>
+        <select name="p4-page-type"><?php
+			foreach ( $terms as $term ) :
+            ?>
+                <option <?php selected($current, $term->slug) ?> value="<?php echo $term->slug ?>"><?php echo $term->name ?></option>
```
mikeselander: These values are still not escaped.

functions.php
```diff
@@ -333,9 +344,87 @@ public function register_p4_page_type_taxonomy() {
...
+
+		get_post_meta( $object->ID );
+
+		$current_term = get_the_terms( $object, 'p4-page-type' );
+		$current = ( $current_term && ! is_wp_error( $current_term ) ) ? $current_term[0]->slug : -1;
+
+		$terms = get_terms( 'p4-page-type', [ 'hide_empty' => false ] );
+		wp_nonce_field( basename( __FILE__ ), 'p4-page-type-nonce' );
+		?>
+        <div>
```
mikeselander: Point of curiosity: why the extra div here?

functions.php
```diff
@@ -333,9 +344,87 @@ public function register_p4_page_type_taxonomy() {
...
+        <select name="p4-page-type"><?php
+			foreach ( $terms as $term ) :
```
mikeselander: This whole PHP line should be on its own line i.e.: `<?php foreach ( $terms as $term ) : ?>`

functions.php
```diff
@@ -333,9 +344,87 @@ public function register_p4_page_type_taxonomy() {
...
+            ?>
+                <option <?php selected($current, $term->slug) ?> value="<?php echo $term->slug ?>"><?php echo $term->name ?></option>
+			<?php
+            endforeach;
+			?>
+            <option value="-1" <?php selected(-1, $current) ?> >none</option>
+        </select>
+        </div><?php
```
mikeselander: The beginning PHP tag should be on it's own line. These are small standards items, but help with overall readability.

Kluny:
Gonna pull from the developer branch again and carefully put my changes back in.

PR #112, opened 2017-12-05
PLANET-1550 Update header block and related changes from page.twig an…

sagarsdeshmukh:
…d base.twig

### DIFF
--- a/templates/blocks/header.twig
+++ b/templates/blocks/header.twig
@@ -1,38 +1,39 @@
 {% if ( header_title ) %}
-	<div class="clearfix page-content-wrap">
-		<div class="fixed-element clearfix">
-			<div class="container">
-				<div class="page-header col-md-8 col-lg-8 col-sm-12 pl-0 pr-0">
-					{% if ( issue or campaigns ) %}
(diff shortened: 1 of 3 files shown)


templates/blocks/header.twig
```diff
@@ -1,38 +1,39 @@
...
+			{% if ( header_subtitle ) %}
+				<h3 class="page-header-subtitle">{{ header_subtitle }}</h3>
+			{% endif %}
+			{% if ( header_description ) %}
+				<p class="page-header-content">{{ header_description|e('wp_kses_post')|raw }}</p>
+			{% endif %}
+			{% if ( header_button_title ) %}
+				<div class="row">
+					<div class="col-md-4">
+						<a href="{{ header_button_link }}" class="btn btn-primary btn-medium page-header-btn">{{ header_button_title }}</a>
```
kirdia: Maybe you can use default here as you did in issue, in case the link is not defined
` issue.link|default('#')`

koyan: If the link is not defined, we should not be showing anything at all! (Add an IF block)

PR #113, opened 2017-12-05
Added extra capabilities on author role

koyan:


functions.php
```diff
@@ -116,6 +116,7 @@ protected function hooks() {
 		add_action( 'admin_enqueue_scripts',  array( $this, 'enqueue_admin_assets' ) );
 		add_action( 'wp_enqueue_scripts',     array( $this, 'enqueue_public_assets' ) );
 		add_filter( 'wp_kses_allowed_html',   array( $this, 'set_custom_allowed_attributes_filter' ) );
+		add_action( 'switch_theme',           'add_role_capabilities' );
```
Kluny: Just wondering why use switch_themes as the hook instead of eg. init?

functions.php
```diff
@@ -140,6 +141,21 @@ private function services( $services = array() ) {
...
+		// gets the author role
+		$role = get_role( 'author' );
+		// This only works, because it accesses the class instance.
+		$role->add_cap( 'edit_others_posts' );
+		$role->add_cap( 'delete_others_posts' );
+		$role->add_cap( 'delete_private_posts' );
+		$role->add_cap( 'edit_private_posts' );
+		$role->add_cap( 'read_private_posts' );
+	}
+
```
Kluny: Looks good

functions.php
koyan: @Kluny I wanted a hook that would run only very rarely (when we install the theme). If I used "init", the function would be called in every page load. Since the add_cap does the changes in the database, it only needs to run once. Running it on every page load would be a waste of resources. (I of course may be wrong, in which case please explain to me).

Kluny: Nope, just curious. That makes sense. 👍

PR #118, opened 2017-12-06
Bugfix/make nav bar link dynamic

sagarsdeshmukh:


templates/navigation-bar.twig
```diff
@@ -26,12 +26,11 @@
...
                                         </div>
                                     </li>
-                                    <li class="nav-item active act-nav-link">
-                                        <a href="{{ data_nav_bar.act_url }}">{{ __( 'Act', domain ) }}</a>
-                                    </li>
-                                    <li class="nav-item explore-nav-link">
-                                        <a href="{{ data_nav_bar.explore_url }}">{{ __( 'Explore', domain ) }}</a>
-                                    </li>
+                                    {% for key,item in navbar_menu.get_items %}
+                                        <li class="nav-item{% if key == 0 %} act-nav-link{% else %} explore-nav-link{% endif %}">
```
kirdia: Just a suggestion, to be more readable, I think it is better to move the if outside and set a variable for the class name.
`{% if key == 0 %}`
`  {% set class_name = 'act-nav-link' %}`
`{% else %}`
`  {% set class_name = 'explore-nav-link' %}`

koyan:
Looks good 👍

PR #120, opened 2017-12-07
Add filters to dequeue devicepx and jitm jetpack scripts.

kirdia:
Add filters to dequeue devicepx and jitm jetpack scripts that are not needed by photon module.


koyan:
Looks good 👍

PR #122, opened 2017-12-08
Applied latest markup to post template.

kirdia:
Changed single post template according to latest markup.
Changed comment form elements according to latest markup.
Top background image is disabled for the moment.

### DIFF
--- a/templates/comments_section.twig
+++ b/templates/comments_section.twig
@@ -1,5 +1,5 @@
 <div class="container page-section">
-	<h2 class="page-section-header">{{ __( 'Comments', domain ) }}</h2>
+	<h2 class="page-section-header">{{ __( 'Comments' ) }}</h2>
 	<div class="comments-block">
 		<div class="form-section">
 			{{ fn('comment_form', comments_args) }}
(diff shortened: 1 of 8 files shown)


templates/comments_section.twig
```diff
@@ -1,16 +1,13 @@
-<div class="container">
+<div class="container page-section">
+	<h2 class="page-section-header">{{ __( 'Comments', domain ) }}</h2>
```
sagarsdeshmukh: I guess ` __( 'Comments', domain )` - here we need to use `__( 'Comments', 'planet4-master-theme' )` ?
Please confirm @koyan @angtheod

koyan: Comments exists as a localised message in wordpress, so we don't need to retranslate it.

koyan: But I am not sure if we get that by using "domain" there. Have you confirmed that @kirdia ?

kirdia: No, I haven't checked if the text domain was working properly, but you are right. We do not need to retranslate common words like 'Comments', 'Comment' etc. I will remove the planet4 master theme text domain for those.

PR #133, opened 2017-12-15
Feature/1071 content submission mail notification

sagarsdeshmukh:


functions.php
```diff
@@ -373,6 +374,36 @@ public function p4_save_page_type( $post_id ) {
...
+		if( $emails ) {
+			$headers  = 'MIME-Version: 1.0' . "\r\n";
+			$headers .= 'Content-type: text/html; charset=iso-8859-1' . "\r\n";
+			$name     = $post->post_name;
+			$subject  = __( 'New post published', 'planet4-master-theme' );
+			$body     = sprintf( 'Hello Team, <br><br>' );
+			$body    .= sprintf( 'The post <strong> %s </strong> has been published. <br>', $name );
+			$body    .= sprintf( 'Post link : %s <br><br>', get_permalink( $post ) );
+			$body    .= sprintf( 'Thanks<br>' );
+			$body    .= sprintf( 'Greenpeace IT' );
```
angtheod: @sagarsdeshmukh Should these strings be translatable?

angtheod:
@koyan On Jira ticket description it mentions that web editors should be notified of new content based on their notification preferences. How do we apply that?

angtheod:
@sagarsdeshmukh I'd suggest that we start separating concerns included in the P4_Master_Site class.
It is getting bigger and bigger and it will be harder to maintain later on.
We could add separate functionality in a new P4_<name> class and add it as a service inside the array that the P4_Master_Site class takes as an argument. What do you think?

functions.php
sagarsdeshmukh: Yes it is but I assume its just an initial basic version of mail content and in future, we will change this mail template  @koyan : Do you have any suggestion for this ?

sagarsdeshmukh:
Hi @angtheod I do agree and I will update accordingly.

koyan:
> @koyan On Jira ticket description it mentions that web editors should be notified of new content 
> based on their notification preferences. How do we apply that?
Hmm, I missed that. Currently there are no notification settings for wp admin users. 
We should ask the product owner if she wants to extend this to creating notification settings, or if for now all "Editors" should receive notifications.

functions.php
27Bslash6: And should this test be in UTF instead of ISO8859?

PR #140, opened 2017-12-19
bugfix - copyright text with html was not inserting with html tags

sagarsdeshmukh:
Issue fixed in Admin >> planet4 settings, copyright content html tags were not saving properly.


angtheod:
@sagarsdeshmukh @koyan 
not sure what is the exact problem here, but in case we want to achieve having more fancy copyright text
then we could make it like the Cookies text

p4-settings

koyan:
Agree with @angtheod . 
@sagarsdeshmukh please remove the sanitize solution, and use a rich text field as we did in the cookies.

sagarsdeshmukh:
Yes I will do as per Angelos suggestion. The problem here was - previously we have added copyright text which is having a texts with href tag inside and we have assigned textbox for this. When we moved general additional settings to planet4 - I used cmb2 but noticed that its not inserting html tag with normal "text" type so this needs to be sanitized so it can accept the href tag.

I will changed it like cookie field.

PR #142, opened 2017-12-19
Latest markup for tag template.

kirdia:


templates/tag.twig
```diff
@@ -1,35 +1,26 @@
...
+				<img src="{{ tag_image }}" class="page-header-image">
+			</div>
+		{% endif %}
 		<div class="container">
-			<div class="page-header col-md-8 col-lg-8 col-sm-12 pl-0 pr-0">
-				<a href="{{ category_link|default('#') }}">{{ category_name|e('wp_kses_post')|raw }}</a>
-				<h1>#{{ tag_name|e('wp_kses_post')|raw }}</h1>
-				<p>{{ tag_description }}</p>
+			<div class="left">
+				<h1 class="page-header-title">{{ category_name|e('wp_kses_post')|raw }}</h1>
```
angtheod: @kirdia @koyan 
So, we don't want to have a link for the Issue anymore here? This had been design request.

kirdia: I did not find a link in the presentation layer and I removed the link from the previous version because it changes styling. But if it is required I can add it.

angtheod: Yes, was asked by Laura I think so that they can navigate back to the Issue. It should go in the presentation as well I believe. @mikeselander

koyan: @kirdia please add it, and we see how it behaves.

kirdia: Ok, I added as it was in the previous version.

PR #152, opened 2018-01-01
Feature/1574 control panel

angtheod:
Hello all,
after asked to implement a way for us to easily check redis server connection in the WP admin panel, I implemented a way to be able to do this inside the WP Dashboard which can also be used with an automated functionality test.
The requests are asynchronous (ajax). I have only 2 of those in the current PR to make code review easier, but I have 2 more ready to be added after PR passes.

Feedback on something like a control panel (see screenshot) is much appreciated.

p4-control-panel


classes/class-p4-control-panel.php
```diff
@@ -0,0 +1,175 @@
...
+		/**
+		 * Adds a check cache button to check connectivity to the Redis server.
+		 */
+		public function check_cache() {
+			if ( wp_doing_ajax() ) {
+				if ( ! current_user_can( 'manage_options' ) ) {
+					return;
+				}
+
+				if ( wp_verify_nonce( $_GET['_wpnonce'], 'cp-action' )  // CSRF protection.
```
27Bslash6: Is there any chance $_GET['_wpnonce'] would be unset, leading to a notice errror?

classes/class-p4-control-panel.php
```diff
@@ -0,0 +1,175 @@
...
+				     && isset( $_GET['cp-action'] )
+				     && 'check_cache' === $_GET['cp-action']
+				) {
+					$messages = [];
+					$info     = wp_redis_get_info();
+
+					if ( $info instanceof WP_Error ) {
+						if ( $info->errors['wp-redis'] ) {
+							foreach ( $info->errors['wp-redis'] as $index => $error ) {
```
27Bslash6: $index is unused here

assets/js/dashboard.js
```diff
@@ -0,0 +1,32 @@
...
+
+// Dashboard page.
+$(function() {
+	$('.btn-cp-action').off('click').on('click', function (event) {
+		event.preventDefault();
+		var $btn      = $(this);
+		var $response = $( '.cp-subitem-response', $btn.parent() );
+
+		$.ajax({
+			url: ajaxurl,
```
27Bslash6: `ajaxurl` is undefined here. Is it defined outside this event, should it be built in this function from the DOM, or passed as an event parameter? https://stackoverflow.com/questions/3273350/jquerys-click-pass-parameters-to-user-function#9467172

classes/class-p4-control-panel.php
```diff
@@ -0,0 +1,175 @@
...
+					}
+				}
+				wp_die();
+			}
+		}
+
+		/**
+		 * Adds a check cache button to check the ENS API.
+		 */
+		public function check_en() {
```
27Bslash6: I'd rename this to `check_engaging_networks()` to avoid any ambiguity, this reads to me like you're checking an english translation.

assets/js/dashboard.js
angtheod: Since WP 2.8 ajaxurl is a global js variable defined by WP itself and points to admin-ajax.php (this is true only for use in the WP admin side) https://codex.wordpress.org/AJAX_in_Plugins

classes/class-p4-control-panel.php
angtheod: This was removed entirely

assets/js/dashboard.js
27Bslash6: Oh I see, that's interesting, thanks.

PR #154, opened 2018-01-02
Made the header block work on homepage

koyan:


page.php
```diff
@@ -75,7 +75,7 @@ function add_body_classes_for_page( $classes ) {
 }
 
 $context['post']                = $post;
-$context['header_title']        = is_front_page() ? '' : ( $page_meta_data['p4_title'][0] ?? $post->title );
+$context['header_title']        = is_front_page() ? ( $page_meta_data['p4_title'][0] ?? '' ) : ( $page_meta_data['p4_title'][0] ?? $post->title );
```
angtheod: Looks like we want the p4_title to show up if it exists regardless if it is front page or not, so we could as well do this in order to avoid unnecessary is_front_page() calls when p4_title exists.

`$page_meta_data['p4_title'][0] ?? is_front_page() ? '' : $post->title;`

koyan: Fixed, please recheck

PR #155, opened 2018-01-02
Functionality for adding DONATE button url field in planet4 settings

sagarsdeshmukh:


functions.php
```diff
@@ -162,6 +162,7 @@ public function add_to_context( $context ) {
 		$context['navbar_menu']  = new TimberMenu( 'navigation-bar-menu' );
 		$context['site']         = $this;
 		$context['sort_options'] = $this->sort_options;
+		$context['donatelink']   = planet4_get_option( 'donate_button' ) ?  planet4_get_option( 'donate_button' ) : '#';
```
angtheod: @sagarsdeshmukh 
I think is better to use the null coalescing operator to avoid double function call. Also, i think its better to use the default WP get_option function to avoid coupling the Theme with CMB2 plugin where it is not needed.

`$context['donatelink']   = get_option( 'donate_button' ) ?? '#';`

PR #156, opened 2018-01-02
Feature/split main js

kirdia:
Splitted search js code to different file.
Changed load more button js for content four column and covers blocks.
Added slick slider css style.

### DIFF
--- a/search.php
+++ b/search.php
@@ -12,7 +12,6 @@
 /**
  * Planet4 - Search functionality.
  */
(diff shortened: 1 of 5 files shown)


search.php
```diff
@@ -12,7 +12,7 @@
 /**
  * Planet4 - Search functionality.
  */
-
+wp_enqueue_script( 'search', get_template_directory_uri() . '/assets/js/search.js', [], '0.1.0', true );
```
angtheod: @kirdia could you place this inside the if statement and at the end of it?

kirdia: Done :+1:

PR #167, opened 2018-01-05
1575 search fixes

angtheod:
Show up available filter options even there are no results found. Made sure the filter items are collapsed or not properly. Also, show up the sorting options in such a case. Removed the Uncategorized category from appearing in filter options. Renamed better some of the class properties. Fixed small issue from last commit.


angtheod:
Might look like it has many changes, but it does not really. Has many indentation changes which makes it look like that.

PR #168, opened 2018-01-05
Issue 114: Remove the Archive choice for launch. Will be enabled again later. Se…

angtheod:
…t default value to constructor arg.


angtheod:
Issue 114

PR #185, opened 2018-01-10
Evergreen page update - Spreadsheet issue 166

angtheod:
Update Evergreen page template to include Issue/Campaigns on top as breadcrumb navigation.

### DIFF
--- a/page-templates/evergreen.php
+++ b/page-templates/evergreen.php
@@ -28,7 +28,6 @@ use Timber\Post as TimberPost;
  */
 function add_body_classes_for_evergreen_page( $classes ) {
 	$classes[] = 'white-bg';
-
 	return $classes;
 }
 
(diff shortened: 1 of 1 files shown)


angtheod:
Also, fixed wrong arguments for function call.

page-templates/evergreen.php
```diff
@@ -33,15 +34,46 @@ function add_body_classes_for_evergreen_page( $classes ) {
...
-$context                        = Timber::get_context();
-$post                           = new TimberPost();
-$page_meta_data                 = get_post_meta( $post->ID );
+$context        = Timber::get_context();
+$post           = new TimberPost();
+$page_meta_data = get_post_meta( $post->ID );
+
+// Retrieve P4 settings in order to check that we add only categories that are children of the Issues category.
+$options  = get_option( 'planet4_options' );
+$category = get_the_category( $post->ID )[0];
```
kirdia: Just a small detail. I think that this would return an empty array if the post does not belong to a category, resulting in a php warning.

angtheod:
@kirdia yep, you are right fixed that.

PR #188, opened 2018-01-15
Added srcset to the header image for pages

koyan:


templates/blocks/header.twig
```diff
@@ -3,7 +3,9 @@
 	<div class="page-header">
 		{% if ( background_image ) %}
 			<div class="page-header-background">
-				<img src="{{ background_image }}" class="page-header-image" />
+				<img src="{{ background_image }}"
+					 srcset="{{ background_image_srcset }}"
```
kirdia: Please add a check here if background_image_srcset is set, in case wp_get_attachment_image_srcset returns `false`.

koyan: Added it.

PR #201, opened 2018-01-19
Feature/1574 control panel

angtheod:
Add 2 more actions to the Control Panel. 1) Flush cache, for flushing the Redis object cache and 2) Check search indexer, for checking if the Search plugin Indexer is stalled or awake. The Control Panel will only be visible to Administrators and the ajax calls will respond only to administrators.


angtheod:
Pushed 1 more commit in order to add more comments and make review easier.

PR #204, opened 2018-01-23
Post template changes for take action boxout block

kirdia:
Removed paragraph splitting in single post template. 
Updated markup in single post template.
Changes made regarding the below PR
[PR presentation layer](https://github.com/greenpeace/planet4-presentation-layer/pull/204)

### DIFF
--- a/templates/single.twig
+++ b/templates/single.twig
@@ -36,29 +36,11 @@
 		<!-- Post Block Start -->
 		<div class="container">
 			<div class="post-content">
+				{% if ( post.take_action_boxout ) %}
+					{{ post.take_action_boxout|shortcodes|raw }}
+				{% endif %}
 				<div class="row">
-					<div class="col-lg-6 post-content-lead">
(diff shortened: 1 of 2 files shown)


templates/single.twig
```diff
@@ -36,29 +36,11 @@
 		<!-- Post Block Start -->
 		<div class="container">
 			<div class="post-content">
+				{% if ( post.take_action_boxout ) %}
+					{{ post.take_action_boxout|shortcodes|raw }}
```
angtheod: Do we need escaping here together with the |raw filter? Or is it already escaped?

kirdia: No its is not escaped. I think we will need escaping here. 
Good catch, I will update the code.

kirdia: Actually escaping here will remove style attribute from top level element in the take action action boxout block (no background image).
I will add escaping in the take action boxout block itself, so I think we are good to go with this.

angtheod: ok

PR #205, opened 2018-01-23
PLANET-1653 Functionality for articles behaviour on post page

sagarsdeshmukh:


single.php
```diff
@@ -56,7 +56,7 @@ function add_body_classes_for_post( $classes ) {
 
 
 // Build the shortcode for articles block.
-if ( ! empty( $articles_title ) ) {
+if ( ! empty( $articles_title ) && $post->include_articles == 'on' ) {
```
angtheod: Please use YODA condition (constant on the left side) as it is the recommended standard by wp and also type checking.
`'on' === $post->include_articles`

PR #210, opened 2018-01-25
PLANET-1660 current post should not display in post article block

sagarsdeshmukh:


single.php
```diff
@@ -57,7 +57,7 @@ function add_body_classes_for_post( $classes ) {
 
 // Build the shortcode for articles block.
 if ( ! empty( $articles_title ) &&  'on' === $post->include_articles ) {
-	$post->articles = "[shortcake_articles article_heading='$articles_title' article_count='$articles_count' /]";
+	$post->articles = "[shortcake_articles post_id='".$post->ID."' article_heading='$articles_title' article_count='$articles_count' /]";
```
angtheod: I think it 'd be better to rename 'post_id' to 'exclude_post_id' so that it is easier to understand what it does at first glance

PR #216, opened 2018-01-26
Corrected the query to pick up correct pages

koyan:
PLANET-1721

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -626,7 +626,8 @@ class P4_Master_Site extends TimberSite {
 			$take_action_pages_args = [
 				'post_type'   => 'page',
 				'post_parent' => $parent_act_id,
-				'numberposts' => -1,
+				'post_status' => 'publish',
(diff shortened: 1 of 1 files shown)


functions.php
```diff
@@ -626,10 +626,10 @@ public function populate_act_page_children_options() {
 			$take_action_pages_args = [
 				'post_type'   => 'page',
 				'post_parent' => $parent_act_id,
+				'numberposts' => -1,
```
angtheod: Generally, using numberposts (or posts_per_page) with -1 is not good idea for production and it is recommended to be avoided by WP Coding Standards. Could we set this to a high enough value instead?

koyan: This is for the backend, and if the editors cannot find an article in that drop down menu, then we have a problem. I have added a filter to include only published articles in the list. And we don't expect multiples of active actions in any NRO.

PR #217, opened 2018-01-27
1649 search optimize

angtheod:
### Search page optimization consists of the following 3 parts (so far):

**1. Implement Lazy-loading (asynchronous) of search results.*****

On page load we find ALL the results but only the first POSTS_PER_LOAD results are sent to be displayed. We use ALL the results in order to create dynamic Filters and calculate result counters for each on of those. Then we load more POSTS_PER_LOAD results asynchronously as the user scrolls down or clicks on the Load More button.
*** Currently this part works only when searching for everything

**2. Cache results in Redis Object Cache.**

When doing search first we look in the object cache for a key that matches our search criteria. If we find it then we load results from the object cache, else we query the primary database and then we add the relevant key/value to the object cache. Each key lives there for DEFAULT_CACHE_TTL (TTL = Time To Live) seconds.
While user scrolls down we get asynchronously from the object cache the next POSTS_PER_LOAD results and add them to the DOM.

A search page key is stored in the following format
`<group>:<term>:<query_string>`
For example if we search for 'action' and then filter by Campaign 'FixFood'
`search:action:s=action&orderby=relevant`
`search:action:s=action&orderby=relevant&f[tag][FixFood]=110`

If a user idles enough time for the object cache key to expire before he/she decides to continue scrolling again then we will get the results again from the primary database and re-add the key to the object cache.

**3. Retrieve just the thumbnail and not the whole featured image of each result.**

I thought that `post.thumbnail.src` did that but it needed to be `post.thumbnail.src('thumbnail')`, but Timber's documentation was not very clear on that one. This was noticed by @kirdia and fixed in #206 

**Notes**
1. Part 1 still needs to be applied for when searching for specific term.
2. There are still few more TODOs to implement to optimize/separate code better.
3. Added the P4_Search class as a service in P4_Master_Site so that it will initialize early enough to catch the necessary 'wp_ajax_' and 'wp_ajax_nopriv_' hooks.
4. The search.js file is not really changed as much as it seems. Changed identation to tabs which makes it look like that in Github. 
5. _Apologies for the big PR but there was no way to split it better without breaking functionality._


assets/js/search.js
```diff
@@ -2,91 +2,130 @@ $ = jQuery;
...
+				type: 'GET',
+				data: {
+					action:          'get_paged_posts',
+					'search-action': 'get_paged_posts',
+					'paged':         next_page,
+					'query-string':  decodeURIComponent( location.search ).substr( 1 )		// Ignore the ? in the search url (first char).
+				},
+				dataType: 'html'
+			})
+			.done(function ( response ) {
```
kirdia: Done callback will run for all cases (success, error). Maybe it is better to separate the actions taken in each case (successful response or error occured during the request ) using success and error callback functions.
```javascript
$.ajax({

	success: function (response) {
	},
	error: function (jqXHR, textStatus, errorThrown) {
	}
}).always(function () {

});
```

assets/js/search.js
```diff
@@ -2,91 +2,130 @@ $ = jQuery;
...
+				// Append the response at the bottom of the results and then show it.
+				$( '.multiple-search-result .list-unstyled' ).append( response );
+				$( '.row-hidden:last' ).removeClass( 'row-hidden' ).show( 'fast' );
+			});
+		} else {
+			var $row = $( '.row-hidden', $load_more_button.closest( '.container' ) );
+
+			if ( 1 === $row.size() ) {
```
kirdia: `size()` function is deprecated as of jquery 1.8 as @comzeradd pointed me in another place. We can use `length` property

classes/class-p4-search.php
```diff
@@ -60,28 +83,104 @@ public function __construct( $search_query, $selected_sort = self::DEFAULT_SORT,
...
 				}
-				$this->posts = $this->get_timber_posts();
+
+				// Set the decoded url query string as key.
+				$query_string = urldecode( filter_input( INPUT_SERVER, 'QUERY_STRING', FILTER_SANITIZE_STRING ) );
+				$group        = 'search';
+				$subgroup     = $this->search_query ? $this->search_query : 'all';
+
+				// Check Object cache for stored key.
+				$this->posts = wp_cache_get( $query_string, "$group:$subgroup" );
```
kirdia: Just a suggestion because I saw this block of code used twice.
You can separate the querying of cache and the adding of key to cache to a separate function.

```php
					$this->posts = wp_cache_get( $query_string, "$group:$subgroup" );
					// TODO - Set the correct filters so that it will work when searching for specific term with filters applied.
					// If cache key expired then retrieve results once again and re-cache them.
					if ( false === $this->posts ) {
						$this->posts = $this->get_timber_posts();
						if ( $this->posts ) {
							wp_cache_add( $query_string, $this->posts, "$group:$subgroup", self::DEFAULT_CACHE_TTL );
						}
					}
```

search.php
```diff
@@ -14,6 +14,8 @@
  */
 if ( is_main_query() && is_search() ) {
 	if ( 'GET' === filter_input( INPUT_SERVER, 'REQUEST_METHOD' ) ) {
+		global $p4;
```
kirdia: I suppose the introduction of global variable `$p4` has to do with the hooks related to ajax requests.
I am not sure if we can avoid it but can we add the hooks to master theme like below? Will this work for our case?
`add_action( 'wp_ajax_get_paged_posts',        array( 'P4_Search', 'get_paged_posts' ) );`

assets/js/search.js
angtheod: Nop, 'done()' executes on 'success()' and 'always()' executes always.
The new recommended way is to chain call done(), fail(), always() (so I guess I should add .fail() and .always() calls).
Moreover, these methods also implement the Promise interface, giving them all the properties, methods, and behavior of a Promise giving more flexibility in the future.

search.php
angtheod: Yes, it is solely related to the ajax requests.
I struggled with this quite a bit. Did not want to use a global either... cause globals are baaaad mkaaaay.... although WP itself uses them quite extensively to make development easier.

I had already tried your suggestion but this did not work in our case, because the get_paged_posts() method can not be static as it needs to access the P4_Search object itself and set/get some of its properties/methods and at that point the already created P4_Search object is not accessible. Thing is I wanted to try it a bit more but somebody whose name begins with K would bite my neck if I delayed more :-)

Your comment though allowed me try it out again, but this time create a new P4_Search object in each ajax request and separate functionality depending whether we are in ajax or normal request. Just tried this and seems to work.
It has a disadvantage of creating a new -local- P4_Search instance (not that big of a disadvantage, we can live with it) but at least we get rid of the global $p4 which is more important.
@kirdia @koyan

angtheod:
@kirdia Pushed changes, please check again

PR #224, opened 2018-01-29
Disallow creation of tags to all roles besides administrator

kirdia:
Add functions to disallow creation of tags to all roles besides administrator.

- Remove default wordpress selection box for all roles besides administrator.
- Add checkboxes list for selecting post's / page's tags for all roles besides administrator.
- Add hook to prevent creation of tags for all roles besides administrator.

[Issue 1690](https://jira.greenpeace.org/browse/PLANET-1690)


koyan:
I tested it also without this and could not find a better way to enfore this at this stage

PR #229, opened 2018-01-30
Dynamically create version of css and js for handling cached files

koyan:
Fingerprinting the css and js files according to: https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching#invalidating_and_updating_cached_responses


comzeradd:
Nice! This should fix all of our version issues.

PR #254, opened 2018-02-08
[PLANET-1501] Add new share buttons to Posts and Take Action pages

comzeradd:


assets/scss/common/_share-buttons.scss
```diff
@@ -0,0 +1,60 @@
...
+
+    &:hover {
+      color: $grey-10;
+    }
+
+    &:active {
+      position: relative;
+      top: 2px;
+      box-shadow: none;
+      color: #e2e2e2;
```
ashish173: @comzeradd New colour addition here? variable? or review with designers?

PR #257, opened 2018-02-09
Bug/1594

ashish173:
Using `hammer.js` for swipe and mobile gestures. @comzeradd thanks for the suggestion. 

Finally, after a lot of procrastination pushing code to master theme :D

### DIFF
--- a/assets/js/custom.js
+++ b/assets/js/custom.js
@@ -412,5 +412,18 @@ $(function() {
 	});
 });
 
+$(function() {
+  /* Carousel header swipe on mobile */
+  var carousel_element = document.getElementsByClassName('carousel-header')[0];
+  var carousel_head_hammer = new Hammer(carousel_element, { recognizers: [] });
+  var hammer = new Hammer.Manager(carousel_head_hammer.element);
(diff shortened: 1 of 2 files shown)


assets/js/custom.js
```diff
@@ -412,5 +412,18 @@ $(function() {
 	});
 });
 
+$(function() {
+  /* Carousel header swipe on mobile */
+  var carousel_element = document.getElementsByClassName('carousel-header')[0];
```
angtheod: @ashish173 @comzeradd This line causes js error
`Uncaught TypeError: Cannot read property 'addEventListener' of undefined`
in search page breaking the rest of the js functionality in the page
https://dev.p4.greenpeace.org/international/?s=&orderby=relevant

comzeradd: @ashish173 you can probably use jquery to do the selection here and then check `carousel_header.length` to run the rest of the code.

ashish173: Thanks @angtheod you are the real bug catcher; @comzeradd thanks for the suggestion #262 is the fix for it.

PR #260, opened 2018-02-09
[PLANET-1752] Use only published posts/pages for searching in all cases.

kirdia:
Issue [1752](https://jira.greenpeace.org/browse/PLANET-1752)


classes/class-p4-search.php
```diff
@@ -223,7 +223,7 @@ protected function get_posts( $all = false, $paged = 1 ) : array {
 				'posts_per_page' => self::POSTS_LIMIT,          // Set a high maximum because -1 will get ALL posts and this can be very intensive in production.
 				'no_found_rows'  => true,                       // This means that the result counters of each filter might not be 100% precise.
 				'post_type'      => 'any',
-				'post_status'    => 'any',
+				'post_status'    => 'publish',
```
angtheod: @kirdia ok, so for this we need to include also 'inherit' post_status in order to include attachments (attachments have 'inherit' as post_status). So, we need an array with both.

PR #263, opened 2018-02-12
[PLANET-1752] Use only published posts/pages for searching in all cases 2

kirdia:
Added also inherit status to include documents in search results.
Issue 1752


classes/class-p4-search.php
```diff
@@ -223,7 +223,7 @@ protected function get_posts( $all = false, $paged = 1 ) : array {
 				'posts_per_page' => self::POSTS_LIMIT,          // Set a high maximum because -1 will get ALL posts and this can be very intensive in production.
 				'no_found_rows'  => true,                       // This means that the result counters of each filter might not be 100% precise.
 				'post_type'      => 'any',
-				'post_status'    => 'publish',
+				'post_status'    => ['publish', 'inherit'],
```
angtheod: @kirdia please leave spaces after array opener and before array closer

PR #264, opened 2018-02-12
PLANET-1751 Added search filter params in Ajax call

sagarsdeshmukh:


classes/class-p4-search.php
```diff
@@ -139,6 +139,29 @@ public function get_paged_posts() {
...
+					        }
+					    }
+					}
+
+					// Validate user input (sort, filters, etc).
+					if ( $search_async->validate( $selected_sort, $filters, $search_async->context ) ) {
+					    $search_async->selected_sort = $selected_sort;
+					    $search_async->filters       = $filters;
+					}
+
```
angtheod: @sagarsdeshmukh I think that this whole part where we set the filters/sort during ajax call, should be placed inside the 'else' statement after the check_cache() call. This should happen only if the results are not already in the cache and we need to search the primary database.

Also, one more thing that needs correction is the issue that we spotted on the Redis db of the staging environment where we have double keys of the same query-string due to the query string params being concatenated in different order. This probably happens because we are using 2 different ways of getting the query string params depending on whether we are on normal or async request.

Check lines 
https://github.com/greenpeace/planet4-master-theme/blob/d8794bf91afe0ae5f67fa0c3dd3713eba0e87de0/classes/class-p4-search.php#L93
https://github.com/greenpeace/planet4-master-theme/blob/d8794bf91afe0ae5f67fa0c3dd3713eba0e87de0/classes/class-p4-search.php#L137

sagarsdeshmukh: @angtheod As in `check_cache` function, it is already searching in DB if no cache found, so we can not move the set filter code block in else part after check cache function. and yes in this case else part is not in use.

I am looking into the second point (double keys of the same query-string).

sagarsdeshmukh: @angtheod I tried to reproduce the staging server Redis cache (double keys of the same query-string due to the query string params being concatenated in a different order) issue on my local, but it's not reproduced. somehow it works for me, After discussion with you (Angelos) I checked query string param in both functions load() and get_paged_posts() but here also I got same results.

Can you please confirm on staging, if the same issue still persists?

angtheod: @sagarsdeshmukh I can not connect to staging server redis for some reason.
I think we do not really need to reproduce it locally.
We could just change the way we the $query_string inside get_paged_posts() method, in order to match the way we get it inside load() method.

Try this way to get the query_string inside the get_paged_posts() method:
`$query_string_full = urldecode( filter_input( INPUT_SERVER, 'QUERY_STRING', FILTER_SANITIZE_STRING ) );
$query_string      = str_replace( '&query-string=', '', strstr( $query_string_full, '&query-string=' ) );`

Issue #129, opened 2017-12-13
Search feature


tolen:
I would like to see _all_ of the Filtering Options displayed at all times.  

When the XX Results are displayed and a Filter option has (0) results, the filter should still be presented. 

![screen shot 2017-12-13 at 19](https://user-images.githubusercontent.com/15688756/33955768-fd2632da-e03c-11e7-9b51-b2c631e3dc5b.png)

------------
Also, on the [Zero Results page](http://dev.p4.greenpeace.org/?s=&orderby=relevant), I would like to see Advanced Search/Filtering options. I will have the Design team mock this up. 

And... there should be default icons displayed when a search result has no Featured Image assigned to it (e.g. Press Releases or Documents). Will have the Design team mock this up too. 

@viniromualdo

PR #270, opened 2018-02-12
Feature/1600 search show all filters

angtheod:
Make sure all filters are displayed even if they have 0 results. Filters with 0 results are disabled. Separated concerns better. Set_filter_context creates the filters and set_results_context sets the results and the filter counters. Because of this change the $all property became obsolete.

This PR takes care the rest of #129 

@27Bslash6 
As a side effect of this PR, the PHP Notices (concerning undefined indexes) that were being generated in the Search page are now all fixed.

### DIFF
--- a/classes/class-p4-search.php
+++ b/classes/class-p4-search.php
@@ -29,8 +29,6 @@ if ( ! class_exists( 'P4_Search' ) ) {
 		protected $search_query;
 		/** @var array|bool|null $posts */
 		protected $posts;
-		/** @var array|bool|null $all_posts */
-		protected $all_posts;
 		/** @var array|bool|null $paged_posts */
 		protected $paged_posts;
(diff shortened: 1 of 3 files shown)


classes/class-p4-search.php
```diff
@@ -399,39 +491,37 @@ protected function set_results_context( &$context ) {
 						if ( 'act' === basename( get_permalink( $post->post_parent ) ) ) {
 							$content_type_text = __( 'ACTION', 'planet4-master-theme' );
 							$content_type      = 'action';
-							$context['posts_data']['found_actions']++;
+							$context['content_types']['0']['results'] ++;
```
kirdia: Just a minor thing.. Please remove the extra space before post-increment operator.
I don't know why, but Phpstorm adds them if you use wordpress coding style.

PR #275, opened 2018-02-13
1648 handle long document names

angtheod:
Handle long Document names to avoid misalignment in search results.

### DIFF
--- a/templates/tease-search.twig
+++ b/templates/tease-search.twig
@@ -44,7 +44,8 @@
 			</div>
 
 			{% if ( is_document ) %}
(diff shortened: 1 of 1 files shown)


templates/tease-search.twig
```diff
@@ -44,7 +44,8 @@
 			</div>
 
 			{% if ( is_document ) %}
-				<a href="{{ post.guid }}" class="search-result-item-headline">{{ post.title|e('wp_kses_post')|raw }}</a>
+				{% set title = post.title|e('wp_kses_post')|raw %}
+				<a href="{{ post.guid }}" class="search-result-item-headline">{{ title|length > 30 ? title|slice(0, 30) ~ '...' : title }}</a>
```
kirdia: This is a bit tricky I think.
If the title contains html, then substringing may result to invalid html (open tags etc..)
`<b>Long Title Long Title Long Title...`

Maybe we can strip tags if the the title is more than 30 characters.

PR #278, opened 2018-02-14
Bug fix PLANET-1740

ashish173:
Centering the campaign image for mobile screen size instead of showing
    just the top.

### DIFF
--- a/assets/scss/common/_campaign-thumbnail-block.scss
+++ b/assets/scss/common/_campaign-thumbnail-block.scss
@@ -78,6 +78,11 @@
         width: 100%;
         max-height: 260px;
         height: auto;
+        margin: -50px 0 -10px 0;
+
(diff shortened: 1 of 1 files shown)


assets/scss/common/_campaign-thumbnail-block.scss
```diff
@@ -78,6 +78,10 @@
         width: 100%;
         max-height: 260px;
         height: auto;
+
+        @media (max-width: #{$medium-width}) {
+          margin: -50px 0 -10px 0;
```
comzeradd: Why not use our responsive mixins? You can add these margins to `img` and use `@include medium-and-up` to reset them.

ashish173: @comzeradd Thanks for the solution.

PR #305, opened 2018-03-01
[PLANET-1877] Added hook for post_link_category

kirdia:
If category is used in permalink structure, wordpress uses the first category (smaller term id) for the permalink construction.
Override this functionality if a post belongs to a category that it is a planet4 page type term.
[Task 1877](https://jira.greenpeace.org/browse/PLANET-1877)


koyan:
Sorry, Kyriakos, this is not what we want. We don't want to continue using the categories at all. 
What we need is to create a new permalink %p4_page_type% so that we can drop completely the categories<->page_type mapping.

kirdia:
Yes, my mistake.
I added it as a temporary solution, because I thought %p4_page_type% solution, would break pages' permalinks, but I was wrong.
I found out that WordPress page permalinks permalinks are not changeable by default.

PR #312, opened 2018-03-02
PLANET4-1874 404 page

angtheod:
Check if we are on child theme or not and use correct background image.

Related to https://github.com/greenpeace/planet4-child-theme-greece/pull/2

### DIFF
--- a/404.php
+++ b/404.php
@@ -25,7 +25,9 @@ add_filter( 'body_class', 'add_body_classes' );
 
 $context = Timber::get_context();
 
-$context['page_notfound_image']       = esc_url( get_template_directory_uri() . '/images/404-header.jpg' );
+$theme_uri = is_child_theme() ? get_stylesheet_directory_uri() : get_template_directory_uri();
+
+$context['page_notfound_image']       = esc_url( $theme_uri . '/images/404-header.jpg' );
 $context['page_notfound_title']       = __( 'Sorry, we can\'t find that page!', 'planet4-master-theme' );
 $context['page_notfound_subheader']   = __( 'The page might be extinct...', 'planet4-master-theme' );
 $context['page_notfound_description'] = __( 'Use the search tool and try your luck again.', 'planet4-master-theme' );


404.php
```diff
@@ -25,7 +25,9 @@ function add_body_classes( $classes ) {
 
 $context = Timber::get_context();
 
-$context['page_notfound_image']       = esc_url( get_template_directory_uri() . '/images/404-header.jpg' );
+$theme_uri = is_child_theme() ? get_stylesheet_directory_uri() : get_template_directory_uri();
```
koyan: Why do you need to check if this code is in the child theme, since you are running it from inside the child theme?

koyan: Also, will get_template_directory_uri return the child theme directory, or the parent theme?

angtheod: This code is inside the master theme and it will run unless there is a 404.php file inside the child theme.
So, if a child theme is active (e.g. Greek site) then $theme_uri will get the value of the child theme uri (get_stylesheet_directory_uri)
else if master theme is active (e.g. International site) it will get the value of the parent theme uri (get_template_directory_uri)

koyan: > This code is inside the master theme
I don't understand the above phrase. You have committed this code in this specific child theme.

Additionally: no site has the master theme active. All sites have their child theme active. 
So, this code makes no sense. If you run this code in the child theme, this code is getting run only if this specific child theme is active, in which case the whole question is irrelevant.

koyan:
Lets not create more work than needed. Coordinate with @comzeradd to make the background image to be controlled by css instead of html, and then the stylesheet can overide it with css.

404.php
angtheod: @koyan 
Actually, you were right on this one, there was no need to check if a child theme is active since always the child theme is active.
It would work simply by changing get_template_directory_uri() to get_stylesheet_directory_uri() in order to get the child theme uri and then add the images/404-header.jpg in the the planet4-child-theme

PR #321, opened 2018-03-08
PLANET-1898 Fix carousel image sliding issue on POST

sagarsdeshmukh:


kirdia:
@sagarsdeshmukh please check which attributes for each of these tags are already allowed, because I think assigning a new array will override the already allowed attributes of the tag (to prevent breaking anything that needs these attributes).

PR #323, opened 2018-03-08
Move all frontend libraries inside the repo

comzeradd:
This will (hopefully) speed things up a bit, since now we do http requests to 4-5 different CDNs. It will also help local development in offline environment.

This also add a simple gulp script for handling any new future frontend assets. I updated the README docs to reflect that. I also checked-in `yarn.lock` to the repo. It's a good practice to have it here, since it contains hashes for the libraries used.


27Bslash6:
Just curious if there are there any tests for improved time to render with this change? Since it's highly likely that several of these libraries will be cached at ISP or even user when they're delivered from the original CDNs, particularly in the case of bootstrap, it might be regressive in terms of performance to deliver them from our own CDN as well as incurring additional Akamai bandwidth fees.

comzeradd:
No, it's not easy to produce such metrics in a local environment. We can test things on dev more reliably. But it's a common best practice (also suggested by many speed test optimization suites) to minimize requests and especially external ones. My suggestion actually would be to go one step further after this is merged and have all js and css bundled in one minimized and compressed file.

Having said, if we want to avoid putting all these things to Akamai and we prefer 3rd party CDNs, we should examine a solution that hosts all (or at least most) of our libraries. It is unlikely that ISPs have local cache for jsdelivr, jquery or even bootstrap's CDN. Cloudflare is usually the only option if we want to achieve that.

27Bslash6:
While it used to be advisable to combine all resources when delivering over HTTP, we're now delivering via solely over H2, so combining all resources is now an anti-pattern as multiple resources are delivered via the same connection, actively discouraging concatenation.

You can also get metrics on time to DOM render via the dev tools in your browser, and compare that previous versions, as well as to the metrics given by NewRelic for end-user at https://rpm.newrelic.com/accounts/301342/browser/40430909#/lite?percentile=50&_k=abcy0l 

Further, you could employ the planet4-docker-compose repository, feed it a new NEWRELIC_APPLICATION_NAME and license key then run tests with dom-parsing load tester, thus building statistically relevant data.

I also find it quite unlikely that, given the extremely long TTL on CDN content and the prevalence of Bootstrap in modern development, we have nothing to gain from using the official CDNs. Do you have data for that?  Even a slight cost or time-to-render reduction is something and would seem to outweigh any benefit obtained by this commit.

Finally, I'm curious as to your statement that reducing external requests is beneficial?  Browsers have (had on HTTP) limits on the number of concurrent requests to the same domain, but that is on a per-domain basis.  Indeed, a common technique for improving performance when hosting content locally is to shard your local domains so browsers send multiple requests.

See: https://stackoverflow.com/questions/26192897/should-i-use-bootstrap-from-cdn-or-make-a-copy-on-my-server

comzeradd:
I'm not a neywork expert, but I think that in http/2 requests happen through the same TCP connection only when they happen towards the same origin. So having resources in multiple domains we actually lose that http/2 feature. While having everything under our domain means that the browsers will open only one TCP connection for all css/js files and request them in parallel using multiplexing.

Regarding Bootstrap's official CDN, I don't have any metrics to be honest. But this is just one of our assets. If everything was coming from that CDN then maybe the high TTL value could indeed be beneficial. But at the moment we can't do that.

I do use docker-compose locally, but the dev edition, where http/2 is not enabled since we don't use https. So I don't considered this very reliable, since it's running from localhost and you can see a lot of queuing due to tcp connections limits, but for what it's worth here is a screenshot below from the Network tab. The first example is using local assets and the second load everything from the CDNs we currently use.

Do we use http/2 on our dev instance? We could easily revert this if we identify that it's more slow.



network

27Bslash6:
Yes, H2 requests are parallelised to the same domain, but that doesn't stop the browser from also requesting from other domains at the same time. They are not mutually exclusive.

Yes, all content is delivered over H2 on dev. See:
image

Of course you're seeing queues when using http, those graphs are not relevant to production.

comzeradd:
One additional note, Akamai tends to [peer directly](https://www.peeringdb.com/net/2) with Country level Exchange point and ISPs. Similar to what Cloudflare does. None of the CDNs we currently use applies such practices.

ashish173:
A lot of discussions has already gone into it. Assuming the paths are correct.

27Bslash6:
@comzeradd all the CDNs we're using are listed on that peering site.  MaxCDN, jsdeliver (Cloudflare), Google.  So far you haven't listed a single benefit that is true.

comzeradd:
That's a bit harsh. We already have 10 requests less by this change, but I'm not the right person to judge if the benefits overcome the cost.

I thought we didn't want to use Cloudflare (for non-technical reasons). If that's not the case anymore we can easily use cdnjs, which is the largest CDN and hosts all of our current frontend libraries. That will cost us less and use just one 3rd party request for everything.

PR #324, opened 2018-03-08
PLANET-1888 open graph markup

angtheod:
Add Facebook Open Graph and Twitter meta info for better sharing of Posts/Pages
(In the 1st commit had added info also for Campaign pages but removed it as it is not in the requirements).

### DIFF
--- a/single.php
+++ b/single.php
@@ -38,6 +38,7 @@ $page_meta_data              = get_post_meta( $post->ID );
 $page_terms_data             = get_the_terms( $post, 'p4-page-type' );
 $context['author_override']  = $page_meta_data['p4_author_override'][0] ?? '';
 $context['background_image'] = $page_meta_data['p4_background_image_override'][0] ?? '';
+$context['post_image_id']    = $page_meta_data['p4_background_image_override_id'][0] ?? $page_meta_data['_thumbnail_id'][0];
 $take_action_page            = $page_meta_data['p4_take_action_page'][0] ?? '';
 $context['page_type']        = $page_terms_data[0]->name ?? '';
 $context['page_term_id']     = $page_terms_data[0]->term_id ?? '';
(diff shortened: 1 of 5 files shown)


templates/html-header.twig
```diff
@@ -5,6 +5,28 @@
 	<title>{% if wp_title %}{{ wp_title|e('esc_html')|raw }} - {{ site.name }}{% else %}{{ site.name }}{% endif %}</title>
 	<meta name="description" content="{{ site.description }}">
 
+	{% if post %}
+		{% set description   = post.post_excerpt %}
+		{% set attachment_id = post.background_image_id ?? fn('get_post_thumbnail_id', post) %}
```
sagarsdeshmukh: @angtheod Just a question - can we move `post.background_image_id ?? fn('get_post_thumbnail_id', post) %}`  this logic to `page.php`?

sagarsdeshmukh:
Please confirm if it works on home page for fields `og:title` , `og:image` , `og:image:width`,  `og:image:height`, `twitter:title`  and `twitter:image`.

templates/html-header.twig
angtheod: @sagarsdeshmukh Why should we do this?

sagarsdeshmukh: I think we can keep the conditional part in page.php

angtheod:
@sagarsdeshmukh Which image should be chosen for home page? There is no instruction/requirement for this as far as I know.

sagarsdeshmukh:
While checking on home page, I got above fields blank, but honestly saying I am also not sure which image should appear here :)

angtheod:
@sagarsdeshmukh Updated

single.php
```diff
@@ -38,6 +38,7 @@ function add_body_classes_for_post( $classes ) {
 $page_terms_data             = get_the_terms( $post, 'p4-page-type' );
 $context['author_override']  = $page_meta_data['p4_author_override'][0] ?? '';
 $context['background_image'] = $page_meta_data['p4_background_image_override'][0] ?? '';
+$context['post_image_id']    = $page_meta_data['p4_background_image_override_id'][0] ?? $page_meta_data['_thumbnail_id'][0];
```
kirdia: I'm not sure, was this meant to be p4_background_image_override instead of p4_background_image_override_id?

angtheod: nop, I wanted the id

PR #325, opened 2018-03-09
PLANET-1879 Add p4 page type as an option in permalink structure

kirdia:
It seems like a big PR, because phpunit tests (and related files) were added to the repo.
The changes that you should focus on are in functions.php

**Changes**
Adds p4 page type as a permalink structure in Settings -> Permalinks.
If p4 page type permalink structure is used, then rewrite rules using p4 page type terms are generated.
Also contains some hooks for p4 page type term creation/edit/deletion. When a p4 page type term is created/edited/deleted then rewrite rules are regenerated and flushed.

**Testing**
You can try testing this branch locally also.
There are some instructions [here](https://github.com/greenpeace/planet4-master-theme/blob/feature/1879-p4-page-type-permalinks/TESTING.md) about the phpunit tests


functions.php
```diff
@@ -959,6 +964,105 @@ private function render_partial( $path, $args = [] ) {
 		}
 		include( locate_template( $path . '.php' ) );
 	}
+
+	/**
+	 * Add p4_page_type structure in available permalink tags for Settings -> Permalinks page.
+	 * available_permalink_structure_tags filter.
+	 *
+	 * @param $tags
```
angtheod: missing param doc descriptions (same below)

functions.php
```diff
@@ -959,6 +964,105 @@ private function render_partial( $path, $args = [] ) {
...
+	 * Replace p4_page_type placeholder with the p4_page_type term for posts permalinks.
+	 * Filter for post_link.
+	 *
+	 * @param $permalink
+	 * @param $post_id
+	 * @param $leavename
+	 *
+	 * @return mixed
+	 */
+	public function p4_page_type_permalink( $permalink, $post_id, $leavename ) {
```
angtheod: Methods of a class that has prefixed name and its own namespace do not need to have prefixed names as well.

functions.php
```diff
@@ -959,6 +964,105 @@ private function render_partial( $path, $args = [] ) {
...
+	/**
+	 * Add p4_page_type structure in available permalink tags for Settings -> Permalinks page.
+	 * available_permalink_structure_tags filter.
+	 *
+	 * @param $tags
+	 *
+	 * @return mixed
+	 */
+	public function add_p4_page_type_permalink_structure( $tags ) {
+		$tags['p4_page_type'] = __( 'P4 page type (A p4 page type term.)' );
```
angtheod: Text domain as 2nd argument missing?

angtheod:
@kirdia I think that functions.php and the P4_Master_Site class in particular is getting too big and has too many concerns as time passes and we keep adding functionality there.
Some of the theme's functionality has been isolated to its own classes inside the classes directory.
I would suggest from now on we do this more often in order to reduce the size of P4_Master_Site class and its complexity and make it easier to debug/maintain.
Perhaps we could add this new parmalink functionality in its own class and then add it in the classes given to the P4_Master_Site constructor. What you think?

PS. There are ofcourse more cases where we could do that in order to reduce the side of this class which is slowly becoming mega-class

kirdia:
@angtheod Yes it is a great idea. I'll try to move most of custom taxonomy related functionality to a separate class.

classes/class-p4-custom-taxonomy.php
```diff
@@ -0,0 +1,233 @@
...
+		 *
+		 * @param WP_Post $post The WordPress that will be filtered/edited.
+		 */
+		public function create_taxonomy_metabox_markup( WP_Post $post ) {
+			$attached_type = get_the_terms( $post, self::TAXONOMY );
+			$current_type  = ( is_array( $attached_type ) ) ? $attached_type[0]->slug : - 1;
+			$all_types     = get_terms( self::TAXONOMY, [ 'hide_empty' => false ] );
+			wp_nonce_field( 'p4-save-page-type', 'p4-page-type-nonce' );
+			?>
+			<select name="<?php echo self::TAXONOMY; ?>" disabled>
```
angtheod: Could you escape the name attribute please. I know it is a constant that you output but still please 
`echo esc_attr( self::TAXONOMY );` just to stop CS from complaining about standards.

classes/class-p4-custom-taxonomy.php
```diff
@@ -0,0 +1,233 @@
...
+
+		/**
+		 * Add a dropdown to choose planet4 post type.
+		 *
+		 * @param WP_Post $post The WordPress that will be filtered/edited.
+		 */
+		public function create_taxonomy_metabox_markup( WP_Post $post ) {
+			$attached_type = get_the_terms( $post, self::TAXONOMY );
+			$current_type  = ( is_array( $attached_type ) ) ? $attached_type[0]->slug : - 1;
+			$all_types     = get_terms( self::TAXONOMY, [ 'hide_empty' => false ] );
```
angtheod: Please, when using assoc arrays put each key-value pair in each own line, to comply with WP coding standards.

angtheod:
Much better now (functions.php is now 150 lines smaller).
Apart from 2 small comments, looks good.

PR #332, opened 2018-03-12
PLANET 1861 Changing header description tag from <p> to <div>.

ashish173:


assets/scss/common/_page-header.scss
```diff
@@ -164,24 +164,27 @@
...
+  @include x-large-and-up {
+    margin-top: 36px;
   }
 }
 
 .page-header-content {
   display: none;
-
+  hyphens: none;
+  font-family: "Lora", serif;
```
comzeradd: Why don't you use `$lora` here?

ashish173: Ohh yeah good catch, I'll make the change.

ashish173:
Yeah, I have been asking that feature for long, I'll squash now but I sincerely hope someone hears me next time I say it.

PR #337, opened 2018-03-12
Feature/1889 assign page to category

sagarsdeshmukh:


classes/class-p4-taxonomy-page.php
```diff
@@ -0,0 +1,146 @@
...
+		 * @param string $output The html to be applied to each row of the $column.
+		 * @param string $column The name of the column to be managed.
+		 * @param int    $term_id The ID of the WP_Term object that is added or edited.
+		 *
+		 * @return string The new html to be applied to each row of the $column.
+		 */
+		public function manage_taxonomy_custom_column( $output, $column, $term_id ) : string {
+			if ( 'category_page' === $column ) {
+				$category_page = get_term_meta( $term_id, 'category_page', true );
+				$output        = get_the_title( $category_page );
```
angtheod: May I suggest we put a link to the Edit page of the relevant page here instead of just the title, in order to make better use of this column.
We could output:
`				$output        = '<a href="' . get_edit_post_link( $category_page ) . '" target="_blank">' . get_the_title( $category_page ) . '</a>';
`

classes/class-p4-taxonomy-page.php
```diff
@@ -0,0 +1,146 @@
...
+
+		/**
+		 * Add custom field(s) to taxonomy form.
+		 *
+		 * @param WP_Term $wp_tag The object passed to the callback when on Edit Tag page.
+		 */
+		public function add_taxonomy_form_fields( $wp_tag ) {
+
+			$args = [
+				'hierarchical'     => 1,
```
angtheod: I believe default is 1 for hierarchical.
Do we need this?

classes/class-p4-taxonomy-page.php
```diff
@@ -0,0 +1,146 @@
...
+			}
+
+			if ( isset( $wp_tag ) && $wp_tag instanceof WP_Term ) {
+				$category_page = get_term_meta( $wp_tag->term_id, 'category_page', true );
+				if ( $category_page ) {
+					$args['selected'] = $category_page;
+				} ?>
+				<tr class="form-field edit-wrap term-category-page-wrap">
+					<th>
+						<label><?php echo __( 'Select Category Page', 'planet4-master-theme' ); ?></label>
```
angtheod: Please use esc_html__ to escape output.

classes/class-p4-taxonomy-page.php
```diff
@@ -0,0 +1,146 @@
...
+					</th>
+					<td>
+						<?php wp_dropdown_pages( $args ); ?>
+						<p class="description"><?php echo __( 'Associate this category with a page.', 'planet4-master-theme' ); ?></p>
```
angtheod: Please use esc_html__ to escape output.

classes/class-p4-taxonomy-page.php
```diff
@@ -0,0 +1,146 @@
...
+				$category_page = get_term_meta( $wp_tag->term_id, 'category_page', true );
+				if ( $category_page ) {
+					$args['selected'] = $category_page;
+				} ?>
+				<tr class="form-field edit-wrap term-category-page-wrap">
+					<th>
+						<label><?php echo __( 'Select Category Page', 'planet4-master-theme' ); ?></label>
+					</th>
+					<td>
+						<?php wp_dropdown_pages( $args ); ?>
```
angtheod: WPCS complains about wp_dropdown_pages args not being consistently escaped by default (some are but some are not). We could make it sure like this
`wp_dropdown_pages( array_map( 'esc_attr', $args ) );`

Same for line 70

classes/class-p4-taxonomy-page.php
```diff
@@ -0,0 +1,146 @@
...
+				'name'             => 'category_page',
+				'id'               => 'category-page-id',
+				'class'            => 'category-page-class',
+			];
+
+			// Filter issue pages
+			$options          = get_option( 'planet4_options' );
+			$explore_page_id  = $options['explore_page'];
+			if ( $explore_page_id ) {
+			    $args['child_of'] = $explore_page_id;
```
angtheod: Indent with tabs

classes/class-p4-taxonomy-page.php
```diff
@@ -0,0 +1,146 @@
...
+			$args = [
+				'hierarchical'     => 1,
+				'post_type'        => 'page',
+				'show_option_none' => 'None',
+				'name'             => 'category_page',
+				'id'               => 'category-page-id',
+				'class'            => 'category-page-class',
+			];
+
+			// Filter issue pages
```
angtheod: Single line comments should end with .

classes/class-p4-taxonomy-page.php
```diff
@@ -0,0 +1,146 @@
...
+					</th>
+					<td>
+						<?php wp_dropdown_pages( $args ); ?>
+						<p class="description"><?php echo __( 'Associate this category with a page.', 'planet4-master-theme' ); ?></p>
+					</td>
+				</tr>
+			<?php } else { ?>
+				<div class="form-field add-wrap term-category-page-wrap">
+					<label><?php esc_html_e( 'Select Category Page', 'planet4-master-theme' ); ?></label>
+					<?php wp_dropdown_pages( $args ); ?>
```
angtheod: Right now in the dropdown menu we are getting only the pages that have a category assigned to them.
I don't think this is problematic really, but thought its worth mentioning in case someone else thinks otherwise.

koyan: @sagarsdeshmukh is that correct? Do you restrict to only pages with a category assigned to them? If yes why?

sagarsdeshmukh: @koyan  - No, I just put a condition to show child pages of explore page.

angtheod: @sagarsdeshmukh @koyan 
My mistake, it does not matter if a category is assigned or not. Did not read well my pages categories locally.

classes/class-p4-taxonomy-page.php
sagarsdeshmukh: Agreed

classes/class-p4-taxonomy-page.php
sagarsdeshmukh: You are right. We don't need this.

sagarsdeshmukh:
Did the changes as per the comments.

classes/class-p4-taxonomy-page.php
```diff
@@ -0,0 +1,147 @@
...
+				$category_page = get_term_meta( $wp_tag->term_id, 'category_page', true );
+				if ( $category_page ) {
+					$args['selected'] = $category_page;
+				} ?>
+				<tr class="form-field edit-wrap term-category-page-wrap">
+					<th>
+						<label><?php echo esc_html__( 'Select Category Page', 'planet4-master-theme' ); ?></label>
+					</th>
+					<td>
+						<?php wp_dropdown_pages( $args ); ?>
```
angtheod: @sagarsdeshmukh 
Actually, could you do this right when you do output (lines 64, 71) because CS is still complaining about not
escaping output.
So, it could be `wp_dropdown_pages( $array_map( 'esc_attr', $args ) );`

PR #338, opened 2018-03-13
PLANET 1858 Including expiry time in document cookie

ashish173:
Setting the expiry string to 1 year in create cookie function.

### DIFF
--- a/assets/js/custom.js
+++ b/assets/js/custom.js
@@ -180,8 +180,10 @@ $(function() {
   }
 });
 
-function createCookie(name,value,days) {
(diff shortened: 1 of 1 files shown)


assets/js/custom.js
```diff
@@ -204,7 +204,9 @@ $(function() {
 
 $("#hidecookie").click(function () {
   $(".cookie-block").slideUp("slow");
-  createCookie('greenpeace', 'policy-accepted');
+  var expiryDate = new Date();
+  expiryDate.setMonth(expiryDate.getMonth() + 12);
```
koyan: This creates a Date() js object.
What you need to pass in your function is a number of days. 
This does not work. Fix it.

PR #344, opened 2018-03-14
PLANET-1921 Eslint support

comzeradd:
This is similar to the `stylelint` work, but for js code this time. The first commit breaks down `custom.js` to partial files, same way we had them on presentation-layer and fixes syntax and indentation errors on current code.

The important commit is the [second one](77f831523b3f96a62247c5eb9b3b57bdc3ae8934) that adds the eslint support.

This should be merged/deployed along with greenpeace/planet4-base#29


ashish173:
@comzeradd can you please fix the conflict.

comzeradd:
@ashish173 fixed :)

PR #345, opened 2018-03-15
PLANET 1704 Fixing alignment of button on small screen size.

ashish173:
Ticket [link](https://jira.greenpeace.org/browse/PLANET-1704) 

As per feedback from the design team changing the width of the button from `90%` to `92%`.
cc: @tolen 

button92


ashish173:
I want them to review it once and take that once they have seen it once. This way they'll be able to suggest the exact changes to be done.

PR #354, opened 2018-03-19
PLANET-1772 - Add issue to Posts breadcrumbs

comzeradd:


single.php
```diff
@@ -29,6 +29,24 @@ function add_body_classes_for_post( $classes ) {
...
 
+// Get Post categories
+$categories       = get_the_category( $post->ID );
+
+// Handle navigation links.
+if ( $categories ) {
+	foreach ( $categories as $category ) {
+		if ( $category && ( $category->name !== $post->post_title ) ) {     // Do not add links inside the Issue page itself.
+			// Get Issue.
+			$issue = get_page_by_title( $category->name );                  // Category and Issue need to have the same name.
```
angtheod: @koyan This way to get Issue page from Category name is to be changed?

koyan: If the final point is just linking to the issue page, then there is no point mixing the category there. By definition everything bellow the defined "Explore" page is an issue page.

angtheod: @koyan Yes, it can work like that, just tested it. Its a bit more complex but it works without this assumption about the names.

PR #355, opened 2018-03-20
PLANET-1926 Carousel header responsive images

kirdia:
Change carousel header js to use current loaded image from srcset for img tags.
Add event listeners to change css background images in a carousel header block as window is resized.
Change retina-large image size not to crop images.

[Issue 1926](https://jira.greenpeace.org/browse/PLANET-1926)

### DIFF
--- a/assets/js/partials/carousel.js
+++ b/assets/js/partials/carousel.js
@@ -3,6 +3,15 @@
 $(document).ready(function() {
   'use strict';
 
+  // Add event listeners when window is resized, change the carousel's images background attribute to the current loaded ones.
+  $("#carousel-wrapper-header").find('img').on('load', function () {
+    var current_img_src = $(this).get(0).currentSrc;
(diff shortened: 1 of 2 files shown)


assets/js/partials/carousel.js
```diff
@@ -3,6 +3,15 @@
 $(document).ready(function() {
   'use strict';
 
+  // Add event listeners when window is resized, change the carousel's images background attribute to the current loaded ones.
+  $("#carousel-wrapper-header").find('img').on('load', function () {
+    var current_img_src = $(this).get(0).currentSrc;
+    var current_bg_img = $(this).parent().css('background-image').replace(/.*\s?url\([\'\"]?/, '').replace(/[\'\"]?\).*/, '');
+    if (current_img_src !== current_bg_img) {
```
comzeradd: Why not change the background image anyway? It would probably be faster and more reliable.

kirdia: I did not want to change the background image every time the load event was triggered on the img, because every time that you resize the window this event is triggered (but the current img has not changed necessarily)

PR #357, opened 2018-03-20
PLANET-1868 - Make 4 column blocks responsive

comzeradd:
Required for greenpeace/planet4-plugin-blocks#243

### DIFF
--- a/assets/js/partials/pubslider.js
+++ b/assets/js/partials/pubslider.js
@@ -1,26 +1,30 @@
 $(document).ready(function() {
   'use strict';
 
-  $('.publications-slider').slick({
-    infinite:       false,
-    mobileFirst:    true,
-    slidesToShow:   2.2,
-    slidesToScroll: 1,
-    arrows:         false,
-    dots:           false,
-    responsive: [
-      {
-        breakpoint: 992,
-        settings: { slidesToShow: 4 }
-      },
-      {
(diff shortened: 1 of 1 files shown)


kirdia:
@nikos are we removing slick from the 4 column block? Removing this will not apply slick to the first row of this block.
You can slickify method to apply slick to this row also.

ashish173:
I am not sure if I fully understand why we are removing this file to ensure responsiveness.

comzeradd:
This script adds inline fixed widths to these columns and also breaks the flex grid. So for instance, if we have only 3 publications these won't expand to cover full width of the page.

I can tweak it to only run this script when we have at least 4 elements. I did some local tests without the script and the "load more" functionality seems to work the same. I'll do another run locally with the proposed tweak.

comzeradd:
I amended the commit so we don't apply the slick code only when we have less than 4 publications. It's when we actually need a more responsive design.

kirdia:
@comzeradd I think slick js was added to provide swiping functionality on smaller screen sizes.
 If functionality is not affected in case we have less than 4 publications, then it is fine :+1:

comzeradd:
Yes, good point. Since we know the breakpoint for this I made a small change just to be sure we don't break slick functionality on small screen sizes. I think it's good to go now.

PR #358, opened 2018-03-21
PLANET-1823 auto set featured image

angtheod:
Simplify the process of auto selecting the featured image while allowing the editor to override it at the same time.

@koyan You were right. I thought this would not work the very first time a new Post is created and there is an image inside the post content and a User selected featured image. But it does.

Still will need some testing, but I think it will be ok like that.

### DIFF
--- a/assets/admin/js/edit_post.js
+++ b/assets/admin/js/edit_post.js
@@ -49,40 +49,4 @@ $(document).ready(function () {
       }
     }
   });
-
-  // Custom handling of Featured Image selection, so that we know if Editor tried to override the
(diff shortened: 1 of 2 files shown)


sagarsdeshmukh:
As we have already checked if the user has set featured image then we can remove else part -
```
if ( isset( $result[0] ) ) {
	$attachment_id = $result[0];
	set_post_thumbnail( $post_id, $attachment_id );
} else {
	// If no image was found inside the post's content then unset the featured image.
	update_post_meta( $post_id, '_thumbnail_id', '' );
}
```

PR #359, opened 2018-03-22
PLANET-1963 smarter post breadcrumps

angtheod:
Implement smarter logic for loading navigation links of Issue pages that are relevant to current P4_Post (page, post, evergreen) categories. Moved logic inside P4_Post class from where it can be accessed by all pages.

- This way we no longer need to match Category Names with Issue page names. -

### DIFF
--- a/classes/class-p4-post.php
+++ b/classes/class-p4-post.php
@@ -49,5 +49,52 @@ if ( ! class_exists( 'P4_Post' ) ) {
 
 			return in_array( $this->id, $pages );
 		}
+
+
+		/**
+		 * Loads in context information on the navigation links for Issue pages relevant to current Post's categories.
+		 *
+		 * @param array $context An indexed array with all data needed to render current page.
+		 */
+		public function load_nav_issues_links( &$context ) {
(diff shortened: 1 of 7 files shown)


classes/class-p4-post.php
```diff
@@ -49,5 +49,52 @@ public function is_take_action_page() {
...
+				if ( $categories_ids && $explore_page_id ) {
+					$args   = [
+						'post_parent'  => $explore_page_id,
+						'post_type'    => 'page',
+						'post_status'  => 'publish',
+					];
+					if ( count( $categories_ids ) > 1 ) {
+						$args['category__in'] = $categories_ids;
+					} elseif ( 1 === count( $categories_ids ) ) {
+						$args['cat'] = (int) $categories_ids[0];
```
koyan: Doesn't the above category__in work even if there is only one category?

angtheod:
Post's navigation data of the P4_Post's Issues can now be accessed from the object's properties.
We should do the same for the navigation data of the P4_Post's Campaigns.

PR #363, opened 2018-03-23
PLANET-1954 Fix issue of page type auto assigned to post_type PAGE

sagarsdeshmukh:
The issue was if no taxonomy is assigned to page/post, it will take first page type taxonomy as default. Here now, we have added condition for post type check, which solved an issue for new records. We need to find a solution for already added pages with default page_type.

### DIFF
--- a/classes/class-p4-custom-taxonomy.php
+++ b/classes/class-p4-custom-taxonomy.php
@@ -24,7 +24,7 @@ if ( ! class_exists( 'P4_Custom_Taxonomy' ) ) {
 			add_action( 'created_term',                       array( $this, 'trigger_rewrite_rules' ), 10, 3 );
 			add_action( 'edited_term',                        array( $this, 'trigger_rewrite_rules' ), 10, 3 );
 			add_action( 'delete_term',                        array( $this, 'trigger_rewrite_rules' ), 10, 3 );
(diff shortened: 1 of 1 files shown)


classes/class-p4-custom-taxonomy.php
```diff
@@ -216,6 +216,11 @@ public function save_taxonomy_page_type( $post_id ) {
 				return;
 			}
 
+			// Check if post type is PAGE.
+			if ( 'page' === $post->post_type ) {
```
kirdia: Good catch @sagarsdeshmukh !! :+1: I completely missed that.
But can you change it to continue execution only when post_type is 'post' ? 
Since we really only want posts to have terms of this taxonomy.

sagarsdeshmukh: @kirdia Do you mean making it `if ( 'post' != $post->post_type )` ?

angtheod: @sagarsdeshmukh I think he means making it 'post' === $post->post_type

kirdia: Well both will do the job, but I preferred the first one 
```if ( 'post' != $post->post_type ) {return;}```
It makes no big difference.

PR #370, opened 2018-03-27
PLANET-1983 Hide search appended checkboxes

kirdia:
[Issue 1983](https://jira.greenpeace.org/browse/PLANET-1983)


angtheod:
No idea how this bug happened, but this should fix it.

PR #401, opened 2018-04-15
PLANET-2027 search adjustments

angtheod:
Adjust sort by date to work with ES. Adjust js code to make sure we get filtered and/or sorted results from ES when the initial search was performed via right-click. Same goes for when clearing filters.

### DIFF
--- a/assets/js/search.js
+++ b/assets/js/search.js
@@ -5,6 +5,7 @@ $ = jQuery; //eslint-disable-line no-global-assign
 // Search page.
 $(function() {
   var $search_form      = $( '#search_form' );
+  var $es               = $('#es', $search_form);
   var $load_more_button = $( '.btn-load-more-click-scroll' );
   var load_more_count   = 0;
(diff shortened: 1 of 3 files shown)


assets/js/search.js
```diff
@@ -20,17 +21,39 @@ $(function() {
...
+	if ( $es.length > 0 ) {
+	  is_elastic_search = true;
+    }
     $( '#orderby', $search_form ).val( $( this ).val() ).parent().submit();
     return false;
   });
 
   // Submit form on Filter click event or on Apply button click event.
   $( 'input[name^="f["]:not(.modal-checkbox), .applybtn' ).off( 'click' ).on( 'click', function() {
+    var $es = $('#es', $search_form);
```
kirdia: This block of code (determining if it is elastic search ) appears sometimes, maybe it could be extracted in a function.

angtheod: Applied suggestions

PR #402, opened 2018-04-17
PLANET-1941 Feature/1941 sitemap

angtheod:
Implement Sitemap page. Add Sitemap page in theme so that it can be accessed on /sitemap.


angtheod:
@comzeradd Niko, I tag you in case you want to start working on the sitemap markup.

PR #403, opened 2018-04-19
PLANET-1941 Feature/1941 sitemap

angtheod:
Implement Sitemap page. Add Sitemap page in theme so that it can be accessed via /sitemap.

Note:
Its the same code changes as #402 , but merging this into develop this time

### DIFF
--- a/classes/class-p4-sitemap.php
+++ b/classes/class-p4-sitemap.php
@@ -1,0 +1,147 @@
+<?php
+
+if ( ! class_exists( 'P4_Sitemap' ) ) {
+
+	/**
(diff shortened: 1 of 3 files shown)


comzeradd:
Shouldn't we merge the sitemap branch, when the frontend is also ready?

koyan:
We trust Nikos to deliver it. 
And we are not doing a release without the whole thing completed.

angtheod:
Ok, btw @comzeradd to test it you need to create a new Page and select Sitemap as Template in the Page Attributes metabox and then add a footer menu item with this URL.

PR #411, opened 2018-04-20
PLANET4-2028 adjust search filtering

angtheod:
Adjust filters to work with ES.
Fixed Content Type -> Page filter to not include Action Pages in its results.
Fixed Content Type -> Document filter to include only P4_Search::DOCUMENT_TYPES in its results. Still, the Document filter counter needs adjustment to include only the defined mime types.
Fixed a bug where Pages where as counted into Category filters counters, while only Posts should have the p4-page-type taxonomy.
Fixed a php notice.

### DIFF
--- a/classes/class-p4-search.php
+++ b/classes/class-p4-search.php
@@ -293,6 +293,8 @@ if ( ! class_exists( 'P4_Search' ) ) {
 								];
 								break;
 							case 'ptype':
+								// This taxonomy is used only for Posts.
+								$args['post_type']   = 'post';
 								$args['tax_query'][] = [
 									'taxonomy' => 'p4-page-type',
 									'field'    => 'term_id',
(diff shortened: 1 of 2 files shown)


classes/class-p4-search.php
```diff
@@ -601,14 +630,16 @@ protected function set_results_context( &$context ) {
...
-				if ( $page_types ) {
-					foreach ( (array) $page_types as $page_type ) {
-						// p4-page-type filters.
-						$context['page_types'][ $page_type->term_id ]['term_id'] = $page_type->term_id;
-						$context['page_types'][ $page_type->term_id ]['name']    = $page_type->name;
-						$context['page_types'][ $page_type->term_id ]['results']++;
+				// Page Type <-> Category. This taxonomy is used only for Posts.
+				if ( 'post' === $post->post_type ) {
+					$page_types = get_the_terms( $post->ID, 'p4-page-type' );
+					if ( $page_types ) {
```
kirdia: @angtheod would it be better if we check that the variable is an array here? Incase get_the_terms returns a WP_Error object.

PR #412, opened 2018-04-23
PLANET-2067 Wpml - planet4 settings multilingual

kirdia:
Add wpml plugin hook to make planet4 settings multilingual.


classes/class-p4-settings.php
```diff
@@ -302,6 +307,14 @@ public function option_metabox() {
 				'fields'     => $this->fields,
 			];
 		}
+
+		/**
+		 * Hook for wpml plugin.
+         * Enables the possibility to save a different value per language for the theme options using WPML language switcher.
```
koyan: Fix spacing (begging of lines is with spaces instead of tabs)

PR #416, opened 2018-04-25
PLANET-2095 Use planet4 settings for take action card button text in search results

kirdia:
[Issue 2095](https://jira.greenpeace.org/browse/PLANET-2095)

### DIFF
--- a/templates/tease-search.twig
+++ b/templates/tease-search.twig
@@ -59,7 +59,7 @@
 
 			{% if ( is_action ) %}
 				{%  set link_text = settings['take_action_covers_button_text'] %}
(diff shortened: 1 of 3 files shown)


templates/tease-search.twig
```diff
@@ -58,7 +58,8 @@
 			<p class="search-result-item-content">{{ post.preview()|truncate( 25, true )|raw }}</p>
 
 			{% if ( is_action ) %}
-				<a href="{{ post.link }}" class="btn btn-primary btn-medium btn-small">{{ __( 'Take Action', domain ) }}</a>
+				{%  set link_text = settings['take_action_covers_button_text'] %}
```
angtheod: @kirdia do we need here the default 'Take Action' value in case the setting is not supplied?

PR #429, opened 2018-05-08
PLANET-2146 - Adjust navbar to fit a Language selector flag

comzeradd:
This removes custom hard-coded class names for the nav items, that we don't use anyway.
Adds a raw filter, which should be safe for item titles. in order to render language flag image.
The current language won't be displayed at all, so in a bilingual scenario we would only have one flag there.

lang

### DIFF
--- a/assets/scss/common/_navbar.scss
+++ b/assets/scss/common/_navbar.scss
@@ -346,7 +346,7 @@ $menu-height-large: rem(60);
     }
 
     .nav-link {
-      padding: 0 20px;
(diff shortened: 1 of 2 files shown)


comzeradd:
@koyan let me know when this goes live on Indian staging so I can adjust WPML admin settings. We should probably add a section on the handbook about this.

koyan:
I have switched gp india staging to the develop branch of master-theme , and I have updated their site. Please check if it is as you expected it.

PR #433, opened 2018-05-11
PLANET-1661 First version of posts report page.

kirdia:
[Issue 1661](https://jira.greenpeace.org/browse/PLANET-1661)

### DIFF
--- a/assets/js/posts_report.js
+++ b/assets/js/posts_report.js
@@ -4,22 +4,21 @@ var postsView, pagesView;
 $ = jQuery;
 
 $(document).ready(function () {
-    var dateFormat = "yy-mm-dd",
-        from = jQuery("#from")
-            .datepicker({
-                dateFormat: "yy-mm-dd",
-                defaultDate: "-3w",
-                changeMonth: true,
-                numberOfMonths: 2
-            }),
-        to = jQuery("#to").datepicker({
-            dateFormat: "yy-mm-dd",
-            defaultDate: "-1w",
+    var from = jQuery("#from")
+        .datepicker({
+            dateFormat: 'yy-mm-dd',
+            defaultDate: '-3w',
             changeMonth: true,
             numberOfMonths: 2
-        })
+        });
+    var to = jQuery('#to').datepicker({
+        dateFormat: 'yy-mm-dd',
+        defaultDate: '-1w',
+        changeMonth: true,
+        numberOfMonths: 2
+    });
 
-    $("#posts-filter").off('click').on('click', function () {
+    $('#posts-filter').off('click').on('click', function () {
 
         var filters = {};
         var from = $('#from').datepicker().val();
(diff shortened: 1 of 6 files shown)


comzeradd:
It seems that `assets/js/posts_report.js` has many indendation and linting issues. Check out [Travis output](https://travis-ci.org/greenpeace/planet4-master-theme/jobs/377670639).

functions.php
```diff
@@ -905,6 +911,84 @@ public function override_img_caption_shortcode( $output, $attr, $content ) {
...
+			]
+		);
+		wp_enqueue_script( 'posts-report' );
+		wp_register_style(
+			'jquery-ui-datepicker',
+			'https://cdnjs.cloudflare.com/ajax/libs/jqueryui/1.12.1/themes/overcast/jquery-ui.min.css'
+		);
+		wp_enqueue_style( 'jquery-ui-datepicker' );
+		include 'posts_report.php';
+	}
```
angtheod: @kirdia What do you think about moving the functionality about the post's report page to its own class under the classes directory as we have done in other cases? This way we can keep the functions.php as small as possible and the P4_Master_Site class without too many concerns.

kirdia: Yes I agree, I'll move the functionality to a new class.

posts_report.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+use Timber\Timber;
+
+Timber::render( [ 'posts_report.twig' ], [] );
+
+include dirname( __FILE__ ) . '/underscore_templates/posts-reports.tpl.php';
```
angtheod: The file is named `posts_reports.tpl.php` (with underscore), so it can not find it

functions.php
```diff
@@ -904,12 +904,12 @@ public function override_img_caption_shortcode( $output, $attr, $content ) {
...
 	}
-
 }
 
 new P4_Master_Site( [
 	'P4_Custom_Taxonomy',
 	'P4_Taxonomy_Image',
 	'P4_Settings',
 	'P4_Control_Panel',
+	'P4_Post_Report_Controller',
```
angtheod: @kirdia could we avoid loading this controller here and instead create it within the posts_report.php so that this functionality actually gets loaded only in the desired backend page instead.

angtheod: ps. at some point, we could do this for the other classes as well (taxonomy_image, settings, control_panel, etc...)

assets/js/posts_report.js
```diff
@@ -0,0 +1,173 @@
...
+    template: wp.template('p4-post'),
+    tagName: 'tr',
+
+    prepare: function () {
+      return this.model.toJSON();
+    }
+  });
+
+  p4.initialize = function () {
+    postCollection = new wp.api.collections.Posts();
```
angtheod: I get `Uncaught TypeError: wp.api.collections.Posts is not a constructor` here, which prevents results from loading.
Am I missing something?

functions.php
kirdia: Unfortunately it can't be placed in posts_report.php. There are some wordpress hooks that need to run early on the theme load process.

assets/js/posts_report.js
kirdia: I can't reproduce it. Is wp-api.js being loaded?

angtheod: fixed by calling the posts_report.tpl.php call

PR #439, opened 2018-05-15
PLANET-2146 - Adapt language switcher to the new design

comzeradd:
This implements the new design as seen at [PLANET-1978](https://jira.greenpeace.org/browse/PLANET-1978).

Please check that the way I'm passing a new context variable is actually the right way to do it ;)


comzeradd:
I wrapped the new context variable inside a `function_exists` check for instances that won't have WPML installed.

PR #441, opened 2018-05-17
PLANET-2191 Sitemap fix

kirdia:
Fix sitemap class methods returning null values.
[Issue 2191](https://jira.greenpeace.org/browse/PLANET-2191)

### DIFF
--- a/templates/sitemap.twig
+++ b/templates/sitemap.twig
@@ -4,36 +4,47 @@
 	<div class="page-sitemap">
 		<div class="container">
 			<div class="row">
-				<div class="col-md-7 order-md-1 col-lg-4">
-					<h5 class="mb-3 mt-5 mt-md-0">{{ actions_title }}</h5>
-					{% for action in actions %}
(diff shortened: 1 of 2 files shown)


templates/sitemap.twig
```diff
@@ -4,36 +4,47 @@
...
+				{% if (actions) %}
+					<div class="col-md-7 order-md-1 col-lg-4">
+						<h5 class="mb-3 mt-5 mt-md-0">{{ actions_title }}</h5>
+						{% for action in actions %}
+							<a href="{{ action.link }}">{{ action.title|e('wp_kses_post')|raw }}</a><br/>
+						{% endfor %}
+					</div>
+				{% endif %}
+				{% if (issues) %}
+					<div class="col-md-7 order-md-3 order-lg-2 col-lg-3">
```
kirdia: @comzeradd Will these classes (order-*) break up design if some of the divs in the row does not exist?

PR #445, opened 2018-05-18
PLANET-2223 Add the copyright symbol in images in posts if it does not exist already

koyan:


sagarsdeshmukh:
Note : If `$image_credit` variable is empty it will produce undefined var warning. @koyan can you initialize it.

PR #457, opened 2018-05-23
PLANET-2240 Prevent cookies setting

kirdia:
[Issue 2240](https://jira.greenpeace.org/browse/PLANET-2240)
- Prevent happy point iframe loading if cookie policy is not accepted.
- Prevents google tag manager js embed if cookie policy is not accepted.
- Remove iframes/embeds from post content (could be changed to remove only iframes/embeds that come from specific urls).
- Set google login plugin cookie only in wp-login page.
- Set cookie expire time to 1 year.

### DIFF
--- a/assets/js/partials/cookies.js
+++ b/assets/js/partials/cookies.js
@@ -19,7 +19,7 @@ $(document).ready(function() {
   function createCookie(name, value) {
     var date = new Date();
     date.setTime(date.getTime() + (365 * 24 * 60 * 60 * 1000));
-    document.cookie = encodeURI(name) + '=' + encodeURI(value) + ';domain=.' + document.domain + ';path=/;' + ";expires=" + date.toGMTString();
+    document.cookie = encodeURI(name) + '=' + encodeURI(value) + ';domain=.' + document.domain + ';path=/;' + ';expires=' + date.toGMTString();
   }
 
 
--- a/assets/js/partials/cookies.js
+++ b/assets/js/partials/cookies.js
@@ -1,24 +1,27 @@
+function readCookie(name) {
+  var nameEQ = name + '=';
+  var ca = document.cookie.split(';');
+  for (var i = 0; i < ca.length; i++) {
+    var c = ca[i];
+    while (c.charAt(0) == ' ') {
+      c = c.substring(1, c.length);
+    }
+    if (c.indexOf(nameEQ) == 0) {
+      return c.substring(nameEQ.length, c.length);
+    }
+  }
+  return null;
+}
+
 $(document).ready(function() {
   'use strict';
 
   function createCookie(name, value) {
-    document.cookie = encodeURI(name) + '=' + encodeURI(value) + ';domain=.' + document.domain + ';path=/;';
+    var date = new Date();
+    date.setTime(date.getTime() + (365 * 24 * 60 * 60 * 1000));
+    document.cookie = encodeURI(name) + '=' + encodeURI(value) + ';domain=.' + document.domain + ';path=/;' + ";expires=" + date.toGMTString();
   }
 
-  function readCookie(name) {
-    var nameEQ = name + '=';
-    var ca = document.cookie.split(';');
-    for(var i=0;i < ca.length;i++) {
-      var c = ca[i];
-      while (c.charAt(0)==' ') {
-        c = c.substring(1,c.length);
-      }
-      if (c.indexOf(nameEQ) == 0) {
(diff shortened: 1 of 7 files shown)


assets/js/partials/cookies.js
```diff
@@ -1,24 +1,27 @@
...
+}
+
 $(document).ready(function() {
   'use strict';
 
   function createCookie(name, value) {
-    document.cookie = encodeURI(name) + '=' + encodeURI(value) + ';domain=.' + document.domain + ';path=/;';
+    var date = new Date();
+    date.setTime(date.getTime() + (365 * 24 * 60 * 60 * 1000));
+    document.cookie = encodeURI(name) + '=' + encodeURI(value) + ';domain=.' + document.domain + ';path=/;' + ";expires=" + date.toGMTString();
```
comzeradd: Please use single quotes along the whole string here.

assets/js/partials/happy_point.js
```diff
@@ -19,11 +19,16 @@ $(document).ready(function() {
     }
   }
 
+  var cookie = readCookie('greenpeace');
```
comzeradd: Please add a `/* global readCookie */` on the top of this file to make Travis happy :)

functions.php
```diff
@@ -260,10 +260,16 @@ public function add_to_context( $context ) {
...
-		$context['website_navbar_title']  = $options['website_navigation_title'] ?? __( 'International (English)', 'planet4-master-theme' );
+
+		// Do not embed google tag manager js if 'greenpeace' cookie is not set.
+		$cookie_consent = isset( $_COOKIE['greenpeace'] );
+		$gtm            = $options['google_tag_manager_identifier'] ?? '';
+
+		$options                         = get_option( 'planet4_options' );
+		$context['donatelink']           = $options['donate_button'] ?? '#';
+		$context['google_tag_value']     = ! empty( $gtm ) && $cookie_consent ? $gtm : '';
+		$context['google_tag_value']     = $options['google_tag_manager_identifier'] ?? '';
```
sagarsdeshmukh: We don't need `$context['google_tag_value']     = $options['google_tag_manager_identifier'] ?? '';` this line. right?

kirdia: Yeap, you are right. I'll remove it.

classes/class-p4-cookies.php
```diff
@@ -0,0 +1,165 @@
...
+		 */
+		public function filter_iframes_and_embeds( $content ) {
+
+			// Regex for iframes and embeds inside content.
+			$pattern = '~<embed.*>[^>]*</embed>|<iframe.*>[^>]*</iframe>~';
+
+			preg_match_all( $pattern, $content, $matches );
+
+			foreach ( $matches[0] as $match ) {
+				$replacement = '<div class="">' . __( $this->filtered_text, 'planet4-master-theme' ) . '</div>';
```
angtheod: the translation function expects a string literal as first arg. Otherwise it can register it for translation

classes/class-p4-cookies.php
```diff
@@ -0,0 +1,165 @@
...
+					if ( false !== stristr( $url, $purl ) ) {
+						$allowed = false;
+					}
+				}
+			}
+			if ( $allowed ) {
+				return $html;
+			}
+
+			return '<div class="">' . __( $this->filtered_text, 'planet4-master-theme' ) . '</div>';
```
angtheod: Same here as in line 122

assets/js/partials/cookies.js
```diff
@@ -1,24 +1,27 @@
+function readCookie(name) {
+  var nameEQ = name + '=';
+  var ca = document.cookie.split(';');
+  for (var i = 0; i < ca.length; i++) {
+    var c = ca[i];
+    while (c.charAt(0) == ' ') {
```
angtheod: can we use also type checking here (and in line 9) with '==='

assets/js/partials/cookies.js
```diff
@@ -32,6 +34,6 @@ $(document).ready(function() {
   $('#hidecookie').click(function () {
     $('.cookie-block').slideUp('slow');
     $('footer').css('margin-bottom', '0');
-    createCookie('greenpeace', 'policy-accepted');
+    createCookie('greenpeace', '1', 365);
```
angtheod: should the value here be '2' ?
On firefox I am getting the iframe and the 'content is blocked' message at the same time

angtheod: screenshot-2018-5-24 development planet 4 website

PR #461, opened 2018-05-25
PLANET-2231 override imagick method

angtheod:
Override imagick's thumbnail_image method.


classes/class-p4-image-compression.php
```diff
@@ -5,5 +5,27 @@
 	/**
 	 * Class P4_Image_Compression
 	 */
-	class P4_Image_Compression extends WP_Image_Editor_Imagick {}
+	class P4_Image_Compression extends WP_Image_Editor_Imagick {
+		/**
```
kirdia: Just a new line here needed.

PR #493, opened 2018-06-18
PLANET-2347 Imagestest

angtheod:
https://greenpeace.github.io/planet4-imagestest-results/

The solution applied in the 6th image produces much better quality than P4 currently produces
and at the same time -47% size on retina-large images, -38% size on large images and -17% on medium images (on large and medium images results about the size of the 2 tested images are conflicting though).
```
Resize: Lanczos
Sharpening: unsharpMask( Radius 1, Sigma 0.45, Amount 3, Threshold 0 )
Compression: Interlace_Plane
Quality: 60
```

### DIFF
--- a/classes/class-p4-image-compression.php
+++ b/classes/class-p4-image-compression.php
@@ -26,7 +26,24 @@ if ( ! class_exists( 'P4_Image_Compression' ) ) {
 				$filter_name = $this->filter;
 			}
 			parent::thumbnail_image( $dst_w, $dst_h, $filter_name, $strip_meta );
+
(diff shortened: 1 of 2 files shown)


kirdia:
@angtheod I agree with Nikos. Do we need all these commits? Since they are mostly tests I don't think they offer anything having them as git history.

angtheod:
@kirdia It happened this way because the main Jira issue was divided into 5 sub-issues and so I had to do separate PRs for each one. I could re-create the branch and add only the final code changes in it if you want.

kirdia:
@angtheod ok it's no big deal. After all they are mostly changes to a new introduced file.

PR #494, opened 2018-06-18
PLANET-2259 cfc-link-colors

angtheod:
Change Content Four Column block's link color.

### DIFF
--- a/assets/scss/common/_content-four-column.scss
+++ b/assets/scss/common/_content-four-column.scss
@@ -62,11 +62,11 @@
         font-weight: 500;
         padding: 0;
         margin: 0;
+        color: $dark-blue;
 
         @include medium-and-up {
           font-size: rem(18);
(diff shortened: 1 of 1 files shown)


assets/scss/common/_content-four-column.scss
```diff
@@ -66,7 +66,7 @@
         @include medium-and-up {
           font-size: rem(18);
           font-weight: 500;
-          color: $black;
+          color: $dark-blue;
```
comzeradd: Can you move the color on the `h4` block above. It seems that on mobile it has the wrong blue, so let's fix it for all devices.

comzeradd:
:+1: 

Can you squash the commits into one? It seems that squashing is not enabled on this repo.

angtheod:
This is to be postponed according to the issue creator. Closing PR for now.

PR #509, opened 2018-06-22
PLANET-2350 fix campaign page cfc bug

angtheod:
Fixes issue with content four column block not displaying on NL site. Issue had to do with page types named in Dutch. Added options within the Edit Tag page for selecting the page type(s) of the posts that will be displayed within the CFC block. If no page type is selected within the Edit Tag page then display Publications by default.

### DIFF
--- a/classes/class-p4-campaigns.php
+++ b/classes/class-p4-campaigns.php
@@ -1,13 +1,15 @@
 <?php
 
-if ( ! class_exists( 'P4_Taxonomy_Image' ) ) {
+if ( ! class_exists( 'P4_Campaigns' ) ) {
 	/**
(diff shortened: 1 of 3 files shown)


kirdia:
@angtheod Looks good, but do NROs need to set the p4 page types/terms for every tag? Maybe we can add an extra p4 setting (like default p4 page type) for a term that will reflect to 'Publications' term.

koyan:
@kirdia there is a finite number of tags. It is not like posts that new ones get added every day. So, I don't feel like a generic setting is really needed in this context.

angtheod:
@kirdia This way they have the option to display different page types in various tags

PR #531, opened 2018-07-26
PLANET-2450 Move the check 'Enforce cookies' logic to client side

sagarsdeshmukh:


templates/html-header.twig
```diff
@@ -30,29 +30,57 @@
 	<meta name="twitter:image" content="{{ image_data[0] }}">
 
 	{% if google_tag_value %}
-		<script>
+	<script>
+          function readCookie(name) {
```
kirdia: If we put readCookie function here, we should remove it then from cookies.js .
Or we can do a quicker check with
`document.cookie.split(';').filter(function(c) {return c.indexOf('greenpeace=2') >= 0}).length`

templates/html-header.twig
```diff
@@ -30,29 +30,42 @@
 	<meta name="twitter:image" content="{{ image_data[0] }}">
 
 	{% if google_tag_value %}
-		<script>
+	<script>
+          var cookie_consent = document.cookie.split(';').filter(function(c) {return c.indexOf('greenpeace=2') >= 0}).length;
+          var google_tag_value = '{{ google_tag_value }}';
+          var enforce_cookies_policy = '{{ enforce_cookies_policy }}';
+
+          if ( enforce_cookies_policy && 1 !== parseInt( cookie_consent ) ) {
```
kirdia: @sagarsdeshmukh also here I believe enforce_cookies_policy will always be interpreted as true since it will hold a value at all times. Probably better to check if it is true.

angtheod: @kirdia ye, I checked that too, but it works because false is interpreted as '' (emtpy string) and so it doesn't enter the if statement.

angtheod:
I think we should check also here the cookie consent because this part is loaded even if the cookie is not accepted (while GDPR is active).

sagarsdeshmukh:
@angtheod - I agree on your above point but the `<noscript>` tag is intend to load in case of javascript is disable. so I don't found the way to add cookie consent check here.

angtheod:
ok, I think it is not so important to be honest.

PR #553, opened 2018-08-30
PLANET-2581 Fix php warnings in P4_Post

kirdia:
Refactor get_social_accounts function in P4_Post to prevent php warnings

### DIFF
--- a/classes/class-p4-post.php
+++ b/classes/class-p4-post.php
@@ -105,7 +105,7 @@ if ( ! class_exists( 'P4_Post' ) ) {
 		 *
 		 * @return array Associative array with the social media accounts.
 		 */
-		public function get_social_accounts( $social_menu ): array {
+		public function get_social_accounts( $social_menu ) : array {
 			$social_accounts = [];
 			if ( isset( $social_menu ) ) {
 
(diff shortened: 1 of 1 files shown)


classes/class-p4-post.php
```diff
@@ -105,19 +105,22 @@ public function set_issues_links() {
...
-					} elseif ( false !== strpos( $social_menu_item->url, 'twitter' ) ) {
-						$social_accounts['twitter'] = end( explode( '/', rtrim( $social_menu_item->url, '/' ) ) );
-					} elseif ( false !== strpos( $social_menu_item->url, 'youtube' ) ) {
-						$social_accounts['youtube'] = end( explode( '/', rtrim( $social_menu_item->url, '/' ) ) );
-					} elseif ( false !== strpos( $social_menu_item->url, 'instagram' ) ) {
-						$social_accounts['instagram'] = end( explode( '/', rtrim( $social_menu_item->url, '/' ) ) );
+					foreach ( $brands as $brand ) {
+						if ( false !== strpos( $social_menu_item->url, $brand ) ) {
+							$temp                      = explode( '/', rtrim( $social_menu_item->url, '/' ) );
+							$social_accounts[ $brand ] = count( $temp ) > 0 ? $temp[ count( $temp ) - 1 ] : '';
```
angtheod: You can move the explode call before the 2nd foreach and
also can do `$social_accounts[ $brand ] = end( $temp );` for faster and more compact.
Also, naming $temp to $url_parts would be an idea.

kirdia: Yeap, good catch for the first idea. I'll move it in the outer loop.
I can't use end(), it produces the warnings.

PR #566, opened 2018-09-27
PLANET-2517 - Use gulp to process frontend code

comzeradd:


comzeradd:
I added a commit to replace `Font-Awesome` with `Fork-Awesome`, an open-source fork that is available on the CDN we use. That way we unify the way we handle our external frontend dependencies.

gulpfile.js
```diff
@@ -1,54 +1,71 @@
...
+    .pipe(sourcemaps.init())
+    .pipe(concat('main.js'))
+    .pipe(uglify())
+    .pipe(sourcemaps.write(path_dest))
+    .pipe(gulp.dest(path_dest));
+});
+
+gulp.task('watch', function () {
+  gulp.watch(path_scss, ['sass']);
+  gulp.watch(path_dest, ['uglify']);
```
kirdia: Is this meant to be `path_js` ?

comzeradd:
@kirdia good catch! I updated the commit

PR #584, opened 2018-10-18
PLANET-2417 - Fix long background images overflow

comzeradd:


comzeradd:
Yes, that rule removed the double scrollbar and the empty space below the footer.

Notice that I applied that only to body element, not html too.

PR #586, opened 2018-10-22
PLANET-2614: Refined results page

pablocubico:
Hi! Starting this PR to show you what I've found so far.

Please read the issue for context: https://jira.greenpeace.org/browse/PLANET-2614

I created a new `page_type` template for the `?p4-page-type=[page_type}` archive requests.
This is the URL provided by WP Admin when clicking "View" on a Page Type.

To reach this page with a friendly URL like `/story` a Redirect rule has to be set in the Redirection plugin as:

```
Source: ^/(press|publication|story)/?$
Target: index.php?p4-page-type=$1
```

Another method would be to add a `'rewrite' => [ 'slug' => 'page_type' ]` parameter to the custom taxonomy's `rewrite` arg, which will resolve to URLs like `/page_type/story`, but I don't know if this could have side effects.

I changed the slug from all the places I've found it, except for the search page, where a click is being simulated in JS to automatically check the proper checkbox and submit the form. 

Please have a look and let me know what you think!

### DIFF
--- a/archive.php
+++ b/archive.php
@@ -33,6 +33,9 @@ if ( is_day() ) {
 } else if ( is_post_type_archive() ) {
 	$context['title'] = post_type_archive_title( '', false );
 	array_unshift( $templates, 'archive-' . get_post_type() . '.twig' );
+} else if ( is_tax('p4-page-type') ) {
+	$context['page_type'] = get_queried_object();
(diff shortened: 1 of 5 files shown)


pablocubico:
Closing, it was duplicated by #587 while trying to avoid Github errors.

PR #587, opened 2018-10-22
PLANET-2614: Refined results page

pablocubico:
Please read the issue for context: https://jira.greenpeace.org/browse/PLANET-2614

Added a `page_type` slug parameter to the custom taxonomy's `register_taxonomy` method.

Using `taxonomy.php` as a catcher for taxonomies requests using the default behavior of WP template hierarchy.

I changed the Page Type slug from most pages from being a filter url to a pure slug except for the search page, where the slug simulates a click in the "Page Type" checkbox.

Let me know what you think!

page-types-posts

### DIFF
--- a/taxonomy.php
+++ b/taxonomy.php
@@ -1,0 +1,16 @@
+<?php
+/**
+ * The template for displaying Taxonomy pages.
+ *
+ * Used to display taxonomy-type pages
+ * 
+ * Learn more: http://codex.wordpress.org/Template_Hierarchy
+ */
+
+$templates = array( 'page_type.twig', 'index.twig' );
+
+$context = Timber::get_context();
+$context['page_type'] = get_queried_object();
+$context['posts'] = Timber::get_posts();
+
+Timber::render( $templates, $context );
--- a/templates/single.twig
+++ b/templates/single.twig
@@ -14,7 +14,7 @@
 			<div class="container">
 				<div class="top-page-tags">
 					{% if ( page_type ) %}
-						<a class="tag-item tag-item--main page-type" href="{{ filter_url }}">{{ page_type|e('wp_kses_post')|raw }}</a>
+						<a class="tag-item tag-item--main page-type" href="{{ fn('get_term_link', page_type.term_id) }}">{{ page_type|e('wp_kses_post')|raw }}</a>
 					{% endif %}
 
 					{% if ( post.issues_nav_data ) %}
--- a/templates/tease-author.twig
+++ b/templates/tease-author.twig
@@ -8,7 +8,7 @@
 				{{ post.get_author_override }}
 
 				{% for page_type in post.get_custom_terms %}
-					<a href="/?s=&orderby=relevant&f[ptype][{{ page_type.name }}]={{ page_type.term_id }}" class="search-result-item-head no-btn tag-item tag-item--main page-type">{{ page_type.name|e('wp_kses_post')|raw }}</a>
+					<a href="{{ fn('get_term_link', page_type.term_id, 'p4-page-type') }}" class="search-result-item-head no-btn tag-item tag-item--main page-type">{{ page_type.name|e('wp_kses_post')|raw }}</a>
 				{% endfor %}
 
 				{% if (post.tags) %}
--- a/templates/tease-page-type.twig
+++ b/templates/tease-page-type.twig
@@ -1,0 +1,32 @@
+<li id="result-row-{{ post.ID }}" class="media search-result-list-item">
+	<img class="d-flex search-result-item-image" src="{{ post.thumbnail.src('thumbnail') }}"
+		alt="{{ fn('esc_attr', post.thumbnail.alt|default( post.title ) )|raw }}" />
+
+	<div id="tease-{{ post.ID }}" class="media-body search-result-item-body tease tease-{{ post.post_type }}">
+		<div class="search-result-tags top-page-tags">
+			{{ post.get_author_override }}
+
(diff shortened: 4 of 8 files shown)


koyan:
Without having done code review, I like the above result. I would proceed to do the 
>  We could also add a [ 'slug' => 'page_type' ] parameter to the
>  rewrite argument of register_taxonomy, which will provide URLs
>  as /page_type/story but I don't know if this could cause any side effects.
Please do this as well before the UAT.

koyan:
Without having done code review, I like the above result. I would proceed to do the 
>  We could also add a [ 'slug' => 'page_type' ] parameter to the
>  rewrite argument of register_taxonomy, which will provide URLs
>  as /page_type/story but I don't know if this could cause any side effects.
Please do this as well before the UAT.

archive.php
```diff
@@ -33,6 +33,9 @@
 } else if ( is_post_type_archive() ) {
 	$context['title'] = post_type_archive_title( '', false );
 	array_unshift( $templates, 'archive-' . get_post_type() . '.twig' );
+} else if ( is_tax('p4-page-type') ) {
```
pablocubico: Not sure about this hardcoded string... Should I use a constant here? something else?

templates/tease-author.twig
```diff
@@ -8,7 +8,7 @@
 				{{ post.get_author_override }}
 
 				{% for page_type in post.get_custom_terms %}
-					<a href="/?s=&orderby=relevant&f[ptype][{{ page_type.name }}]={{ page_type.term_id }}" class="search-result-item-head no-btn tag-item tag-item--main page-type">{{ page_type.name|e('wp_kses_post')|raw }}</a>
+					<a href="/page_type/{{ page_type.slug }}" class="search-result-item-head no-btn tag-item tag-item--main page-type">{{ page_type.name|e('wp_kses_post')|raw }}</a>
```
kirdia: @pablocubico please make these links either relative permalinks (for example using wp_make_link_relative) or use the full link for the term (in production we have base urls like greenpeace.org/international/ ).

templates/tease-page-type.twig
```diff
@@ -0,0 +1,29 @@
+<li id="result-row-{{ post.ID }}" class="media search-result-list-item">
+	<img class="d-flex search-result-item-image" src="{{ post.thumbnail.src('thumbnail') }}"
+		alt="{{ fn('esc_attr', post.thumbnail.alt|default( post.title ) )|raw }}" />
+
+	<div id="tease-{{ post.ID }}" class="media-body search-result-item-body tease tease-{{ post.post_type }}">
+		<div class="search-result-tags top-page-tags">
+			{{ post.get_author_override }}
+
+			<a href="/page_type/{{ page_type.slug }}" class="search-result-item-head no-btn tag-item tag-item--main page-type">{{ page_type.name|e('wp_kses_post')|raw }}</a>
```
kirdia: Same here for relative/full urls

templates/single.twig
```diff
@@ -14,7 +14,7 @@
 			<div class="container">
 				<div class="top-page-tags">
 					{% if ( page_type ) %}
-						<a class="tag-item tag-item--main page-type" href="{{ filter_url }}">{{ page_type|e('wp_kses_post')|raw }}</a>
+						<a class="tag-item tag-item--main page-type" href="/page_type/{{ page_type_slug }}">{{ page_type|e('wp_kses_post')|raw }}</a>
```
kirdia: Same here for relative/full urls

taxonomy.php
```diff
@@ -0,0 +1,25 @@
...
+ * @subpackage  Timber
+ * @since   Timber 0.2
+ */
+
+$templates = array( 'page_type.twig', 'index.twig' );
+
+$context = Timber::get_context();
+
+$context['page_type'] = get_queried_object();
+array_unshift( $templates, 'page_type.twig' );
```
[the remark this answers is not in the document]
pablocubico: Oh right! this is vestigial from the `archives-` version.

pablocubico:
@kirdia Updated! But I was just checking these changes from Sagar here: https://github.com/greenpeace/planet4-plugin-blocks/pull/374/files

I think we may need to check for a trailing slash here too.

templates/single.twig
```diff
@@ -14,7 +14,7 @@
 			<div class="container">
 				<div class="top-page-tags">
 					{% if ( page_type ) %}
-						<a class="tag-item tag-item--main page-type" href="{{ fn('get_home_url') ~ '/page_type/' ~ page_type.slug }}">{{ page_type|e('wp_kses_post')|raw }}</a>
+						<a class="tag-item tag-item--main page-type" href="{{ fn('rtrim', fn('get_home_url'), '/') ~ '/page_type/' ~ page_type.slug }}">{{ page_type|e('wp_kses_post')|raw }}</a>
```
pablocubico: We could probably encapsulate this into a helper but maybe it's out of the scope of this issue.

kirdia: Actually I saw that with the rewrite rule in place for p4-page-type taxonomy we can use built-in WordPress functions.

[get_term_link](https://developer.wordpress.org/reference/functions/get_term_link/)
will return the full url for the term (and it will work with wpml enabled also)

https://www.planet4.test/en/page_type/story/
https://www.planet4.test/page_type/story/

so we can use `fn('get_term_linkl', page_term_id)` passing the term_id/object/slug as parameter.

pablocubico: @kirdia Good point, updated! Also in the blocks plugin: https://github.com/greenpeace/planet4-plugin-blocks/pull/375

PR #590, opened 2018-10-23
PLANET-2476 Custom open graph fields

kirdia:
Removed all body class filters in php templates and passed the body classes straight to the context.
Add cmb2 metabox for open graph custom fields, to override title, description and image in a page/post.
Add tests for custom open graph values.
Add dev dependency for stevegrunwell/phpunit-markup-assertions that provides markups assertions.


koyan:
> Removed all body class filters in php templates and passed the body classes straight to the context.
Why is this relevant to this Jira issue?

kirdia:
> > Removed all body class filters in php templates and passed the body classes straight to the context.
> 
> Why is this relevant to this Jira issue?

It is not relevant with the context of the issue, but I needed to remove them for the tests to run properly (plus it is a cleaner solution). The functions are redeclared when running the test suite resulting in failure.

functions.php
```diff
@@ -748,6 +748,45 @@ public function register_header_metabox() {
...
+			'id'   => $prefix . 'og_title',
+			'type' => 'text_medium',
+		] );
+
+		$p4_open_graph->add_field( [
+			'name'    => __( 'Description', 'planet4-master-theme-backend' ),
+			'desc'    => __( 'Enter description if you want to override the open graph description', 'planet4-master-theme-backend' ),
+			'id'      => $prefix . 'og_description',
+			'type'    => 'wysiwyg',
+			'options' => array(
```
sagarsdeshmukh: please use short hand for array declaration([]).

classes/class-p4-post.php
```diff
@@ -179,5 +179,62 @@ public function get_custom_type() {
...
+				}
+			}
+
+			return $og_title;
+		}
+
+		/**
+		 * Get value for open graph description meta.
+		 *
+		 * @return mixed
```
sagarsdeshmukh: Would it be a string return type?

kirdia:
Rebased onto release branch and applied code review suggestions.

functions.php
```diff
@@ -748,6 +748,45 @@ public function register_header_metabox() {
...
+			'desc'         => __( 'Upload an image or select one from the media library to override the open graph image', 'planet4-master-theme-backend' ),
+			'id'           => $prefix . 'og_image',
+			'type'         => 'file',
+			'options'      => [
+				'url' => false,
+			],
+			'text'         => [
+				'add_upload_file_text' => __( 'Add Image', 'planet4-master-theme-backend' ),
+			],
+			'preview_size' => 'large',
```
sagarsdeshmukh: We can add `query_args`, so that the image selection popup will only list image options.
```
'query_args'   => array(
	'type' => 'image',
),
```

PR #596, opened 2018-10-25
PLANET-2530: Include visual regression tests in the automated testing

pablocubico:
Starting this PR so we can move the conversation here.

I added/updated a few URLs according to the default content, let me know which ones you think we should keep / update / add, etc.

I also added the option to create a CI report, I just wanted to test it as I saw it was present in Nikos's `planet-2530` branch. Do we need it?

### DIFF
--- a/backstop.js
+++ b/backstop.js
@@ -2,7 +2,7 @@
 
 var arguments = require('minimist')(process.argv.slice(2));
 
-var url = 'https://dev.p4.greenpeace.org/international';
+var url = 'https://k8s.p4.greenpeace.org/defaultcontent/';
 if (arguments.url) {
   url = arguments.url;
 }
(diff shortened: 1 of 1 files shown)


pablocubico:
@comzeradd @27Bslash6 let me know your thoughts!

comzeradd:
Yes, the "CI" option will create a report in junit format, which is what CircleCI wants.

I wouldn't add all the footer links on paths, since all of them use the same template. One should be enough (eg. Copyright).

pablocubico:
Ok! Removed those URLs, I thought about keeping the Privacy and Cookies page since it has a slightly different layout (a "submenu block") that I guess may be affected for other changes. Let me know if you think it's unnecessary, I'll remove it too.

The script currently takes a URL argument, let me know if we should change that to an environment variable or some other method. What else do you  need from me to close this?

pablocubico:
This became outdated and superseded by the `planet-uitests` branch.

PR #598, opened 2018-10-26
PLANET-2805 - Load More functionality - Decide and document approach

pablocubico:
WIP - opening this PR to chat about implementations.

Disregard the hardcoded rows per page and the fact that I inlined the script into a template and the CamelCase classnames, I guess working with React just tricked my brain into having html + behavior on the same file and _componentesque_ names. This is just for demoing.

I did a few tests on the WP REST API, but I wanted to know what your previous thoughts were on this. If we use the REST API and the JSON response to render on the client side, we'll need to use some sort of template (whether is jQuery, EJS, etc) to render the posts rows (`tease-page-type.twig`, or `tease-author.twig`), which will duplicate the code of those templates, so we'll need to change code in both places on every updated.

On the other hand, I tried a more Wordpress-like approach using server-side rendering (which is back since progressive web apps!), the very same template with a query var for the page and some of Timber's tools (Pagination, PostsQuery). This way we would only be maintaining our typical templates.

Attaching the result. Let me know your thoughts.

load-more-ssr

### DIFF
--- a/templates/page_type.twig
+++ b/templates/page_type.twig
@@ -21,12 +21,38 @@
 
 		<div class="row">
 			<div class="col-lg-8 multiple-search-result">
-				<ul class="list-unstyled">
+				<ul class="list-unstyled PostsList">
 					{% for post in posts %}
 						{% include 'tease-page-type.twig' %}
 					{% endfor %}
 				</ul>
 			</div>
 		</div>
+
+		<div class="row">
+			<div class="col-lg-8">
+				<button data-page="1" class="LoadMoreButton btn btn-medium btn-small btn-secondary mb-5">SHOW 5 MORE RESULTS</button>
+			</div>
+		</div>
 	</div>
+
+	<script>
+		$ = jQuery;
+		const loadMore = $('.LoadMoreButton')
+
+		loadMore.on('click', (e) => {
+			const nextPage = parseInt(loadMore[0].dataset.page) + 1
+			const totalPages = {{ posts.pagination.total }}
+			const url = `{{ fn('get_term_link', page_type.term_id, 'p4-page-type') }}?page=${ nextPage }`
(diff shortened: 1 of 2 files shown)


templates/page_type.twig
```diff
@@ -21,12 +21,38 @@
...
+		$ = jQuery;
+		const loadMore = $('.LoadMoreButton')
+
+		loadMore.on('click', (e) => {
+			const nextPage = parseInt(loadMore[0].dataset.page) + 1
+			const totalPages = {{ posts.pagination.total }}
+			const url = `{{ fn('get_term_link', page_type.term_id, 'p4-page-type') }}?page=${ nextPage }`
+			loadMore[0].dataset.page = nextPage
+
+			$.get(url, (response) => {
```
pablocubico: TODO: Add proper error handling.

27Bslash6:
+1 to this general approach rather than trying to rebuild our templates client-side.  All the tools we need are baked into the current frameworks we have.

27Bslash6:
First thoughts: instead of the button being 'load more', it should by default be 'next page' - if javascript is disabled the button still works and we're shown the next X responses (whatever `posts_per_page` is set to, via select / link ), but if the request is via AJAX ( https://developer.wordpress.org/reference/functions/wp_doing_ajax/ ) then the results are retrieved and appended, with an `e.preventDefault` in the loadMore/nextPage event handler.

comzeradd:
Nice! :+1: 

I like this approach. I did some reading on WP's API, mostly out of curiosity, but that would indeed add more complexity to our codebase. Better to render templates on the backend.

I also like the fact that this can work even without js. But regarding the text on the button, we could ask for some input from the design team. We have these buttons in various places and it makes sense to make them consistent.

PR #600, opened 2018-10-30
PLANET-2193 media search list

angtheod:
Fixed issue that prevented searching the WP Media Library in List view mode from returning results.


angtheod:
@koyan should we push this one in 1.31 as well?

PR #615, opened 2018-11-06
PLANET-2637: Author Profile: missing posts

pablocubico:
Starting this PR to polish & finish our method to get posts for [Author](https://jira.greenpeace.org/browse/PLANET-2637) and [Page Type](https://jira.greenpeace.org/browse/PLANET-2822) pages.

I noticed we could be using the search class, but the `get_paged_posts` method is tied to the `search` pages templates. Maybe if we could separate the core functionality from the render, we could benefit from using the search logic already implemented in the search class, which looks more robust  in terms of validation, sanitization and caching. But maybe that's out of the scope of this issue.

### DIFF
--- a/assets/js/author.js
+++ b/assets/js/author.js
@@ -1,0 +1,25 @@
+var $ = jQuery;
+const loadMore = $('button.load-more');
+loadMore.on('click', (e) => {
+  e.preventDefault();
+
+  const nextPage = parseInt(loadMore[0].dataset.page) + 1;
+  const totalPages = loadMore[0].dataset.total;
+  const url = loadMore[0].dataset.url + `?page=${ nextPage }`;
+  loadMore[0].dataset.page = nextPage;
+
+  $.ajax({
+    url: url,
+    type: 'GET',
+    dataType: 'html'
+  }).done(function ( response ) {
+    // Append the response at the bottom of the results and then show it.
+    $( '.multiple-search-result .list-unstyled' ).append( response );
+  }).fail(function ( jqXHR, textStatus, errorThrown ) {
+    console.log(errorThrown); //eslint-disable-line no-console
+  });
+
+  if (nextPage == totalPages) {
+    loadMore.fadeOut();
+  }
+});
--- a/assets/js/author.js
+++ b/assets/js/author.js
@@ -1,12 +1,11 @@
-var $ = jQuery;
 const loadMore = $('button.load-more');
-loadMore.on('click', (e) => {
+loadMore.on('click', function (e) {
   e.preventDefault();
 
-  const nextPage = parseInt(loadMore[0].dataset.page) + 1;
-  const totalPages = loadMore[0].dataset.total;
-  const url = loadMore[0].dataset.url + `?page=${ nextPage }`;
-  loadMore[0].dataset.page = nextPage;
+  const nextPage = parseInt(this.dataset.page) + 1;
+  const totalPages = this.dataset.total;
+  const url = this.dataset.url + `?page=${ nextPage }`;
+  this.dataset.page = nextPage;
 
   $.ajax({
     url: url,
--- a/author.php
+++ b/author.php
@@ -8,36 +8,39 @@
  * @subpackage  Timber
  * @since    Timber 0.1
  */
+
 wp_register_script( 'author', get_template_directory_uri() . '/assets/js/author.js', [ 'jquery' ], '0.0.1', true );
 wp_enqueue_script( 'author' );
 
-$context = Timber::get_context();
+$context          = Timber::get_context();
 $context['posts'] = Timber::get_posts( false, 'P4_Post' );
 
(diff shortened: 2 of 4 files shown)


author.php
```diff
@@ -8,13 +8,36 @@
...
 $context = Timber::get_context();
 $context['posts'] = Timber::get_posts( false, 'P4_Post' );
+
+$post_args = [
+    'posts_per_page' => 10,
+    'post_type' => 'post',
+    'paged' => 1
+];
+
 if ( isset( $wp_query->query_vars['author'] ) ) {
```
pablocubico: Is this check necessary? 🤔

templates/author.twig
```diff
@@ -39,5 +39,17 @@
 				</ul>
 			</div>
 		</div>
+
+		{% if posts.pagination.total > 1 %}
+			<div class="row">
+				<div class="col-lg-8">
+					<button
+						data-page="1"
```
pablocubico: I'm using data attributes here but we could use a different method, for example, calling a JS function using twig variables as in:

```
<script>
  setupAuthorLazyLoad('{{ author.path }}', '{{ posts.pagination.total }}')
</script>
```

For now, I think the `data` method is fine.

author.php
```diff
@@ -8,13 +8,36 @@
...
  */
-global $wp_query;
+wp_register_script( 'author', get_template_directory_uri() . '/assets/js/author.js', [ 'jquery' ], '0.0.1', true );
+wp_enqueue_script( 'author' );
 
 $context = Timber::get_context();
 $context['posts'] = Timber::get_posts( false, 'P4_Post' );
+
+$post_args = [
+    'posts_per_page' => 10,
```
sagarsdeshmukh: Please add comma after last array element, also align the array double arrow.

author.php
```diff
@@ -8,13 +8,36 @@
...
+    'post_type' => 'post',
+    'paged' => 1
+];
+
 if ( isset( $wp_query->query_vars['author'] ) ) {
 	$author = new TimberUser( $wp_query->query_vars['author'] );
 	$context['author'] = $author;
-	$context['title'] = 'Author Archives: ' . $author->name();
+    $context['title'] = 'Author Archives: ' . $author->name();
```
sagarsdeshmukh: Please align equal sign with surrounding assignments (PHPCS checks).

author.php
```diff
@@ -8,13 +8,36 @@
...
+    $post_args['author'] = $wp_query->query_vars['author'];
+}
+
+if ( get_query_var('page') ) {
+    $templates = [ 'tease-author.twig' ];
```
sagarsdeshmukh: Same here, equal sign alignment.

sagarsdeshmukh:
Please use WP default setting of date format at here - https://github.com/greenpeace/planet4-master-theme/blob/bug/planet-2637/templates/tease-author.twig#L26
(This change is not part of this issue)
(https://github.com/greenpeace/planet4-master-theme/pull/618/commits/beccc6c5624d5edd06c02caafee01e0ee23ce33b)

author.php
pablocubico: Done!

author.php
pablocubico: Done!

pablocubico:
@sagarsdeshmukh All checks fixed, thanks! 💯

assets/js/author.js
```diff
@@ -0,0 +1,25 @@
+var $ = jQuery;
+const loadMore = $('button.load-more');
+loadMore.on('click', (e) => {
```
kirdia: please replace the arrow function with function() so we can use `this` and refer to the actual element that the event was called on.

assets/js/author.js
```diff
@@ -0,0 +1,25 @@
...
+  e.preventDefault();
+
+  const nextPage = parseInt(loadMore[0].dataset.page) + 1;
```
kirdia: we can replace loadMore[0] with `this`, dropping the arrow function

templates/author.twig
kirdia: I think it is good for now also! :+1:

assets/js/author.js
```diff
@@ -0,0 +1,25 @@
+var $ = jQuery;
```
kirdia: we can drop this if we add main js as dependency

author.php
```diff
@@ -8,13 +8,39 @@
...
  * @since    Timber 0.1
  */
-global $wp_query;
 
-$context = Timber::get_context();
+wp_register_script( 'author', get_template_directory_uri() . '/assets/js/author.js', [ 'jquery' ], '0.0.1', true );
+wp_enqueue_script( 'author' );
+
+$context          = Timber::get_context();
 $context['posts'] = Timber::get_posts( false, 'P4_Post' );
```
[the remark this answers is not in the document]
pablocubico: Removed

PR #616, opened 2018-11-06
PLANET-2820 Fix author override issue on new page type pages

sagarsdeshmukh:


taxonomy.php
```diff
@@ -7,10 +7,22 @@
  * Learn more: http://codex.wordpress.org/Template_Hierarchy
  */
```
angtheod: Please add `use Timber\Timber;`, for our editors to find the class.

PR #624, opened 2018-11-12
PLANET-2559

angtheod:
Add more functionality in P4_Post class to make it easier to present data for those posts.

### DIFF
--- a/classes/class-p4-post.php
+++ b/classes/class-p4-post.php
@@ -10,6 +10,14 @@ if ( ! class_exists( 'P4_Post' ) ) {
 
 		/** @var array $issues_nav_data */
 		protected $issues_nav_data;
+		/** @var array $tags */
+		protected $tags;
+		/** @var string $page_type */
+		protected $page_type;
+		/** @var int $page_type_id */
+		protected $page_type_id;
+		/** @var string $content_type */
+		protected $content_type;
 
 		/**
 		 * Checks if post is the act page.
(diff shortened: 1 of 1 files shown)


classes/class-p4-post.php
```diff
@@ -156,28 +164,76 @@ public function get_author_override() {
 		}
 
 		/**
-		 * Get post/page custom planet4 type.
+		 * Gets the tags for this P4_Post.
+		 */
+		public function set_tags() {
+			$tags = wp_get_post_tags( $this->id );
```
kirdia: @angtheod you can get the post tags with timber also.
https://timber.github.io/docs/reference/timber-post/#terms
It returns Timber\Term objects which contain the tag link also.
https://timber.github.io/docs/reference/timber-term/#link

classes/class-p4-post.php
```diff
@@ -156,28 +164,76 @@ public function get_author_override() {
...
+		 */
+		public function get_tags() {
+			$this->tags;
+		}
+
+		/**
+		 * Sets the page type for this P4_Post.
+		 */
+		public function set_page_types() {
+			$page_type_data = get_the_terms( $this->id, P4_Custom_Taxonomy::TAXONOMY );
```
kirdia: Maybe you can use timber's terms() that wrap the objects with Timber\Term and you get the term link also.

angtheod: @kirdia I know, but I did it this way so that it is consistent with the pre-existing code [here](https://github.com/greenpeace/planet4-plugin-blocks/blob/941d4c69b97a349e6083fbd5951307d4f335945d/classes/controller/blocks/class-articles-controller.php#L357) and the way that the [template](https://github.com/greenpeace/planet4-plugin-blocks/pull/393/files#diff-4356b2eb546b563b92942ed0b711c6ddR30) expects the data to render.

angtheod: If I change it, then I will need to change also the populate_post_items() method in the controller.

classes/class-p4-post.php
angtheod: I agree overall, but then I will need to change also the way that the controller passes those data during page load to the template. Should I do this in this PR?

koyan: No. It is outside the scope of this. Please open a new jira issue for it.

classes/class-p4-post.php
koyan: We are in the end of the release. Please open a jira issue for the follow up. IF this works at the moment.

classes/class-p4-post.php
kirdia: Yes, you are right. it needs change there also because we use arrays there.
Because is big change, let's address it in a different issue.

PR #631, opened 2018-11-19
PLANET-2822 If author override is enabled, do not show author profile…

sagarsdeshmukh:
… link

### DIFF
--- a/templates/single.twig
+++ b/templates/single.twig
@@ -39,7 +39,11 @@
 						<div class="single-post-meta">
 							<address class="single-post-author">
 								{{ __( 'by', 'planet4-master-theme' ) }}
-								<a href="{{ post.author.path }}">{{ post.author.name }}</a>
+								{% if not ( post.get_author_override ) %}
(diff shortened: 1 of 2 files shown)


templates/single.twig
```diff
@@ -39,7 +39,11 @@
 						<div class="single-post-meta">
 							<address class="single-post-author">
 								{{ __( 'by', 'planet4-master-theme' ) }}
-								<a href="{{ post.author.path }}">{{ post.author.name }}</a>
+								{% if not ( post.get_author_override ) %}
```
angtheod: Just a notice here as well to be sure you saw this:
P4_Post and P4_User objects are (after 1.24.0) more complete and P4_Post has since its creation a P4_User in it which actually is a Timber/User. We can deprecate `get_author_override` and `get_custom_terms` and use `post.author` and `post.page_types` directly in the templates having the full data provided by Timber for Timber/Post and Timber/User objects.
#629 
@sagarsdeshmukh @kirdia @pablocubico

PR #636, opened 2018-11-20
PLANET-2637 authorpage remove overridden posts

angtheod:
Remove overridden posts from Author page. This happened because we were using TimberPost instead of P4_Post. TimberPost does not have post.get_author_override nor post.author.is_fake

### DIFF
--- a/author.php
+++ b/author.php
@@ -9,6 +9,9 @@
  * @since    Timber 0.1
  */
 
+use Timber\Timber;
+use Timber\PostQuery;
+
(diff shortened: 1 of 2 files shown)


pablocubico:
Looks good, but there is a minor concern: as the Post is not really removed from the query, the pagination takes the overridden posts into account. 

Is there a way to filter out the overridden posts in the query itself? filtering by custom fields perhaps?

angtheod:
The PR can still be merged since it only improves things

PR #647, opened 2018-11-26
PLANET-1799 Create a Category template

Wakkatela:
Category template was missing. I created a `category.php`.


category.php
```diff
@@ -0,0 +1,44 @@
+<?php
+/**
+ * The template for displaying Taxonomy pages.
```
pablocubico: I guess this comments should say "Category" instead of Taxonomy!

category.php
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * Learn more: http://codex.wordpress.org/Template_Hierarchy
+ *
+ * @package  WordPress
+ * @subpackage  Timber
+ */
+
+use Timber\Timber;
+
+$templates = [ 'page_type.twig', 'index.twig' ];
```
pablocubico: This is using the `page_type` template, I guess the `category` pages should look more or less the same, but maybe just copying the `page_type` template to a new one could be better for adapting to tiny differences.

pablocubico: ...OR we could change the `page_type`, and `tease-page-type` files into something more generic, like `taxonomy` and `tease-taxonomy`, given that a `category` is a `taxonomy` too... I'll wait for other opinions on this.

koyan: The correct way (throughout the project) would be that all common elements are defined on their own. (For example `article_excerpt.twig`), and then included in other common elements (for example articles_list.twig), and then included in any template that needs them. 
(For example inside page_type.twig, and in categories.twig etc). 
Like that we would have consistency (changing the design in one place should change in all). 

For now, if they have common parts, then 
1) find a name describing the common part, 
2) create a template for it
3) create a category.twig
4) Inside the category.twig call the common part
5) Inside the page_type.twig call the common part
6) change the above line 15 to call the category.twig

comzeradd:
Please make sure you'll squash the second commit before merge.

category.php
```diff
@@ -0,0 +1,43 @@
...
+
+$context              = Timber::get_context();
+$context['page_type'] = get_queried_object();
+$context['wp_title']  = $context['page_type']->name;
+
+wp_register_script( 'load_more', get_template_directory_uri() . '/assets/js/load_more.js', [ 'jquery', 'main' ], '0.0.1', true );
+wp_enqueue_script( 'load_more' );
+
+$post_args = [
+	'posts_per_page' => 10,
```
kirdia: @Wakkatela there is no query argument for the category here
Use the queried object (I think it is WP_Term object in this case also) to query only for posts of the queried category

category.php
```diff
@@ -0,0 +1,43 @@
...
+ * @package  WordPress
+ * @subpackage  Timber
+ */
+
+use Timber\Timber;
+
+$templates = [ 'category.twig', 'index.twig' ];
+
+$context              = Timber::get_context();
+$context['page_type'] = get_queried_object();
```
kirdia: Please change 'page_type' to 'category' for consistency

templates/category.twig
```diff
@@ -0,0 +1,43 @@
+{% extends "base.twig" %}
+
+{% block content %}
+
+	<div class="clearfix"></div>
+	<div class="skewed-overlay"></div>
+
+	<header class="page-header">
+		<div class="container">
+			<h1 class="page-header-title">{{ page_type.name }}</h1>
```
kirdia: this should be changed to category.name since we changed the variable that is being passed to the template
`category.name`
https://github.com/greenpeace/planet4-master-theme/pull/647/files#diff-538c51b64adf70254db21dbfa4b24e44R16

templates/category.twig
```diff
@@ -0,0 +1,43 @@
...
+		</div>
+
+		{% if posts.pagination.total > 1 %}
+			<div class="row">
+				<div class="col-lg-8">
+					<button
+						data-content=".multiple-search-result .list-unstyled"
+						data-page="1"
+						data-total="{{ posts.pagination.total }}"
+						data-url="{{ fn('get_term_link', page_type.term_id, 'p4-page-type') }}"
```
kirdia: also here please change this to generate category link
`data-url="{{ fn('get_category_link', category.term_id ) }}"`

templates/category.twig
```diff
@@ -0,0 +1,43 @@
...
+{% block content %}
+
+	<div class="clearfix"></div>
+	<div class="skewed-overlay"></div>
+
+	<header class="page-header">
+		<div class="container">
+			<h1 class="page-header-title">{{ page_type.name }}</h1>
+			<div class="page-header-content">
+				<p>{{ page_type.description }}</p>
```
kirdia: same here 
`category.description`

templates/category.twig
```diff
@@ -0,0 +1,43 @@
...
+	</header>
+
+	<div class="page-template">
+		<h3>{{ __( 'Results', 'planet4-master-theme' ) }}</h3>
+
+		<div class="row">
+			<div class="col-lg-8 multiple-search-result">
+				<ul class="list-unstyled">
+					{% for post in posts %}
+						{% include 'tease-page-type.twig' %}
```
kirdia: this fails because there are some specific variables for page type used in the template
https://github.com/greenpeace/planet4-master-theme/blob/develop/templates/tease-page-type.twig#L7

PR #650, opened 2018-11-28
PLANET-2579: Search bar: apply adjusted design

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-2579

Result: 
captura de pantalla 2018-11-27 a la s 23 12 37


Wakkatela:
Everything seems ok, just for a future discussion:
`.open`is a very common selector and we could repeat it and get unwanted CSS properties conflicts. We may use some naming convention to avoid this: BEM is my favorite for CSS. 

If we do that, we could reuse properties and avoid conflicts:
```
//block
.toggle {
 ...
}
   //Element
  .toggle__active {
     //sample code
     + toggle__show {
              opacity: 1;
         }
     }
   //Modifier
   .toggle--footer {
       background-color: $color-brand;
    }
```

But this is maybe something for the refactor and definition of CSS conventions to use (we need some of it =P).

PR #651, opened 2018-11-28
PLANET-2857 User Profile: HTML not appearing correctly on mobile devices (M,S)

Wakkatela:
- Added specificity to remove floats from share buttons.
- Removed unnecessary markup.
- Removed duplicated user info block.

Still need to check that fits design. Is there a guide where I can get designs for layouts?

In this case, in desktop the share buttons are top right, alone and away from the main content while description is below the picture.  I moved the buttons top of description under author photo, but need to be sure is that way and not the description that needs to be top right too.

### DIFF
--- a/templates/author.twig
+++ b/templates/author.twig
@@ -10,22 +10,25 @@
 			<h1 class="page-header-title">{{ author.name }}</h1>
 			<div class="row">
 				<div class="col-lg-7 col-md-8 col-sm-12 author-info">
-					<figure class="author-image">
-						{% if ( author.avatar ) %}
+					{% if ( author.avatar ) %}
+						<figure class="author-image">
 							<img itemprop="image" class="author-pic"
-								src="{{ fn('get_avatar_url', author.id, {'size' : 300}) }}"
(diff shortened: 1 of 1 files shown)


comzeradd:
Are these changes related to [2857](https://jira.greenpeace.org/browse/PLANET-2857)?

As far as I understand, the bug report is about the Author bio not rendered properly on mobile when it has html tags (check the [screenshot](https://drive.google.com/file/d/1SxFPL2ng7YoK6rla-lcrBu_c3y12n2Q9/view)).

comzeradd:
Ah right! Now, I get it :+1:

assets/scss/pages/_author.scss
```diff
@@ -21,6 +21,10 @@
         }
       }
     }
+
+    .share-buttons {
+      float: none;
```
comzeradd: That affects L, XL screens as well. We need the share buttons to float right there. See `_share-buttons.scss`.

kirdia:
I removed them. Yes, they weren't really necessary .

PR #652, opened 2018-11-29
PLANET-2574: IdeaPush with broken sidebar

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-2574

Ok, this is a long shot but here it goes: I don't have a local working Handbook DB for testing, so I just did some research on the surface, but my spider sense is telling me this could be the cause of the sidebar issue.

Looks like the `ip-tag` (IdeaPush tag) templates include `sidebar.php` in the render pipeline, which is a surprisingly old file:

captura de pantalla 2018-11-28 a la s 23 29 58

IdeaPush includes it by calling: 

image

This is an example page: https://planet4.greenpeace.org/ip-tag/contribution/

Removing the file causes this error: 
```
Notice: Theme without sidebar.php is deprecated since version 3.0.0 with no alternative available. Please include a sidebar.php template in your theme. in /app/source/public/wp-includes/functions.php on line 3984
```

So, keeping the file but removing the output seems to do the trick, as we are using a different method than `get_sidebar()` to include the sidebar.

Also, not included in this PR, but: is this `sidebar` variable still in use? is it ever defined? I can't find its source.

https://github.com/greenpeace/planet4-master-theme/commit/bdd2b2a93a02b9e53de1195937d9c2e1307415fc#diff-165325163f96a207bc47e67bbe480772R17

Can someone with a working local Handbook test it? 🙏 Just comment the Timber `sidebar.php` line out.

### DIFF
--- a/sidebar.php
+++ b/sidebar.php
@@ -1,10 +1,9 @@
 <?php
 /**
- * The Template for displaying all single posts
+ * Placeholder file to avoid this error:
+ * "Notice: Theme without sidebar.php is deprecated since
+ *  version 3.0.0 with no alternative available.
+ *  Please include a sidebar.php template in your theme."
  *
- *
- * @package  WordPress
- * @subpackage  Timber
+ * No output.
  */
-
-Timber::render( array( 'sidebar.twig' ), $data );


Wakkatela:
That indeed solves the sidebar problem: no error and now I can see the sidebar. (At least in my local environmnet).

Still some plugin errors, but I guess they are not related.

koyan:
I have synced our k8s handbook environments so that you can test it easier, and have made it take the branch you are working on: 
https://github.com/greenpeace/planet4-handbook/blob/develop/composer-local.json#L7

You can see that the navbar is no longer broken at: https://k8s.p4.greenpeace.org/handbook/ip-tag/ux-ui/

koyan:
@kirdia @comzeradd Any suspicion that this may break something else?

kirdia:
> @kirdia @comzeradd Any suspicion that this may break something else?

I don't know. I don't think we ever used the sidebar php template. 
It was transferred when master-theme was created from timber starter theme.

Will have to check if it is loaded in the wordpress loop, but it does not output anything (even though we have a reference in base twig template).

We can try this change in handbook child theme, if it is solely related with ideapush

koyan:
> We can try this change in handbook child theme, if it is solely related with ideapush
Yes, I thought about that, but we are going to meet the problem with other child-themes eventually in the future, so better to solve it in the master (there are going to be campaign or cms non-nro p4 sites which need a sidebar)

kirdia:
I have checked it. It is safe to do the change.
sidebar php template is never called with current code.

It needs to register a sidebar and call the template for the sidebar php template to be rendered.
We don't have any of these, so it is safe to move on.
```
register_sidebar()
get_sidebar( 'primary' )
```

PR #658, opened 2018-12-06
PLANET-3004 Fix truncate post excerpt for character based languages

kirdia:
Truncate and excerpt are pretty identical, but excerpt uses [wp_trim_words](https://developer.wordpress.org/reference/functions/wp_trim_words/) which splits the content for character based languages correctly.

### DIFF
--- a/templates/tease-author.twig
+++ b/templates/tease-author.twig
@@ -23,7 +23,7 @@
 			<span class="search-result-item-date">{{ post.post_date|date }}</span>
 		</div>
 
-		<p class="search-result-item-content">{{ post.post_excerpt|truncate( 30, true )|e('wp_kses_post')|raw }}</p>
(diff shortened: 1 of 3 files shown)


sagarsdeshmukh:
@kirdia - It won't work on my local.
image
Do I miss something?

koyan:
Sagar, is it the case that that article already has the excerpt field filled in? If yes, delete the contents of the existing excerpt field and see again.

kirdia:
> @kirdia - It won't work on my local.
> Do I miss something?

It does not depend on the content really, it depends on the site language setting. 
Try changing the Settings -> General -> Site Language to another language (Japanese for example)

sagarsdeshmukh:
@kirdia , @koyan - Yes, Its working perfect after installing the Japanese Language. Thanks both of you :)

PR #659, opened 2018-12-06
Upgrade gulp-livereload to avoid flatmap issue

tamlyn:
Currently on `develop` the yarn install fails because flatmap-stream [has been removed from npm](https://www.npmjs.com/package/flatmap-stream).

gulp-livereload 4.0.1 [pins event-stream to 3.3.4](https://github.com/vohof/gulp-livereload/pull/140) which predates the flatmap-stream security vulnerability.


comzeradd:
Thanks @tamlyn :100:

PR #663, opened 2018-12-11
PLANET-2964 - Switch from Font icons to SVG

comzeradd:
I added also [this snippet](https://github.com/greenpeace/planet4-master-theme/pull/663/files#diff-0b6561d6fb690bf3612eb9e8544876e6R5) of code for backwards compatibility. So even with the current class names this will work.


pablocubico:
This looks good as far as I can tell, but since we are engaging in this refactor my only suggestion is that as the new syntax is too verbose, maybe we could add a helper to ease the adding of new icons in a compact way and at the same time ensure that the markup is consistent, and in case we want to change something in the future, we only have to change it in one place. And we could opt-out of the helper for particular cases. Here is a working example:

```diff
diff --git a/functions.php b/functions.php
index c01c9a1..0ac8dfa 100644
--- a/functions.php
+++ b/functions.php
@@ -289,10 +289,20 @@ class P4_Master_Site extends TimberSite {
         */
        public function add_to_twig( $twig ) {
                $twig->addExtension( new Twig_Extension_StringLoader() );
+               $twig->addFilter( new Twig_SimpleFilter( 'svgicon', [ $this, 'svgicon' ] ) );

                return $twig;
        }

+       public function svgicon( $name ) {
+               $svgIconTemplate = <<<SVGICON
+<svg viewBox="0 0 32 32" class="icon">
+       <use xlink:href="{$this->theme_images_dir}symbol/svg/sprite.symbol.svg#$name"></use>
+</svg>
+SVGICON;
+               return new \Twig_Markup( $svgIconTemplate, 'UTF-8' );
+       }
+
        /**
         * Set attributes that should be allowed for posts filter
         * Allow img srcset and sizes attributes.
diff --git a/templates/navigation-bar.twig b/templates/navigation-bar.twig
index c3495cc..c63b48b 100644
--- a/templates/navigation-bar.twig
+++ b/templates/navigation-bar.twig
@@ -62,9 +62,7 @@
                                   value="{{ data_nav_bar.search_query|e('wp_kses_post')|raw }}" name="s" aria-label="Search">
                        <input id="orderby" type="hidden" name="orderby" value="{{ selected_sort ?? default_sort }}"/>
                        <button class="top-nav-search-btn" type="submit">
-                               <svg viewBox="0 0 32 32" class="icon">
-                                       <use xlink:href="{{ data_nav_bar.images }}symbol/svg/sprite.symbol.svg#search"></use>
-                               </svg>
+                               {{ 'search'|svgicon }}
                                <span class="screen-reader-text">{{ __( 'Search', 'planet4-master-theme' ) }}</span>
                        </button>
                </form>
```

comzeradd:
@pablocubico nice :100: 

I added a custom twig filter and changed the templates accordingly.

pablocubico:
Great, it looks neat! I left a comment about the `data:image/svg+xml...` backgrounds on the blocks plugin PR, the same applies here: do you think we can turn those into just `background: url(someimage.svg)` since they don't come from the image sprite? Is that possible? Maybe I'm missing something.

comzeradd:
Yes, good catch. I did that before I thought of using `mask-image` to achieve the desired color. Commit updated.

PR #674, opened 2018-12-20
Update circleci config

kirdia:
A followup for #673 
- Install repo's dependencies in circleci
- Use repo's phpcs version

Thoughts/Questions:
Phpunit version should be defined by repo's dependencies and not in ci config (leaving it as it is now for the moment)
Do we need the multi-site testing?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,8 +25,10 @@ job-references:
       sudo -E docker-php-ext-install mysqli
       sudo sh -c "printf '\ndeb http://ftp.us.debian.org/debian sid main\n' >> /etc/apt/sources.list"
       sudo apt-get update && sudo apt-get install mysql-client-5.7
-      sudo apt-get update && sudo apt-get install npm
+      curl -sL https://deb.nodesource.com/setup_10.x | sudo -E bash -
+      sudo apt-get update && sudo apt-get install build-essential nodejs
       sudo npm install -g stylelint eslint
+      composer install --dev
 
   php_job: &php_job
     environment:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -36,23 +36,28 @@ job-references:
       - checkout
       - run: *setup_environment
       - run: *install_dependencies
+      - run:
```
27Bslash6: Wouldn't it make more sense to install these dependencies in the `install_dependencies` mixin?

.circleci/config.yml
```diff
@@ -36,23 +36,28 @@ job-references:
...
+          name: "composer"
+          command: |
+            sudo mkdir ../../plugins
```
27Bslash6: What path does this resolve as, and why do you need to create it via sudo? 
You are `circleci` user at this point, and should be in the `/home/circleci/project` folder, so you're creating directories at `/` here, why?

.circleci/config.yml
```diff
@@ -36,23 +36,28 @@ job-references:
...
+            sudo chown circleci:circleci ../../plugins
```
27Bslash6: As above, what is this for?

.circleci/config.yml
```diff
@@ -36,23 +36,28 @@ job-references:
...
           when: always
           command: |
             mkdir -p /tmp/test-results/phpunit /tmp/test-results/phpunit-multisite
-            composer global require "phpunit/phpunit=5.7.*"
             rm -rf $WP_TESTS_DIR $WP_CORE_DIR
             bash bin/install-wp-tests.sh wordpress_test root '' 127.0.0.1 latest
-            phpunit --log-junit /tmp/test-results/phpunit/$CIRCLE_STAGE.xml
-            WP_MULTISITE=1 phpunit --log-junit /tmp/test-results/phpunit-multisite/$CIRCLE_STAGE.xml
+            vendor/bin/phpunit --log-junit /tmp/test-results/phpunit/$CIRCLE_STAGE.xml
+            WP_MULTISITE=1 vendor/bin/phpunit --log-junit /tmp/test-results/phpunit-multisite/$CIRCLE_STAGE.xml
```
27Bslash6: I think yes we could remove WP_MULTISITE testing for now. Perhaps another PR after this one builds successfully.

.circleci/config.yml
```diff
@@ -36,23 +36,28 @@ job-references:
       - checkout
       - run: *setup_environment
       - run: *install_dependencies
+      - run:
+          name: "composer"
+          command: |
+            sudo mkdir ../../plugins
+            sudo chown circleci:circleci ../../plugins
+            composer install --dev
+            composer require "phpunit/phpunit=5.7.*"
```
27Bslash6: Why not place the phpunit dependencies directly in `composer.json` instead of CI config?

.circleci/config.yml
kirdia: It needs sudo, because it creates a new folder in /home.
This comes from the master-theme composer config
https://github.com/greenpeace/planet4-master-theme/blob/develop/composer.json#L30
But I can't recall why we have introduced this installer path.

.circleci/config.yml
kirdia: Yes, that would be ideal.
I'll update the composer file

.circleci/config.yml
27Bslash6: Okay, this smells bad. Could you please test removing these lines and seeing what/if breaks?

.circleci/config.yml
27Bslash6: Thanks

PR #682, opened 2019-01-08
Planet 2561 - Campaign Creation

mehulkaklotar:
Ref: https://jira.greenpeace.org/browse/PLANET-2561
- New campaigns CPT
- Campaigns design templates
- CSS changes


assets/scss/campaigns/child/theme-oil.scss
```diff
@@ -0,0 +1,545 @@
...
+  .content-two-column-block {
+    @include content-two-column-block($anton);
+  }
+
+  .video-block {
+    @include video-block($oil-gradient-standard);
+  }
+
+  /** IE/Edge only **/
+  _:-ms-lang(x), .page-header {
```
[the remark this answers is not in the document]
mehulkaklotar: @comzeradd I have added ignore types for those 2 IE selectors in `.stylelintrc`

comzeradd:
Looking at the `mixins` file you added, I think that all of these can safely go to the `base/_mixins` file. I don't see any conflict. Some of those are nice, and can be used in the rest of the master-theme at some point. So no need to have a separate mixins partial.

*But* let's avoid adding a new set of media queries mixins. You can use the ones already [in place](https://github.com/greenpeace/planet4-master-theme/blob/planet-2561/assets/scss/base/_mixins.scss) and adjust your code to be mobile first.

functions.php
```diff
@@ -406,7 +406,6 @@ public function enqueue_public_assets() {
 		// CSS files.
 		wp_enqueue_style( 'bootstrap', 'https://cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/4.1.1/css/bootstrap.min.css', array(), '4.1.1' );
 		wp_enqueue_style( 'slick', 'https://cdnjs.cloudflare.com/ajax/libs/slick-carousel/1.9.0/slick.min.css', array(), '1.9.0' );
-		wp_enqueue_style( 'fork-awesome', 'https://cdnjs.cloudflare.com/ajax/libs/fork-awesome/1.1.1/css/fork-awesome.min.css', array(), '1.1.1' );
```
koyan: Why are you removing fork-awesome here?

classes/class-p4-campaigns.php
```diff
@@ -285,5 +293,137 @@ public function enqueue_admin_assets() {
...
+
+		/**
+		 * callback function for campaign page template selection
+		 *
+		 * @param $post
+		 */
+		public function campaign_page_templates_meta_box_callback( $post ) {
+
+			// Add a nonce field so we can check for it later.
+			wp_nonce_field( 'campaign_page_template_nonce', 'campaign_page_template_nonce' );
```
mikelittle: It's usually better to include something unique (e.g. 'edit_campaign-' . $post->ID) to this post in the nonce action. Otherwise, theoretically, this nonce could be used to modify a different post.

templates/single-campaigns.twig
```diff
@@ -0,0 +1,93 @@
...
+							{% endfor %}
+						</div>
+					{% endif %}
+				</div>
+				<h1 class="page-header-title">{{ post.title|raw }}</h1>
+				<div class="row">
+					<div class="col-md-6">
+						<div class="single-post-meta">
+							<address class="single-post-author">
+								{{ __( 'by', 'planet4-master-theme' ) }}
```
mikelittle: Constructing the 'by' line like this is not really translatable. For example if a particular language puts the equivalent of 'by' after the author name. It may be better to do a sprintf in the php and add it to the context.

templates/single-campaigns.twig
```diff
@@ -0,0 +1,93 @@
...
+								{% if not ( post.get_author_override ) %}
+									<a href="{{ post.author.path }}">{{ post.author.name }}</a>
+								{% else %}
+									{{ post.author.name }}
+								{% endif %}
+							</address>
+							<time class="single-post-time" pubdate>{{ post.post_date|date }}</time>
```
mikelittle: As far as I am aware, the 'pupdate' atribute was dropped from the HTML 5 spec in 2012 https://w3c.github.io/html/textlevel-semantics.html#the-time-element

templates/single-campaigns.twig
```diff
@@ -0,0 +1,93 @@
...
+							{% if ( show_comments ) %}
+								<span class="separator">|</span>
+								<span id="comments-link" class="comment-link">
+									<img src="{{ data_nav_bar.images }}speech_bubble.svg">
+									<a class="quantity"> {{ post_comments_count }} <span class="display-text">{{ __( 'Comments' ) }}</span> </a>
```
mikelittle: This link element has no 'href'. Is it a Javascript only thing? If so, that stops it being accessible. It could probably be a fragment link to an appropriate id in the comments template.

PR #690, opened 2019-01-16
PLANET-3069: The "Got It" button in P4 MENA - Cookies Pop-Up Banner

pablocubico:
This change fixes some styling issues in the MENA child theme (see: https://jira.greenpeace.org/browse/PLANET-3069), but also decouples / dedupes some duplicated code between the master theme and the blocks plugin.

I changed the name of the bottom cookie panel (the one with the "GOT IT" legend), to `cookie-notice`, to avoid confusion with the `cookies-block` from the Blocks plugin. 

So:
* cookies block -> the cookies block from the Blocks plugin
* cookie notice -> the bottom legend with the "GOT IT" button.

Notes:
- The styles for `.cookie-block` were living in the blocks plugin, those styles belong to the `cookie-notice`, so I moved them to the master theme.
- The "cookies block" class is actually `.cookies-block` ("cookies", in plural), so it was not affected by those styles.
- There is some JS in the blocks plugin which show/hide the `.cookie-notice` depending on the selection made inside the cookies block, see: https://github.com/greenpeace/planet4-plugin-blocks/pull/447


kirdia:
Looks good! :+1: 
Please fix the conflict.

pablocubico:
Done!

PR #697, opened 2019-01-21
PLANET-3088 - Posts: contain zoomed images on the current viewport

comzeradd:
This makes sure that we show the whole picture regardless of user's viewport (see screenshots below).

Also:
* Balance the close button, since the right attribute starts from the left part of the element.
* Reverse close button on RTL.

Ref: https://jira.greenpeace.org/browse/PLANET-3088


pablocubico:
@comzeradd I was just about to merge it but it has conflicts now, I guess from the CSS refactor. Please update and merge!

PR #702, opened 2019-01-22
PLANET-3059: Implement new gutter system on master-theme scss code

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-3059

Using variables instead of hardcoded values for margins and paddings.

Notes:
- `style.css` has not changed, the map did: this is not an error, it is actually what we want!
- I noticed there are `padding()` and `margin()` mixins, but I don't really see the benefit of using them, as they are too verbose compared to the shorthand syntax, and using the variables is enough for achieving consistency and flexibility. Using the mixins seems redundant, in my opinion.


tjerkheringa:
Hi Pablo,

that is quite a valid remark.
I added the option as an option.
It is up top GP whether this is aomething the want to use in future or not.

Kind regards,
Tjerk

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

________________________________

pablocubico:
@tjerkheringa Thanks Tjerk! We are definitely improving our SCSS code. 💪

pablocubico:
@comzeradd @tjerkheringa Rebased to latest develop to resolve conflicts.

PR #703, opened 2019-01-24
Added Custom Campaign Setting Metaboxes

onyekaa:
Set up styling options in the Campaign post types, including:

- Setting page header fonts and background color
- Setting primary and secondary button colors
- Setting body fonts.
- Setting campaign logo (e.g arctic, forests)
- Setting nav type (full/minimal with only logo) and background color based on preset palette.
- Setting logo color (black or white) depending on the nav background
- Most of the above are set using custom CSS injected into the head or by conditionally modifying the .twig files.

Pending:

- Setting the page header fonts.


onyekaa:
Updated formatting on the the files, deleted and squashed the old styling documents that were clogging up the history. 
@mikelittle  I addressed your comments.

templates/html-header.twig
```diff
@@ -85,10 +85,17 @@
         </script>
 	{% endif %}
 
-	<link href="https://fonts.googleapis.com/css?family=Roboto" rel="stylesheet">
+	<link href="https://fonts.googleapis.com/css?family=Lora|Roboto" rel="stylesheet">
 	<link rel="pingback" href="{{ site.pingback_url }}">
```
comzeradd: This line is probably redundant and should be completely removed. We already have these fonts on [scss code](https://github.com/greenpeace/planet4-master-theme/blob/develop/assets/scss/base/_fonts.scss).

single-campaigns.php
```diff
@@ -1,46 +1,106 @@
...
+$custom_styles['nav_type']             = $post->campaign_nav_type;
+$custom_styles['logo_color']           = $post->campaign_logo_color;
+$custom_styles['css']['header_bg']     = $post->campaign_header_color ? ".page-header { background: {$post->campaign_header_color} !important;}" : null;
+$custom_styles['css']['header_serif']  = $post->campaign_header_serif ? ".page-header { font-family: {$post->campaign_header_serif}!important;}" : null;
+$custom_styles['css']['header_sans']   = $post->campaign_header_sans ? ".page-header { font-family: {$post->campaign_header_sans} !important;}" : null;
+$custom_styles['css']['body_font']     = $post->campaign_body_font ? "body { font-family: '{$post->campaign_body_font}' !important;}" : null;
+$custom_styles['css']['btn_primary']   = $post->campaign_primary_color ? ".btn-primary { background: {$post->campaign_primary_color} !important; border-color: {$post->campaign_primary_color} }" : null;
+$custom_styles['css']['btn_secondary'] = $post->campaign_secondary_color ? ".btn-secondary { background: {$post->campaign_secondary_color} !important; border-color: {$post->campaign_primary_color} }" : null;
```
comzeradd: Shouldn't these style live inside a scss file instead?

assets/css/campaign.css
```diff
@@ -0,0 +1,34 @@
...
+    height: 30px !important;
+    width: 30px !important;
+    float: none;
+  }
+.palette-only .iris-palette-container {
+    position: relative !important;
+    left: 0 !important;
+    bottom: 0 !important;
+    padding: 8px 10px 4px;
+  }
```
comzeradd: Where is this css code being used?

comzeradd: Indentation seems of in the last two rules. Is this code here in order to adjust some part of the administration dashboard?

single-campaigns.php
onyekaa: Not sure if that's the way to go? These are meant to be individual overrides of each campaign theme's default styles. So the user can pick a different font or different nav colour from the post's edit page. Or what do you think?

assets/css/campaign.css
onyekaa: Fixed. Yes it's for the color picker in the dashboard. I was trying to disable a part of the color picker used in CMB2, but it inherits from Wordpress's default colorpicker and doesn't come with that setting by default, so I figured it would be less messy to just hide it and fix the appearance a bit.

classes/class-p4-campaigns.php
```diff
@@ -297,13 +296,57 @@ public function validate_page_types( $selected_page_types ) : bool {
...
-			if ( ! is_admin() || strpos( get_current_screen()->taxonomy, $this->taxonomy ) === false ) {
+			if ( ! is_admin() || strpos( get_current_screen()->post_type, $this->post_type ) === false ) {
 				return;
 			}
 			wp_register_script( $this->taxonomy, get_template_directory_uri() . "/assets/js/$this->taxonomy.js", [ 'jquery' ], null, true );
 			wp_localize_script( $this->taxonomy, 'localizations', $this->localizations );
 			wp_enqueue_script( $this->taxonomy );
 			wp_enqueue_media();
 		}
+
```
kirdia: Please remove this whole function.
The campaign post type is registered in P4_Post_Campaign class

classes/class-p4-campaigns.php
```diff
@@ -297,13 +296,57 @@ public function validate_page_types( $selected_page_types ) : bool {
 		 * Load assets.
 		 */
 		public function enqueue_admin_assets() {
-			if ( ! is_admin() || strpos( get_current_screen()->taxonomy, $this->taxonomy ) === false ) {
+			if ( ! is_admin() || strpos( get_current_screen()->post_type, $this->post_type ) === false ) {
```
kirdia: Please revert this. It was probably left from the first commits for the campaigns feature

single-campaigns.php
```diff
@@ -1,46 +1,108 @@
 <?php
 /**
- * The Template for displaying all single campaign posts
+ * Template Variables for Campaigns.
```
kirdia: This file seems that it was checked out earlier.
There are many changes in planet-2561 (that contains the 1st PR merged for the campaigns feature).
Please synchronize with the one currently on the branch.

58-105 lines were also removed from this template in that 1st PR. Please remove them from here also

single-campaigns.php
```diff
@@ -1,46 +1,108 @@
...
+ *
+ * @var P4_Post $post
+ * */
 $post            = Timber::query_post( false, 'P4_Post' );
 $context['post'] = $post;
 
+// Save custom style settings.
+$custom_styles = [];
+
+$custom_styles['css']['nav_color']     = $post->campaign_nav_color ? ".navbar { background-color: {$post->campaign_nav_color} !important;}" : null;
```
kirdia: I don't see any custom fields set for the custom post type. I guess this would be part of another PR?

single-campaigns.php
kirdia: I'm thinking about pre-builting the custom css file(s) (naming it per postid/postname) for each specific post/campaign  in a hook like 'save_post' and then enqueing it here. What do you think?

onyekaa: @kirdia Or I can make inline-classes for the options? It should be fine for the fonts but things like colors will be a lot of classes for the colours. Is that okay?

onyekaa: Actually @comzeradd  is probably right. I'll add these as variables to the SCSS. Would you prefer I make a new file or add to the existing ones?

kirdia: @onyekaa I did not get this. How would you pass the values to the css variables? Shouldn't these be customizable per campaign page?

single-campaigns.php
onyekaa: It was part of another branch, but I've merged into this one. I'm assuming you mean the metabox fields?

single-campaigns.php
onyekaa: Sorry mistake. I was suggesting making classes for the settings, so for instance, header serif will have a class of `.header-serif`, colors can uses SASS maps for all the color options like `.bg-green-light`. But that feels excessive. 

All the settings currently generate about 5 classes which I just put in a style tag in the head. For example:
`<style>
.navbar { background-color: #8cbf4f !important;}
.page-header { background: #ffffff !important;}
body { font-family: 'lora' !important;}
.btn-primary { background: #fcd00a !important; border-color: #fcd00a }
.btn-secondary { background: #ffffff !important; border-color: #fcd00a }		
</style>`

classes/class-p4-post-campaign.php
```diff
@@ -29,24 +29,25 @@ public function __construct() {
 		 * Class hooks.
 		 */
 		private function hooks() {
-
 			add_action( 'init', [ $this, 'register_campaigns_cpt' ] );
+			add_action( 'cmb2_admin_init', [ $this, 'register_campaigns_metaboxes' ] );
 			add_action( 'add_meta_boxes', [ $this, 'campaign_page_templates_meta_box' ] );
+			add_action( 'admin_enqueue_scripts', [ $this, 'enqueue_admin_assets' ] );
 			add_action( 'save_post_campaigns', [ $this, 'save_campaign_page_templates_meta_box_data' ] );
```
kirdia: You should change this hook `save_post_campaign`, since post type name has been changed

single-campaign.php
```diff
@@ -0,0 +1,66 @@
...
+
+// Get the cmb2 custom fields data
+// Articles block parameters to populate the articles block
+// p4_take_action_page parameter to populate the take action boxout block
+// Author override parameter. If this is set then the author profile section will not be displayed.
+$page_meta_data                 = get_post_meta( $post->ID );
+$campaign_template              = ! empty( $page_meta_data['_campaign_page_template'][0] ) ? $page_meta_data['_campaign_page_template'][0] : 'antarctic';
+$page_terms_data                = get_the_terms( $post, 'p4-page-type' );
+$context['background_image']    = $page_meta_data['p4_background_image_override'][0] ?? '';
+$take_action_page               = $page_meta_data['p4_take_action_page'][0] ?? '';
```
kirdia: Please remove this line

single-campaign.php
```diff
@@ -0,0 +1,66 @@
...
+$post->set_issues_links();
+
+
+// Get the cmb2 custom fields data
+// Articles block parameters to populate the articles block
+// p4_take_action_page parameter to populate the take action boxout block
+// Author override parameter. If this is set then the author profile section will not be displayed.
+$page_meta_data                 = get_post_meta( $post->ID );
+$campaign_template              = ! empty( $page_meta_data['_campaign_page_template'][0] ) ? $page_meta_data['_campaign_page_template'][0] : 'antarctic';
+$page_terms_data                = get_the_terms( $post, 'p4-page-type' );
```
kirdia: Please remove this line

single-campaign.php
```diff
@@ -0,0 +1,66 @@
...
+$context['background_image']    = $page_meta_data['p4_background_image_override'][0] ?? '';
+$take_action_page               = $page_meta_data['p4_take_action_page'][0] ?? '';
+$context['page_type']           = $page_terms_data[0]->name ?? '';
```
kirdia: Please remove this line and anything related with p4-page-type taxonomy and it's terms. They are not needed for this post type

PR #709, opened 2019-01-31
PLANET-3172 - Tags: Add option for a custom page redirection

comzeradd:
In order to keep the slug unchanged, instead of redirecting the Tag page fetches and renders the custom page content.

Ref: https://jira.greenpeace.org/browse/PLANET-3172

### DIFF
--- a/classes/class-p4-campaigns.php
+++ b/classes/class-p4-campaigns.php
@@ -84,8 +84,27 @@ if ( ! class_exists( 'P4_Campaigns' ) ) {
 				$happypoint_attachment_url   = $happypoint_image_attributes ? $happypoint_image_attributes[0] : '';
 
 				$happypoint_bg_opacity = get_term_meta( $wp_tag->term_id, 'happypoint_bg_opacity', true );
-				$happypoint_bg_opacity = $happypoint_bg_opacity ?? '30'; ?>
+				$happypoint_bg_opacity = $happypoint_bg_opacity ?? '30';
 
+				$redirect_page = get_term_meta( $wp_tag->term_id, 'redirect_page', true );
+				$dropdown_args = [
+					'show_option_none' => __( 'Select Page', 'planet4-master-theme-backend' ),
+					'hide_empty'       => 0,
+					'hierarchical'     => true,
+					'selected'         => $redirect_page,
+					'name'             => 'redirect_page',
+				];
+				?>
+
+				<tr class="form-field edit-wrap">
+					<th>
+						<label><?php echo __( 'Redirect Page', 'planet4-master-theme-backend' ); ?></label>
+					</th>
+					<td>
+						<?php wp_dropdown_pages( $dropdown_args ); ?>
+						<p class="description"><?php echo __( 'Leave this empty if you want to use the automated Tag page. Otherwise pick a page to redirect this Tag to.', 'planet4-master-theme-backend' ); ?></p>
+					</td>
+				</tr>
 				<tr>
 					<th colspan="2">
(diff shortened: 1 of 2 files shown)


tag.php
```diff
@@ -16,103 +16,152 @@
...
-	$tag_image_id                   = get_term_meta( $context['tag']->term_id, 'tag_attachment_id', true );
-
-	$context['og_description'] = $context['tag_description'];
-	if ( $tag_image_id ) {
-		$context['og_image_data'] = wp_get_attachment_image_src( $tag_image_id, 'full' );
-	}
+	$redirect_id = get_term_meta( $context['tag']->term_id, 'redirect_page', true );
+	if ( $redirect_id ) {
+
+		$redirect_page = get_post( $redirect_id );
```
kirdia: Could you try this? to avoid duplicating the page template (although it's hack also) 😰 
```
global $wp_query;
$wp_query->queried_object = $redirect_page;
$wp_query->queried_object_id = $redirect_page->ID;
include 'page.php';
```

comzeradd: I amended the commit with this suggestion and it seems to work fine. It's definitely more clean that the original approach.

comzeradd:
One of the tests keeps failing, but it doesn't seem related to this commit.

koyan:
The failing test is not caused by this PR, but by a webdriver issue. Fixed (or going around it) at https://github.com/greenpeace/planet4-base-fork/pull/16

classes/class-p4-campaigns.php
```diff
@@ -84,8 +84,27 @@ public function add_taxonomy_form_fields( $wp_tag ) {
 				$happypoint_attachment_url   = $happypoint_image_attributes ? $happypoint_image_attributes[0] : '';
 
 				$happypoint_bg_opacity = get_term_meta( $wp_tag->term_id, 'happypoint_bg_opacity', true );
-				$happypoint_bg_opacity = $happypoint_bg_opacity ?? '30'; ?>
+				$happypoint_bg_opacity = $happypoint_bg_opacity ?? '30';
 
+				$redirect_page = get_term_meta( $wp_tag->term_id, 'redirect_page', true );
+				$dropdown_args = array(
```
sagarsdeshmukh: Could you please use shorthand of array syntax here ([])?

tag.php
```diff
@@ -16,103 +16,116 @@
...
-			'engaging_network_id' => $options['engaging_network_form_id'] ?? '',
-			'opacity'             => $opacity,
-			'mailing_list_iframe' => 'true',
-		]
-	);
-
-	$campaign->view();
+		$templates = [ 'tag.twig', 'archive.twig', 'index.twig' ];
+
+
```
sagarsdeshmukh: please remove extra empty line.

sagarsdeshmukh:
Overall looks good to me except few small changes. 👍

comzeradd:
@sagarsdeshmukh thanks! commit amended.

classes/class-p4-campaigns.php
```diff
@@ -198,6 +230,11 @@ public function save_taxonomy_meta( $term_id ) {
 			if ( $this->validate( $happypoint_bg_opacity ) ) {
 				update_term_meta( $term_id, $field_id, $happypoint_bg_opacity );
 			}
+
+			$redirect_page = $_POST['redirect_page'] ?? 0;
+			if ( $this->validate_page_types( $redirect_page ) ) {
```
kirdia: Validation is useful here, but this would validate the input against the p4 page types.
You can validate if the passed id corresponds to an actual page (maybe with `get_post()`)

PR #711, opened 2019-02-07
PLANET-3022: MENA center search icon

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-3022


assets/scss/layout/_navbar.scss
```diff
@@ -466,7 +466,8 @@ $menu-height-large: 3.75rem;
 
     .icon {
       vertical-align: middle;
-      font-size: 1.3125rem;
+			font-size: 1.3125rem;
+			top: 0;
```
comzeradd: Indentation seems off here

PR #712, opened 2019-02-08
PLANET-3172 - Fix validation for tag redirection page

comzeradd:
As per @kirdia's [comment](https://github.com/greenpeace/planet4-master-theme/pull/709#discussion_r254691296).

### DIFF
--- a/classes/class-p4-campaigns.php
+++ b/classes/class-p4-campaigns.php
@@ -232,7 +232,7 @@ if ( ! class_exists( 'P4_Campaigns' ) ) {
 			}
 
 			$redirect_page = $_POST['redirect_page'] ?? 0;
-			if ( $this->validate_page_types( $redirect_page ) ) {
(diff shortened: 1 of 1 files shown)


classes/class-p4-campaigns.php
```diff
@@ -232,7 +232,7 @@ public function save_taxonomy_meta( $term_id ) {
 			}
 
 			$redirect_page = $_POST['redirect_page'] ?? 0;
-			if ( $this->validate_page_types( $redirect_page ) ) {
+			if ( get_post( $redirect_page ) ) {
```
kirdia: Looks good, but better check the return value

```suggestion
			if ( ! is_null( get_post( $redirect_page ) ) ) {
```

comzeradd:
@kirdia commit amended.

kirdia:
Good! thank you! :+1:

PR #715, opened 2019-02-12
PLANET-2364: Search: Don't display 'Show 5 more' button when there are less than 5 results

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-2364

Note: as we don't have lazy loading for all types of search, I consider this to be a temporary solution until we fix the TODO's in the current search code. That said, it works!.

### DIFF
--- a/classes/class-p4-search.php
+++ b/classes/class-p4-search.php
@@ -384,7 +384,7 @@ if ( ! class_exists( 'P4_Search' ) ) {
 
 			// This happens when we search for everything or when filtering for attachments, since WP_Query does not support searching within rich text documents.
 			if ( ! $this->search_query || ( isset( $args['post_type'] ) && 'attachment' !== $args['post_type'] ) ) {
-				if ( $posts ) {
+				if ( isset( $posts ) ) {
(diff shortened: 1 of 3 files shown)


classes/class-p4-search.php
```diff
@@ -384,7 +388,7 @@ protected function get_posts( $paged = 1 ) : array {
 
 			// This happens when we search for everything or when filtering for attachments, since WP_Query does not support searching within rich text documents.
 			if ( ! $this->search_query || ( isset( $args['post_type'] ) && 'attachment' !== $args['post_type'] ) ) {
-				if ( $posts ) {
+				if ( isset($posts) ) {
```
pablocubico: This change is unrelated to the issue, but the line was causing: `Notice: Undefined variable: posts in /app/source/public/wp-content/themes/planet4-master-theme/classes/class-p4-search.php on line 387`.

pablocubico:
Sorry about the force pushes, had to remove some debug code and make PHP Code Sniffer happy... 
`¯\_(ツ)_/¯`

PR #719, opened 2019-02-21
PLANET-3211: Sitemap: About Greenpeace section shows up only 5 pages

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-3211

### DIFF
--- a/classes/class-p4-sitemap.php
+++ b/classes/class-p4-sitemap.php
@@ -13,6 +13,11 @@ if ( ! class_exists( 'P4_Sitemap' ) ) {
 	class P4_Sitemap {
 
 		/**
+		 * Limit the number of evergreen pages
+		 */
+		const MAX_EVERGREEN_PAGES = 100;
+
+		/**
 		 * Gets data for the Action pages.
 		 *
 		 * @return array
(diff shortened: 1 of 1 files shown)


classes/class-p4-sitemap.php
```diff
@@ -112,6 +112,7 @@ public function get_evergreen_pages() : array {
 				'meta_key'         => '_wp_page_template',
 				'meta_value'       => 'page-templates/evergreen.php',
 				'suppress_filters' => false,
+				'numberposts'      => -1,
```
kirdia: It's good to be an upper limit on this (maybe 100 posts) just in case. You can use a constant for this like 
https://github.com/greenpeace/planet4-plugin-blocks/blob/develop/classes/controller/blocks/class-articles-controller.php#L32

Or even better retrieve only the fields needed for this (id, permalink) using fields query argument.
https://developer.wordpress.org/reference/classes/wp_query/parse_query/.
I know this is out of scope :) just a suggestion

pablocubico: Added the limit, as for the `fields` part, I tried adding a `fields` key to `args` but it didn't change anything, so I left the current query as is, as I assumed that will need refactoring the current code into using something different than `get_posts`. 🤔

kirdia: No problem, we can revisit later.
Thank you! :+1:

PR #727, opened 2019-03-01
PLANET-1068: Help button in admin that leads to GP specific content documentation

pablocubico:
See: https://jira.greenpeace.org/browse/PLANET-1068


comzeradd:
`phpcs` is happy again :+1: we have a unit test failing, but it doesn't seem relevant to this commit.

PR #732, opened 2019-03-12
PLANET-3267 Add Google Metadata on Planet4 websites

sagarsdeshmukh:
[JIRA # 3267](https://jira.greenpeace.org/browse/PLANET-3267)


classes/class-p4-master-site.php
```diff
@@ -1103,4 +1106,34 @@ public function add_help_sidebar() {
...
+
+		// Filter social menu links.
+		$footer_social_menu = wp_get_nav_menu_items( 'Footer Social' );
+		$social_menu_links  = [];
+
+		foreach ( $footer_social_menu as $menu_details ) {
+			$social_menu_links[] = $menu_details->url;
+		}
+
+		$p4_structured_data  = '<script type=\'application/ld+json\'>';
```
kirdia: @sagarsdeshmukh the string should not be constructed here. Please use a twig template, pass the needed variables to context and then use them to include or not the new template for the structured data.

classes/class-p4-master-site.php
```diff
@@ -1103,4 +1106,34 @@ public function add_help_sidebar() {
...
+	 */
+	public function generate_p4_structured_data() {
+		$p4_site_url   = get_site_url();
+		$p4_site_title = get_bloginfo( 'name' );
+
+		// Filter social menu links.
+		$footer_social_menu = wp_get_nav_menu_items( 'Footer Social' );
+		$social_menu_links  = [];
+
+		foreach ( $footer_social_menu as $menu_details ) {
```
kirdia: Please check also the failed unit tests, here should be a check that there are indeed menu items for example

sagarsdeshmukh:
@kirdia - Yes I agree with your comment. I updated the functionality accordingly. Please have a look.

templates/blocks/p4_structured_data.twig
```diff
@@ -0,0 +1,8 @@
+{% set social_menu_links = [] %}
+{% if footer_social_menu %}
+	{% for menu_details in footer_social_menu %}
+		{% set social_menu_links = social_menu_links|merge([menu_details.url]) %}
+	{% endfor %}
+{% endif %}
+<script type='application/ld+json'>{"@context":"http:\/\/schema.org","@type":"WebSite","url":"{{ site.link }}","name":"{{ site.name }}","alternateName":"Greenpeace","potentialAction":{"@type":"SearchAction","target":"{{ site.link }}/?s={search_term_string}&orderby=_score","query-input":"required name=search_term_string"}}</script>
```
kirdia: Backslashes are not needed here, `http://schema.org`

koyan: Since you want to parse Json, I would prefer if you created your json in a more structured and less error prone way, like: 
```
$arr = [
    '@context' => 'http://schema.org', 
    '@type' => 'Website', 
    'url' => site.link, 
    'd' => 4, '
    e' => 5
];

$myjson = json_encode($arr);
```

sagarsdeshmukh: The previous commit is amended. Please have a look.

templates/blocks/p4_structured_data.twig
```diff
@@ -0,0 +1,31 @@
...
+{% set website_info = {
+	'@context': 'http://schema.org',
+	'@type': 'WebSite',
+	'url': site.link,
+	'name': site.name,
+	'alternateName': 'Greenpeace',
+	'potentialAction': {
+		'@type': 'SearchAction',
+		'target': ( site.link ~ '/?s={search_term_string}&orderby=_score' ),
+		'query-input': 'required name=search_term_string',
```
angtheod: I think the {search_term_string} is a var defined in the USA website. We probably need to do
something similar like
```
{% set search_term_string = fn('get_search_query') %}
'target': ( site.link ~ '/?s=' ~ search_term_string ~ '&orderby=_score' ),
```
As for the `query-input` are you sure this is the correct value for it? Might need to check here https://schema.org/SearchAction

PR #736, opened 2019-03-18
PLANET-3268 object cache issue

angtheod:
Flush object cache after every page/post update (due to Nginx-Helper plugin 4-hour-caching not allowing editors to view their changes when not logged in).
Also, revert previous changes that did not help.

*Note*
Tried to delete only the nginx keys from cache but did not find a way to get multiple keys based on a phrase or wild card, and also the wp_cache_delete_group function works only if the group is empty... so...

### DIFF
--- a/classes/class-p4-master-site.php
+++ b/classes/class-p4-master-site.php
@@ -202,12 +202,27 @@ class P4_Master_Site extends TimberSite {
 	 * @param WP_Post $post_before Whether this is an existing post being updated or not.
 	 */
 	public function clean_post_cache( $post_id, $post_after, $post_before ) {
-
 		// Ignore autosave.
 		if ( defined( 'DOING_AUTOSAVE' ) && DOING_AUTOSAVE ) {
 			return;
 		}
-		clean_post_cache( $post_id );
+		// Lua Script.
+		$lua = <<<LUA
+local k =  0
+for i, name in ipairs(redis.call('KEYS', KEYS[1]))
+do
+    redis.call('DEL', name)
+    k = k+1
+end
+return k
+LUA;
+
+		$group = 'nginx-cache';
+		$redis = new Redis();
+		if ( $redis->connect( '127.0.0.1', 6379, 5 ) ) {
+			$redis->eval( $lua, [ $group . '*' ], 1 );
+			wp_cache_delete_group( $group );
+		}
 	}
 
 	/**
--- a/classes/class-p4-taxonomy-campaign.php
+++ b/classes/class-p4-taxonomy-campaign.php
@@ -6,7 +6,6 @@
  */
 
 use Timber\Timber;
-use Timber\Loader as TimberLoader;
 
 if ( ! class_exists( 'P4_Taxonomy_Campaign' ) ) {
 
@@ -66,7 +65,6 @@ if ( ! class_exists( 'P4_Taxonomy_Campaign' ) ) {
 		 * View the Campaign template.
 		 */
 		public function view() {
-			( new TimberLoader() )->clear_cache_timber();
 			Timber::render( $this->templates, $this->context );
 		}
 	}


27Bslash6:
@angtheod This is really not an ideal solution.  If my read is correct from https://developer.wordpress.org/reference/classes/WP_Object_Cache/flush/ , you're deleting the entire contents of the redis database on each post update, including full page cache, object, transient and as PHP is using redis for session data, that goes too?

The question though is under what circumstances does the FPC cache not get cleared?

It works for these cases:

On k8s.p4.greenpeace.org/flibble:
- edit homepage, update slider type > check in different browser, not loggged in > see changes
- edit post title > check diff browser > see changes on the homepage and on the post
- edit post content > homepage cache not affected (good) > check post in diff browser > see changes

It's not working for www.greenpeace.org/flibble:
- edit homepage, update slider type > check in different browser, not logged in > no change

First glance at redis-cli shows for www.greenpeace.org that ALL keys are stored as `nginx-cache:httpsGETmaster.k8s.p4.greenpeace.org/<path>`, yet Wordpress thinks the domain is www.greenpeace.org.  So it would seem there's a disconnect between what nginx is storing cache keys as and what Wordpress nginx-helper is trying purge.

This seems to have been newly introduced in one of the infrastructure updates. Unfortunately due to Akamai not providing CIDR ranges to trust upstream X_FORWARDED_ headers I can't apply a fix in nginx for this, so finding a solution in application is required.

The prefix in redis for full page caching is `nginx-cache:httpsGET<domain>/<path>` so there should be nothing stopping you from clearing the specific page from FPC on update, as linked in #p4-daily-standup https://chat.greenpeace.org/channel/p4-techtrack-dailystandup?msg=JamEGZcRyEC7XWj49

If you can do this in PHP then we can likely drop nginx-helper as it's only useful on non-prod sites.

27Bslash6:
Code used by nginx-helper for clearing wildcard redis keys: https://github.com/greenpeace/planet4-nginx-helper/blob/master/admin/class-phpredis-purger.php#L212-L221

27Bslash6:
My guess is if we switch to using the forked version of nginx-helper I've just created, and swith all instances of `$parse['host']` to use `$_SERVER['HTTP_X_FORWARDED_HOST']` instead then it will work.

@koyan How do we add packages to our wpackagist?

koyan:
I have added it to packagist (had to change its package name first): 
https://packagist.org/packages/greenpeace/planet4-nginx-helper#dev-feature/PLANET-3268-fix-server-host
I have also done the changes you suggested at a branch : feature/PLANET-3268-fix-server-host
You can add it to test it in your site's composer like this: 
"greenpeace/planet4-nginx-helper " : "dev-feature/PLANET-3268-fix-server-host"

angtheod:
@27Bslash6 
> The prefix in redis for full page caching is `nginx-cache:httpsGET<domain>/<path>` so there should be nothing stopping you from clearing the specific page from FPC on update, as linked in #p4-daily-standup https://chat.greenpeace.org/channel/p4-techtrack-dailystandup?msg=JamEGZcRyEC7XWj49
> 
> If you can do this in PHP then we can likely drop nginx-helper as it's only useful on non-prod sites.

The problem was not with clearing the specific page/post. It was with code generated pages. For example, if I add a new Tag to a Post and visit the relevant campaign page, this key will be stored
redis-nginx-keys

So, when you visit the campaign page, you do not see (inside the Covers Block) that an Action Page was added/removed. That was one of the issues that were reported.
And there is no cool way for me to know in what way the Page was updated so that I can form that key string. So, I was trying to delete the whole group with `wp_cache_delete_group` but this one works only if the group is empty of keys.

angtheod:
@koyan @27Bslash6 
tried it using nginx helper plugin functions and works fine
```
$prefix = $nginx_helper_admin->options['redis_prefix'];
( new PhpRedis_Purger() )->delete_keys_by_wildcard( $prefix . '*' );
wp_cache_delete_group( substr( $prefix, 0, -1 ) );```

Will  try now to do the same without using nginx helper plugin, so that we can drop it in future if we wish.

File: classes/class-p4-master-site.php [hunk 1]
```diff
@@ -202,12 +202,27 @@ public function set_featured_image( $post_id, $post, $update ) {
...
+do
+    redis.call('DEL', name)
+    k = k+1
+end
+return k
+LUA;
+
+		$group = 'nginx-cache';
+		$redis = new Redis();
+		if ( $redis->connect( '127.0.0.1', 6379, 5 ) ) {
```
angtheod: @27Bslash6 I know this won't work for docker compose (since the redis container has other IP).
Any ideas on this one?

27Bslash6:
> @koyan @27Bslash6
> tried it using nginx helper plugin functions and works fine
> 
> ```
> $prefix = $nginx_helper_admin->options['redis_prefix'];
> ( new PhpRedis_Purger() )->delete_keys_by_wildcard( $prefix . '*' );
> wp_cache_delete_group( substr( $prefix, 0, -1 ) );
> ```
> Will try now to do the same without using nginx helper plugin, so that we can drop it in future if we wish.

Again, this is deleting EVERYTHING in the full page cache, which is a bit overkill.

classes/class-p4-master-site.php
27Bslash6: The redis server is configured in multiple places: it's a global variable in wp-config.php ( https://github.com/greenpeace/planet4-docker/blob/master/src/planet-4-151612/wordpress/wp-config.php.tmpl#L83 ), as a database option `rt_wp_nginx_helper_options`, and according to https://github.com/greenpeace/planet4-nginx-helper/blob/feature/PLANET-3268-fix-server-host/admin/class-nginx-helper-admin.php#L305 it's defined as a constant.

27Bslash6:
Confirmed that after updating the `$parse['host']` variable, cache purges are once again successful on

- https://k8s.p4.greenpeace.org/flibble 
- https://www.greenpeace.org/flibble
- https://release.k8s.greenpeace.org/flibble

Tag pages are purged for newly added tags, though it doesn't clear _removed_ tag cache keys (obviously), they will clear in the cache TTL. This is before and after adding 'coal' to a post which already included 'climate':

image

Also confirming the top 'Purge cache' button clears all keys successfully.

I'm going to modify base-fork to use the forked plugin and redeploy all sites prior to release. Unfortunately until we can trust X_FORWARDED_FOR headers this is the best we can do.

angtheod:
@27Bslash6 
> Tag pages are purged for newly added tags, though it doesn't clear _removed_ tag cache keys (obviously), they will clear in the cache TTL.

Problem is that the TTL is 4 hours and for these 4 hours there is confusing (kinda) info because the actual page cache key is removed but the tag page cache key is not. If we can live with that, then it is totally fine by me.

page without climate tag

climate tag page

PR #739, opened 2019-03-25
Planet-3256: Add "Open in a new tab" for Page Header Link fields

dpivo:
Checkbox added to header to choose whether to open header link in new tab or not.

Ticket requested to add checkbox alongside Header Button Title but that wasn't feasible so it was added below.

Ref: https://jira.greenpeace.org/browse/PLANET-3256

### DIFF
--- a/templates/blocks/header.twig
+++ b/templates/blocks/header.twig
@@ -43,7 +43,12 @@
 			{% if ( header_button_title and header_button_link ) %}
 				<div class="row">
 					<div class="col-md-4">
-						<a href="{{ header_button_link }}" class="btn btn-primary btn-medium page-header-btn">{{ header_button_title }}</a>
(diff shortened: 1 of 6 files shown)


templates/blocks/header.twig
```diff
@@ -43,7 +43,12 @@
 			{% if ( header_button_title and header_button_link ) %}
 				<div class="row">
 					<div class="col-md-4">
-						<a href="{{ header_button_link }}" class="btn btn-primary btn-medium page-header-btn">{{ header_button_title }}</a>
+						<p>{{ header_button_link_checkbox }}</p>
+						<a href="{{ header_button_link }}"
+						    {% if ( header_button_link_checkbox ) %}
```
sagarsdeshmukh: please remove spaces from line indentation.(use tab instead)

PR #740, opened 2019-03-26
PLANET-3311: Improve twig rendering times in search results

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-3311

### DIFF
--- a/classes/class-p4-search.php
+++ b/classes/class-p4-search.php
@@ -102,6 +102,7 @@ if ( ! class_exists( 'P4_Search' ) ) {
 		 * Initialize the class. Hook necessary actions and filters.
 		 */
 		protected function initialize() {
+			$this->start = TimberHelper::start_timer();
 			$this->localizations = [
 				// The ajaxurl variable is a global js variable defined by WP itself but only for the WP admin
 				// For the frontend we need to define it ourselves and pass it to js.
@@ -737,7 +738,8 @@ if ( ! class_exists( 'P4_Search' ) ) {
 		 * View the Search page template.
 		 */
 		public function view() {
-			Timber::render( $this->templates, $this->context );
+			Timber::render( $this->templates, $this->context, 600 );
+			echo '<!-- Timber timer ' . TimberHelper::stop_timer($this->start) . ' -->';
 		}
 
 		/**
--- a/classes/class-p4-search.php
+++ b/classes/class-p4-search.php
@@ -102,7 +102,6 @@ if ( ! class_exists( 'P4_Search' ) ) {
 		 * Initialize the class. Hook necessary actions and filters.
 		 */
 		protected function initialize() {
-			$this->start = TimberHelper::start_timer();
 			$this->localizations = [
 				// The ajaxurl variable is a global js variable defined by WP itself but only for the WP admin
 				// For the frontend we need to define it ourselves and pass it to js.
@@ -738,8 +737,7 @@ if ( ! class_exists( 'P4_Search' ) ) {
 		 * View the Search page template.
 		 */
 		public function view() {
-			Timber::render( $this->templates, $this->context, 600 );
-			echo '<!-- Timber timer ' . TimberHelper::stop_timer($this->start) . ' -->';
+			Timber::render( $this->templates, $this->context, self::DEFAULT_CACHE_TTL );
 		}
 
 		/**
@@ -760,7 +758,7 @@ if ( ! class_exists( 'P4_Search' ) ) {
 					} else {
 						$paged_context['first_of_the_page'] = false;
 					}
-					Timber::render( [ 'tease-search.twig' ], $paged_context );
+					Timber::render( [ 'tease-search.twig' ], $paged_context, self::DEFAULT_CACHE_TTL );
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


pablocubico:
@angtheod @27Bslash6 @kirdia I'm looking for opinions on this subject.

I tested a few methods to improve the render time on the first render of the search template (details on the JIRA issue). The best results were achieved by adding a expiration time for the render method of Timber so it caches anything it can cache. 

I also changed `preview()` to `post_excerpt` in the `tease-search` template which seems to be faster. 

For some reason I can't read the XDebug cache dumps on my PHPStorm installation, so I could only do some benchmarking using Timber's timer functions. 

This is of course not addressing the "300 results on the first load issue", which I assume is needed for the rendering of the counters on the filters bar (the count for matches in each filter). So I guess it will not be a big improvement, but using Timber's cache seems like a good idea in general.

I don't know if there is another reason than the filter counters for loading more than the first 5 results in the first render, I guess Angelos know more of the history of this feature, and I think he mentioned he found some improvements for that query in the p4-dev channel. 

Another way to accomplish that would be to get the filter counters asynchronously after the first load, by hitting a dedicated endpoint via JS after loading the first results. 

Thoughts? Ideas? Comments?

27Bslash6:
What are the numbers for the changes on seaches with 300 hits? Can you confirm post edits are reflected in search output successfully, ie no additional caching problems are introduced by these changes?

If we can be confident that Timber is properly hashing the context and renewing its cache on changed content then yes, I think this is something that we can look into delivering not only here but in other controllers.

pablocubico:
Ok, after some testing with the Canada database and trying a few methods from the Timber docs, I think this is a improvement.

The rendering times went down from:
- **First render**: from **~1.50** to **~0.24**
- **Ajax renders**: from **~0.35** to **~0.09** on each "load more" call.

Apart from adding a third argument with the default TTL for the `render` methods, I tried several other things, none of them seem to have a noticeable impact on the rendering times, either alone or combined with the current method:

- Using the Twig `cache` block (the one Ray tested on a different PR).
- Caching the collection of `TimberPosts` in the search class:

```php
protected function get_timber_posts( $paged = 1 ) : array {
  $timber_posts = TimberHelper::transient( 'search_results', function() use ($paged) {
    $cached_timber_posts = [];
    $posts = $this->get_posts( $paged );
    // Use Timber's Post instead of WP_Post so that we can make use of Timber within the template.
    if ( $posts ) {
      foreach ( $posts as $post ) {
        $cached_timber_posts[] = new TimberPost( $post->ID );
      }
      return $cached_timber_posts;
    } else {
      return [];
    }
  }, self::DEFAULT_CACHE_TTL);

  return (array) $timber_posts;
}
```

- Adding a global Twig cache in `functions.php`: 
```php
if ( class_exists( 'Timber' ) ){
  Timber::$cache = true;
}
```

This last setting is recommended for production in the Timber docs, it creates a `cache` folder within `plugins/timber-library`,  which stores the compiled Twig-to-PHP code. Sounds like something we could try, but it would require some process to ensure any changes to the Twig templates are there after deploying (for example, clearing the cache folder on deploy).

As far as I tested, Timber picks up the changes to content, so the hashing seems to be smart enough to handle changes. I tested disabling the `wp_cache_get/add` methods on the class and any changes to the content were instantly there. 

Do note that this is only dealing with the rendering of templates, and it can be quite cumbersome to debug since we have a few caching layers working at the same time, I think we should document this in a dedicated **Performance/Caching** section on the Handbook.

I guess we could put this on a live site like `koyansync` in case you want to test further.

templates/tease-search.twig
```diff
@@ -64,7 +64,7 @@
 				</div>
 			</div>
 
-			<p class="search-result-item-content">{{ post.preview().read_more('')|excerpt( 25 )|raw }}</p>
+			<p class="search-result-item-content">{{ post.post_excerpt|excerpt( 25 )|e('wp_kses_post')|raw }}</p>
```
angtheod: We need to check if this re-introduces that weird bug we used to have in the past when the actual excerpt was less than 25 words. The `.read_more('')` was added to solve that issue which caused the next search results to appear misaligned with broken markup.

classes/class-p4-search.php
```diff
@@ -737,7 +737,7 @@ public function add_load_more( $args = null ) {
 		 * View the Search page template.
 		 */
 		public function view() {
-			Timber::render( $this->templates, $this->context );
+			Timber::render( $this->templates, $this->context, self::DEFAULT_CACHE_TTL );
```
angtheod: Wanted to add this as well since we do not have FPC for search results, but I had not tested it...

I think we could also add the 4th arg to set cache_mode as `CACHE_OBJECT`.
If we don't then it tries to `get_transient( $key )` which does check if an external object cache is being used in which case leads again to checking redis cache instead of going for sql. So, I think we can directly go for checking redis db by setting the 4th arg. What you think?

pablocubico: Ahhh great, I wasn't sure about that, I'll add it and test.

templates/tease-search.twig
pablocubico: Oh, didn't know about that, will check it further then.

pablocubico:
@angtheod Added CACHE_OBJECT, I think it works even faster now by a few milliseconds. I restored the original `post_preview()` call since it works best for all post types and it doesn't seem to have a noticeable impact, `post_excerpt` returned a few empty excerpts.

PR #742, opened 2019-03-26
PLANET-3324 Move campaign fields (dataLayer) to the P4 page backend

sagarsdeshmukh:
- Added new option(block) on P4 pages backend for campaign fields
- All fields are optional (no mandatory validation)
- The dataLayer push code is added in HTML `<head>` before GTM code
- The old dataLayer push code is removed from EN form block in separate PR
image

### DIFF
--- a/classes/class-p4-master-site.php
+++ b/classes/class-p4-master-site.php
@@ -792,6 +792,101 @@ class P4_Master_Site extends TimberSite {
 				'preview_size' => 'large',
 			]
 		);
+
+		// P4 Datalayer/Campaign fields.
+		$p4_campaign_fields = new_cmb2_box(
+			[
+				'id'           => $prefix . 'campaign_fields',
+				'title'        => __( 'Campaign information (dataLayer)', 'planet4-master-theme-backend' ),
+				'object_types' => [ 'page' ],
+				'closed'       => true,  // Keep the metabox closed by default.
+			]
+		);
+
+		$campaign_options = [
+			0                                    => __( '- Select Campaign -', 'planet4-master-theme-backend' ),
+			'Local Campaign'                     => 'Local Campaign',
+			'Ocean Sanctuaries'                  => 'Ocean Sanctuaries',
+			'Climate Emergency Response'         => 'Climate Emergency Response',
+			'All Eyes on the Amazon'             => 'All Eyes on the Amazon',
+			'Asia Energy Transition'             => 'Asia Energy Transition',
+			'BrAndino: Hold the Line'            => 'BrAndino: Hold the Line',
+			'Break Free'                         => 'Break Free',
+			'Climate Justice Liability'          => 'Climate Justice Liability',
+			'Congo Basin Forests'                => 'Congo Basin Forests',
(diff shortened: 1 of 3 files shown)


templates/html-header.twig
```diff
@@ -56,14 +56,28 @@
           }
 
           if ( google_tag_value ) {
-			window.dataLayer = window.dataLayer || [];
+			var dataLayer = window.dataLayer = window.dataLayer || [];
```
angtheod: is this intended? I mean can't it be `var dataLayer = window.dataLayer || [];`

angtheod:
@sagarsdeshmukh Just checking... in an Issue page like `/explore/energy/`, I get these values. Is everything correct here? The pageType in this case should have the Issue (category)?
Also, there are escaped characters.

PLANET-3324

templates/html-header.twig
```diff
@@ -56,14 +56,28 @@
...
 				'signedIn' : 'false',
 				'visitorType' : '',
 				'userID' : '',
 				'post_tags': '{{ post_tags }}'
 			});
+
+			var cf_campaign_name = '{% if cf_campaign_name is defined and cf_campaign_name is not null %}{{ cf_campaign_name }}{% endif %}';
+			var cf_basket_name = '{% if cf_basket_name is defined and cf_basket_name is not null %}{{ cf_basket_name }}{% endif %}';
+			var cf_scope = '{% if cf_scope is defined and cf_scope is not null %}{{ cf_scope }}{% endif %}';
+			var cf_department = '{% if cf_department is defined and cf_department is not null %}{{ cf_department }}{% endif %}';
```
angtheod: please align the equal signs here, to make it easier to read

templates/html-header.twig
sagarsdeshmukh: Yes it is intended.(https://www.bounteous.com/insights/2016/03/21/gtm-data-layer-best-practices/#proper-datalayer-instantiation)
Julia also recommended to use it as `var dataLayer = window.dataLayer = window.dataLayer || [];`

sagarsdeshmukh:
> @sagarsdeshmukh Just checking... in an Issue page like `/explore/energy/`, I get these values. Is everything correct here? The pageType in this case should have the Issue (category)?
> Also, there are escaped characters.
> 
> PLANET-3324

Indeed there is a problem. (Julia already opened a ticket for [same](https://jira.greenpeace.org/browse/PLANET-3355))

classes/class-p4-master-site.php
```diff
@@ -792,6 +792,101 @@ public function register_header_metabox() {
...
+			[
+				'id'           => $prefix . 'campaign_fields',
+				'title'        => __( 'Campaign information (dataLayer)', 'planet4-master-theme-backend' ),
+				'object_types' => [ 'page' ],
+				'closed'       => true,  // Keep the metabox closed by default.
+			]
+		);
+
+		$campaign_options = [
+			0                                    => __( '- Select Campaign -', 'planet4-engagingnetworks' ),
```
angtheod: please change this also to `planet4-master-theme-backend`,
same for lines 846, 867

PR #746, opened 2019-03-29
PLANET-3329 - Add cookie to store last visited NRO

comzeradd:


pablocubico:
@comzeradd Just tested locally and the cookie is created. I am using Canada's database, so the site shows a Language Selector page when visiting `https://www.planet4.test`, the cookie value is set to:

`https://www.planet4.test/en//`, it includes the language and a trailing slash, maybe we could use PHP's native `parseurl` like:

`data-nro="{{ fn('parse_url', site.link ).host }}"` (output: `www.planet4.test`)

To strip the protocol, language and pretty much everything else.

Other than that, looks good to me. 👍

comzeradd:
We need that information (protocol and language), because in the end the landing page needs the final url to redirect to. So for instance, we need to know that it's the french version of the canada site.

I amended the commit to remove the duplicate trailing slash.

PR #750, opened 2019-04-04
PLANET-3419 - Enable Timber template caching

27Bslash6:
Sets `Timber::$cache` to the inverse of `WP_DEBUG`

Enables Timber .twig > .php blob template caching for all non-development environments.

See: https://github.com/greenpeace/planet4-docker/blob/master/src/planet-4-151612/wordpress/wp-config.php.tmpl#L60

If APP_ENV matches `develop` then WP_DEBUG is true, therefore template caching is disabled and edits _should_ be immediately visible without cache clear.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -24,6 +24,13 @@ if ( ! class_exists( 'Timber' ) ) {
 	);
 
 	return;
+} else {
+	# Enable Timber template cache unless this is a debug environment
+	if (defined('WP_DEBUG') && is_bool(WP_DEBUG)) {
+		Timber::$cache=!WP_DEBUG;
+	} else {
+		Timber::$cache=true;
+	}
 }
 
 require_once( __DIR__ . '/classes/class-p4-master-site.php' );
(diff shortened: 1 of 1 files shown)


pablocubico:
Hey @27Bslash6  I tried this before for the search results, but couldn't really measure the performance impact locally. 

Just a note: this adds a `cache` folder inside `plugins/timber-library`, do we need to clear that on production? maybe flushing the cache or something to make sure it picks up any change in the Twig files?

27Bslash6:
No there's no need to flush changes. Without any additional considerations, the containers will be deployed with blank timber cache folders and recreate them on first use.

Secondly, the helm charts have been adapted to mount this folder as tmpfs - a ramdrive which persists only as long as the container lifetime: https://github.com/greenpeace/planet4-helm-wordpress/blob/master/templates/deployment-php.yaml#L49

While you may not have been able to measure a difference, it's a recommended step for production anyway - there's no need for twig to recreate the php logic from twig on every pageview.

PR #751, opened 2019-04-04
Add Campaign type and export functionality

comzeradd:


templates/html-header.twig
```diff
@@ -101,10 +101,14 @@
...
 	{% include 'blocks/p4_structured_data.twig' %}
 
-	<link href="https://fonts.googleapis.com/css?family=Roboto" rel="stylesheet">
 	<link rel="pingback" href="{{ site.pingback_url }}">
 	<meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no">
 	<link rel="shortcut icon" type="image/ico" href="{{ theme.uri }}/favicon.ico"/>
 	<link rel="dns-prefetch" href="//act.greenpeace.org">
 	{{ wp_head }}
+	{% if custom_styles %}
+		<style>
```
kirdia: Just for keeping with standards we can add type attribute here
`type="text/css"`

templates/html-header.twig
```diff
@@ -101,10 +101,14 @@
...
+		{% for style in custom_styles.css %}{{ style|raw }}{% endfor %}
```
kirdia: slight indentation fix needed here

single-campaign.php
```diff
@@ -0,0 +1,83 @@
...
+$custom_styles['css']['header_serif']            = $post->campaign_header_serif ? " .page-header { font-family: {$post->campaign_header_serif}!important;}" : null;
+$custom_styles['css']['header_sans']             = $post->campaign_header_sans ? " .page-header { font-family: {$post->campaign_header_sans} !important;}" : null;
+$custom_styles['css']['body_font']               = $post->campaign_body_font ? " body, p { font-family: '{$post->campaign_body_font}' !important;}" : null;
+$custom_styles['css']['btn_primary']             = $post->campaign_primary_color ? " .btn-primary { background: {$post->campaign_primary_color} !important; border-color: {$post->campaign_primary_color} !important;}" : null;
+$custom_styles['css']['btn_secondary']           = $post->campaign_secondary_color ? " .btn-secondary:hover { background: $post->campaign_secondary_color !important; border-color: {$post->campaign_primary_color}; }" : null;
+$custom_styles['css']['anchor']                  = $post->campaign_secondary_color ? " a { color: {$post->campaign_secondary_color } !important; }" : null;
```
kirdia: not a big issue, but the rules here override wp admin's rules

` a { color: #25784f !important; }		`

image

templates/html-header.twig
kirdia: Also these custom styles are not added when previewing blocks, so the blocks are previewed using the 'casual' styling

single-campaign.php
```diff
@@ -0,0 +1,83 @@
...
+$custom_styles['campaign_logo']                  = $post->campaign_logo ?? null;
+
+
+// Set Navigation Issues links.
+$post->set_issues_links();
```
kirdia: this is not needed, can be removed

single-campaign.php
```diff
@@ -0,0 +1,83 @@
...
+
+$header_font_style = [
+	'Montserrat'       => "font-family: 'Montserrat' !important; font-weight: 900 !important;",
+	'Montserrat_Light' => "font-family: 'Montserrat' !important; font-weight: 500 !important;",
+];
+
+if ( $pf && array_key_exists( $pf, $header_font_style ) ) {
+	$header_font = $header_font_style[ $pf ];
+} else {
+	$header_font = `font-family: {$pf} !important;`;
```
kirdia: I get a phpcs warning for this
`use of the backtick operator is forbidden`

PR #759, opened 2019-04-10
PLANET-3425 Hide the campaigns menu item until it is ready to be shown

koyan:
As per product owners requirements, the campaigns functionality should be hidden until it is ready to be delivered.

### DIFF
--- a/classes/class-p4-post-campaign.php
+++ b/classes/class-p4-post-campaign.php
@@ -79,7 +79,7 @@ if ( ! class_exists( 'P4_Post_Campaign' ) ) {
 				'public'             => true,
 				'publicly_queryable' => true,
 				'show_ui'            => true,
-				'show_in_menu'       => true,
+				'show_in_menu'       => false,
 				'query_var'          => true,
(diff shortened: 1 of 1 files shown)


classes/class-p4-post-campaign.php
```diff
@@ -400,5 +401,14 @@ public function enqueue_admin_assets() {
 			wp_register_style( 'cmb-style', get_template_directory_uri() . '/assets/css/campaign.css' );
 			wp_enqueue_style( 'cmb-style' );
 		}
+
+		/**
+		 * Hide the menu item until the product owner is happy to share it.
+		 * Related Jira issue: PLANET-3425
+		 */
+		public function remove_campaigns_menu() {
+		    remove_menu_page( 'edit.php?post_type=campaign' );
```
comzeradd: It seems you have mixed indentation here. Both tabs and spaces.

koyan: This code is not there anymore, as I followed the method that Kyriakos suggested

PR #762, opened 2019-04-11
PLANET-3390 Add prepare-commit-msg git hook

kirdia:


.githooks/prepare-commit-msg
```diff
@@ -0,0 +1,33 @@
...
+if [ -z "$BRANCHES_TO_SKIP" ]; then
+  BRANCHES_TO_SKIP=(master develop)
+fi
+
+
+BRANCH_NAME=$(git symbolic-ref --short HEAD)
+PLANET_BRANCH_NAME=$(git symbolic-ref --short HEAD | grep -Po "(PLANET|planet)-[0-9]*" |  tr '[:lower:]' '[:upper:]')
+BRANCH_NAME="${BRANCH_NAME##*/}"
+BRANCH_EXCLUDED=$(printf "%s\n" "${BRANCHES_TO_SKIP[@]}" | grep -c "^$BRANCH_NAME$")
+ISSUE_IN_COMMIT=$(grep -cP "(PLANET|planet)-[0-9]*" $1)
```
27Bslash6:
```
In .githooks/prepare-commit-msg line 23:
ISSUE_IN_COMMIT=$(grep -cP "(PLANET|planet)-[0-9]*" $1)
                                                    ^-- SC2086: Double quote to prevent globbing and word splitting.
```

.githooks/prepare-commit-msg
```diff
@@ -0,0 +1,33 @@
...
+
+# If planet-xxx is not included in the branch name or the commit message quit.
+# If it is included in the branch name and not in the commit message, prepend it to the commit message.
+if ! [[ -z "$PLANET_BRANCH_NAME" ]] && ! [[ $BRANCH_EXCLUDED -eq 1 ]] && ! [[ $ISSUE_IN_COMMIT -ge 1 ]]; then
+  sed -i.bak -e "1s/^/$PLANET_BRANCH_NAME /" $1;
```
27Bslash6:
```
In .githooks/prepare-commit-msg line 28:
  sed -i.bak -e "1s/^/$PLANET_BRANCH_NAME /" $1;
                                             ^-- SC2086: Double quote to prevent globbing and word splitting.
```

.githooks/prepare-commit-msg
```diff
@@ -0,0 +1,33 @@
...
+
+BRANCH_NAME=$(git symbolic-ref --short HEAD)
+PLANET_BRANCH_NAME=$(git symbolic-ref --short HEAD | grep -Po "(PLANET|planet)-[0-9]*" |  tr '[:lower:]' '[:upper:]')
+BRANCH_NAME="${BRANCH_NAME##*/}"
+BRANCH_EXCLUDED=$(printf "%s\n" "${BRANCHES_TO_SKIP[@]}" | grep -c "^$BRANCH_NAME$")
+ISSUE_IN_COMMIT=$(grep -cP "(PLANET|planet)-[0-9]*" $1)
+
+# If planet-xxx is not included in the branch name or the commit message quit.
+# If it is included in the branch name and not in the commit message, prepend it to the commit message.
+if ! [[ -z "$PLANET_BRANCH_NAME" ]] && ! [[ $BRANCH_EXCLUDED -eq 1 ]] && ! [[ $ISSUE_IN_COMMIT -ge 1 ]]; then
```
27Bslash6: Prefer `-n` to `! -z` here

.githooks/prepare-commit-msg
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
```
27Bslash6: Please add exit on error and pipefail options.

`set -eo pipefail`

.githooks/prepare-commit-msg
27Bslash6: Again, please rewrite to use cross-platform grep pattern matching

.githooks/prepare-commit-msg
```diff
@@ -0,0 +1,33 @@
...
+# If planet-xxx is not included in the branch name or the commit message quit.
+# If it is included in the branch name and not in the commit message, prepend it to the commit message.
+if ! [[ -z "$PLANET_BRANCH_NAME" ]] && ! [[ $BRANCH_EXCLUDED -eq 1 ]] && ! [[ $ISSUE_IN_COMMIT -ge 1 ]]; then
+  sed -i.bak -e "1s/^/$PLANET_BRANCH_NAME /" $1;
+elif [[ -z "$PLANET_BRANCH_NAME" ]] && [[ $ISSUE_IN_COMMIT -eq 0 ]]; then
+  echo "Error! Issue number is not part of the branch name or the commit message.";
+  echo "Commit message should start with PLANET-xxx or branch name should contain planet-xxx or PLANET-xxx.";
+  exit 1;
+fi
```
27Bslash6: There's a lot of if conditions here which could probably be simplified, for example instead of an elif, test for the conditions in line 29 first and then use a separate if statement afterwards.  This reduces [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) and increases maintanability.

27Bslash6:
Also, do all developers consistently use gulp while working? If so perhaps that would be a good place to hook the pre-commit hook into the .git/hooks folder? Else making explicit in documentation that something like https://github.com/greenpeace/planet4-builder/blob/master/Makefile#L65 this needs to be performed manually in order for the pre-commit hook to be wired in.

comzeradd:
Could this also be part of the `make dev` target? Since it's an initial step all developers take in order to setup the repos properly.

kirdia:
@27Bslash6 did most of the changes.
did not add exit on error and pipefail, it conflicts with the grep matching for the variables assignments.

kirdia:
> Could this also be part of the `make dev` target? Since it's an initial step all developers take in order to setup the repos properly.

Yes, I can add in docker-compose repo

.githooks/prepare-commit-msg
```diff
@@ -0,0 +1,48 @@
...
+PLANET_BRANCH_NAME=$(echo "$BRANCH_NAME" | grep -oE "(PLANET|planet)-[0-9]*" | tr '[:lower:]' '[:upper:]')
+
+# Is current branch in excluded list?
+BRANCH_EXCLUDED=$(printf "%s\\n" "${BRANCHES_TO_SKIP[@]}" | grep -c "^$BRANCH_NAME$")
+
+# Check if planet4 branch exists in commit message.
+ISSUE_IN_COMMIT=$(grep -cE "(PLANET|planet)-[0-9]*" "$1")
+
+# If current branch is in the excluded list, nothing to do here, exit successfully.
+if [[ $BRANCH_EXCLUDED -eq 1 ]]; then
```
27Bslash6: Test exit codes directly instead of assigning them to a variable that isn't used elsewhere, ie removing line 25 and replacing this conditional with:

```
if printf "%s\\n" "${BRANCHES_TO_SKIP[@]}" | grep -q "^$BRANCH_NAME$"
then
  exit 0
fi
```

Tests in `if` statements don't cause set -e to exit the script.  grep returns 0 (success) if it matches and 1 if it fails, so you don't need to count number of hits just use the return code.

Then we can turn on `set -eo pipefail` https://explainshell.com/explain?cmd=set+-euxo%20pipefail with confidence.

.githooks/prepare-commit-msg
```diff
@@ -0,0 +1,48 @@
...
+  exit 0;
+fi
+
+# If commit message already includes planet4 issue, nothing to do here, exit successfully.
+if [[ $ISSUE_IN_COMMIT -ge 1 ]]; then
```
27Bslash6: As above.

27Bslash6:
> @27Bslash6 did most of the changes.
> did not add exit on error and pipefail, it conflicts with the grep matching for the variables assignments.

This indicates we're not done refactoring. See: https://gist.github.com/outro56/4a2403ae8fefdeb832a5

```
Skip [[ ... ]] in your if-expression if you can test for exit code instead.
```

.githooks/prepare-commit-msg
27Bslash6: Could (and should) also add `set -u`, but that introduces another syntax change which I'll leave as an exercise for the intrepid.

kirdia: I used variables to make the code more readable, it gives a hint to whoever reading it about the pipeline.

Replaced it with awk.

27Bslash6: The number of hits is irrelevant, the test is only for whether the string is present at all, and this is where using the exit code from `grep -q` is the correct tool. Adding more complexity with awk is unwarranted.

https://github.com/progrium/bashstyle/blob/master/README.md#best-practices-and-tips
"Skip [[ ... ]] in your if-expression if you can test for exit code instead."

I don't think it improves readability more than a comment, and adding variables that are only used as a test once is an anti-pattern.

koyan:
Closed by mistake

.githooks/prepare-commit-msg
```diff
@@ -0,0 +1,53 @@
...
+BRANCH_NAME=$(git symbolic-ref --short HEAD)
+
+# Get planet4 branch.
+PLANET_BRANCH_NAME=$(echo "$BRANCH_NAME" | awk 'match($0,/(PLANET|planet)-[0-9]*/) {print toupper(substr($0,RSTART,RLENGTH))}')
+
+# Is current branch in excluded list?
+BRANCH_EXCLUDED=$(printf "%s\\n" "${BRANCHES_TO_SKIP[@]}" | awk -v branch=$BRANCH_NAME "BEGIN{count=0} /^branch$/ {count++} END{print count}")
+
+# Check if planet4 branch exists in commit message.
+ISSUE_IN_COMMIT=$(grep -cE "(PLANET|planet)-[0-9]*" "$1" || true)
```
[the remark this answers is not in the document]
koyan: This has now been done differently

PR #763, opened 2019-04-11
PLANET-3412 fixed languague selection vertical bars

dpivo:
Fixed language selection vertical bar appearing after last language selection

ref: https://jira.greenpeace.org/browse/PLANET-3412


comzeradd:
I'm probably missing something, but how is this solving the problem reported in the screenshot. As far as I can the problem is caused because the `#country-list` is inside the wpml menu, so the second language is no longer the last child in the div.

See the language selector on Belgium for instance: https://www.greenpeace.org/belgium/fr/

Screenshot from 2019-04-11 16-39-11

dpivo:
> I'm probably missing something, but how is this solving the problem reported in the screenshot. As far as I can the problem is caused because the `#country-list` is inside the wpml menu, so the second language is no longer the last child in the div.
> 
> See the language selector on Belgium for instance: https://www.greenpeace.org/belgium/fr/
> 
> Screenshot from 2019-04-11 16-39-11

Yes that was the problem, but it only has to be the last child inside the <li></li> not the <div>. So the changes forces the countries_html to be appended inside the first <li></li> with the dropdown button instead of into the wpml-menu <li></li>

comzeradd:
Ah got it, thanks :+1:

PR #770, opened 2019-04-18
PLANET-3435 Restructure P4_Master_Site class code

sagarsdeshmukh:
[Jira link 3435](https://jira.greenpeace.org/browse/PLANET-3435)


classes/class-p4-register-metabox.php
```diff
@@ -0,0 +1,704 @@
+<?php
+/**
+ * P4 Register Metabox Class
+ *
+ * @package P4MT
+ */
+
+if ( ! class_exists( 'P4_Register_Metabox' ) ) {
```
kirdia: I think it is time to get rid of these checks since autoloading is handled by master_site class.
It adds extra indentation without any reason.

Also class names should be nouns to define better their existance. A better name would be P4_Metabox_Register or P4_Metabox_Registrator

classes/class-p4-register-metabox.php
```diff
@@ -0,0 +1,704 @@
...
+			static $added = false;
+
+			// Only add this to the page once (not for every metabox).
+			if ( $added ) {
+				return;
+			}
+
+			$added = true;
+			?>
+			<script type="text/javascript">
```
angtheod: I think it'd be better if we put this in a js file and use `src=""` attribute.

classes/class-p4-register-metabox.php
sagarsdeshmukh: yes It adds extra indentation, removed the `class_exists()` check and renamed the class name to `P4_Metabox_Register`

classes/class-p4-register-metabox.php
sagarsdeshmukh: Good idea, I moved the js code to separate file and include it.

admin/js/cmb2_form_validation.js
```diff
@@ -0,0 +1,74 @@
+jQuery(document).ready(function($) {
+
+  $form       = $( document.getElementById( 'post' ) );
+  $htmlbody   = $( 'html, body' );
+  $toValidate = $( '[data-validation]' );
```
angtheod: please use `let` to define these variables

classes/class-p4-metabox-register.php
```diff
@@ -0,0 +1,625 @@
...
+
+		if ( 0 !== absint( $parent_act_id ) ) {
+			$take_action_pages_args = [
+				'post_type'        => 'page',
+				'post_parent'      => $parent_act_id,
+				'post_status'      => 'publish',
+				'orderby'          => 'menu_order',
+				'order'            => 'ASC',
+				'suppress_filters' => false,
+				'numberposts'      => -1,
```
angtheod: could you please use here a big enough integer in order to follow WP coding standards?

classes/class-p4-metabox-register.php
```diff
@@ -0,0 +1,625 @@
...
+		$this->register_meta_box_header();
+		$this->register_meta_box_post();
+		$this->register_meta_box_open_graph();
+		$this->register_meta_box_campaign();
+	}
+
+	/**
+	 * Register Page Header meta box.
+	 */
+	public function register_meta_box_header() {
```
angtheod: Even though you created the new functions, you forgot to remove their code from the big one, so this one
is still 300 lines long.

sagarsdeshmukh: Ops, It was my miss :( 
Thanks for pointing it out.

sagarsdeshmukh:
While merging this PR, need to "**Squash and merge**".

classes/class-p4-metabox-register.php
```diff
@@ -0,0 +1,420 @@
...
+	 */
+	public function cmb2_after_form_do_js_validation( $post_id, $cmb ) {
+		static $added = false;
+
+		// Only add this to the page once (not for every metabox).
+		if ( $added ) {
+			return;
+		}
+
+		$added = true;
```
angtheod: Could use [wp_script_is](https://codex.wordpress.org/Function_Reference/wp_script_is) to check if it is already loaded
`if ( ! wp_script_is( 'cmb2_form_validation' ) ) {` but if that works as expected then I guess its ok.

sagarsdeshmukh: If we don't have [this](https://github.com/greenpeace/planet4-master-theme/pull/770/files#diff-03bce058cc5b9903fb8e557c6e4c4ff0R413) check then it would be the good fit.

PR #778, opened 2019-04-30
PLANET-2668: Automate testing of new versions of plugins using RIPPs on the cloud

pablocubico:
Add a config entry to automate the test of the master theme in RIPS.

Ref: https://jira.greenpeace.org/browse/PLANET-2668

See the related PR on the p4-builder repo: https://github.com/greenpeace/planet4-builder/pull/27

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,8 +12,6 @@ workflows:
       - rips-test:
           context: rips-scans
           filters:
-            branches:
-              ignore: languages
             tags:
               only: /^v\p{Digit}+\.\p{Digit}+\.\p{Digit}+.*/
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,6 +12,8 @@ workflows:
       - rips-test:
           context: rips-scans
           filters:
+            branches:
+              ignore: /.*/
             tags:
               only: /^v\p{Digit}+\.\p{Digit}+\.\p{Digit}+.*/
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,6 +9,14 @@ workflows:
           filters:
             branches:
               ignore: languages
+      - rips-test:
+          context: rips-scans
+          filters:
+            branches:
+              ignore: languages
+            tags:
+              only: /^v\p{Digit}+\.\p{Digit}+\.\p{Digit}+.*/
+
 version: 2
 
 job-references:
@@ -136,3 +144,12 @@ jobs:
           name: Build - Notify failure
           when: on_fail
           command: notify-job-failure.sh
+
+  rips-test:
+    docker:
+      - image: rips/rips-cli:circleci
+    environment:
+      RIPS_APPLICATION_ID: 32
+    steps:
+      - checkout
+      - run: rips-cli -vvv rips:scan:start -a $RIPS_APPLICATION_ID -p ~/project/ -t 1


.circleci/config.yml
```diff
@@ -136,3 +144,17 @@ jobs:
...
+
+  rips-test:
+    docker:
+      - image: gcr.io/planet-4-151612/p4-builder:latest
+    working_directory:  /home/circleci/
+    environment:
+      RIPS_SCAN_REPO: planet4-master-theme
+      RIPS_SCAN_APP_ID: 32
+    steps:
+      - setup_remote_docker:
```
27Bslash6: This job doesn't require a docker machine, I believe. Removing this line will save up to a minute of waiting for a docker machine to be provisioned.

.circleci/config.yml
```diff
@@ -9,6 +9,14 @@ workflows:
           filters:
             branches:
               ignore: languages
+      - rips-test:
+          context: org-global
```
27Bslash6: Is the global context necessary for this workflow? What variables are required?

27Bslash6:
Can be set mergable after https://github.com/greenpeace/planet4-builder/pull/27 is merged to at least develop and CI passes in this repo.

.circleci/config.yml
pablocubico: I defined three variables on the CircleCI job for the `planet4-master-theme`, but I'm not sure if that's the "global" context or not, the variables are: `RIPS_BASE_URI`, `RIPS_USERNAME` and `RIPS_PASSWORD`, which are required for `rips-cli` to authenticate.

image

Will the variables still be available if we remove this line?

27Bslash6: I think a new context specifically for RIPS test jobs might be the way to go - https://circleci.com/gh/organizations/greenpeace/settings#contexts

Adding these vars to a new context and assign that instead of org-global will reduce the number of secrets we're unnecessarily exposing to environments where they are not required.

pablocubico:
@27Bslash6 updated, docker line removed, env variables gone, using a new `rips-scans` context instead of `org-global`.

27Bslash6:
>Heya pablo...

so I've had a thought.
There's really nothing in the rips scan script which requires any dependencies from either the base or the builder images, is there?

And rips-phar only needs PHP
I'm now thinking we don't put this in builder at all
in CI, we will always have the variables:
```
  CIRCLE_PROJECT_REPONAME=planet4-kube-hunter
  CIRCLE_PROJECT_USERNAME=greenpeace

```
so replace the master-theme rips job with something like:
```
  rips-test:
    docker:
      - image: circleci/php:7.3
    environment:
      RIPS_APPLICATION_ID: 42
    steps:
      - run:
          name: Scan Planet4 Master Theme using RIPS
          command: |
            ZIPBALL_URL=$(curl -s "https://api.github.com/repos/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}/tags" | jq -r '.[0].zipball_url')
            wget "${ZIPBALL_URL}" -O "${CIRCLE_PROJECT_REPONAME}".zip

            wget https://github.com/rips/rips-cli/releases/download/1.2.1/rips-cli.phar -O ./rips-cli
            chmod 755 ./rips-cli

            ./rips-cli rips:scan:start -a "${RIPS_APPLICATION_ID}" -p "${CIRCLE_PROJECT_REPONAME}.zip"


```
Using one of Circle's pre-baked images instead of builder will save between 30s and 2 minutes waiting for it to download the image

Because if we're not using any of the tools in base or builder, and we're not storing the rips-cli.phar file during image creation, there's really no compelling reason to include this in builder at all

To get tricky about it, maybe define a new var in the RIPS context `RIPS_VERSION=1.2.1`
and use that in the wget line then we only need to update that context var and all jobs will immediately use the latest version

.circleci/config.yml
pablocubico: Using the new `rips-scans` context.

27Bslash6:
Great solution, well found.

Is there junit output for the tests, so we can see a summary of issues?  https://circleci.com/docs/2.0/collect-test-data/

pablocubico:
@27Bslash6 there is an "export" command on the CLI, but it does not export in JUnit format. It supports PDF, CSV and "JiraCSV" (?) formats. Maybe we can store one of those as artifacts.

https://kb.ripstech.com/doc/2.0/tooling/rips-cli

27Bslash6:
Okay, well I'll sign off on the CI code and config as is - tests are successfully run, and the contest approach is good. In future we should probably switch to Circle orbs for re-use later, as it's only one 

I think merge approval should come from @greenpeace/planet-4-developers in general though for business rules - I don't think limiting the workflow to only running on tagged releases is enough, this won't catch errors introduced in development until AFTER you've pushed a release, which means it's possible these errors will be picked up by running containers.

pablocubico:
@27Bslash6 thanks Ray! 

Then I'll leave you and @koyan to decide on the tag limitation. Of course there is a failing check which is the `rips-test` itself. Other than that, this is ready to merge I guess.

koyan:
The functionality we wanted works fine. We will merge it and start working on the issues reported by RIPPS (I have already opened tickets).

PR #780, opened 2019-04-30
Planet 2658 Campaign Export and Import with attachments

sagarsdeshmukh:
[JIRA 2658](https://jira.greenpeace.org/browse/PLANET-2658)

On export of campaign, export all attachment data in XML file.

On campaign import, import attachments and update all source image ids with newly imported attachment ids in post content.

The `planet-2658` branch is pinned to `koyansync` and `campaign` instances.

Example -
Exported campaign from `campaigns` instance :
https://k8s.p4.greenpeace.org/campaigns/campaign/test-campaign-export/
Imported campaign on `koyansync` instance:
https://k8s.p4.greenpeace.org/koyansync/campaign/test-campaign-export/

### DIFF
--- a/exporter-helper.php
+++ b/exporter-helper.php
@@ -8,45 +8,53 @@
 /**
  * Returns all attachment ids from campaign post content.
  *
- * @param integer $post_ids Post ID.
+ * @param string $post_ids Post IDs.
+ * @return string  $post_ids Post IDs.
  */
 function get_campaign_attachments( $post_ids ) {
 
 	global $wpdb;
 
+	// phpcs:disable
 	$attachments = $wpdb->get_results( "SELECT ID, guid, post_parent FROM {$wpdb->posts} WHERE post_type = 'attachment'", OBJECT_K );
+	// phpcs:enable
+
 	if ( empty( $attachments ) ) {
 		return $post_ids;
 	}
 
-	$ids = array();
+	$ids = [];
 
 	/**
 	 * Post thumbnails
 	 */
 	if ( $post_ids ) {
+		// phpcs:disable
 		$ids = $wpdb->get_col( sprintf( "SELECT meta_value FROM {$wpdb->postmeta} WHERE meta_key = '_thumbnail_id' AND post_id IN(%s)", implode( ',', $post_ids ) ) );
+		// phpcs:enable
 	}
 
 	/**
 	 * Uploaded to (post_parent)
 	 */
 	foreach ( $attachments as $id => $att ) {
-		if ( in_array( $att->post_parent, $post_ids ) ) {
+		if ( in_array( $att->post_parent, $post_ids, true ) ) {
 			$ids[] = $id;
 		}
 	}
 
-	foreach ( $wpdb->get_col( "SELECT post_content FROM {$wpdb->posts} WHERE ID IN( " . implode( ',', $post_ids ) . " ) AND post_content REGEXP '((wp-image-|wp-att-)[0-9][0-9]*)|\\\[gallery |href=|src='" ) as $text ) {
+	// phpcs:disable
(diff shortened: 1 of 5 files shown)


exporter-helper.php
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * Post Exporter Helper Function
+ *
+ * @package P4MT
+ */
+
+/**
+ * Returns all attachment ids from campaign post content.
+ *
+ * @param string $post_ids Post IDs.
```
angtheod: please correct doc comment since $post_ids is an array

functions.php
```diff
@@ -42,6 +42,7 @@ function( $template ) {
 		'P4_Campaigns',
 		'P4_Post_Campaign',
 		'P4_Campaign_Exporter',
+		'P4_Campaign_Importer',
```
angtheod: Do we have to load this always? This way it will always create this object. Can we instantiate this only when needed?

angtheod:
Overall looks good, just couple comments.

functions.php
angtheod: We should do this also for other classes as well (P4_Post_Report_Controller, P4_Control_Panel, etc) and make sure they load only in wp admin section and if possible only in the specific pages that are actually needed.

I will create a ticket again for this (since this is still relevant after all).

angtheod:
Getting a few notices and warning in Edit Campaign page
```
Notice: Trying to get property of non-object in /app/source/public/wp-content/plugins/planet4-plugin-blocks/classes/controller/blocks/class-splittwocolumns-controller.php on line 295

Notice: Trying to get property of non-object in /app/source/public/wp-content/plugins/planet4-plugin-blocks/classes/controller/blocks/class-splittwocolumns-controller.php on line 295

Warning: DOMDocument::loadHTML(): htmlParseEntityRef: expecting ';' in Entity, line: 6 in /app/source/public/wp-content/plugins/planet4-plugin-blocks/classes/controller/blocks/class-submenu-controller.php on line 202

Warning: DOMDocument::loadHTML(): htmlParseEntityRef: expecting ';' in Entity, line: 6 in /app/source/public/wp-content/plugins/planet4-plugin-blocks/classes/controller/blocks/class-submenu-controller.php on line 202

Warning: DOMDocument::loadHTML(): htmlParseEntityRef: expecting ';' in Entity, line: 6 in /app/source/public/wp-content/plugins/planet4-plugin-blocks/classes/controller/blocks/class-submenu-controller.php on line 202

Notice: Undefined variable: tags in /app/source/public/wp-content/plugins/planet4-plugin-blocks/classes/controller/blocks/class-takeactionboxout-controller.php on line 125```

File: functions.php [hunk 3]
```diff
@@ -35,18 +35,30 @@ function( $template ) {
...
+	'P4_Settings',
+	'P4_Control_Panel',
+	'P4_Post_Report_Controller',
+	'P4_Cookies',
+	'P4_Dev_Report',
+];
+
+// Load `P4_Campaign_Exporter` class on admin campaign listing page and campaign export only.
+if ( is_admin() && ( 'campaign' === filter_input( INPUT_GET, 'post_type', FILTER_SANITIZE_STRING ) || 'export_data' === filter_input( INPUT_GET, 'action', FILTER_SANITIZE_STRING ) ) ) {
+	array_splice( $services, 4, 0, 'P4_Campaign_Exporter' );
```
angtheod: Do we need to add them in specific order? Can we just `$services[] = 'P4_Campaign_Exporter'` for simplicity?
Also, please add both in 1 `if ( is_admin() )` since we will be adding more classes under this `if` check.

sagarsdeshmukh: I am not sure about the order of classes will impact or not, I will test and confirm it.

PR #786, opened 2019-05-03
PLANET-2402 Use wp api load promise to initiliaze post report js

kirdia:
Use wp api load promise to wait for wp backbone js client to load, before initializing post report backbone views/collections.

### DIFF
--- a/assets/js/posts_report.js
+++ b/assets/js/posts_report.js
@@ -1,3 +1,5 @@
+/* global jQuery, wp */
+
 var postCollection,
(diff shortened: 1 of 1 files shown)


angtheod:
@kirdia Hmm looks like there are 18 failing tests. Is this caused by another issue?

koyan:
> @kirdia Hmm looks like there are 18 failing tests. Is this caused by another issue?

There was a problem on Friday, but it was solved. Rerunning the tests.

PR #789, opened 2019-05-07
PLANET-3355 datalayer-page-type

angtheod:
Customize GTM Data Layer pageType value depending on the current page.


sagarsdeshmukh:
@angtheod - On my local the sitemap page is showing blank `'pageType'` value, does it need to show `'Default Page'`?
Remaining all looks good 💯

sagarsdeshmukh:
Just for note: On `search page` the `pageType` value is blank.

sagarsdeshmukh:
@angtheod - The `[404 page](https://github.com/greenpeace/planet4-master-theme/blob/develop/404.php#L18)` is using the old `page_category` var for `page_type`.

sagarsdeshmukh:
The Post page is showing `page_type` value above post title. looks like `page_type` variable used on other part of `single.twig` file.
image

single.php
```diff
@@ -32,7 +32,7 @@
 $page_terms_data                = get_the_terms( $post, 'p4-page-type' );
 $context['background_image']    = $page_meta_data['p4_background_image_override'][0] ?? '';
 $take_action_page               = $page_meta_data['p4_take_action_page'][0] ?? '';
-$context['page_type']           = $page_terms_data[0]->name ?? '';
+$context['page_type']           = 'Post Page';
 $context['page_term_id']        = $page_terms_data[0]->term_id ?? '';
 $context['page_category']       = $category->name ?? __( 'Post page', 'planet4-master-theme' );
```
sagarsdeshmukh: As we are using new `page_type` var for datalayer, so we don't need `page_category` var anymore. right?

angtheod: Changed it back to page_category to avoid name collision.

PR #791, opened 2019-05-09
PLANET-3517 Code cleanup: Sanitize form input, fix phpcs errors

sagarsdeshmukh:


classes/class-p4-campaign-exporter.php
```diff
@@ -29,9 +29,9 @@ public function single_post_export_data() {
 			if ( ! ( isset( $_GET['post'] ) || isset( $_POST['post'] ) || ( isset( $_REQUEST['action'] ) ) ) ) {
 				wp_die( 'No post to export has been supplied!' );
 			}
-			$post_id = ( isset( $_GET['post'] ) ? $_GET['post'] : $_POST['post'] );
+			$post_id = ( isset( $_GET['post'] ) ? filter_input( INPUT_GET, 'post', FILTER_SANITIZE_NUMBER_INT ) : filter_input( INPUT_POST, 'post', FILTER_SANITIZE_NUMBER_INT ) );
```
angtheod: No need to check if 'post' key isset since we are going to use filter funciton to access it.
Is it coming from a GET or a POST request? It may come from both? Then we can use INPUT_REQUEST to make sure we get it right. And we can even remove the first `if` statement.
`$post_id = filter_input( INPUT_REQUEST, 'post', FILTER_SANITIZE_NUMBER_INT );`
```
if ( ! empty( $post_id ) && isset( $_REQUEST['action'] ) ) {
	include get_template_directory() . '/exporter.php';
} else {
	wp_die( 'No post to export has been supplied!' );
}
```

classes/class-p4-campaign-exporter.php
```diff
@@ -101,7 +99,7 @@ public function add_import_button() {
 			?>
 			<script>
 				jQuery(function(){
-					jQuery("body.post-type-campaign .wrap .page-title-action").after('<a href="admin.php?import=wordpress" class="page-title-action"><?php _e( 'Import', 'planet4-master-theme-backend' ); ?></a>');
+					jQuery("body.post-type-campaign .wrap .page-title-action").after('<a href="admin.php?import=wordpress" class="page-title-action"><?php esc_attr_e( 'Import', 'planet4-master-theme-backend' ); ?></a>');
```
angtheod: Here you probably want `esc_html_e()` instead of `esc_attr_e()` since we are not echoing within an attribute.

classes/class-p4-campaign-exporter.php
```diff
@@ -41,12 +41,11 @@ public function single_post_export_data() {
...
 		 */
 		public function single_post_export_bulk() {
-			if ( ! empty( $_GET['post_type'] ) &&
-			'campaign' == $_GET['post_type'] &&
+			if ( 'campaign' === filter_input( INPUT_GET, 'post_type', FILTER_SANITIZE_STRING ) &&
 			current_user_can( 'edit_posts' ) ) { ?>
 			<script type="text/javascript">
 				jQuery(document).ready(function () {
-					jQuery('<option>').val('export').text('<?php _e( 'Export', 'planet4-master-theme-backend' ); ?>').appendTo("select[name='action']");
+					jQuery('<option>').val('export').text('<?php esc_attr_e( 'Export', 'planet4-master-theme-backend' ); ?>').appendTo("select[name='action']");
```
angtheod: same here

classes/class-p4-campaign-exporter.php
```diff
@@ -41,12 +41,11 @@ public function single_post_export_data() {
 		 * Export multiple data
 		 */
 		public function single_post_export_bulk() {
-			if ( ! empty( $_GET['post_type'] ) &&
-			'campaign' == $_GET['post_type'] &&
+			if ( 'campaign' === filter_input( INPUT_GET, 'post_type', FILTER_SANITIZE_STRING ) &&
```
angtheod: You can use here the const that I added [here](https://github.com/greenpeace/planet4-master-theme/blob/planet-3517/classes/class-p4-post-campaign.php#L17)
`P4_Post_Campaign::POST_TYPE`

classes/class-p4-campaign-exporter.php
```diff
@@ -60,18 +59,18 @@ public function single_post_export_bulk_action() {
...
 			$action          = $wp_list_table->current_action();
 			$allowed_actions = array( 'export' );
-			if ( ! in_array( $action, $allowed_actions ) ) {
+			if ( ! in_array( $action, $allowed_actions, true ) ) {
 				return false;
 			}
 			switch ( $action ) {
 				case 'export':
-					$sendback = 'admin.php?action=export_data&post=' . join( ',', $_REQUEST['post'] );
+					$sendback = 'admin.php?action=export_data&post=' . join( ',', array_map( 'absint', $_REQUEST['post'] ) );
```
angtheod: should use `filter_input()` here as well

classes/class-p4-campaign-exporter.php
sagarsdeshmukh: I was also looking for `INPUT_REQUEST` type. but I didn't found it in available type options in documentation.(https://www.php.net/manual/en/function.filter-input.php).

classes/class-p4-campaign-exporter.php
sagarsdeshmukh: I tried it but the `post` param is a array and filter_input function isn't provide the filtered array data.

classes/class-p4-campaign-exporter.php
angtheod: Oh ok, I saw it is defined (in php7.3 at least) and thought we could use it.
But still no need to check if isset since the filter_input function does that anyway.
We could also validate the 'action' key if we know what value it must have.

classes/class-p4-campaign-exporter.php
```diff
@@ -82,8 +81,7 @@ public function single_post_export_bulk_action() {
 		 * @param object $post Post object.
 		 */
 		public function single_post_export( $actions, $post ) {
-			if ( ! empty( $_GET['post_type'] ) &&
-				'campaign' == $_GET['post_type'] &&
+			if ( 'campaign' === filter_input( INPUT_GET, 'post_type', FILTER_SANITIZE_STRING ) &&
```
angtheod: same here for 'campaign'

classes/class-p4-campaign-exporter.php
```diff
@@ -60,18 +59,18 @@ public function single_post_export_bulk_action() {
...
 				return false;
 			}
 			switch ( $action ) {
 				case 'export':
-					$sendback = 'admin.php?action=export_data&post=' . join( ',', $_REQUEST['post'] );
+					$sendback = 'admin.php?action=export_data&post=' . join( ',', array_map( 'absint', $_REQUEST['post'] ) );
 					break;
 
 				default:
 					return;
```
angtheod: Not part of the PR, but please make it return false here (like in line 63), so that inspection does not complain that there is no returning value.

classes/class-p4-campaign-exporter.php
```diff
@@ -82,8 +81,7 @@ public function single_post_export_bulk_action() {
 		 * @param object $post Post object.
```
angtheod: Again not part of this PR, but please fix doc comments. $actions is `array`, $post is `WP_Post` and add @return array
and `public function single_post_export( $actions, $post ) : array {` in line 83

classes/class-p4-campaign-exporter.php
angtheod: In this case I think it might be better to use [array_filter](https://php.net/manual/en/function.array-filter.php) to validate each one of the post ids and exclude those who are not valid integers.
Can you try something like this?
```
$validated_posts = array_filter(
	$_REQUEST['post'],
	function ( $element ) {
		return filter_var( $element, FILTER_VALIDATE_INT );
	}
);
```

sagarsdeshmukh: looks perfect 👍

PR #792, opened 2019-05-09
PLANET-3520 Code cleanup: fix uncaught Exception

sagarsdeshmukh:
[JIRA Ticket 3520](https://jira.greenpeace.org/browse/PLANET-3520)


footer.php
```diff
@@ -10,8 +10,12 @@
  */
 
 $timber_context = $GLOBALS['timberContext'];
-if ( ! isset( $timber_context ) ) {
-	throw new \Exception( 'Timber context not set in footer.' );
+try {
```
kirdia: Do we use that file anywhere? @angtheod any thoughts?

tests/bootstrap.php
```diff
@@ -9,9 +9,12 @@
 if ( ! $_tests_dir ) {
 	$_tests_dir = '/tmp/wordpress-tests-lib';
 }
-
-if ( ! file_exists( $_tests_dir . '/includes/functions.php' ) ) {
-	throw new Exception( "Could not find $_tests_dir/includes/functions.php, have you run bin/install-wp-tests.sh ?" );
+try {
+	if ( ! file_exists( $_tests_dir . '/includes/functions.php' ) ) {
+		throw new Exception( "Could not find $_tests_dir/includes/functions.php, have you run bin/install-wp-tests.sh ?\n" );
```
kirdia: You can stop execution here without needing to catch the exception

footer.php
sagarsdeshmukh: I didn't found this file used somewhere.

kirdia: Can we remove this and the page-plugin twig template?
@angtheod do you recall, this comes from timber starter theme?

angtheod: @kirdia Yes, page-plugin.twig comes from Timber. We are not really using those files `header.php footer.php page-plugin.twig` since we are using Timber.

sagarsdeshmukh: As we are not using `header.php footer.php page-plugin.twig` files, I am removing them.

PR #794, opened 2019-05-09
PLANET-3320: Add a new dataLayer field (Goal) on Form Block backend

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-3320

### DIFF
--- a/classes/class-p4-metabox-register.php
+++ b/classes/class-p4-metabox-register.php
@@ -359,6 +359,26 @@ class P4_Metabox_Register {
 			]
 		);
 
+		$goal_options = [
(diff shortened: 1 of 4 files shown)


sagarsdeshmukh:
@pablocubico - If I am not wrong then this data layer push is intend to happen on [EN form submit](https://github.com/greenpeace/planet4-plugin-engagingnetworks/blob/develop/assets/js/partials/enform_submit.js#L165) and not on page load.

I was waiting for Jira to check ticket details, that's why I was late to review the PR. sorry :(

pablocubico:
@sagarsdeshmukh ohhhhh good catch, I didn't catch that part from the ticket description. Now I get it. I'll send another fix. Thanks!

PR #805, opened 2019-05-15
PLANET-3293 Use the field comment_author instead of comment.author.name

koyan:
Since the comment.author.name returns always the original author we change it to use the field comment_author instead of comment.author.name


koyan:
Before the fix, the reattributed comment still shows the original author: 
3293-before

After the fix:
3293-after

PR #806, opened 2019-05-17
PLANET-1884 Rename description of settings fields

dpivo:
- Settings field descriptions renamed.
- .pot file updated and .po files synced
- Relating id's and variables changed to be consistent with field descriptions.


kirdia:
@dpivo the task was to change the fields descriptions not the ids of the relevant options.
This PR would break the configuration of these fields

Also no need to update the .po files. The NRO users have given instructions to sync their language template with the master template before updating their strings.

kirdia:
@dpivo Can you squash the commits?

PR #810, opened 2019-05-24
PLANET-3563 - Adjust blocks vertical spacing for M and S devices

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-3563

This also removes unused `.page-section` class.

Related to greenpeace/planet4-plugin-blocks#569


assets/scss/layout/_blocks.scss
```diff
@@ -1,14 +1,29 @@
...
+    margin-bottom: $space-xxl;
+  }
 }
 
 .block-footer {
+  margin-top: $space-lg;
   margin-bottom: 0;
+
+  @include large-and-up {
+    margin-top: $space-lg;
```
pablocubico: I think this rule can be removed... unless it's overriding something I'm not seeing by specificity, it looks like it's the same as the one above. Ignore me if it's intended.

comzeradd: Ah yes, you are right. I added this to make the rule more explicit, like on `block-header`. But in this case it's redundant. I'll amend before merge.

PR #816, opened 2019-05-30
PLANET-2723 date removed from search results

dpivo:
Date removed from post teasers in search results.

### DIFF
--- a/templates/tease-search.twig
+++ b/templates/tease-search.twig
@@ -46,7 +46,7 @@
 						<a href="{{ post.link() }}" class="search-result-item-headline">{{ post.title|e('wp_kses_post')|raw }}</a>
 					{% endif %}
 
-					{% if ( not is_action ) %}
+					{% if ( 'page' != post.post_type ) %}
 						<div class="search-result-item-info">
 							<span class="search-result-item-date">{{ post.post_date|date('d/m/Y') }}</span>
 						</div>


comzeradd:
Doesn't that remove dates from all results?

My understanding from the ticket is that the request is to hide date only from pages with certain templates, because it's not useful information.

templates/tease-search.twig
```diff
@@ -46,7 +46,7 @@
 						<a href="{{ post.link() }}" class="search-result-item-headline">{{ post.title|e('wp_kses_post')|raw }}</a>
 					{% endif %}
 
-					{% if ( not is_action ) %}
+					{% if ( not is_action and 'default' != post._wp_page_template ) %}
```
comzeradd: shouldn't we also add 'evergreen' here?

angtheod: So, since we do not show dates for Action pages already, does that mean that we do not want visible dates for any Page result?
Can we then simplify the whole `if` statement by checking that it is not a page post_type?
@dpivo @comzeradd

comzeradd: Yes, good point. I guess if we block all these page types there is no page actually left, only posts.

kirdia: Yes, I was thinking the same.
We could hide dates for all WordPress 'pages'.
This could help to check against the post type
https://timber.github.io/docs/reference/timber-post/#type

dpivo: Noted, I'll add the changes

comzeradd:
@dpivo can you squash the two commits?

PR #819, opened 2019-06-04
PLANET-3539 optimize loading classes

angtheod:
Load certain files/classes only when they are actually needed.


angtheod:
> Looks good!
> We could extract that functionality to a 'loader' class (preferably singleton) which makes these decisions and keeps a reference to master-site instance also.
> What do you think?
Yes, I will separate the loading functionality from the actual theme functionality.

angtheod:
@kirdia Can you check again?

classes/class-p4-loader.php
```diff
@@ -0,0 +1,140 @@
...
+	 */
+	public static function get_instance( $services = [] ) : P4_Loader {
+		if ( ! isset( self::$instance ) ) {
+			self::$instance = new self( $services );
+		}
+		return self::$instance;
+	}
+
+	/**
+	 * P4_Master_Site constructor.
```
sagarsdeshmukh: please change the comment.

functions.php
```diff
@@ -33,34 +35,5 @@ function( $template ) {
...
-	if ( 'wordpress' === filter_input( INPUT_GET, 'import', FILTER_SANITIZE_STRING ) ) {
-	// phpcs:enable
-		$services[] = 'P4_Campaign_Importer';
-	}
-}
-
-new P4_Master_Site(
-	$services
-);
+require_once( __DIR__ . '/classes/class-p4-loader.php' );
```
sagarsdeshmukh: `phpcs` is complaining about parentheses.

sagarsdeshmukh:
@angtheod - Looks nice to me. 💯  only small 2 changes needed from my side. Well done 👍

PR #821, opened 2019-06-05
Make the whole 'Login with Google' button clickable.

27Bslash6:
So you don't have to aim precisely for the text

### DIFF
--- a/style-login.css
+++ b/style-login.css
@@ -28,6 +28,8 @@ body.login h1 a {
 body.login div#login form#loginform p.galogin a {
   text-decoration: none;
   font-weight: normal;
+  diplay: block;
+  line-height: 3rem;
 }
 
 body.login div#login form#loginform p.galogin-powered {
@@ -58,10 +60,10 @@ form#loginform p.galogin {
   text-align: center;
   vertical-align: middle;
   border-radius: 0px;
-  padding: 4px;
-  height: 27px;
+  padding: 0;
+  margin: 0;
+  height: auto;
   font-size: 14px;
-  margin-bottom: 6px;
 }
 
 body.login #login_error, body.login .message, body.login .success {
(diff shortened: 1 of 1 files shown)


style-login.css
```diff
@@ -28,6 +28,8 @@ body.login h1 a {
 body.login div#login form#loginform p.galogin a {
   text-decoration: none;
   font-weight: normal;
+  diplay: block;
```
kirdia: there is a typo error here (display)

27Bslash6: Nicely spotted, fixed.

style-login.css
```diff
@@ -28,6 +28,8 @@ body.login h1 a {
 body.login div#login form#loginform p.galogin a {
   text-decoration: none;
   font-weight: normal;
+  display: block;
+  line-height: 3rem;
```
comzeradd: Since you changed the display to block, you can just add some padding instead of changing the line-height. A 5px padding feels close to the [required design](https://jira.greenpeace.org/secure/attachment/27966/login_adjusted.jpg).

style-login.css
```diff
@@ -58,10 +60,10 @@ form#loginform p.galogin {
   text-align: center;
   vertical-align: middle;
   border-radius: 0px;
-  padding: 4px;
-  height: 27px;
+  padding: 0;
+  margin: 0;
+  height: auto;
   font-size: 14px;
-  margin-bottom: 6px;
```
comzeradd: Check the design from above. We actually need some margin between the button and the "powered" text.

style-login.css
27Bslash6: image
Without modifying line-height, and adding 5px, the result is unbalanced. I've modified the line-height to 2rem to bring it into alignment with the design guide.

style-login.css
27Bslash6: I'd removed the 'powered by' text entirely from local test copy, pending questions about whether it's actually required.

Reverted change for now.

PR #831, opened 2019-07-02
PLANET-3639 - JS: Use only let or const for variables

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-3639

* Add no-var rule to eslint
* Move all admin assets under admin folder
* Added admin js code to eslint gulp paths
* Add babel for compatibility


admin/js/posts_report.js
```diff
@@ -1,13 +1,13 @@
...
+let postCollection;
+let pageCollection;
+let postsView;
+let pagesView;
 
-var p4 = p4 || {};
-var p4_data = p4_data || {};
-var wp = window.wp || {};
+const p4 = p4 || {};
+const p4_data = p4_data || {};
```
kirdia: I think we define this `p4_data` in global scope in another case.
I get this error locally

```
Uncaught SyntaxError: Identifier 'p4_data' has already been declared
```

kirdia: hmm or maybe not in master-theme, but in another plugin

pablocubico: @kirdia @comzeradd sorry for the delay, I added a commit to move the declarations into the closure and use "window." to reference the globals, since JS was complaining about these `const p4 = p4 || {};` statements, which seemed to trigger a "Cannot access 'p4' before initialization" exception. Apparently, you can't use the null coalesce operator with `let` and `const`, didn't know that, lesson learned. 
The posts report looks fine with this change.

classes/class-p4-search.php
```diff
@@ -793,16 +793,5 @@ public function view_paged_posts() {
...
 			}
 		}
-
-		/**
-		 * Load assets only on the search page.
-		 */
-		public function enqueue_public_assets() {
-			if ( is_search() ) {
-				wp_register_script( 'search', get_template_directory_uri() . '/assets/js/search.js', [ 'jquery' ], '0.2.8', true );
-				wp_localize_script( 'search', 'localizations', $this->localizations );
```
kirdia: this breaks some parts in the search, because the global var 'localizations' is not populated
https://github.com/greenpeace/planet4-master-theme/blob/b01d4569122e680f880c70123117198175928fd1/assets/js/search.js#L133

kirdia:
Looks good overall :100:  apart from the broken search js

classes/class-p4-search.php
comzeradd: Yes, indeed. The reason I included `search.js` to the main bundle is to pipe it through babel. But we miss the 'localizations' variable that way.

PR #832, opened 2019-07-08
PLANET-3726 fix control panel

angtheod:
Fix broken P4 Control Panel links. Ajax was failing. Update Check EN link to work once again after ENS API code was updated. Add check Elasticsearch connectivity link. Add Sync ES link. Remove old Check SearchWP indexer link.

### DIFF
--- a/classes/class-p4-control-panel.php
+++ b/classes/class-p4-control-panel.php
@@ -6,6 +6,7 @@
  */
 
 use P4EN\Controllers\Ensapi_Controller as ENS_API;
+use ElasticPress\Elasticsearch as ES;
 
 if ( ! class_exists( 'P4_Control_Panel' ) ) {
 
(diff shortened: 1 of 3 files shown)


classes/class-p4-control-panel.php
```diff
@@ -247,28 +253,15 @@ public function check_search_indexer() {
...
-							$response['message'] = __( 'Indexer has stalled', 'planet4-master-theme-backend' );
-							$response['class']   = 'cp-error';
-						} else {
-							$response['message'] = __( 'Indexer is awake', 'planet4-master-theme-backend' );
-							$response['class']   = 'cp-success';
-						}
+				if ( wp_verify_nonce( $cp_nonce, 'cp-action' ) && 'check_elasticsearch' === $cp_action ) {
+					$response = [];
+
+					if ( ep_elasticsearch_can_connect() ) {
```
kirdia: Looks good! This does not seem to work with latest elasticpress version, or is it happening only for me?

sagarsdeshmukh: The "Check Elasticsearch" link is failing on my local as well - 
image

angtheod: Fixed that. I had not updated to latest EP version, so it was working with 2.8.*

angtheod: image

PR #833, opened 2019-07-09
PLANET-3619 Add filter for embed_oembed_html

kirdia:
Filter youtube.com domain for oembeds to youtube-nocookie.com

Use embed_oembed_html filter to cover for cached oembeds also (which WordPress stores in db)
https://developer.wordpress.org/reference/hooks/embed_oembed_html/

save_post hook won't work entirely, because it breaks post preview

### DIFF
--- a/classes/class-p4-master-site.php
+++ b/classes/class-p4-master-site.php
@@ -135,6 +135,7 @@ class P4_Master_Site extends TimberSite {
 		add_filter( 'login_headertext', [ $this, 'add_login_logo_url_title' ] );
 		add_action( 'login_enqueue_scripts', [ $this, 'add_login_stylesheet' ] );
 		add_filter( 'comment_form_submit_field', [ $this, 'gdpr_cc_comment_form_add_class' ], 150, 2 );
+		add_filter( 'embed_oembed_html', [ $this, 'filter_youtube_oembed_nocookie' ], 10, 2 );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


classes/class-p4-master-site.php
```diff
@@ -783,4 +784,24 @@ public function gdpr_cc_comment_form_add_class( $submit_field, $args ) {
...
 	}
+
+	/**
+	 * Filter function for embed_oembed_html.
+	 * Transform youtube embeds to youtube-nocookie.
+	 *
+	 * @see https://developer.wordpress.org/reference/hooks/embed_oembed_html/
+	 *
+	 * @param $cache mixed  The cached HTML result, stored in post meta.
+	 * @param $url   string The attempted embed URL.
```
comzeradd: I think these should be reversed, like `@param string $url` and phpcs is complaining.

classes/class-p4-master-site.php
```diff
@@ -783,4 +784,24 @@ public function gdpr_cc_comment_form_add_class( $submit_field, $args ) {
...
+	 * Transform youtube embeds to youtube-nocookie.
+	 *
+	 * @see https://developer.wordpress.org/reference/hooks/embed_oembed_html/
+	 *
+	 * @param mixed  $cache The cached HTML result, stored in post meta.
+	 * @param string $url   The attempted embed URL.
+	 *
+	 * @return mixed
+	 */
+	function filter_youtube_oembed_nocookie( $cache, $url ) {
```
sagarsdeshmukh: Please declare function visibility here.

sagarsdeshmukh:
We should consider the youtube short urls as well like - https://youtu.be

kirdia:
> We should consider the youtube short urls as well like - https://youtu.be

Done

classes/class-p4-master-site.php
```diff
@@ -783,4 +784,29 @@ public function gdpr_cc_comment_form_add_class( $submit_field, $args ) {
...
+	 * @param string $url   The attempted embed URL.
+	 *
+	 * @return mixed
+	 */
+	public function filter_youtube_oembed_nocookie( $cache, $url ) {
+		if ( ! empty( $url ) ) {
+			if ( strpos( $url, 'youtube.com' ) !== false ) {
+				$cache = str_replace( 'youtube.com', 'youtube-nocookie.com', $cache );
+			}
+			if ( strpos( $url, 'youtu.be' ) !== false ) {
```
angtheod: make it an `elseif` better?

classes/class-p4-master-site.php
```diff
@@ -783,4 +784,29 @@ public function gdpr_cc_comment_form_add_class( $submit_field, $args ) {
...
+				$cache = str_replace( 'youtube.com', 'youtube-nocookie.com', $cache );
```
angtheod: should replace `youtu.be` part here?

kirdia: nope. at this point the html from the embed is filtered, which contains the `youtube.com` domain although the original embed url is `youtu.be`

angtheod: can we merge it then into one if? since we are doing the same thing in both if/else

PR #838, opened 2019-07-11
PLANET-3744 - Hide CTA links from page header on admin

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-3744

This is a small counterpart for greenpeace/planet4-plugin-blocks#599

### DIFF
--- a/classes/class-p4-metabox-register.php
+++ b/classes/class-p4-metabox-register.php
@@ -95,28 +95,31 @@ class P4_Metabox_Register {
 
 		$p4_header->add_field(
 			[
-				'name' => __( 'Header Button Title', 'planet4-master-theme-backend' ),
-				'desc' => __( 'Header button title comes here', 'planet4-master-theme-backend' ),
-				'id'   => $this->prefix . 'button_title',
-				'type' => 'text_medium',
+				'name'       => __( 'Header Button Title', 'planet4-master-theme-backend' ),
+				'desc'       => __( 'Header button title comes here', 'planet4-master-theme-backend' ),
+				'id'         => $this->prefix . 'button_title',
+				'type'       => 'text_medium',
+				'show_on_cb' => [ $this, 'is_not_campaign_post' ],
 			]
 		);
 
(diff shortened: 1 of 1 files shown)


kirdia:
I see it could be done in cmb2 settings also, could a little bit more clear in php

https://github.com/CMB2/CMB2/wiki/Field-Parameters#show_on_cb

sagarsdeshmukh:
@comzeradd - Please add resource version [here](https://github.com/greenpeace/planet4-master-theme/blob/develop/classes/class-p4-post-campaign.php#L385), so that it load the updated css .

Also better we will add hint for web editor on backend about "On Campaigns the CTA buttons are hidden" or the CMB2 field callback function to hide the field conditionally. 
(Example - https://github.com/greenpeace/planet4-master-theme/blob/develop/classes/class-p4-metabox-register.php#L409 , https://github.com/greenpeace/planet4-master-theme/blob/develop/admin/js/cmb2_form_validation.js#L36)

comzeradd:
I see that `campaign.css` gets a version number. Currently `5.2.2` on stage, but not sure where this comes from.

The cmb2 callback filter seems indeed more clean solution. I tried it out, but I couldn't make it work. Regardless what the callback returns the fields are always visible. Any help would be appreciated.

kirdia:
Ammended the commit to do the post type check where cmb2 fields are registered.

sagarsdeshmukh:
@kirdia - I tested the call back function fix on local, It works just perfect 💯 . well done 👍 
Just small thing could you please also remove the `New Tab` field as well -
image

kirdia:
I tried to add it that too, but now all of them are shown.
I don't know why it is inconsistent.
Do we change the visibility in other place also?
Can you retry creating a new campaign post?

sagarsdeshmukh:
@kirdia - The latest changes working fine on my local.
image

kirdia:
Yeap, it was the function's visibility on my local indeed.
Thanx for checking.

PR #839, opened 2019-07-11
PLANET-3773: Finish docs for migrating Blocks to Gutenberg

pablocubico:
Testers wanted! Is this breaking any styles on the admin side or the site?

We need to do some changes to render the blocks inside the Gutenberg block editor:
* include (enqueue) the styles in the admin side too.
* add `show_in_rest` to the custom taxonomy (used to populate the "Page type" tokens field, Story, Publication, Press Release)

I think we should merge adaptations like this in advance to make our branches "Gutenberg-ready" now, maybe behind some development flag, or just as it is in this commit, I think it's not generating any conflicts so far.


pablocubico:
Closing as this was already taken care of on the gutenberg-plugin repo, we'll see if we need to make changes to these branches eventually.

classes/class-p4-custom-taxonomy.php
```diff
@@ -230,6 +230,7 @@ public function register_taxonomy() {
 				'show_admin_column' => true,
 				'query_var'         => true,
 				'meta_box_cb'       => [ $this, 'create_taxonomy_metabox_markup' ],
+				'show_in_rest'      => true
```
kirdia: @pablocubico Could you open a PR for this. We still need this change.

PR #840, opened 2019-07-12
Feature/planet 3736 tag page script

koyan:
[Jira Issue 3736](https://jira.greenpeace.org/browse/PLANET-3736)

Adding functionality that automatically creates a wordpress page for each tag that does not have a redirect page assigned to it, and assigns it to the tag


classes/class-p4-campaigns.php
```diff
@@ -222,18 +227,48 @@ public function save_taxonomy_meta( $term_id ) {
...
+
+				$my_post = array(
+					'post_title'   => '#' . $tag_data->name,
+					'post_content' => $post_content,
+					'post_status'  => 'publish',
+					'post_type'    => 'page',
+				);
+
+				// Insert the post into the database.
+				$tag_page_id = wp_insert_post( $my_post );
```
kirdia: A check for return type (if it's an int) would be good here.
Insert post could return WP_Error if it fails for some reason.

classes/class-p4-campaigns.php
```diff
@@ -222,18 +227,48 @@ public function save_taxonomy_meta( $term_id ) {
...
+				$happypoint_bg_opacity = 30;
 			}
 
 			$redirect_page = $_POST['redirect_page'] ?? 0;
-			if ( ! is_null( get_post( $redirect_page ) ) || ! $redirect_page ) {
+			if ( ! is_null( get_post( $redirect_page ) ) && 0 != $redirect_page ) {
 				update_term_meta( $term_id, 'redirect_page', $redirect_page );
+			} else {
+
+				$tag_data     = get_term( $term_id );
```
kirdia: same here for get_term.

Could return `WP_Error`

kirdia:
Looks good, a few checks for return type.
Also could you rebase over release branch? there is a translation file in the commit

koyan:
Updated according to feedback. The translation file is there intentionally: we are passing a string to shortcake with a quote, complicating it unecessarily. I am removing this from the text.

PR #846, opened 2019-07-19
PLANET-3784 Fix article block portrait image full width issue

sagarsdeshmukh:
[JIRA 3784](https://jira.greenpeace.org/browse/PLANET-3784)


assets/scss/pages/_page.scss
```diff
@@ -29,8 +29,9 @@ div.page-template {
   }
 
   img {
```
dpivo: This might effect more images than we intend to. It might be better to add the css changes to: https://github.com/greenpeace/planet4-plugin-blocks/blob/bd214664db8a7a815b9bcdc5a0e3959724a37914/assets/scss/components/_articles.scss#L62

PR #855, opened 2019-07-26
PLANET-3809 Hide meta fields for password protected content

sagarsdeshmukh:
[JIRA 3809](https://jira.greenpeace.org/browse/PLANET-3809)


page.php
```diff
@@ -82,6 +82,11 @@
 if ( post_password_required( $post->ID ) ) {
```
kirdia: It could be better to add a check in html-header template to check if a post is rendered and if it is password protected and then render the meta tags.

something like
`{% if (post and fn('post_password_required', post)) %}`

kirdia:
Looks good! Please remove the changes from the templates, since they are not needed and squash the commits

sagarsdeshmukh:
@kirdia - I remove old changes and squash the commits. please have a look.

PR #858, opened 2019-07-30
PLANET-3781: Gallery Block: Grid style focus point not working when used in Posts

pablocubico:
I had to add a filter to allow the `object-position` property, otherwise, `wp_kses` would still remove the `style` attribute even if it's defined as an allowed attribute in the `set_custom_allowed_attributes` filter.

Ref: https://jira.greenpeace.org/browse/PLANET-3781


kirdia:
Looks good! there are 2 code styling error though

pablocubico:
@kirdia updated! there is still a failing test but it's the failing EN test from Codeception

PR #862, opened 2019-08-06
PLANET-3852: Add external link icon next to links in text

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-3852

This ticket is for adding an external link icon to "all" links in the site, some important notes:
- as the icon is added using a **background SVG in a pseudo-element**, it does not inherit the color of the link, the current color is hardcoded to look like the default master theme link color. I tried using SVG `mask`, `clip-path`, using `currentColor`, forcing the cascade, etc, but it just will not work being an SVG in a background property of a pseudo element, which is clean & clever but also hacky, buried deep into the DOM's internal plumbing.

I see a couple of solutions for this:
   - Use a hardcoded color as the current code is, and just hope it lands correctly on content, using the default link color for A tags.
   - Use a font icon for this case only (?)
   - Use a JS approach and add a regular `<svg>` element to the A tags, not inside a pseudo...

References:

https://codepen.io/simevidas/pen/MKdmeV
https://css-tricks.com/creating-a-maintainable-icon-system-with-sass/


pablocubico:
hey @comzeradd it's not for the image! it's for the `href` rule for creating the links here: https://github.com/greenpeace/planet4-master-theme/pull/862/files#diff-c56b545271598ba7579d4de374ed2497R3

However, i don't know if we are still using this technique, because this way the icon does not inherit the color of the link (check the PR description), if we choose another method (like a JS solution) all this would go away.

pablocubico:
Amended last commit to remove the duplicated Twig include tag.

PR #863, opened 2019-08-06
PLANET-3864 - Wrap Optimize code to an if clause

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-3864

The snippet brackets conflict with twig variable evaluation, so I removed them and used an if clause and a twig evaluation directly into the snippet.


comzeradd:
Closed in favor of #864

PR #868, opened 2019-08-09
Make the master theme compatible with the Gutenberg blocks plugin

pablocubico:
This is just for adapting the master theme to be able to add the parent styles to the Gutenberg blocks plugin, and avoid breaking the admin side styles.


comzeradd:
Looks good. Is there a reason why this should be part of next release?

sagarsdeshmukh:
Looks like BT css hides radio buttons and checkboxes 🤔 
image
image

pablocubico:
> Looks good. Is there a reason why this should be part of next release?

@comzeradd hmmm, we can even skip merging it, and just pin the Gutenberg instance to this master theme branch, but I think it would be sane to merge it in advance so when the Gutenberg plugin is done the master theme is prepared with no conflicts

comzeradd:
We can merge it. But I think it's better to merge it on the `develop` branch. It will give us some extra time to make sure it doesn't affect production, since there is no ticket for doing UAT.

pablocubico:
@sagarsdeshmukh aha! it looks like the `_forms` file in the `styleguide` is affecting the checkboxes... and other inputs, I'll make a PR for that too 🤔 

@comzeradd very right you are sir, I'll change the base

pablocubico:
@sagarsdeshmukh I addressed your comment in this PR in the styleguide: https://github.com/greenpeace/planet4-master-theme/pull/868
@comzeradd base changed, I added a commit to pin the styleguide's commit reference, does that make sense?

comzeradd:
@pablocubico I created a new tag with that change on the styleguide (v0.11). Maybe checkout styleguide submodule to that tag.

PR #890, opened 2019-09-04
Campaigns: limit top padding to first block only

comzeradd:
This was introduced at 243abbe65491eff8d977b78c4a5465010dd23860 to compensate for Campaigns not having a page title. But this affected all blocks. This change limits that change only to the first block.


comzeradd:
@pablocubico you are right :+1: I did a small change.

PR #901, opened 2019-09-13
PLANET-4023 Allow "Take Action" boxout in Posts 

dpivo:
- Strip the take action boxout block from the post content and add it to where the current shortcode would be added.

### DIFF
--- a/single.php
+++ b/single.php
@@ -21,6 +21,17 @@ $context = Timber::get_context();
 $post            = Timber::query_post( false, 'P4_Post' );
 $context['post'] = $post;
 
+// Strip Take Action Boxout block from the post content to add outside the normal block container.
+if ( false !== strpos( $post->post_content, '[shortcake_take_action_boxout' ) ) {
+
+	$take_action_boxout_block_start  = strpos( $post->post_content, '[shortcake_take_action_boxout' );
+	$take_action_boxout_block_end    = strpos( $post->post_content, ']', $take_action_boxout_block_start );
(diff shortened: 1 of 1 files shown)


single.php
```diff
@@ -21,6 +21,16 @@
 $post            = Timber::query_post( false, 'P4_Post' );
 $context['post'] = $post;
 
+// Strip Take Action Boxout block from the post content to add outside the normal block container.
+if ( strpos( $post->post_content, '[shortcake_take_action_boxout' ) ) {
```
kirdia: better check if strpos !== false, cause strpos could return 0 and this won't be evaluated

dpivo: Good point! Changed...

single.php
```diff
@@ -21,6 +21,16 @@
 $post            = Timber::query_post( false, 'P4_Post' );
 $context['post'] = $post;
 
+// Strip Take Action Boxout block from the post content to add outside the normal block container.
+if ( false !== strpos( $post->post_content, '[shortcake_take_action_boxout' ) ) {
+
+	$take_action_boxout_block_start = strpos( $post->post_content, '[shortcake_take_action_boxout' );
+	$take_action_boxout_block_end   = strpos( $post->post_content, ']', $take_action_boxout_block_start );
+	$take_action_boxout_block       = substr( $post->post_content, $take_action_boxout_block_start, $take_action_boxout_block_end );
```
kirdia: this won't work. substr needs length of the string as the 3rd parameter not an index.

dpivo: calculated the length and added used that instead.

PR #909, opened 2019-09-18
PLANET-4063 Add blocks spacing to post images

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-4063

This is added only to images with no alignment, so we don't mess with floating images.


comzeradd:
@pablocubico yes, good point. commit amended.

There is also a difference in class hierarchy when the image has caption. So I did a small addition to catch these cases too.

Can you have another look?

pablocubico:
@comzeradd Ah, I noticed the caption thing but it had some margin so I assumed this didn't apply to that. Works great! Feel free to merge.

PR #911, opened 2019-09-19
PLANET-4084 CPP - Add campaign import capability to Campaigner role

sagarsdeshmukh:
[JIRA 4084](https://jira.greenpeace.org/browse/PLANET-4084)


classes/class-p4-post-campaign.php
```diff
@@ -204,6 +205,15 @@ public function add_campaigner_role() {
 			);
 		}
 
+		/**
+		 * Add Campaign import capabilities to Campaigner User.
+		 */
+		public function add_campaigner_caps_import() {
```
kirdia: better this done on the point when we add our custom role `add_role`, not as an extra step

sagarsdeshmukh: If we add it on `add_role()` it will applicable only for new user creation. By adding separate function, we add import capability to old campaigner users as well.

PR #912, opened 2019-09-19
PLANET-3992 Use opengraph metadata for share buttons

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-3992

Add a function for constructing all share button metadata.


pablocubico:
Hey @comzeradd , when using custom OG fields, everything worked fine, but something weird happened with a post with no custom OG fields:

image

Maybe an escaping issue? 🤔 
image

PR #914, opened 2019-09-20
PLANET-4054: Gutenberg Compatibility: Separate Bootstrap, layout styles and block styles in master theme

pablocubico:
**:warning: HEADS UP! The base is the Gutenberg branch now**

Switching from Gulp to Webpack.

At first I thought of just using a concatenation plugin for Webpack but that's not the way Webpack is intended to be used, so I went for the default Webpack configuration from Wordpress, as in the Gutenberg plugin repo, I'm also replicating its directory structure, so we have consistency between both. I renamed the `react-blocks` dir to `assets` as Dirk suggested since it's more common and makes sense for both repos.

Ref: https://jira.greenpeace.org/browse/PLANET-4054


comzeradd:
There a couple of minor phpcs errors.

On the root folder, we can delete `style.css.map` and maybe link `style.css` to `assets/build/style.min.css`, so all theme metadata are discoverable from wordpress?

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "assets/scss/styleguide"]
 	path = assets/scss/styleguide
 	url = https://github.com/greenpeace/planet4-styleguide.git
+[submodule "assets/src/scss/styleguide"]
+	path = assets/src/scss/styleguide
+	url = https://github.com/greenpeace/planet4-styleguide
```
comzeradd: Let's remove the old submodule deceleration and keep just the new one. I think the url should end with `.git`, but maybe I'm wrong.

PR #919, opened 2019-09-24
PLANET-4050 Refactor campaigner role and capabilities

kirdia:
Extract campaigner role functions to a new class - P4_Campaigner
Add P4_Activator class to run actions on theme activation/deactivation hooks
Load P4_Activator when theme is switched


comzeradd:
Let's hold merging this after tomorrow release.

PR #922, opened 2019-09-27
PLANET-4139 Alignment of social media icons in all footers

dpivo:
- Styled footer to align icons if there are less than the usual 5


comzeradd:
@dpivo I think you forgot to commit the actual css code here or am I missing something?

sagarsdeshmukh:
Related [PR](https://github.com/greenpeace/planet4-styleguide/pull/13)

comzeradd:
Thanks @sagarsdeshmukh :+1: 

So let's wait for that one to merged first and then we can checkout the styleguide submodule to a new tag. Otherwise these changes here will be overwritten next time we re-generate the minified styles.

PR #926, opened 2019-10-02
PLANET-4101 - Adjust page-header spacing when title is hidden

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-4101

Adds a new class to page-header when title is hidden, in order to adjust spacing.

Let's wait for greenpeace/planet4-styleguide#15 before we merge


comzeradd:
This is now checked out to the latest styleguide tag.

PR #935, opened 2019-10-17
[No ticket] Fix image paths and handling by Webpack

pablocubico:
Webpack resolves images as it is intended to embed them either as Base64 in the code, or do some manipulation and "build" them into the build directory, or even just copying them there.

As we don't follow that process (since this wasn't originally a typical single page "Webpacked" app), I removed the URL resolution from the CSS loader, restored the relative paths, and symlinked the minified `style.min.css` so the paths still work with our `images` folder in the root.

Another way around would be to move the images folder into `assets/build`, but that directory could be destroyed by other processes if we are not careful enough.

I also corrected the path for the SVG sprites.


pablocubico:
> Looks good. When I run `npm run build` I get two modified files
> 
> ```
> modified:   assets/build/bootstrap.min.css
> modified:   assets/build/index.js
> ```

Can you test again? I'm not able to reproduce this effect

PR #943, opened 2019-10-28
PLANET-4305: Enable wpautop as Wordpress did before Gutenberg.

pablocubico:
Test with more content if this solves the issue.

Ref: https://jira.greenpeace.org/browse/PLANET-4305

I think WP by default also adds `wpautop` to excerpts, we could mimic this by adding the same filter to the `the_excerpt` hook, but we are using the `excerpt` filter from Timber which is an alias of `wp_trim_words`, it trims words, strips tags, and removes line-breaks internally, so it makes no difference.

Some notes:
https://wpquark.com/kb/misc/wordpress/fixing-themes-custom-wpautop-priority/

### DIFF
--- a/classes/class-p4-master-site.php
+++ b/classes/class-p4-master-site.php
@@ -150,6 +150,9 @@ class P4_Master_Site extends TimberSite {
 			}
 		);
 
+		// wpautop is already added, this just changes its priority.
+		add_filter( 'the_content', 'wpautop', 9 );
+
 		add_action( 'init', [ $this, 'login_redirect' ], 1 );
 	}
 
--- a/classes/class-p4-master-site.php
+++ b/classes/class-p4-master-site.php
@@ -150,8 +150,25 @@ class P4_Master_Site extends TimberSite {
 			}
 		);
 
-		// wpautop is already added, this just changes its priority.
-		add_filter( 'the_content', 'wpautop', 9 );
+		/**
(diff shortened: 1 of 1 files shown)


pablocubico:
# Steps from original post to Gutenberg

Monitoring how it looks in the editor and in the frontend and each step.

-----

Original post:

Captura de pantalla 2019-10-29 a la(s) 11 26 40

Original post frontend:

www planet4 test_uncategorized_25273_a-post-with-a-block-on-it_

### 1. After just turning Gutenberg on (_before_ conversion, notice the counter style seems to change, this is a different bug we should address):

Editor:

Captura de pantalla 2019-10-29 a la(s) 11 27 03

Frontend:

www planet4 test_uncategorized_25273_a-post-with-a-block-on-it_ (1)

-----

### 2. After converting the post using PHP (Conversion step 1):

Editor:

Captura de pantalla 2019-10-29 a la(s) 11 27 32

Frontend **(notice the line-breaks are gone)**:

www planet4 test_uncategorized_25273_a-post-with-a-block-on-it_ (2)

### 3. After re-enabling `wpautop` with a priority of `9`.

www planet4 test_uncategorized_25273_a-post-with-a-block-on-it_ (3)

comzeradd:
@pablocubico looks good. can you fix the `phpcs` errors, and then it's good to be merged.

PR #952, opened 2019-11-12
PLANET-4264 : Add style tags to allowed tags for posts. Fix broken cs…

galousis:
…s issue with meetup embed. 

[Ref](https://jira.greenpeace.org/browse/PLANET-4264)

### DIFF
--- a/classes/class-p4-master-site.php
+++ b/classes/class-p4-master-site.php
@@ -391,6 +391,7 @@ class P4_Master_Site extends TimberSite {
 	 */
 	public function set_custom_allowed_attributes_filter( $allowedposttags ) {
 		// Allow iframes and the following attributes.
+		$allowedposttags['style']  = [];
 		$allowedposttags['iframe'] = [
 			'align'           => true,
 			'width'           => true,


galousis:
Need to fix PHP code style errors

galousis:
Will remove private property $allowed_html, because is not so needed anymore

classes/class-p4-loader.php
```diff
@@ -52,6 +52,89 @@ public static function get_instance( $services = [] ) : P4_Loader {
...
+		add_filter( 'wp_kses_allowed_html', [ $this, 'filter_allowed_html' ], 10, 2 );
+	}
+
+	/**
+	 * Get allowed tags and add some more
+	 *
+	 * @param string $context context.
+	 * @return mixed
+	 */
+	public function filter_allowed_html( $context = '' ) {
```
Inwerpsel: There is already a function that adds allowed tags at `\P4_Master_Site::set_custom_allowed_attributes_filter`, probably better to add these extra tags there.

classes/class-p4-loader.php
```diff
@@ -52,6 +52,89 @@ public static function get_instance( $services = [] ) : P4_Loader {
...
+		global $allowedposttags;
+		$allowed_tags = $allowedposttags;
+		// Add our own tags.
+		$allowed_tags['figure'] = array();
+		$allowed_tags['style']  = array(
+			'background'            => true,
```
Inwerpsel: The attributes here are not for the element's content, only for its attributes. So it should work with just an empty array.

classes/class-p4-loader.php
galousis: It is a lot better in there cause it does not spoil for example twitter & youtube embeds.

galousis:
Despite all the thoughts on **security vs flexibility** issues that had risen upon this [ticket](https://jira.greenpeace.org/browse/PLANET-4264) in the dev channel, i think this ticket is ok to close it. Since all it takes care is to add to allowed html the style tag (so the css does not crash) by not removing the kses_post filter from single twig file for posts.
 
All other thoughts like if we allow script gets stored as _oembed_* meta entry in the postmeta table (like in the case of twitter embed block) either in posts or pages belong to another context that should probably live in another new ticket that will weight everything that has to do with policy on editors, user capabilities, kses filters etc...

galousis:
For reasons i do not know so far the circle keeps failing on 2 certain cept test. They seem quite irrelevant to this ticket though. 

- AddCommentCept

- LoginCept

comzeradd:
Tested fixed :)

So yes, this seems the simplest solution for now and we allow the style tag on pages anyway. 

if @Inwerpsel can have a final look on the code.

classes/class-p4-master-site.php
```diff
@@ -392,6 +392,8 @@ public function set_custom_allowed_css_properties( $allowedproperties ) {
 	 */
 	public function set_custom_allowed_attributes_filter( $allowedposttags ) {
 		// Allow iframes and the following attributes.
+		$allowedposttags['style'] = array();
```
Inwerpsel: Can you replace this with the short array notation `[]`?

comzeradd:
@galousis make sure to squash the two commits into one before merging.

PR #958, opened 2019-11-21
PLANET-4113 Added open graph meta fields on author profile page

sagarsdeshmukh:
[JIRA 4113](https://jira.greenpeace.org/browse/PLANET-4113)

- Updated open graph data on author profile page
- Fix social share button links on author profile page

### DIFF
--- a/author.php
+++ b/author.php
@@ -27,6 +27,21 @@ if ( isset( $wp_query->query_vars['author'] ) ) {
 	$context['author']   = $author;
 	$context['title']    = 'Author Archives: ' . $author->name();
 	$post_args['author'] = $wp_query->query_vars['author'];
+
+	$context['social_accounts'] = P4_Post::filter_social_accounts( $context['footer_social_menu'] );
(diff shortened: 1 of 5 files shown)


templates/blocks/post_meta_fields.twig
```diff
@@ -0,0 +1,45 @@
+{% if (post.post_excerpt) %}
```
Inwerpsel: This template is actually identical to author_meta_fields.twig, except for the variables. Would it make sense here to just have one template with variables? That way we don't have to duplicate all the html and twig logic of escaping the variables.
Screenshot from 2019-11-26 12-09-02

Inwerpsel: So you could do something like 
`{% include 'blocks/author_meta_fields.twig' with {description: post.excerpt|striptags|default(site.description)}%}

sagarsdeshmukh:
Hi @Inwerpsel - I did some change in PR after your approval. could you please have a look one more time.

PR #964, opened 2019-11-28
Campaigns - Added option to choose default logo

alagregor:


comzeradd:
Isn't the image same as [images/gp-logo.svg](https://github.com/greenpeace/planet4-master-theme/blob/develop/images/gp-logo.svg) ?

alagregor:
Is the same, but there are already logo-light and logo-dark, which are fetched by color so I added logo-green which can you choose from campaign post type. Is this ok?

images/gp-logo-green.svg
```diff
@@ -0,0 +1,68 @@
+<?xml version="1.0" encoding="utf-8"?>
+<svg version="1.1" id="Calque_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 viewBox="-50 127.5 183 30" enable-background="new -50 127.5 183 30" xml:space="preserve">
+<title>gp-logo</title>
+<g>
+	<path fill="#66CC00" d="M7.7,134.1c-0.5,0.4-1.4,0.5-5.9,1.3c0,0-0.2,0-0.9,0.4c-1.1,0-2.8,0.8-3.3,1.1c-0.2,0.1-0.5,0.7-0.6,1.8
```
27Bslash6: Instead of using three different files, why not [style the fill via CSS](https://www.sitepoint.com/real-world-use-of-css-with-svg/)?

comzeradd:
So this is the template that handles the logo:
https://github.com/greenpeace/planet4-master-theme/blob/develop/templates/blocks/site_logo.twig

If an editor chooses "Climate Emergency" logo and then color "Green", this will break because there is no green variation for Campaign logos.

PR #968, opened 2019-12-03
PLANET-4389 Add “Credit” custom media metadata fields

sagarsdeshmukh:
[JIRA 4389](https://jira.greenpeace.org/browse/PLANET-4389)

- Add “Credit” custom media metadata fields
-  Remove `action_page_media_buttons ` function as it's not in use anymore
- Add visibility to  `login_redirect`, `add_login_logo_url ` and `add_login_logo_url_title ` function
- Remove `post_gallery ` filter function (In Gutenberg editor, we don't have classic image Gallery feature)
- Remove `img_caption_shortcode`  filter, added new `p4_core_image_block_render` function to append image credit field with image caption on core/image Gutenberg block

### DIFF
--- a/classes/class-p4-master-site.php
+++ b/classes/class-p4-master-site.php
@@ -112,9 +112,7 @@ class P4_Master_Site extends TimberSite {
 			}
 		);
 		add_action( 'after_setup_theme', [ $this, 'add_image_sizes' ] );
-		add_filter( 'post_gallery', [ $this, 'carousel_post_gallery' ], 10, 2 );
 		add_action( 'save_post', [ $this, 'p4_auto_generate_excerpt' ], 10, 2 );
-		add_filter( 'img_caption_shortcode', [ $this, 'override_img_caption_shortcode' ], 10, 3 );
 
 		add_action( 'wp_ajax_get_paged_posts', [ 'P4_ElasticSearch', 'get_paged_posts' ] );
 		add_action( 'wp_ajax_nopriv_get_paged_posts', [ 'P4_ElasticSearch', 'get_paged_posts' ] );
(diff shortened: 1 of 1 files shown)


classes/class-p4-master-site.php
```diff
@@ -868,4 +799,80 @@ public function filter_youtube_oembed_nocookie( $cache, $url ) {
...
+		$meta         = get_post_meta( $image_id );
+		$image_credit = '';
+		if ( isset( $meta['_credit_text'] ) && ! empty( $meta['_credit_text'][0] ) ) {
+			$image_credit = ' ' . $meta['_credit_text'][0];
+			if ( ! is_numeric( strpos( $meta['_credit_text'][0], '©' ) ) ) {
+				$image_credit = ' ©' . $image_credit;
+			}
+		}
+
+		if ( $image_credit ) {
```
Inwerpsel: You could move what is in this if statement to the end of the previous one, as at that point image credit will always have something in it.

classes/class-p4-master-site.php
```diff
@@ -868,4 +799,80 @@ public function filter_youtube_oembed_nocookie( $cache, $url ) {
...
+		$image_id = trim( str_replace( 'attachment_', '', $attributes['id'] ) );
+		$meta     = get_post_meta( $image_id );
+
+		if ( isset( $meta['_credit_text'] ) && ! empty( $meta['_credit_text'][0] ) ) {
+			$image_credit = ' ' . $meta['_credit_text'][0];
+			if ( false === strpos( $meta['_credit_text'][0], '©' ) ) {
+				$image_credit = ' ©' . $image_credit;
+			}
+
+			if ( $image_credit ) {
```
Inwerpsel: This condition will always be true as `$image_credit` comes from `$meta['_credit_text'][0]` which is already guaranteed from the check above to be non-empty. So it can just be removed here.

PR #971, opened 2019-12-11
PLANET-4591 Add dataLayer param - signedIn and visitorType

sagarsdeshmukh:
[JIRA 4591](https://jira.greenpeace.org/browse/PLANET-4591)

### DIFF
--- a/classes/class-p4-master-site.php
+++ b/classes/class-p4-master-site.php
@@ -330,6 +330,16 @@ class P4_Master_Site extends TimberSite {
 		$context['google_optimizer']       = isset( $options['google_optimizer'] ) ? true : false;
 		$context['facebook_page_id']       = $options['facebook_page_id'] ?? '';
 
+		// Datalayer feed.
+		$current_user = wp_get_current_user();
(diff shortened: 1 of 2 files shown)


classes/class-p4-master-site.php
```diff
@@ -330,6 +330,16 @@ public function add_to_context( $context ) {
...
 		$context['facebook_page_id']       = $options['facebook_page_id'] ?? '';
 
+		// Datalayer feed.
+		$current_user = wp_get_current_user();
+		if ( $current_user->ID ) {
+			$context['p4_signedin_status'] = 'true';
+			$context['p4_visitor_type']    = $current_user->roles[0] ?? '';
+		} else {
+			$context['p4_signedin_status'] = 'false';
+			$context['p4_visitor_type']    = '';
```
Inwerpsel: Shouldn't this be "Logged out" according to the ticket description? But maybe something like "Guest" would be more appropriate as "Logged out" implies that the user was logged in before, which is not the case for most visitors :)

PR #979, opened 2019-12-18
PLANET-4456 Fix special chars in campaign import

sagarsdeshmukh:
- Fix special char issue in campaign export/import functionality.
- The EN form background image not exported properly, added small fix in regular exp. to fix this issue.

Example :
Campaign exported from `staging` -
https://release.k8s.p4.greenpeace.org/international/campaign/campaign-page-demo-plastics/
Campaign imported on `sagardev` instance (current PR branch is pinned on sagardev instance) -
https://k8s.p4.greenpeace.org/sagardev/?post_type=campaign&p=27747&preview=true

The EN form bg image issue tested by exporting from sagardev instance and imported on staging.
https://k8s.p4.greenpeace.org/sagardev/campaign/test-en-form/
https://release.k8s.p4.greenpeace.org/international/?post_type=campaign&p=27806&preview=true

### DIFF
--- a/classes/class-p4-campaign-importer.php
+++ b/classes/class-p4-campaign-importer.php
@@ -131,7 +131,7 @@ if ( ! class_exists( 'P4_Campaign_Importer' ) ) {
 					'post_content' => $post_content,
 					'postmeta'     => $campaign_postmeta,
 				];
-				wp_update_post( $updated_post );
+				wp_update_post( wp_slash( $updated_post ) );
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


classes/class-p4-campaign-importer.php
```diff
@@ -125,6 +125,18 @@ public function update_campaign_attachements( $post_id, $original_post_id, $post
 					$campaign_postmeta = $post['postmeta'];
 				}
 
+				// Search and replace the special char in post content.
+				$replace_chars = [
+					'u0026' => '\u0026',
```
Inwerpsel: Shouldn't this replace `'u0026'` only when it is not preceded by a `\`?

sagarsdeshmukh: Here the slash is already there, but on save of post content it cleaned up. the patch added here will add additional slash for `\u0026 , \u0022 , \u003c and \u003e` special chars which for now solves the issue.

PR #984, opened 2019-12-27
PLANET-4603 Register campaign fields instead of metaboxes

Inwerpsel:
* Use `register_post_meta` to add campaign fields instead of cmb2. The
editor interface for these fields will be added to
planet4-plugin-gutenberg-blocks in https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/190
* Move allowed values to json files. These are fetched by the frontend
to populate the values of the editor. Initially all files contain the
same values as were available for all templates in the code.
* Use a wrapper function `P4_Post_Campaign::campaign_field()` to add
these fields for convenience, as they all share some args.
* Create a list of CSS variables. Ultimately these should be used as
actual css variables, however as a first step use these in a twig
template that only adds the rule if a variable is set. This intermediary
step is because a bigger change is needed to handle fallbacks when a
value is not set: the fallback would be dependant on which campaign
template is chosen.

### DIFF
--- a/classes/class-p4-campaign-importer.php
+++ b/classes/class-p4-campaign-importer.php
@@ -24,6 +24,7 @@ if ( ! class_exists( 'P4_Campaign_Importer' ) ) {
 		public function __construct() {
 			add_action( 'wp_import_insert_post', [ $this, 'update_campaign_attachements' ], 10, 4 );
 			add_filter( 'wp_import_post_terms', [ $this, 'filter_wp_import_post_terms' ], 10, 3 );
+			add_filter( 'wp_import_post_meta', [ $this, 'read_old_campaign_template_attribute' ] );
 			add_filter( 'wp_import_post_data_processed', [ $this, 'set_imported_campaigns_as_drafts' ], 10, 2 );
 			add_action( 'import_end', [ $this, 'action_import_end' ], 10, 0 );
 		}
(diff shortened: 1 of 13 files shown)


Inwerpsel:
I'm going to move the configuration options of themes to a separate json file instead of coupling them to the post_meta fields. That way post_meta only functions as storage and isn't aware of which values it can contain for a theme.

Also the endpoint isn't needed then, as well as the current coupling of `withPostMeta` to the allowed campaign values, making that component usable for all post_meta controls.

campaign_templates/forest.json
```diff
@@ -0,0 +1,173 @@
+{
+	"id": "oceans",
```
pablocubico: You probably already know but some of these files have these "oceans" values, please change accordingly!

classes/class-p4-campaign-importer.php
```diff
@@ -192,5 +193,19 @@ public function set_imported_campaigns_as_drafts( $postdata, $post ) {
...
+		/**
+		 * Needed to remove the underscore to expose the field in the API.
+		 * Use the value with underscore for old exports if the new field isn't present.
+		 *
+		 * @param array $post_meta The to be imported post meta fields.
+		 *
+		 * @return array The normalized post meta fields.
+		 */
+		public function read_old_campaign_template_attribute( $post_meta ) {
+			$post_meta['theme'] = $post_meta['theme'] ?? $post_meta['_campaign_page_template'];
```
pablocubico: This caused a bunch of notices (`Undefined index: _campaign_page_template`) while importing. Probably not shown in production. 🤔

PR #986, opened 2020-01-03
PLANET-4429 Give `list_users` capability to editors

Inwerpsel:
This enables them to change the author from the document sidebar. By
default wordpress doesn't require this permission, however the Wordfence
plugin adds a check on this.

This does mean that editors are able to access the `users` menu item and see (but not edit) users of all roles. I asked Gabor in the ticket if this is an issue for us.

### DIFF
--- a/classes/class-p4-activator.php
+++ b/classes/class-p4-activator.php
@@ -39,5 +39,11 @@ class P4_Activator {
 
 		$campaigner = new P4_Campaigner();
 		$campaigner->register_role_and_add_capabilities();
+
+		// Needed to allow the editor rule to change the author of a post in the document sidebar. The users data for that
+		// control is fetched using the REST API, where WordPress by default doesn't perform a permissions check, however
+		// the Wordfence plugin adds this check in `\wordfence::jsonAPIAuthorFilter`.
+		$roles = wp_roles();
+		$roles->add_cap( 'editor', 'list_users' );
 	}
 }


functions.php
```diff
@@ -37,3 +37,9 @@ function( $template ) {
 
 require_once __DIR__ . '/classes/class-p4-loader.php';
 P4_Loader::get_instance();
+
+// Needed to allow the editor rule to change the author of a post in the document sidebar. The users data for that
+// control is fetched using the REST API, where WordPress by default doesn't perform a permissions check, however
+// the Wordfence plugin adds this check in `\wordfence::jsonAPIAuthorFilter`.
+$roles = wp_roles();
```
sagarsdeshmukh: we may add this piece of code at [here](https://github.com/greenpeace/planet4-master-theme/blob/develop/classes/class-p4-activator.php#L32) in function

Inwerpsel: Doesn't that only get run when the theme is switched? see https://github.com/greenpeace/planet4-master-theme/blob/develop/classes/class-p4-activator.php#L25

sagarsdeshmukh: Yes it execute on theme switch as well as once on every deploy (in post deploy execution script) [https://github.com/greenpeace/planet4-base-fork/blob/develop/tasks/post-deploy/01-run-p4-activator.sh#L8]

More details about the post deploy P4 activator class is in [PLANET-4050](https://jira.greenpeace.org/browse/PLANET-4050)

PR #987, opened 2020-01-03
PLANET-4524: CPP: Rogue space at top of page

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-4524

### DIFF
--- a/assets/src/scss/pages/_page.scss
+++ b/assets/src/scss/pages/_page.scss
@@ -64,14 +64,18 @@ div.page-template {
     margin-right: auto;
     display: block;
   }
+}
 
-  .single-campaign & {
-    section:first-child {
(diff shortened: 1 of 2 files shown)


assets/src/scss/pages/_page.scss
```diff
@@ -64,12 +64,16 @@ div.page-template {
     margin-right: auto;
     display: block;
   }
+}
 
-  .single-campaign & {
+.single-campaign {
+  .page-template:not(.no-page-title) {
```
Inwerpsel: Can you include this sunder `.page-template`? like `&:not(.no-page-title`

pablocubico: @Inwerpsel done!

PR #995, opened 2020-01-14
PLANET-4456 CPP export - Replace regex with parse_block()

sagarsdeshmukh:
Code refactor in export campaign functionality.

### DIFF
--- a/exporter-helper.php
+++ b/exporter-helper.php
@@ -78,34 +78,67 @@ function get_campaign_attachments( $post_ids ) {
 	foreach ( (array) $results as $text ) {
 		$text = $text->post_content;
 
-		// Filter attachment ids from caption.
-		preg_match_all( '#(wp-image-|wp-att-|attachment\_)(\d+)#', $text, $matches, PREG_SET_ORDER );
-		foreach ( $matches as $match ) {
-			$attachment_ids[] = $match[2];
-		}
+		$blocks = parse_blocks( $text );
+		foreach ( $blocks as $block ) {
 
-		// Filter attachment ids from shortcake code(shortcake_gallery, shortcake_happy_point, shortcake_media_video).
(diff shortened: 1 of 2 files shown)


exporter-helper.php
```diff
@@ -78,34 +78,64 @@ function get_campaign_attachments( $post_ids ) {
...
 
-		// Filter attachment ids from caption.
-		preg_match_all( '#(wp-image-|wp-att-|attachment\_)(\d+)#', $text, $matches, PREG_SET_ORDER );
-		foreach ( $matches as $match ) {
-			$attachment_ids[] = $match[2];
-		}
+		$blocks = parse_blocks( $text );
+		foreach ( $blocks as $block ) {
+			// Fetch background image id from EN form block.
+			if ( 'planet4-blocks/enform' === $block['blockName'] && isset( $block['attrs']['background'] ) ) {
```
Inwerpsel: You can use a switch statement here (`switch $block['blockName']`)

exporter-helper.php
```diff
@@ -78,34 +78,64 @@ function get_campaign_attachments( $post_ids ) {
...
+			// Fetch image id from Carousel Header block.
+			if ( 'planet4-blocks/carousel-header' === $block['blockName'] && isset( $block['attrs']['slides'] ) ) {
+				foreach ( $block['attrs']['slides'] as $slide ) {
+					$attachment_ids[] = $slide['image'];
+				}
+			}
+
+			// Fetch issue/tag image id from Split Two Columns block.
+			if ( 'planet4-blocks/split-two-columns' === $block['blockName'] && ( isset( $block['attrs']['issue_image'] ) || isset( $block['attrs']['tag_image'] ) ) ) {
+				$attachment_ids[] = $block['attrs']['issue_image'];
```
Inwerpsel: You probably need to use the null coalescing operator here to avoid notices, or you could split it in two if-statements (which is a bit easier when you use a switch for the block name.

PR #999, opened 2020-01-20
PLANET-4157 Fix author bio block readmore link

sagarsdeshmukh:
[JIRA 4157](https://jira.greenpeace.org/browse/PLANET-4157)

- The `wp_trim_words()` function is localized. For languages that count ‘words’ by the individual character (such as East Asian languages), the `num_words` argument will apply to the number of individual characters
- The `wp_is_mobile()` function is used to apply `author_bio_word_limit` as per device size

### DIFF
--- a/templates/blocks/author_profile.twig
+++ b/templates/blocks/author_profile.twig
@@ -6,9 +6,24 @@
 				<h5 rel="author" class="author-block-info-name">
 					<a href="{{ post.author.link }}" itemprop="name">{{ post.author.name }}</a>
 				</h5>
+
+				{% set author_bio_readmore %}
+					...  &nbsp;<a href="{{ post.author.link }}">{{ __( 'Read More', 'planet4-master-theme' ) }}</a>
+				{% endset %}
+
+				{% if ( fn('wp_is_mobile') ) %}
+					{% set author_bio_char_limit = 180 %}
+				{% else %}
+					{% set author_bio_char_limit = 300 %}
+				{% endif %}
+
+				{% set author_bio = fn('substr', post.author.description, 0, fn('strpos', post.author.description ,' ', author_bio_char_limit)) %}
+				{% set author_bio = post.author.description|length > author_bio_char_limit ? author_bio ~ author_bio_readmore : author_bio %}
+
 				<p class="author-block-info-bio" itemprop="description">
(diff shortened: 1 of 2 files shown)


templates/blocks/author_profile.twig
```diff
@@ -6,9 +6,21 @@
...
+
+				{% if ( fn('wp_is_mobile') ) %}
+					{% set author_bio_word_limit = 24 %}
+				{% else %}
+					{% set author_bio_word_limit = 50 %}
+				{% endif %}
+
 				<p class="author-block-info-bio" itemprop="description">
-					{{ fn('wpautop', fn('the_author_meta', 'description', author.id))|e('wp_kses_post')|raw }}
+					{{ fn('wp_trim_words', fn('wpautop', post.author.description ), author_bio_word_limit, author_bio_readmore)|e('wp_kses_post')|raw }}
```
Inwerpsel: Is there a function that allows to specify a number of characters and that then cuts off, but doesn't do so within a word? That way we can have a predictable amount of characters, regardless of how big the words are. Using `wp_trim_words` there is quite a big range of sizes the result could have. That could be problematic in languages that use a lot of long words like Turkish.

sagarsdeshmukh: The `substr($author_bio, 0, strpos($author_bio, ' ', 320));` will work in this case.

templates/blocks/author_profile.twig
```diff
@@ -6,9 +6,25 @@
...
+				{% endset %}
+
+				{% if ( fn('wp_is_mobile') ) %}
+					{% set author_bio_char_limit = 180 %}
+					{% set author_bio = fn('substr', post.author.description, 0, fn('strpos', post.author.description ,' ', author_bio_char_limit)) %}
+					{% set author_bio = post.author.description|length > author_bio_char_limit ? author_bio ~ author_bio_readmore : author_bio %}
+				{% else %}
+					{% set author_bio_char_limit = 300 %}
+					{% set author_bio = fn('substr', post.author.description, 0, fn('strpos', post.author.description ,' ', author_bio_char_limit)) %}
+					{% set author_bio = post.author.description|length > author_bio_char_limit ? author_bio ~ author_bio_readmore : author_bio %}
```
Inwerpsel: You can take the last 2 lines out of the if-statement as they are the same in both cases

PR #1002, opened 2020-01-23
PLANET-4685 CPP - Exclude taxonomies(tags, categories etc.) in export…

sagarsdeshmukh:
… file

[JIRA 4685](https://jira.greenpeace.org/browse/PLANET-4685)

### DIFF
--- a/exporter.php
+++ b/exporter.php
@@ -75,33 +75,6 @@ $post_ids = apply_filters( 'export_post_ids', $wpdb->get_col( $sql ), $args );  
 $cats  = [];
 $tags  = [];
 $terms = [];
-if ( isset( $term ) && $term ) {
-	$cat  = get_term( $term['term_id'], 'category' );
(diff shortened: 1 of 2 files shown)


exporter.php
```diff
@@ -75,33 +75,33 @@
...
-		if ( 0 == $t->parent || isset( $terms[ $t->parent ] ) ) {
-			$terms[ $t->term_id ] = $t;
-		} else {
-			$custom_terms[] = $t;
-		}
-	}
-
-	unset( $categories, $custom_taxonomies, $custom_terms );
-}
+//if ( isset( $term ) && $term ) {
```
comzeradd: Just a minor suggestion, to actually delete the unneeded code instead of commenting it, in order to keep the overall code more clean and readable. I understand that we may need to add taxonomies in CPP at some point, but that's why we have git, so can go back in history ;)

PR #1005, opened 2020-02-11
PLANET-4626 Update composer dev dependencies to match blocks-plugins,

sagarsdeshmukh:
Sync phpcs configuration with blocks-plugins,
Fix PHPCS errors

[JIRA 4626](https://jira.greenpeace.org/browse/PLANET-4626)

### DIFF
--- a/classes/class-p4-post-campaign.php
+++ b/classes/class-p4-post-campaign.php
@@ -40,7 +40,7 @@ if ( ! class_exists( 'P4_Post_Campaign' ) ) {
 		 */
 		public function register_campaigns_cpt() {
 
-			$labels = array(
+			$labels = [
 				'name'               => _x( 'Campaigns', 'post type general name', 'planet4-master-theme-backend' ),
 				'singular_name'      => _x( 'Campaign', 'post type singular name', 'planet4-master-theme-backend' ),
 				'menu_name'          => _x( 'Campaigns', 'admin menu', 'planet4-master-theme-backend' ),
@@ -55,9 +55,9 @@ if ( ! class_exists( 'P4_Post_Campaign' ) ) {
 				'parent_item_colon'  => __( 'Parent Campaigns:', 'planet4-master-theme-backend' ),
 				'not_found'          => __( 'No campaigns found.', 'planet4-master-theme-backend' ),
 				'not_found_in_trash' => __( 'No campaigns found in Trash.', 'planet4-master-theme-backend' ),
-			);
+			];
 
-			$args = array(
+			$args = [
 				'labels'             => $labels,
 				'description'        => __( 'Campaigns', 'planet4-master-theme-backend' ),
 				'public'             => true,
@@ -83,7 +83,7 @@ if ( ! class_exists( 'P4_Post_Campaign' ) ) {
 					// Required to expose meta fields in the REST API.
 					'custom-fields',
 				],
-			);
+			];
 
 			register_post_type( self::POST_TYPE, $args );
 
@@ -171,7 +171,7 @@ if ( ! class_exists( 'P4_Post_Campaign' ) ) {
 		 * Load assets.
 		 */
 		public function enqueue_admin_assets() {
-			wp_register_style( 'cmb-style', get_template_directory_uri() . '/admin/css/campaign.css' );
+			wp_register_style( 'cmb-style', get_template_directory_uri() . '/admin/css/campaign.css', [], '0.1' );
 			wp_enqueue_style( 'cmb-style' );
 		}
 
@@ -196,7 +196,7 @@ if ( ! class_exists( 'P4_Post_Campaign' ) ) {
 				href="#" onclick="openSidebar()"
 				id="new_sidebar_link">
 				<?php
-					echo __( 'Design settings moved to a new sidebar.', 'planet4-master-theme-backend' )
+					esc_html_e( 'Design settings moved to a new sidebar.', 'planet4-master-theme-backend' )
 				?>
 			</a>
 			<script>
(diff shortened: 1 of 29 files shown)


exporter-helper.php
```diff
@@ -45,8 +46,8 @@ function get_campaign_attachments( $post_ids ) {
 		$values         = [];
 		$values[0]      = $wpdb->postmeta;
 		$values         = array_merge( $values, $post_ids );
-		$prepared_sql   = $wpdb->prepare( $sql, $values );     // WPCS: unprepared SQL OK.
-		$results        = $wpdb->get_results( $prepared_sql ); // WPCS: unprepared SQL OK.
+		$prepared_sql   = $wpdb->prepare( $sql, $values ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
```
Inwerpsel: Probably this phpcs:ignore is not needed, as this is not a `query()` call on this line. And I would assume that for the next line it should be able to figure out that `$prepared_sql` has been prepared, so maybe that's also not needed.

exporter.php
```diff
@@ -1,32 +1,33 @@
 <?php
+// phpcs:ignoreFile
```
Inwerpsel: Do we really want to turn of CS for this entire file?

exporter.php
```diff
@@ -331,7 +332,7 @@ function p4_px_single_post_filter_postmeta( $return_me, $meta_key ) {
 			global $wp_query;
 			$wp_query->in_the_loop = true;
 			for ( $i = 0; $i < count( $post_id ); $i ++ ) {
```
Inwerpsel: Not sure if phpcs also flagged this, but here there is also a `count()` in the loop condition. You could extract it, but really the loop doesn't need to be a counter as `$i` is only used to access the current element, so it would be cleaner to instead just use `foreach( $post_ids  as $post_id ) { $next_posts = $post_id; ... }`

Inwerpsel: Also the naming of `$post_id` is wrong and should be `$post_ids`. But I see there is a naming conflict with another variable in this file and changing it might cascade into more refactors, so you can decide if you want to take the time to fix that.

Inwerpsel: By now I understand why you disabled phpcs for this file, as it's probably a lot of work to appropriately fix everything.

search.php
```diff
@@ -15,24 +15,24 @@
...
+		$selected_filters = $_GET['f'] ?? ''; // phpcs:ignore
 		$filters          = [];
 
 		// Handle submitted filter options.
 		if ( $selected_filters && is_array( $selected_filters ) ) {
-			foreach ( $selected_filters as $type => $filter_type ) {
-				foreach ( $filter_type as $name => $id ) {
-					$filters[ $type ][] = [
-						'id'   => $id,
+			foreach ( $selected_filters as $type_name => $filter_type ) {
```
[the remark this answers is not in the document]
sagarsdeshmukh: Yes, The `$type` is a WP global var, it throughs - 
`Overriding WordPress globals is prohibited. Found assignment to $type (WordPress.WP.GlobalVariablesOverride.Prohibited)`

tests/p4-testcase.php
```diff
@@ -13,17 +13,17 @@ class P4_TestCase extends WP_UnitTestCase {
...
-		require_once( get_template_directory() . '/functions.php' );
+		require_once get_template_directory() . '/functions.php';
 	}
 
 
 	/**
 	 * Use wp unit testcase factories to create data in database for the tests.
 	 */
-	function initialize_planet4_data() {
+	public function initialize_planet4_data() {
```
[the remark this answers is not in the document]
sagarsdeshmukh: Adding a linter check on public function usage would be nice idea, if you have some link/details in this context would be very helpful 👍

exporter-helper.php
sagarsdeshmukh: Yes, That's true. but phpcs still complain on this 2 lines.
```
FILE: ...ublic/wp-content/themes/planet4-master-theme/exporter-helper.php
----------------------------------------------------------------------
FOUND 2 ERRORS AFFECTING 2 LINES
----------------------------------------------------------------------
 49 | ERROR | Use placeholders and $wpdb->prepare(); found $sql
    |       | (WordPress.DB.PreparedSQL.NotPrepared)
 50 | ERROR | Use placeholders and $wpdb->prepare(); found
    |       | $prepared_sql
    |       | (WordPress.DB.PreparedSQL.NotPrepared)
----------------------------------------------------------------------
```

exporter.php
sagarsdeshmukh: Fixing all phpcs issues in this file are very prone to cause to break the export xml functionality. I check the top level issues and fixed them. 
Or 
We will make a ticket to refactor this functionality. (I personally prefer a separate task)

partials/tags-box.php
```diff
@@ -7,17 +7,17 @@
 
 ?>
 <ul id="p4_tags_list" style="height: 150px; overflow: auto; border: 1px solid #dfdfdf; margin: 1em 0; padding: .5em;">
-	<?php foreach ( $tags as $tag ) { ?>
+	<?php foreach ( $tags as $tag_obj ) { ?>
```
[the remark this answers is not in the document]
sagarsdeshmukh: I am not sure but I guess some NRO's(EA) are still using the non-Gutenberg P4 instance. If we remove this file will break their sites.
cc: @comzeradd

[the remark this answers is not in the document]
comzeradd: Yes indeed, we have 3 NROs not in Gutenberg yet.

classes/class-p4-post-campaign.php
```diff
@@ -350,8 +350,8 @@ public static function css_vars( array $meta ): array {
 			$footer_theme = $meta['campaign_footer_theme'] ?? null;
 
 			if ( 'white' === $footer_theme ) {
-				$default_footer_links_color = $meta['campaign_nav_color'] ?: '#1A1A1A';
-				$footer_links_color         = $meta['footer_links_color'] ?: $default_footer_links_color;
+				$default_footer_links_color = $meta['campaign_nav_color'] ? $meta['campaign_nav_color'] : '#1A1A1A';
```
Inwerpsel: Weird that phpcs flags this, as `$meta['campaign_nav_color'] ?: '#1A1A1A';` is just a shorthand for `$meta['campaign_nav_color'] ? $meta['campaign_nav_color'] : '#1A1A1A';`. I'm not aware of any issues with that syntax, but there might be?

sagarsdeshmukh: The `wpcs` is complaining about ternary shorthand -
`Using short ternaries is not allowed
 (WordPress.PHP.DisallowShortTernary.Found)`

PR #1006, opened 2020-02-11
PLANET-4603 Dealing with how WP deals with default values

Inwerpsel:
... of post meta fields, which is not very well. This can definitely be
cleaned up later.

Context:
- When you create a new post there will be no entries in the db for the meta fields, until they are set.
- However when in the editor you use the API `dispatch( 'core/editor' ).editPost( { meta: { [ metaKey ]: value } } )`, wordpress will actually create a key for all registered meta fields, however it doesn't use the default value specified in `register_post_meta`. Instead the entries will have an empty value.
- So I moved all the defaults into the code where they are used, which is manageable as they're mostly used in a single place, but does involve a lot of checks.

### DIFF
--- a/classes/class-p4-post-campaign.php
+++ b/classes/class-p4-post-campaign.php
@@ -87,54 +87,19 @@ if ( ! class_exists( 'P4_Post_Campaign' ) ) {
 
 			register_post_type( self::POST_TYPE, $args );
 
-			self::campaign_field(
-				'theme',
-				[ 'default' => '' ]
-			);
-			self::campaign_field(
-				'campaign_logo',
-				[ 'default' => 'greenpeace' ]
-			);
-			self::campaign_field(
-				'campaign_logo_color',
-				[ 'default' => 'light' ]
-			);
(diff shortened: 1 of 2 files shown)


classes/class-p4-post-campaign.php
```diff
@@ -352,34 +317,22 @@ private static function campaign_field(
...
-				'oceans'    => 'Montserrat',
-				'oil'       => 'Anton',
-				'plastic'   => 'Montserrat',
-			];
-
-			$campaign_font = $campaigns_font_map[ $theme ?: 'default' ];
-
-			if ( 'campaign' === $meta['campaign_body_font'] ) {
-				$body_font = $campaign_font;
+			if (
```
Inwerpsel: The campaign fonts map is something I didn't take over from the original campaigns code initially. I had created it so that the setting's value is always used. However all those values come from the css files and there is already a rule there which sets them, so it's a bit silly what I did there and we can just only apply the rule if anything but 'campaign' is chosen.

PR #1008, opened 2020-02-12
[No issue]: Hotfix - Load More in search page

pablocubico:
Updates the lazy-loaded images when clicking "Load More" or scrolling to see more results.


assets/src/js/load_more.js
```diff
@@ -1,6 +1,12 @@
 export const setupLoadMore = function($) {
```
comzeradd: can you add `;lazyLoad` as global here to make eslint happy?

PR #1009, opened 2020-02-13
PLANET-4740: CSS: Remove :not(#hidecookie) rule

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-4740

Styleguide PR: https://github.com/greenpeace/planet4-styleguide/pull/37
Blocks Plugin PR: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/215

Note: this also removes a vestigial animation, which made sense when the cookie bar was not fixed but not anymore.

cookie-animation

### DIFF
--- a/assets/src/js/cookies.js
+++ b/assets/src/js/cookies.js
@@ -27,15 +27,11 @@ export const setupCookies = function($) {
 
   if (cookie == null) {
     $('.cookie-notice').css('display', 'flex');
-    const height = $('.cookie-notice').height();
-    $('footer').css('margin-bottom', height + 'px');
   } else {
     window.createCookie('gp_nro', nro, 365);
   }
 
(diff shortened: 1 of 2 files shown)


assets/src/js/cookies.js
```diff
@@ -34,8 +34,6 @@ export const setupCookies = function($) {
   }
 
   $('#hidecookie').click(function () {
-    $('.cookie-notice').slideUp('slow');
```
comzeradd: Removing this doesn't prevent the cookie bar from being hidden on button click?

pablocubico: Yes, I don't know how I missed that. Replaced the slideUp with a simple fadeOut, not only because the slide was buggy, it also had a performance impact as it was pushing the entire layout.

PR #1012, opened 2020-02-17
PLANET-4745: Campaigns CSS: EN form block title and CTA style

pablocubico:
This is the minimum needed to close the mentioned ticket, I engaged in a few more refactors that might be out of scope, so I moved them to a separate branch, so we can close this ticket in this sprint and keep working on refactoring the campaigns code with proper testing for the next sprint.

Ref: https://jira.greenpeace.org/browse/PLANET-4745

This also needs:
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/218
https://github.com/greenpeace/planet4-plugin-gutenberg-engagingnetworks/pull/42

### DIFF
--- a/templates/css-variables.twig
+++ b/templates/css-variables.twig
@@ -5,7 +5,7 @@ Later we will use these variables in our source scss files (with a fallback to t
 <style>
 	:root {
 		{% if css_vars['header-primary-font'] %}
-			--campaigns-header-primary-font: {{ css_vars['header-primary-font'] }};
+			--header-primary-font: {{ css_vars['header-primary-font'] }};
 		{% endif %}
 	}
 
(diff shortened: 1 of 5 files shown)


pablocubico:
The CSS Vars ponyfill seems to be working on IE 11 in Browserstack, I'll try to test on a Windows box I have too.

Captura de pantalla 2020-02-19 a la(s) 19 07 50

templates/css-variables.twig
```diff
@@ -3,6 +3,12 @@ Later we will use these variables in our source scss files (with a fallback to t
  to expose variables to theme configuration files, thereby removing the need of separate selectors.
  #}
 <style>
+	:root {
+		{% if css_vars['header-primary-font'] %}
+			--campaigns-header-primary-font: {{ css_vars['header-primary-font'] }};
```
Inwerpsel: Just noticed this after approving, but probably better to not prefix this variable (and all other variables) with "campaign".

pablocubico: Indeed! Will change it

pablocubico: Changed here and in the master-theme.

PR #1013, opened 2020-02-17
PLANET-4722 Integrate Elastic Search database with search code

dpivo:
- Replaced OpenSearch retrieval of archived data with ElasticSearch.
- Docs to set up ElasticSearch index and documents: https://documenter.getpostman.com/view/7448731/SzKN12Ha

### DIFF
--- a/classes/class-p4-search.php
+++ b/classes/class-p4-search.php
@@ -170,7 +170,7 @@ if ( ! class_exists( 'P4_Search' ) ) {
 		 * Gets the paged posts that belong to the next page/load and are to be used with the twig template.
 		 */
 		public function get_paged_posts() {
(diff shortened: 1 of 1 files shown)


classes/class-p4-search.php
```diff
@@ -349,66 +349,73 @@ protected function check_cache( $cache_key, $cache_group ) {
...
+							],
+							'should'               => [
+								0 => [
+									'term' => [
+										'title' => $search_query,
+									],
+								],
+								1 => [
+									'term' => [
+										'description' => $search_query,
```
sagarsdeshmukh: I think this should be `excerpt` instead of `description `?

PR #1016, opened 2020-02-18
PLANET-4711 Rename "Campaign" to "Campaign Page" and use constants

Inwerpsel:
* Rename all labels and the class name to campaign page to prepare for
new campaign type we will be introducing.
* Use constants instead of string for all occurrences of the post type
identifier.


Inwerpsel:
Not needed anymore because of change in approach in https://jira.greenpeace.org/browse/PLANET-4712

PR #1019, opened 2020-02-21
PLANET-4712 Multi page campaigns

Inwerpsel:
This required me to change the permalink settings to "post name" instead of the custom structure, else campaigns with a parent would navigate to `/parent-slug/own-slug` but that page would give a 404. So if you run locally you will need to do that as well. For now that is probably breaking other aspects of the site, though I didn't run into them yet.


pablocubico:
Other than that, looks great, tested locally creating multipage campaigns with a few nested pages. 🚀

Inwerpsel:
> This looks good to me, but I wonder if the permalinks change has any side effects, maybe @comzeradd or @sagarsdeshmukh can jump in?

Before we only had the case where the url was of the format `campaign/{campaign_name}`, and that still works. The `use_front` option is used to set whether wordpress tries to apply the rewriting rule set in `wp-admin/options-permalink.php`, so before it was trying to match the url to that, but never did as the value for it on all our sites is `/%p4_page_type%/%post_id%/%postname%/`, which doesn't match urls like `campaign/{campaign_name}`, so had no effect anyway.

PR #1020, opened 2020-02-21
PLANET-4784: Refactor Campaigns themes codebase

pablocubico:
WIP.

_Note: This is based on PLANET-4745._

Using this doc as the ultimate source of truth:
https://docs.google.com/document/d/1eciWb_8FHF4qmwPGNtHFtmf9zGsDCF3bGnwJhcUx13M/edit#

So, I'm not there yet, this is a bit broken, but let's sync before moving on, as I have other tickets to do in the meantime.

These are the ideas behind these changes, based on our conversations with Pieter:
- Turning the `css_vars` method (the function resolving the theme) into a "high level", kinda "pseudo code" function, short and readable, encapsulating whatever logic we currently have into separate functions for each case (footer theme, "passive buttons", etc). So we can review that logic in isolation and deprecate as needed. The general approach is:
  - Read the theme from a JSON file
  - Blend with the meta variables
  - Resolve the defaults and the meta into flat CSS vars.
  - These vars will be used directly in the Planet4 theme.
  - Eventually (and ideally) we'll get rid of the specific SCSS for campaigns.
- Turning the CSS variables Twig file into a dumb for loop. Variables are resolved before in PHP.
- Resolving legacy variable names into new ones before we pass them to the CSS.
- Use the new variable names as needed in the default theme code.
- Move extended customizations (e.g.: gallery slider) into new, possibly hidden variables in the JSON files, so they can describe all cases.
- Also trying to find more readable ways to express the conditions if possible.

Not sure if I'm forgetting something, I'll keep posting here any updates.


assets/src/scss/style.scss
```diff
@@ -79,7 +79,7 @@ Text Domain: planet4-master-theme
 @import "vendors/gtm";
 
 // Campaign overrides
-@import "campaigns/campaign_footer";
+// @import "campaigns/campaign_footer";
```
pablocubico: I'm evaluating what to do with this.

campaign_themes/arctic.json
```diff
@@ -18,7 +18,8 @@
 						}
 					],
 					"default": "greenpeace"
-				}
+				},
+				"default": "minimal"
```
pablocubico: I added a default to the configurations 🤔

campaign_themes/arctic.json
```diff
@@ -186,5 +189,17 @@
 				}
 			}
 		}
-	]
+	],
+	"hidden_fields": {
```
pablocubico: I'm hardcoding these to the JSON files just to make the switch. Not sure if we need more than that for now.

campaign_themes/arctic.json
```diff
@@ -186,5 +189,17 @@
...
+		"gallery": {
+			"padding": "20px",
+			"caption-background": "red",
+			"left-arrow-asset": "arrow-left.svg",
+			"right-arrow-asset": "arrow-right.svg",
+			"arrows-bottom-offset": "30px",
+			"indicators-style": "round",
+			"passive-indicator-color": "blue",
```
pablocubico: These are not the real values yet!

classes/class-p4-post-campaign.php
```diff
@@ -315,98 +315,163 @@ private static function campaign_field(
 			register_post_meta( self::POST_TYPE, $meta_key, $args );
 		}
 
+		private static function get_theme_defaults( $theme_json ) {
```
pablocubico: Pretty dumb function, could get smarter, I'm encapsulating just in case.

classes/class-p4-post-campaign.php
```diff
@@ -315,98 +315,163 @@ private static function campaign_field(
...
-				'default'   => 'lora',
-				'antarctic' => 'sanctuary',
-				'arctic'    => 'Save the Arctic',
-				'climate'   => 'Jost',
-				'forest'    => 'Kanit',
-				'oceans'    => 'Montserrat',
-				'oil'       => 'Anton',
-				'plastic'   => 'Montserrat',
-			];
+			$theme = self::get_theme( $meta );
```
pablocubico: I moved the "theme" vs "_campaign_page_template" thing to a method.

classes/class-p4-post-campaign.php
```diff
@@ -315,98 +315,163 @@ private static function campaign_field(
...
 
-			$theme = empty( $meta['theme'] )
-								? 'default'
-								: $meta['theme'];
+			// TODO: Use wp_safe_remote_get?
+			// TODO: Handle errors
+			$theme_json = json_decode(
+				file_get_contents( __DIR__ . '/../campaign_themes/' . $theme . '.json' ),
```
pablocubico: Using this just for dev now, added a few TODOs above.

classes/class-p4-post-campaign.php
```diff
@@ -315,98 +315,163 @@ private static function campaign_field(
...
-															: $meta['campaign_header_primary'];
+			$defaults = self::get_theme_defaults( $theme_json );
 
-			$body_font = empty( $meta['campaign_body_font'] )
-										? $campaigns_font_map[ $theme ]
-										: $meta['campaign_body_font'];
+			// Replace the defaults with the campaign options where applicable
+			// 'array_intersect_key' removes the keys from $meta that are not present in the $defaults.
+			// 'array_filter' removes empty (falsey?) values from the resulting array.
+			$css_vars = array_filter( array_intersect_key( $meta, $defaults ) );
```
pablocubico: Not sure how we'll handle the defaults. Will there always be a default value? No? What's the best strategy? 🤔

classes/class-p4-post-campaign.php
```diff
@@ -315,98 +315,163 @@ private static function campaign_field(
...
 
-			// TODO: Remove this special case.
-			$header_primary_font = str_replace( 'Montserrat_Light', 'Montserrat', $header_primary_font );
-			$body_font           = str_replace( 'Montserrat_Light', 'Montserrat', $body_font );
+			$css_vars = array_merge( $css_vars, self::get_footer_theme( $meta ) );
```
pablocubico: Isolated all these logic bits into functions that we can review and possibly deprecate in the future.

classes/class-p4-post-campaign.php
```diff
@@ -315,98 +315,163 @@ private static function campaign_field(
...
+		}
+
+		/**
+		 * DEPRECATE: Replaces the legacy mapping to add missing or composed variables,
+		 * returns an iterable flat set of variables.
+		 *
+		 * @param array $css_vars The array containing the CSS variables.
+		 * @return array The variables for the footer theme.
+		 */
+		public static function transform_legacy_mapping( array $css_vars ): array {
```
pablocubico: As these variables may not represent the new scheme, I'll say stop using them at this point (before rendering), so we can transform the meta values accordingly if they are old, but render new names with a polished scheme since now, and use those in the SCSS files.

classes/class-p4-post-campaign.php
```diff
@@ -315,98 +315,163 @@ private static function campaign_field(
...
+			return $css_vars;
+		}
+
+		/**
+		 * Get the theme based on the meta settings.
+		 *
+		 * @param array $meta The meta containing the style settings.
+		 * @return string The identifier of the theme.
+		 */
+		public static function get_theme( array $meta ): string {
```
pablocubico: Moved this to a method.

campaign_themes/default.json
```diff
@@ -43,7 +43,8 @@
 					],
 					"default": "greenpeace"
```
pablocubico: If the assumption is "there is no default theme", then we should remove this file.

Inwerpsel: Well currently this file contains the allowed values for certain style settings if no theme is selected. So it makes sense to have a config file for the default theme (but without "hidden_fields"), else if we wanted to provide these options the default theme would have to store it somewhere in the code in another way than the actual themes.

campaign_themes/arctic.json
Inwerpsel: Be aware that these defaults currently need to be matched with the backend logic, as we still need to clean up handling of default values so that everything can be in one place in the config files.

campaign_themes/arctic.json
Inwerpsel: Maybe it would make sense to not have this as a separate property, but instead combine them with the configurable fields, making the distinction on the fields themselves using an `editable` attribute or something. That will make it easier and more clear in version history when a field changes from being hidden to being configurable.

Inwerpsel: Maybe it would make sense to separate the fields that contain the value of a css variable, and those that contain values that are used in some backend logic (such as logo and navigation type).

Inwerpsel: Then we would have a property which contains only css variables that have the variable name as a key.

classes/class-p4-post-campaign.php
```diff
@@ -315,98 +315,163 @@ private static function campaign_field(
...
-				$footer_color = $meta['campaign_nav_color'] ?? null;
+			if ( isset( $css_vars['campaign_body_font'] ) ) {
+				$css_vars['campaign_body_font'] = str_replace( 'Montserrat_Light', 'Montserrat', $css_vars['campaign_body_font'] );
 			}
 
+			return $css_vars;
+		}
+
+		/**
+		 * DEPRECATE: Get the passive button color based on the meta settings.
```
Inwerpsel: You can use the `@deprecated` annotation, then IDE's will highlight this wherever it's used.

Inwerpsel: Also best to include explanation why it's deprecated and what the alternative is.

pablocubico:
Closing this as its outdated and many of these things were actually done.

PR #1024, opened 2020-02-26
PLANET-4703 Use styleguide with checkbox fix

Inwerpsel:


comzeradd:
This change seems to affect checkboxes on post comments
eg. https://k8s.p4.greenpeace.org/test-venus/story/28643/tackling-air-pollution-through-people-power/

Inwerpsel:
> This change seems to affect checkboxes on post comments
> eg. https://k8s.p4.greenpeace.org/test-venus/story/28643/tackling-air-pollution-through-people-power/

Nice catch. I had checked the twig files where all of the custom checkboxes are in `.custom-control` wrapper, but for the comment section it is added in PHP so I had missed it. Changed that to also add the wrapper class there.

PR #1026, opened 2020-02-28
PLANET-3427 Prevent wordpress autosave recreation

Inwerpsel:
* Due to a bug in wordpress the autosave record gets created and deleted
many times when clicking preview without changes to revisioned
attributes. Add a filter so that the autosave record is only ever
deleted if the post is permanently deleted.

### DIFF
--- a/classes/class-p4-loader.php
+++ b/classes/class-p4-loader.php
@@ -52,6 +52,7 @@ final class P4_Loader {
 	private function __construct( $services ) {
 		$this->load_files();
 		$this->load_services( $services );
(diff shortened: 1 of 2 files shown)


comzeradd:
Is there a ticket on WP for this bug?

Inwerpsel:
> Is there a ticket on WP for this bug?

Yes, I created a core ticket and a gutenberg.
https://core.trac.wordpress.org/ticket/49532#ticket
https://github.com/WordPress/gutenberg/issues/20505

PR #1028, opened 2020-03-03
PLANET-4751 Refactor campaign exporter module

sagarsdeshmukh:
[JIRA 4751](https://jira.greenpeace.org/browse/PLANET-4751)

### DIFF
--- a/exporter-helper.php
+++ b/exporter-helper.php
@@ -1,19 +1,20 @@
 <?php
 /**
- * Post Exporter Helper Function
+ * Campaign Exporter Helper Functions
  *
  * @package P4MT
  */
+
 /**
  * Generate a bunch of placeholders for use in an IN query.
  * Such a shame WordPress doesn't offer a decent way to do bind IN statement params and we have to do this ourselves.
  *
  * @param array $items The items to generate placeholders for.
- * @param int $start_index The start index to use for creating the placeholders.
+ * @param int   $start_index The start index to use for creating the placeholders.
  * @return string The generated placeholders string.
  */
-function generate_list_placeholders(array $items, int $start_index): string {
-	$placeholders   = [];
+function generate_list_placeholders( array $items, int $start_index ): string {
+	$placeholders = [];
 	foreach ( range( $start_index, count( $items ) + $start_index - 1 ) as $i ) {
 		$placeholders[] = "%$i\$d";
 	}
@@ -32,10 +33,8 @@ function get_attachments_used_in_content( string $content ): array {
 	$attachment_ids = [];
 
 	foreach ( $blocks as $block ) {
-
 		// Fetch the attachement id/s from block fields.
 		switch ( $block['blockName'] ) {
-
 			case 'planet4-blocks/enform':
 				$attachment_ids[] = $block['attrs']['background'] ?? '';
 				break;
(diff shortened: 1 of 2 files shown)


exporter-helper.php
```diff
@@ -1,10 +1,101 @@
...
+ *
+ * @param string $content The content to parse.
+ * @return array All attachments used in the blocks.
+ */
+function get_attachments_used_in_content( string $content ): array {
+	$blocks = parse_blocks( $content );
+
+	$attachment_ids = [];
+
+	foreach ( $blocks as $block ) {
```
Inwerpsel: I just found out something using the Inspections PhpStorm plugin: if a foreach contains only a switch statement, you can combine both and only have one set of square brackets instead of two.
`foreach ( $blocks as $block ) switch ( $block['blockName']) {
			case 'planet4-blocks/enform':
				$attachment_ids[] = $block['attrs']['background'] ?? '';
				break;

			case 'core/image':
    // ...
}

Inwerpsel: The nice thing about it is that from the first line you can see the foreach only has a switch in it, whereas otherwise you would need to scroll past the switch to see that there is nothing more in the function.

exporter-helper.php
```diff
@@ -15,139 +106,131 @@ function get_campaign_attachments( $post_ids ) {
...
-		$prepared_sql   = $wpdb->prepare( $sql, $values ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
-		$results        = $wpdb->get_results( $prepared_sql ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
-		$attachment_ids = [];
-		foreach ( (array) $results as $result ) {
-			$attachment_ids[] = $result->meta_value;
-		}
-	}
+	$sql = '
+		SELECT meta_value
+		FROM %1$s
```
Inwerpsel: I'm not sure if there is actually much value in using a placeholder for the table name when making custom queries. Realistically the wordpress table names are never going to change, and the placeholders do obfuscate and complicate the code quite a lot. Though it's done like that throughout the codebase so no need to change it here if you want to stay consistent.

exporter-helper.php
```diff
@@ -15,139 +106,131 @@ function get_campaign_attachments( $post_ids ) {
...
+	}
 
-		$blocks = parse_blocks( $text );
-		foreach ( $blocks as $block ) {
+	$attachment_ids = array_unique( $attachment_ids );
+	sort( $attachment_ids );
 
-			// Fetch the attachement id/s from block fields.
-			switch ( $block['blockName'] ) {
+	// The post ids are reorderd as sort all attachment ids first and then append the post id to array.(Added for simplification of import process).
```
Inwerpsel: Small typo in "reordered" here ;)

exporter.php
```diff
@@ -1,257 +1,42 @@
...
+require_once 'exporter-helper.php';
+
 define( 'CMX_VERSION', '1.0' );
 $sitename = sanitize_key( get_bloginfo( 'name' ) );
 if ( ! empty( $sitename ) ) {
 	$sitename .= '.';
 }
-$filename = $sitename . gmdate( 'Ymdhis' ) . '.xml';
+
+// Added "P4-Campaign" word in XML filename to distinguish between normal WP export and P4 campaign export.
```
Inwerpsel: Awesome, I was going to suggest changing the filename as indeed it was not very descriptive, but you already did that. Maybe we can go even further and add the campaign slug to the filename?

exporter.php
```diff
@@ -1,257 +1,42 @@
...
-	$taxonomies = get_object_taxonomies( $post->post_type );
-	if ( empty( $taxonomies ) ) {
-		return;
-	}
-	$terms = wp_get_object_terms( $post->ID, $taxonomies );
-
-	foreach ( (array) $terms as $term ) {
-		echo "\t\t<category domain=\"{$term->taxonomy}\" nicename=\"{$term->slug}\">" . p4_px_single_post_cdata( $term->name ) . "</category>\n";
-	}
-}
```
Inwerpsel: Nice, moving the these functions out of this file really makes it a lot more manageable :100:

exporter-helper.php
sagarsdeshmukh: I use wp var for table name because wordpress offer custom prefix option for table name at time of installation. if some external entity want to use our campaign feature they may not restrict with hardcoded table names. its an edge case but we cover it :)

exporter.php
sagarsdeshmukh: Thanks Pieter :) I also think about adding slug in file name but we also provide option to export campaign in bulk.(select checkboxes and from top dropdown choose export option)

exporter-helper.php
sagarsdeshmukh: Good catch 👍

exporter-helper.php
sagarsdeshmukh: It looks more cleaner way, but `phpcs` won't like it :(

exporter.php
Inwerpsel: Maybe we can check if they're exporting more than one campaign? And only add the campaign slug if they are exporting exactly one campaign?

exporter.php
```diff
@@ -1,257 +1,51 @@
...
-
-	if ( $args['start_date'] ) {
-		$where .= $wpdb->prepare( " AND {$wpdb->posts}.post_date >= %s", date( 'Y-m-d', strtotime( $args['start_date'] ) ) );
-	}
-
-	if ( $args['end_date'] ) {
-		$where .= $wpdb->prepare( " AND {$wpdb->posts}.post_date < %s", date( 'Y-m-d', strtotime( '+1 month', strtotime( $args['end_date'] ) ) ) );
-	}
+	// Add "P4-Campaign" word in XML filename to distinguish between normal WP export and P4 campaign export.
+	$filename = $sitename . 'P4-Campaign.' . gmdate( 'Y-m-d' ) . '.xml';
```
Inwerpsel: Maybe now we have the name for exports of single campaigns, we can make this plural?

PR #1031, opened 2020-03-09
PLANET-4561: Make sure P4 Spreadsheet embed (New Block) respects all Campaign Styles

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-4561


campaign_themes/antarctic.json
```diff
@@ -152,7 +153,21 @@
 						{ "color": "#52c9f4" }
 					]
 				}
-			}
+			},
+			"default": "white"
+		},
+		{
+			"id": "spreadsheet",
+			"children": [
```
pablocubico: Not sure about the naming, but this is to create a set of fields. The CSS Vars are a flat list but the JSON file doesn't need to be, so this is to avoid adding a `spreadsheet-` prefix to all of these values.

classes/class-p4-post-campaign.php
```diff
@@ -338,100 +338,203 @@ private static function campaign_field(
...
+			);
 
-			$theme = $meta['theme'] ?? $meta['_campaign_page_template'] ?? null;
-			$theme = $theme ? $theme : 'default';
+			$defaults = self::get_theme_defaults( $theme_json );
 
-			$header_primary_font = empty( $meta['campaign_header_primary'] )
-															? $campaigns_font_map[ $theme ]
-															: $meta['campaign_header_primary'];
+			// $body_font = $meta['campaign_body_font'] ?? null;
```
pablocubico: I forgot about the Body font!

campaign_themes/antarctic.json
pablocubico: After discussing via chat with @Inwerpsel I'm removing this key and the flattenizing function, I propose we stop any refactors until we settle what the structure of this file will be via this ADR: https://docs.google.com/document/d/1eciWb_8FHF4qmwPGNtHFtmf9zGsDCF3bGnwJhcUx13M/edit# ,  when we have more input from the product side. cc @comzeradd

assets/src/scss/style.scss
```diff
@@ -79,7 +79,7 @@ Text Domain: planet4-master-theme
 @import "vendors/gtm";
 
 // Campaign overrides
-@import "campaigns/campaign_footer";
+// @import "campaigns/campaign_footer";
```
comzeradd: if we don't need this, let's delete the line completely.

classes/class-p4-post-campaign.php
```diff
@@ -338,100 +338,182 @@ private static function campaign_field(
...
+		}
+
+		/**
+		 * DEPRECATE: Replaces the legacy mapping to add missing or composed variables,
+		 * returns an iterable flat set of variables.
+		 *
+		 * @param array $css_vars The array containing the CSS variables.
+		 * @return array The variables for the footer theme.
+		 */
+		public static function transform_legacy_mapping( array $css_vars ): array {
```
comzeradd: Some adjustments to the equal signs spacing needed here to make `phpcs` happy

PR #1036, opened 2020-03-17
PLANET-4860 Fix a bug catch by Sentry Issue: 908

sagarsdeshmukh:
[JIRA 4860](https://jira.greenpeace.org/browse/PLANET-4860)

Sentry Issue: [908](https://sentry.greenpeace.org/greenpeace/planet4/issues/908/)

### DIFF
--- a/exporter-helper.php
+++ b/exporter-helper.php
@@ -71,7 +71,7 @@ function get_attachments_used_in_content( string $content ): array {
 			case 'planet4-blocks/columns':
 				if ( isset( $block['attrs']['columns'] ) ) {
 					foreach ( $block['attrs']['columns'] as $column ) {
-						$attachment_ids[] = $column['attachment'];
+						$attachment_ids[] = $column['attachment'] ?? '';
 					}
 				}
 				break;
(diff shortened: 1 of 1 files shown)


exporter-helper.php
```diff
@@ -118,7 +118,13 @@ function get_campaign_attachments( $post_ids ) {
 		$sql, // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
 		array_merge( [ $wpdb->posts ], $post_ids )
 	);
-	$attachment_ids = $wpdb->get_results( $prepared_sql, OBJECT_K ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+	$results        = $wpdb->get_results( $prepared_sql ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
```
Inwerpsel: Maybe you can instead use a function that ensures the result set comtains an integer and not an object? Then you wouldn't need the `array_map` below.

sagarsdeshmukh: @Inwerpsel - can we move with current changes for this release and work on improvements in coming sprint?

PR #1038, opened 2020-03-19
PLANET-4869 Remove unnecessary css code from login page

comzeradd:
This comes as a follow-up from upgrade the Google Apps Login plugin, since the link text was replaced by an image.

This also move the css file into the admin stylesheets folder. And uses a better planet4 image.


comzeradd:
A [live example](https://k8s.p4.greenpeace.org/test-titan/wp-login.php) of this

admin/css/login.css
```diff
@@ -52,26 +35,19 @@ body.login div#login form#loginform {
...
-  vertical-align: middle;
-  border-radius: 0px;
   padding: 0;
-  height: auto;
-  font-size: 14px;
-  margin-bottom: 6px;
+}
+
+form#loginform p.galogin a {
+  line-height: initial;
```
pablocubico: Hey @comzeradd this seems to cause the text to be not centered vertically (at least in Chrome), I guess you could make it `27px` which is the exact height of the button. I would also add `text-decoration: none` to remove the underline.

pablocubico: Ignore this, my plugin was outdated..

PR #1039, opened 2020-03-23
PLANET-4827 Archive as a post type

Inwerpsel:
NOTE TO REVIEWER:
This is a pretty big PR, but most of the changes are in `classes/class-p4-search.php`. I suggest first going through the other files, as those are easier to understand in isolation, and then proceed to `classes/class-p4-search.php`.

I removed the caching of search results, as that was mainly for reducing the footprint of loading all results in memory, which is not needed anymore as aggregations are used instead of looping through all posts to get the totals data.

* Add custom post type for archive.
* Setup search for that post type.
* Make decay params configurable (mainly for making it easier to find good values for these during development, but maybe we can preserve these for NROs to configure).
* Add settings for including archived content for logged in users only,
or publicly.

* Convenience function `simple_value_filter` for when a filter is used just to provide a
value, so that this value can be provided without having to pass a
function.

* Specify which fields to return from ElasticSearch. Otherwise all
data, including post_content is loaded in memory even though it's
not used.

* Use ElasticSearch aggregations for tag and category data, so that we
can limit the amount of posts that are fetched to the a mount per page.

* Remove the filter for the main query. We should ensure that only one
query is used (by either using the main query for search, or by ensuring
the main query is not executed), but that can be taken up later.

* Fix decaying not being applied in ajax requests.
* Fix date ordering not working in ajax requests.
* Fix mime type filter not being applied in ajax requests.

### DIFF
--- a/classes/class-p4-master-site.php
+++ b/classes/class-p4-master-site.php
@@ -85,7 +85,6 @@ class P4_Master_Site extends TimberSite {
 		add_filter( 'get_twig', [ $this, 'add_to_twig' ] );
 		add_action( 'init', [ $this, 'register_taxonomies' ], 2 );
 		add_action( 'init', [ $this, 'register_oembed_provider' ] );
-		add_action( 'pre_get_posts', [ $this, 'add_search_options' ] );
 		add_action( 'admin_enqueue_scripts', [ $this, 'enqueue_admin_assets' ] );
 		add_action( 'wp_enqueue_scripts', [ $this, 'enqueue_public_assets' ] );
 		add_filter( 'safe_style_css', [ $this, 'set_custom_allowed_css_properties' ] );
@@ -114,6 +113,9 @@ class P4_Master_Site extends TimberSite {
 		add_action( 'after_setup_theme', [ $this, 'add_image_sizes' ] );
 		add_action( 'save_post', [ $this, 'p4_auto_generate_excerpt' ], 10, 2 );
 
+		if ( wp_doing_ajax() ) {
+			P4_Search::add_general_filters();
+		}
 		add_action( 'wp_ajax_get_paged_posts', [ 'P4_ElasticSearch', 'get_paged_posts' ] );
 		add_action( 'wp_ajax_nopriv_get_paged_posts', [ 'P4_ElasticSearch', 'get_paged_posts' ] );
 
@@ -639,20 +641,6 @@ class P4_Master_Site extends TimberSite {
 	 */
 	public function register_oembed_provider() {
 		wp_oembed_add_provider( '#https?://(?:www\.)?[^/^\.]+\.carto(db)?\.com/\S+#i', 'https://services.carto.com/oembed', true );
-	}
-
-	/**
-	 * Add custom options to the main WP_Query.
-	 *
-	 * @param WP_Query $wp The WP Query to customize.
-	 */
-	public function add_search_options( WP_Query $wp ) {
-		if ( ! $wp->is_main_query() || ! $wp->is_search() ) {
-			return;
-		}
-
-		$wp->set( 'posts_per_page', P4_Search::POSTS_LIMIT );
-		$wp->set( 'no_found_rows', true );
 	}
 
 	/**
--- a/classes/class-p4-post-archive.php
+++ b/classes/class-p4-post-archive.php
@@ -1,0 +1,90 @@
+<?php
+/**
+ * A custom post type for P3 posts that were archived.
+ *
+ * @package P4MT
+ */
+
+/**
+ * A custom post type for P3 posts that were archived.
+ */
+class P4_Post_Archive {
+
+	public const POST_TYPE = 'archive';
+
+	/**
+	 * The constructor.
+	 */
+	public function __construct() {
+		$this->hooks();
+	}
+
+	/**
+	 * Class hooks.
+	 */
+	private function hooks() {
+		add_action( 'init', [ $this, 'register_archive_cpt' ] );
+		add_action( 'add_meta_boxes_archive', [ $this, 'add_archive_link' ] );
+	}
+
+	/**
+	 * Register the custom post type.
+	 */
+	public function register_archive_cpt() {
+		$args = [
+			'labels'             => [
+				'name'               => _x( 'Archive', 'post type general name', 'planet4-master-theme-backend' ),
+				'singular_name'      => _x( 'Archive', 'post type singular name', 'planet4-master-theme-backend' ),
+				'menu_name'          => _x( 'Archive', 'admin menu', 'planet4-master-theme-backend' ),
+				'name_admin_bar'     => _x( 'Archive', 'add new on admin bar', 'planet4-master-theme-backend' ),
+				'add_new'            => _x( 'Add New', 'archive', 'planet4-master-theme-backend' ),
+				'add_new_item'       => __( 'Add New archive', 'planet4-master-theme-backend' ),
+				'new_item'           => __( 'New archive', 'planet4-master-theme-backend' ),
+				'edit_item'          => __( 'Edit archive', 'planet4-master-theme-backend' ),
+				'view_item'          => __( 'View archive', 'planet4-master-theme-backend' ),
+				'all_items'          => __( 'All archived posts', 'planet4-master-theme-backend' ),
+				'search_items'       => __( 'Search archives', 'planet4-master-theme-backend' ),
+				'parent_item_colon'  => __( 'Parent archive:', 'planet4-master-theme-backend' ),
+				'not_found'          => __( 'No archives found.', 'planet4-master-theme-backend' ),
+				'not_found_in_trash' => __( 'No archives found in Trash.', 'planet4-master-theme-backend' ),
+			],
+			'description'        => __( 'Archive', 'planet4-master-theme-backend' ),
+			'public'             => true,
+			'publicly_queryable' => false,
+			'show_ui'            => true,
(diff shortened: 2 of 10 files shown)


classes/class-p4-search.php
```diff
@@ -305,14 +353,13 @@ public function get_posts( $paged = 1 ) : array {
 		 */
 		protected function get_general_args( $paged ): array {
 			$args = [
-				'posts_per_page' => self::POSTS_LIMIT,          // Set a high maximum because -1 will get ALL posts and this can be very intensive in production.
-				'no_found_rows'  => true,                       // This means that the result counters of each filter might not be 100% precise.
```
Inwerpsel: "might not be 100% precise"... well that's an understatement :D

classes/class-p4-search.php
```diff
@@ -617,15 +704,15 @@ protected function set_filters_context( &$context ) {
 			}
 
 			// Sort associative array with filters alphabetically.
-			if ( $context['categories'] ) {
+			if ( $context['categories'] ?? false ) {
```
Inwerpsel: Probably doesn't occur in prod, but I set up a new language which didn't have categories and this was producing warnings.

classes/class-p4-search.php
```diff
@@ -265,10 +286,29 @@ protected function get_timber_posts( $paged = 1 ) : array {
...
+						$archive_post->title = $post->post_title;
+						$archive_post->link = $post->guid;
+						$archive_post->post_type = P4_Post_archive::POST_TYPE;
+						$archive_post->post_date = $post->post_date_gmt;
+						$timber_posts[] = $archive_post;
+					} else {
+						// TODO: Not use TimberPost as that will query the data again. We just called ElasticSearch,
+						// so if we need more data, we just need to pull it from there (might require setting up more
+						// indexing.
+						$timber_post = new TimberPost( $post->ID );
```
Inwerpsel: Notice how the `TimberPost` that is passed to the template is created with only the ID from ElasticSearch as input. This means that, even though we got most of the data from ES, this will just reload it from the db again, which kinda defeats the point of using ElasticSearch. After the change in this PR this luckily only happens for the 5 posts that are being displayed. 

But currently, in production, if you search for something that matches over 300 results, it will fetch all 300 of them again, one query at a time, even though it already has all data.

classes/class-p4-master-site.php
```diff
@@ -85,7 +85,6 @@ protected function hooks() {
 		add_filter( 'get_twig', [ $this, 'add_to_twig' ] );
 		add_action( 'init', [ $this, 'register_taxonomies' ], 2 );
 		add_action( 'init', [ $this, 'register_oembed_provider' ] );
-		add_action( 'pre_get_posts', [ $this, 'add_search_options' ] );
```
Inwerpsel: This only affects the main query, however our search doesn't use the main query (even though it's still executed) so there's no use in adding this action any more.

templates/search.twig
```diff
@@ -12,6 +12,9 @@
 			<div class="without-search-result row clearfix">
 				<div class="col-md-12">
 					<h2 class="result-statement">{{ page_title|e('wp_kses_post')|raw }}</h2>
+					{% if query_time %}
```
Inwerpsel: Mainly for checking the performance of these queries on develop sites, only shown for logged in users but we probably will remove this again.

classes/class-p4-settings.php
```diff
@@ -18,13 +18,6 @@ class P4_Settings {
 		 */
 		private $key = 'planet4_options';
 
-		/**
-		 * Array of metaboxes/fields
-		 *
-		 * @var array
-		 */
-		protected $option_metabox = [];
-
```
Inwerpsel, line 27: This was not used anymore.

classes/class-p4-search.php
```diff
@@ -27,12 +24,6 @@ abstract class P4_Search {
 		const DEFAULT_CACHE_TTL     = 600;
 		const DUMMY_THUMBNAIL       = '/images/dummy-thumbnail.png';
 		const EXCLUDE_FROM_SEARCH   = 'p4_do_not_index';
-		const POST_TYPES            = [
```
Inwerpsel, line 30: Replaced this with https://github.com/greenpeace/planet4-master-theme/pull/1039/files#diff-8332426ee2026a248f5fc7cf275911c3R515

classes/class-p4-search.php
```diff
@@ -293,7 +401,25 @@ public function get_posts( $paged = 1 ) : array {
 			// Set Engine Query arguments.
 			$this->set_engines_args( $args );
 
-			$posts = ( new WP_Query( $args ) )->posts;
+			add_filter(
+				'ep_es_query_results',
+				function ( $result ) {
+					$this->total_matches = $result['found_documents'];
```
Inwerpsel, line 407: check if this produces warning if ES fails

classes/class-p4-search.php
```diff
@@ -665,80 +901,66 @@ protected function set_results_context( &$context ) {
...
+						$content_type_text = __( 'Archive', 'planet4-master-theme' );
+						$content_type      = 'archive';
 						break;
 					default:
 						continue 2;     // Ignore other post_types and continue with next $post.
 				}
-				$context['posts_data'][ $post->ID ]['content_type_text'] = $content_type_text;
-				$context['posts_data'][ $post->ID ]['content_type']      = $content_type;
+
+				if ( ! isset( $post->ID ) ) {
```
Inwerpsel, line 929: can probably be removed

sagarsdeshmukh:
Initially when we hit the search page(https://k8s.p4.greenpeace.org/international/?s=&orderby=_score), 
- The top search count show’s “0 results”
- All left side bar counters are zero and no checkboxes are clickable
- “Show 5 more results” option is not visible

sagarsdeshmukh:
The elastic search sync is throwing error -
Screen Recording 2020-04-08 at 12.03.15 PM.mov
(https://k8s.p4.greenpeace.org/international/wp-admin/admin.php?page=elasticpress)

sagarsdeshmukh:
Need phpcs fixes - https://circleci.com/gh/greenpeace/planet4-master-theme/6943?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

Inwerpsel:
> The elastic search sync is throwing error -
> https://drive.google.com/file/d/14TW_tEhX9kGdDdEBj494l-EmiKQhx2Tz/view?usp=sharing
> (https://k8s.p4.greenpeace.org/international/wp-admin/admin.php?page=elasticpress)

Ah it needs a check around something for WPML for sites that don't have it. Adding it.

Inwerpsel:
> Initially when we hit the search page(https://k8s.p4.greenpeace.org/international/?s=&orderby=_score),
> 
> * The top search count show’s “0 results”
> * All left side bar counters are zero and no checkboxes are clickable
> * “Show 5 more results” option is not visible

The reason for this was there was some code that only ran if there's no search query and that produced an invalid query. This happens in the same way in production, where it silently falls back to using mysql. I removed that code for now as it can never work, but looks like it was intended to apply a filter for mime types so we might want to look at this afterwards.
```
 
-                       // Get only DOCUMENT_TYPES from the attachments.
-                       if ( ! $this->search_query && ! $this->filters ) {
-                               add_filter(
-                                       'ep_formatted_args',
-                                       function ( $formatted_args ) use ( $args ) {
-                                               // TODO - Fix parsing exception in EP API call to Elasticsearch.
-                                               $formatted_args['post_mime_type'] = self::DOCUMENT_TYPES;
-
-                                               return $formatted_args;
-                                       },
-                                       10,
-                                       1
-                               );
-                       }
-
```

classes/class-p4-settings.php
```diff
@@ -256,6 +249,38 @@ public function __construct() {
 					'id'   => 'campaigns_import_exclude_style',
 					'type' => 'checkbox',
 				],
+				[
+					'name'    => 'Include archived content in search for',
```
sagarsdeshmukh, line 253: string translation function [__()] needs for strings

classes/class-p4-post-archive.php
```diff
@@ -0,0 +1,87 @@
...
+		];
+
+		register_post_type( self::POST_TYPE, $args );
+
+	}
+
+	public function add_archive_link() {
+		add_meta_box(
+			'archive-url',
+			__( 'Archive URL', 'sitepoint' ),
```
sagarsdeshmukh, line 77: textdomain need to change with 'planet4-master-theme-backend'

classes/class-p4-search.php
```diff
@@ -116,9 +123,132 @@ protected function initialize() {
...
+			if ( ( ! isset( $_GET['orderby'] ) || '_score' === $_GET['orderby'] ) && wp_doing_ajax() ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+				add_filter(
+					'ep_formatted_args',
+					[ new \ElasticPress\Feature\Search\Search(), 'weight_recent' ],
+					10,
+					2
+				);
+			}
+			add_filter( 'posts_where', [ self::class, 'edit_search_mime_types' ] );
+			remove_filter(
```
Inwerpsel, line 244: This was causing the issue where documents show up for other types. If document search is enabled in ElasticPress this filter added this to the query, after our code has determined the post_types. We can safely remove this filter as we set attachment as a type ourselves.

classes/class-p4-search.php
Inwerpsel: Could also get this for the query which also works without ES

PR #1042, opened 2020-03-25
Hotfix: Restore old body font logic in a method.

pablocubico:
Restoring the body font logic and replacing a call to `wp_safe_remote_get` to `file_get_contents`.

### DIFF
--- a/classes/class-p4-post-campaign.php
+++ b/classes/class-p4-post-campaign.php
@@ -385,7 +385,7 @@ if ( ! class_exists( 'P4_Post_Campaign' ) ) {
 			// TODO: Handle errors.
 			$theme_json = json_decode(
 				// Ignoring the PHPCS error in the next line because it's a local file, not a remote request.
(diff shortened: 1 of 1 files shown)


classes/class-p4-post-campaign.php
```diff
@@ -397,14 +397,42 @@ public static function css_vars( array $meta ): array {
 
 			$css_vars = array_merge( $css_vars, self::get_footer_theme( $meta ) );
 			$css_vars = array_merge( $css_vars, self::get_passive_button_color( $meta, $defaults ) );
-			$css_vars = array_merge( $css_vars, self::get_body_font( $meta, $defaults ) );
+			$css_vars = array_merge( $css_vars, self::get_body_font( $meta ) );
```
Inwerpsel: Sure we don't need the defaults?

pablocubico: No, in fact the above call does not need the defaults either.

PR #1043, opened 2020-03-25
[No issue] Minor refactor in campaign variables

pablocubico:
No JIRA issue, we shouldn't need the campaigns font map anymore since the defaults are now present in the JSON files.


pablocubico:
Closing this as it is outdated and the requirements are quite likely to change.

PR #1048, opened 2020-03-27
PLANET-4820: P4CG - Footer color should not be applied when "Main website navigation" is selected

pablocubico:
Note: despite of the issue description, I'm fixing other defaults and rules in this PR.

Ref: https://jira.greenpeace.org/browse/PLANET-4820

Styleguide PR: https://github.com/greenpeace/planet4-styleguide/pull/42
Blocks PR: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/247

### DIFF
--- a/classes/class-p4-post-campaign.php
+++ b/classes/class-p4-post-campaign.php
@@ -16,6 +16,8 @@ if ( ! class_exists( 'P4_Post_Campaign' ) ) {
 		 */
 		const POST_TYPE = 'campaign';
 
+		const DEFAULT_NAVBAR_THEME = 'planet4';
+
 		public const META_FIELDS = [
 			'theme',
 			'campaign_logo',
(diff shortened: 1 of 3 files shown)


templates/blocks/site_logo.twig
```diff
@@ -4,7 +4,8 @@
 	{% if logo == 'greenpeace' %}
 		<img src="{{ data_nav_bar.images }}gp-logo-{{ logo_color }}.svg" alt="Greenpeace">
 	{% else %}
-		<img src="{{ data_nav_bar.images }}{{ logo }}/logo-{{ logo_color }}.png" alt="Greenpeace">
+		{% set color = logo_color == 'green' ? 'light' : logo_color %}
```
pablocubico: This is caused by another issue, it was added as a separate commit so we can revert it when the issue is solved.

pablocubico: Details in the follow up ticket: https://jira.greenpeace.org/browse/PLANET-4901

comzeradd:
Checking the branch locally I get some warnings when viewing a campaign that was already there

```
Notice: Undefined index: campaign_nav_type in single-campaign.php on line 79
Notice: Undefined index: campaign_nav_type in class-p4-post-campaign.php on line 494
Notice: Undefined index: campaign_nav_color in class-p4-post-campaign.php on line 516
Notice: Undefined index: campaign_nav_type in class-p4-post-campaign.php on line 521
Notice: Undefined index: campaign_nav_color in class-p4-post-campaign.php on line 535
```

classes/class-p4-post-campaign.php
```diff
@@ -392,13 +394,14 @@ public static function css_vars( array $meta ): array {
 			$defaults = self::get_theme_defaults( $theme_json );
 
 			// Replace the defaults with the campaign options where applicable.
-			$intersect = array_intersect_key( $defaults, $meta );
+			$intersect = array_intersect_key( array_filter( $meta ), $defaults );
```
Inwerpsel, line 397: This line is to ensure that we only use keys from the meta that occur in the defaults, right? Maybe you can add a short comment like `// Use only meta keys that exist in the defaults` to make this a bit? Or maybe change the naming of the variable to something like `$metas_for_theme` to make this a bit more obvious?

classes/class-p4-post-campaign.php
```diff
@@ -427,10 +428,11 @@ public static function get_body_font( $meta ): array {
 					'plastic'   => 'Montserrat',
 				];
 				$theme              = self::get_theme( $meta );
-				$body_font          = $campaigns_font_map[ $theme ];
+
+				$css_vars['campaign_body_font'] = $campaigns_font_map[ $theme ];
 			}
 
-			return [ 'campaign_body_font' => $body_font ];
+			return $css_vars;
```
Inwerpsel, line 435: If the `$css_vars` is passed by reference then this function doesn't need to return it as it's not used. Though personally I prefer not using passing by reference, as in function calls you can't really see that that is happening, and assigning the return value makes it a bit more obvious what is happening, though the extra assignment does make the syntax a bit more verbose.

classes/class-p4-post-campaign.php
pablocubico: 👍  I just shamelessly copy-pasted your comment...

classes/class-p4-post-campaign.php
pablocubico: Good point, I'll revert to the assignments.

PR #1052, opened 2020-04-08
PLANET-4855 Get Global Standards values from a SmartSheet

Inwerpsel:
* Add client for SmartSheet API.
* Get API key from config, if it's not there from environment variable
`SMARTSHEET_API_KEY`.
* Get global standards values from the API response (sheet ID still
hardcoded).
* Cache the global standards values for 5 minutes.
* Fall back to the hardcoded values if the API request fails.

### DIFF
--- a/classes/class-p4-analytics-values.php
+++ b/classes/class-p4-analytics-values.php
@@ -1,0 +1,287 @@
+<?php
+/**
+ * Data class for post editor analytics values.
+ *
+ * @package P4MT
+ */
+
+/**
+ * Data class for post editor analytics values.
+ */
+final class P4_Analytics_Values {
+	private const CACHE_KEY = 'analytics_global_projects';
+
+	/**
+	 * @var string[] List of global projects.
+	 */
+	private $global_projects;
+
+	/**
+	 * @var string[] List of local projects.
+	 */
+	private $local_projects;
+
+	/**
+	 * P4_Analytics_Values constructor.
+	 *
+	 * @param string[]      $global_projects A list of the global projects.
+	 * @param string[]|null $local_projects A list of the local projects.
+	 */
+	private function __construct( array $global_projects, ?array $local_projects = null ) {
+		$this->global_projects = $global_projects;
+		$this->local_projects  = $local_projects;
+	}
+
+	/**
+	 * Extract values from smartsheets.
+	 *
+	 * @param P4_Smartsheet      $global_smartsheet Smartsheet containing global projects data.
+	 * @param P4_Smartsheet|null $local_smartsheet Smartsheet containing local projects data.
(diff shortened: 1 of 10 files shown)


page.php
```diff
@@ -76,6 +76,7 @@
 
 // P4 Campaign/dataLayer fields.
 $context['cf_campaign_name'] = $page_meta_data['p4_campaign_name'][0] ?? '';
```
Inwerpsel, line 78: Probably we want to deduplicate these 3 files here, maybe I'll add that to this PR too.

sagarsdeshmukh:
Need PHPCS fixes - https://circleci.com/gh/greenpeace/planet4-master-theme/7034?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

classes/class-p4-analytics-values.php
```diff
@@ -0,0 +1,143 @@
...
+		return [
+			'global_projects' => $this->global_projects,
+			'local_projects'  => $this->local_projects,
+		];
+	}
+
+	public static function from_cache_or_api_or_hardcoded(): self {
+		$cache = wp_cache_get( self::CACHE_KEY );
+
+		$cache = false;
```
sagarsdeshmukh, line 49: looks like a leftover from testing.

classes/class-p4-metabox-register.php
```diff
@@ -442,4 +414,18 @@ public function cmb2_after_form_do_js_validation( $post_id, $cmb ) {
...
+			return;
+		}
+		if ( $action === 'removed' ) {
+			update_post_meta( $field->object_id(), 'p4_global_project_tracking_id', null );
+
+			return;
+		}
+
+		$project_id = P4_Analytics_Values::from_cache_or_api_or_hardcoded()->get_id_for_global_project( $field->value() );
+		update_post_meta( $field->object_id(), 'p4_global_project_tracking_id', $project_id );
```
sagarsdeshmukh, line 429: The value of `p4_global_project_tracking_id` not saved. 
Tested at https://k8s.p4.greenpeace.org/test-atlas/test-project-id/

Inwerpsel: Ah looks like `get_env` is returning false instead of null here https://github.com/greenpeace/planet4-master-theme/pull/1052/files#diff-51e0c855392eaf41efec35a871fa2dc4R62 , which breaks the null coalescing. Changed it to using `$_ENV` which should do the same thing as `getenv` except it does return null.

Inwerpsel: Hmm something else was going on too, the hidden field actually overwrote what was set here :facepalm: so I removed that field

PR #1053, opened 2020-04-08
PLANET-4919: Ensure all variables & rules work as expected in the Sidebar.

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-4919

Blocks counterpart: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/249

### DIFF
--- a/campaign_themes/arctic.json
+++ b/campaign_themes/arctic.json
@@ -79,11 +79,12 @@
 			"configurations": {
 				"minimal": {
 					"options": [
-						{ "color":  "#ffffff"},
-						{ "color":  "#000000"},
-						{ "color": "#035880"},
-						{ "color": "#00b6c2"}
-					]
+						{ "value": "#ffffff", "label": "White" },
+						{ "value": "#000000", "label": "Black" },
+						{ "value": "#035880", "label": "Blue" },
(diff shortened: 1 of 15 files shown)


campaign_themes/arctic.json
```diff
@@ -79,11 +79,12 @@
...
-						{ "color":  "#000000"},
-						{ "color": "#035880"},
-						{ "color": "#00b6c2"}
-					]
+						{ "value": "#ffffff" },
+						{ "value": "#000000" },
+						{ "value": "#035880" },
+						{ "value": "#00b6c2" }
+					],
+					"default": "#035880"
```
Inwerpsel, line 87: Not necessarily something we should do in this PR, but maybe it would be nicer to maintain if we marked default values by setting a `default` property on the object of that option.
```
					"options": [
						{ "value": "#ffffff" },
						{ "value": "#000000" },
						{ "value": "#035880", "isDefault": true },
						{ "value": "#00b6c2" }
					],
```

Inwerpsel:
Checked these latest changes (https://github.com/greenpeace/planet4-master-theme/pull/1053/files/ce100205a7503a0f2ee643eb7e2d8c93670e652e..HEAD), still looks good :+1:

PR #1054, opened 2020-04-09
PLANET-4927 Planet 4 Bug Report: Page, Evergreen: Spacing issue when using background image

dpivo:
- mask-position doesn't work well on chrome browser
- positioned the overlay with 'top' and the value of the height of the navbar

Ref: https://jira.greenpeace.org/browse/PLANET-4927

### DIFF
--- a/assets/src/scss/pages/_evergreen.scss
+++ b/assets/src/scss/pages/_evergreen.scss
@@ -25,11 +25,15 @@
     -webkit-mask-image: linear-gradient(170deg, rgba(0, 0, 0, 1), rgba(0, 0, 0, .2), rgba(0, 0, 0, 0), rgba(0, 0, 0, 0), rgba(0, 0, 0, 0));
     mask-size: 100%;
     mask-repeat: no-repeat;
-    mask-position: left top;
 
     html[dir="rtl"] & {
       mask-image: linear-gradient(-170deg, rgba(0, 0, 0, 1), rgba(0, 0, 0, .2), rgba(0, 0, 0, 0), rgba(0, 0, 0, 0), rgba(0, 0, 0, 0));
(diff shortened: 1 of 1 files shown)


comzeradd:
The difference I see with other (non-evergreen) pages is that `.skip-links` has some margins, inherited by lists, that push everything a bit lower. If you check on the inspector the Act page for instance, you'll see thsa `skewed-overlay` start from the very top.

So maybe instead of using navbar's height you can make the [skip-links](https://github.com/greenpeace/planet4-styleguide/blob/master/src/components/_skip-links.scss) margins a bit more explicit to be `0`.


Screenshot from 2020-04-14 14-29-39

assets/src/scss/pages/_evergreen.scss
```diff
@@ -25,12 +25,16 @@
...
-    mask-position: left top;
 
     html[dir="rtl"] & {
       mask-image: linear-gradient(-170deg, rgba(0, 0, 0, 1), rgba(0, 0, 0, .2), rgba(0, 0, 0, 0), rgba(0, 0, 0, 0), rgba(0, 0, 0, 0));
       -webkit-mask-image: linear-gradient(-170deg, rgba(0, 0, 0, 1), rgba(0, 0, 0, .2), rgba(0, 0, 0, 0), rgba(0, 0, 0, 0), rgba(0, 0, 0, 0));
     }
+
+    top: $menu-height-small;
+    @include medium-and-up {
+      top: $menu-height-large;
```
comzeradd, line 36: Now that styleguide margin PR is merged, do you think we still need `top` here?

dpivo: Ah no, we can get rid of this PR then.

PR #1057, opened 2020-04-14
PLANET-4859 Pull "Local Project" values from each NRO smartsheet and pass it to DataLayer

sagarsdeshmukh:
[JIRA 4859](https://jira.greenpeace.org/browse/PLANET-4859)

- Get local project smartsheet id from Planet4 settings, if it's not there from environment variable
`ANALYTICS_LOCAL_SMARTSHEET_ID`.

The change are available for testing at [here](https://k8s.p4.greenpeace.org/test-pluto/test-local-project-id-analytics-tracking-field/)

### DIFF
--- a/classes/class-p4-analytics-values.php
+++ b/classes/class-p4-analytics-values.php
@@ -41,7 +41,7 @@ final class P4_Analytics_Values {
 	 */
 	public static function from_smartsheets(
 		P4_Smartsheet $global_smartsheet,
-		?P4_Smartsheet $local_smartsheet = null
+		P4_Smartsheet $local_smartsheet = null
 	): self {
 		$project_name_column = $global_smartsheet->get_column_index( 'Global Project Standard' );
 		$approved_column     = $global_smartsheet->get_column_index( 'Standard Approved' );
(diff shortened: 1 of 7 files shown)


classes/class-p4-analytics-values.php
```diff
@@ -265,6 +279,26 @@ function ( $project ) {
...
 
+	/**
+	 * Returns local(NRO) project smartsheet options.
+	 *
+	 * @return array
+	 */
+	public function local_projects_options(): array {
+		$local_projects_options = [];
+		if ( $this->local_projects ) {
+			$names                  = array_map(
```
Inwerpsel, line 290: This spacing is a bit of a nasty side-effect of the equal sign alignment rule we have, where the result is actually less readable. You can avoid it by leaving a blank line between.

classes/class-p4-metabox-register.php
```diff
@@ -433,29 +443,30 @@ public static function save_global_project_id( $updated, $action, CMB2_Field $fi
 	 * If the post has a value that is not in the sheet, keep it in the dropdown so that metaboxes save doesn't set
 	 * it to another value, but mark it with `[DEPRECATED]` prefix.
 	 *
-	 * @param array $global_project_options The list of supported global projects.
+	 * @param array  $options_array The list of supported global/local projects.
+	 * @param string $field_name The meta field name.
 	 * @return array The list with maybe the current post value.
 	 */
-	private static function maybe_add_current_post_value( array $global_project_options ): array {
+	private static function maybe_add_current_post_value( array $options_array, $field_name ): array {
```
Inwerpsel, line 450: Nice generalization :100:

page.php
```diff
@@ -76,7 +76,8 @@
 
 // P4 Campaign/dataLayer fields.
 $context['cf_campaign_name'] = $page_meta_data['p4_campaign_name'][0] ?? '';
-$context['cf_project_id']    = $page_meta_data['p4_global_project_tracking_id'][0] ?? '';
+$context['cf_project_id']    = $page_meta_data['p4_global_project_tracking_id'][0] ?? 'not set';
+$context['cf_local_project'] = $page_meta_data['p4_local_project'][0] ?? '';
```
Inwerpsel, line 80: Shouldn't this also fall back to `'not set'`?

sagarsdeshmukh: by default if no smartsheet ID configured, the default value would be set as 'not set' from "-Select Local Project-" dropdown option, but we will also add one more level of fallback at here.

PR #1060, opened 2020-04-16
PLANET-4700 TA boxout excerpt is copied to post excerpt

dpivo:
- Changed auto excerpt generator to first look for a paragraph to get excerpt from, and then the rest of the content.

Ref: https://jira.greenpeace.org/browse/PLANET-4700

### DIFF
--- a/classes/class-p4-master-site.php
+++ b/classes/class-p4-master-site.php
@@ -674,7 +674,11 @@ class P4_Master_Site extends TimberSite {
 			remove_action( 'save_post', [ $this, 'p4_auto_generate_excerpt' ], 10 );
 
 			// Generate excerpt text.
-			$post_excerpt   = strip_shortcodes( $post->post_content );
+			$post_excerpt = strip_shortcodes( $post->post_content );
+
+			preg_match( '<p>(.*?)</p>', $post_excerpt, $match_paragraph );
+
+			$post_excerpt   = count( $match_paragraph ) > 0 ? $match_paragraph[1] : $post_excerpt;
 			$post_excerpt   = apply_filters( 'the_content', $post_excerpt );
 			$post_excerpt   = str_replace( ']]>', ']]&gt;', $post_excerpt );
 			$excerpt_length = apply_filters( 'excerpt_length', 30 );


classes/class-p4-master-site.php
```diff
@@ -674,9 +674,14 @@ public function p4_auto_generate_excerpt( $post_id, $post ) {
 			remove_action( 'save_post', [ $this, 'p4_auto_generate_excerpt' ], 10 );
 
 			// Generate excerpt text.
-			$post_excerpt   = strip_shortcodes( $post->post_content );
-			$post_excerpt   = apply_filters( 'the_content', $post_excerpt );
-			$post_excerpt   = str_replace( ']]>', ']]&gt;', $post_excerpt );
+			preg_match( "'<p>(.*?)</p>'", $post->post_content, $match_paragraph );
```
Inwerpsel, line 677: Seems to make sense to do this. Just checking, are we sure there are no cases where there could be text in something else than a `<p>` tag that we still want to be part of the excerpt?

Inwerpsel: Shouldn't this regex just be inside single quotes? Currently it will try to match everything that is inside the first `<p>` tag that is inside single quotes.

dpivo: There could be. We can have a look at the other blocks and what they produce.

classes/class-p4-master-site.php
```diff
@@ -674,9 +674,14 @@ public function p4_auto_generate_excerpt( $post_id, $post ) {
...
+			if ( count( $match_paragraph ) > 0 ) {
+				$post_excerpt = $match_paragraph[1];
```
Inwerpsel, line 679: This will match only the first paragraph, which will work in most cases but maybe there's cases where the first paragraph is only a couple of words and we want to include more in the excerpt?

classes/class-p4-master-site.php
```diff
@@ -674,9 +674,15 @@ public function p4_auto_generate_excerpt( $post_id, $post ) {
 			remove_action( 'save_post', [ $this, 'p4_auto_generate_excerpt' ], 10 );
 
 			// Generate excerpt text.
-			$post_excerpt   = strip_shortcodes( $post->post_content );
-			$post_excerpt   = apply_filters( 'the_content', $post_excerpt );
-			$post_excerpt   = str_replace( ']]>', ']]&gt;', $post_excerpt );
+			preg_match( '<p>(.*?)</p>', $post->post_content, $match_paragraph );
+			if ( count( $match_paragraph ) > 0 ) {
+				$post_excerpt = $match_paragraph[1];
+				$post_excerpt = apply_filters( 'the_content', $post_excerpt );
```
Inwerpsel, line 680: You could move this to after the if-statement so we only have to call it once. Not a problem I think if `str_replace( ']]>', ']]&gt;', $post_excerpt )` is also called then for both cases

classes/class-p4-master-site.php
sagarsdeshmukh: I agree on @Inwerpsel point, in that case the except would be very small.

classes/class-p4-master-site.php
```diff
@@ -674,7 +674,9 @@ public function p4_auto_generate_excerpt( $post_id, $post ) {
 			remove_action( 'save_post', [ $this, 'p4_auto_generate_excerpt' ], 10 );
 
 			// Generate excerpt text.
-			$post_excerpt   = strip_shortcodes( $post->post_content );
+			preg_match( '<p>(.*?)</p>', $post->post_content, $match_paragraph );
+
+			$post_excerpt   = count( $match_paragraph ) > 0 ? $match_paragraph[1] : strip_shortcodes( $post->post_content );
```
Inwerpsel, line 679: Are we sure there can be no shortcodes inside the first paragraph? If there could be then you could put this ternary inside the `strip_shortcodes` call.

dpivo: Good point.

PR #1061, opened 2020-04-16
PLANET-4827 Fix document type filtering and page count

Inwerpsel:
* Put the document type filter on all queries, by making it an OR query:
either the post type is something else than attachment, or the mime type
is one of the supported types.
* Don't show numbers lower than 0 for calculated page amount. We
subtract the Act pages from this as they have their own metric, however
in some cases the counts are not reliable and we might end up below 0.
Probably we want to replace this calculation with an aggregation that
already subtracts the Act pages.


Inwerpsel:
Closed this and opened separate ticket https://github.com/greenpeace/planet4-master-theme/pull/1062

PR #1062, opened 2020-04-17
PLANET-4975 Fix document type filtering and page count

Inwerpsel:
* Put the document type filter on all queries, by making it an OR query:
either the post type is something else than attachment, or the mime type
is one of the supported types. Remove previous filters that were in
multiple places as the new filter can be used on any result set.
* Don't show numbers lower than 0 for calculated page amount. We
subtract the Act pages from this as they have their own metric, however
in some cases the counts are not reliable and we might end up below 0.
Probably we want to replace this calculation with an aggregation that
already subtracts the Act pages.

### DIFF
--- a/classes/class-p4-search.php
+++ b/classes/class-p4-search.php
@@ -366,12 +366,18 @@ if ( ! class_exists( 'P4_Search' ) ) {
 				foreach ( $posts as $post ) {
 					if ( P4_Post_Archive::POST_TYPE === $post->post_type ) {
 						$archive_post               = new stdClass();
+						$archive_post->id           = $post->ID;
 						$archive_post->post_title   = $post->post_title;
 						$archive_post->link         = $post->guid;
 						$archive_post->post_type    = P4_Post_Archive::POST_TYPE;
 						$archive_post->post_date    = $post->post_date_gmt;
 						$archive_post->post_excerpt = $post->post_excerpt;
-						$template_posts[]           = $archive_post;
+
+						if ( current_user_can( 'edit_posts' ) ) {
+							$archive_post->edit_link = get_edit_post_link( $post->ID );
+						}
+
+						$template_posts[] = $archive_post;
 					} else {
 						$template_post                = $post;
 						$template_post->id            = $post->ID;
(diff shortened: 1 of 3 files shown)


templates/tease-search.twig
```diff
@@ -90,6 +90,9 @@
 			</div>
 
 			<p class="search-result-item-content">{{ post.post_excerpt|default(post.post_content)|excerpt( 25 )|raw }}</p>
+			{% if user.can('edit_posts') %}
```
sagarsdeshmukh, line 93: Does this link works with load more button results?

Inwerpsel: It does not currently, good catch! :100:

templates/tease-search.twig
```diff
@@ -90,6 +90,9 @@
 			</div>
 
 			<p class="search-result-item-content">{{ post.post_excerpt|default(post.post_content)|excerpt( 25 )|raw }}</p>
+			{% if is_archive and fn('is_user_logged_in') %}
+				<a href="{{ fn('get_edit_post_link', post.id)|raw }}">{{ __('Edit this', 'planet4-master-theme') }}</a>
```
sagarsdeshmukh, line 94: The "Edit this" link is visible in load more results for non logged-in users as well.
image

Inwerpsel: Updated it so that it checks the permissions. Great catch :+1:

classes/class-p4-search.php
```diff
@@ -366,12 +366,18 @@ protected function get_posts( $paged = 1 ) : array {
 				foreach ( $posts as $post ) {
 					if ( P4_Post_Archive::POST_TYPE === $post->post_type ) {
 						$archive_post               = new stdClass();
+						$archive_post->id           = $post->ID;
```
sagarsdeshmukh, line 369: Do we need to keep post id in archive data?

sagarsdeshmukh:
The new changes tested on local env, Looks good to me 👍

classes/class-p4-search.php
Inwerpsel: Not necessarily but might make sense to keep it as the other post types also have this set. During development I at first forgot to add this and accessed the ID in the template and it actually didn't fail but produced a wrong link, which is easy to overlook. So I guess that would be prevented from happening again by keeping it.

PR #1064, opened 2020-04-20
PLANET-4953 Fix Sentry Issue: PLANET4-D0

sagarsdeshmukh:
[JIRA 4953](https://jira.greenpeace.org/browse/PLANET-4953)

### DIFF
--- a/classes/class-p4-user.php
+++ b/classes/class-p4-user.php
@@ -69,7 +69,7 @@ if ( ! class_exists( 'P4_User' ) ) {
 		 *
 		 * @return string
 		 */
-		public function name() : string {
+		public function name() : ?string {
 			if ( $this->is_fake ) {
 				return (string) $this->display_name;
 			} else {
--- a/templates/single.twig
+++ b/templates/single.twig
@@ -37,14 +37,16 @@
 				<div class="row">
 					<div class="col-md-6">
 						<div class="single-post-meta">
-							<address class="single-post-author">
-								{{ __( 'by', 'planet4-master-theme' ) }}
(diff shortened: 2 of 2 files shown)


templates/single.twig
```diff
@@ -38,10 +38,11 @@
 					<div class="col-md-6">
 						<div class="single-post-meta">
 							<address class="single-post-author">
-								{{ __( 'by', 'planet4-master-theme' ) }}
```
sagarsdeshmukh, line 41: This change is made because of issue found on below link -
https://k8s.p4.greenpeace.org/test-venus/story/28598/why-you-should-care-where-your-seafood-comes-from/
image

classes/class-p4-user.php
```diff
@@ -73,7 +73,7 @@ public function name() : string {
 			if ( $this->is_fake ) {
 				return (string) $this->display_name;
 			} else {
-				return parent::name();
+				return parent::name() ?? '';
```
Inwerpsel, line 76: You could also solve this by making the return type nullable (`name(): ?string`), which is maybe more straightforward as then the signature is the same as `name()` in the parent.

templates/single.twig
Inwerpsel: Shouldn't you wrap this in `{% if post.author %}`? On the test instance it looks like this post doesn't have an author set, so with the current change this will still be showing as it happens in both branches of the if-else below.

sagarsdeshmukh: You are right, I updated the if condition. 👍

PR #1065, opened 2020-04-20
PLANET-4738 Fix Sentry Issue: PLANET4-2J

sagarsdeshmukh:
[JIRA 4738](https://jira.greenpeace.org/browse/PLANET-4738)

### DIFF
--- a/classes/class-p4-search.php
+++ b/classes/class-p4-search.php
@@ -322,6 +322,9 @@ if ( ! class_exists( 'P4_Search' ) ) {
 					// Handle submitted filter options.
 					if ( $selected_filters && is_array( $selected_filters ) ) {
 						foreach ( $selected_filters as $type => $filter_type ) {
+							if ( ! is_array( $filter_type ) ) {
+								continue;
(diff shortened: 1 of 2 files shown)


classes/class-p4-search.php
```diff
@@ -322,11 +322,13 @@ public function get_paged_posts() {
 					// Handle submitted filter options.
 					if ( $selected_filters && is_array( $selected_filters ) ) {
 						foreach ( $selected_filters as $type => $filter_type ) {
-							foreach ( $filter_type as $name => $id ) {
-								$filters[ $type ][] = [
-									'id'   => $id,
-									'name' => $name,
-								];
+							if ( is_array( $filter_type ) ) {
```
Inwerpsel, line 325: It's a bit easier to read if you `continue` when it's not an array, so the rest of the loop doesn't need to be wrapped in a long if-statement.

sagarsdeshmukh: yes, I agree with you 👍

PR #1068, opened 2020-04-23
PLANET-4937: Search results post type should use an a tag instead of a button that behaves like an a tag

pablocubico:
Changed the `button` tag for post type into an `a` tag in the Search page.

While doing this, I found a weird thing about the `search.js` script, it was triggering a 404 locally. It looks like it's loaded from the search class here: https://github.com/greenpeace/planet4-master-theme/blob/master/classes/class-p4-search.php#L1097 (which actually does nothing, because the `setupSearch` function is never invoked) but also built in: https://github.com/greenpeace/planet4-master-theme/blob/master/assets/src/js/app.js

So I removed the script being loaded from the Search class, and moved the `localizations` variables to the place where we enqueue `index.js`, the main script.

### DIFF
--- a/assets/src/js/search.js
+++ b/assets/src/js/search.js
@@ -8,6 +8,11 @@ if (!window.lazyLoad) {
 
 // Search page.
 export const setupSearch = function($) {
+  const isSearch = !!$('body.search').length;
+  if (!isSearch) {
+    return;
+  }
+
   const $search_form      = $( '#search_form' );
   const $load_more_button = $( '.btn-load-more-click-scroll' );
   let load_more_count   = 0;
@@ -52,11 +57,6 @@ export const setupSearch = function($) {
   });
 
   let $search_results = $( '.multiple-search-result' );
-  // Add filter by clicking on the page type label inside a result item.
-  // Delegate event handler to the dynamically created descendant elements.
-  $search_results.off( 'click', '.search-result-item-head' ).on( 'click', '.search-result-item-head', function() {
-    window.location.href = $( this ).data( 'href' );
-  });
 
   // Navigate to the page of the search result item when clicking on it's thumbnail image.
   // Delegate event handler to the dynamically created descendant elements.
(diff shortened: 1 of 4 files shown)


templates/tease-search.twig
```diff
@@ -42,7 +42,7 @@
 
 					<div class="search-result-tags top-page-tags">
 						{% for page_type in post.p4_page_types %}
-							<button class="search-result-item-head no-btn tag-item tag-item--main page-type" data-href="{{ fn('get_term_link', page_type.term_id) }}">{{ page_type.name|e('wp_kses_post')|raw }}</button>
+							<a class="search-result-item-head no-btn tag-item tag-item--main page-type" data-href="{{ fn('get_term_link', page_type.term_id) }}">{{ page_type.name|e('wp_kses_post')|raw }}</a>
```
Inwerpsel, line 45: If we change the `data-href` to just `href`, we don't need the following script to open the link. https://github.com/greenpeace/planet4-master-theme/blob/2a5faa966c2ecbfba8eb983d1344dae4eba69330/assets/src/js/search.js#L58-L58

pablocubico: Ugh, right, this misses the whole point! Amending...

pablocubico: @Inwerpsel Updated!

assets/src/js/search.js
```diff
@@ -8,6 +8,9 @@ if (!window.lazyLoad) {
 
 // Search page.
 export const setupSearch = function($) {
+  const isSearch = !!$('body.search').length;
+  if (!isSearch) return;
```
sagarsdeshmukh, line 12: Need to add curly brackets here, eslint is complaining about it -
`12:3  error  Expected { after 'if' condition  curly`

pablocubico: @sagarsdeshmukh fixed!

PR #1071, opened 2020-04-24
PLANET-4968 Hide Scope dropdown from Analytics & Tracking fields

sagarsdeshmukh:
[JIRA 4968](https://jira.greenpeace.org/browse/PLANET-4968)

### DIFF
--- a/classes/class-p4-context.php
+++ b/classes/class-p4-context.php
@@ -59,10 +59,30 @@ class P4_Context {
 	public static function set_campaign_datalayer( &$context, $meta ) {
 		$context['cf_campaign_name'] = $meta['p4_campaign_name'] ?? '';
 		$context['cf_basket_name']   = $meta['p4_basket_name'] ?? '';
-		$context['cf_scope']         = $meta['p4_scope'] ?? '';
 		$context['cf_department']    = $meta['p4_department'] ?? '';
 		$context['cf_project_id']    = $meta['p4_global_project_tracking_id'] ?? 'not set';
 		$context['cf_local_project'] = $meta['p4_local_project'] ?? 'not set';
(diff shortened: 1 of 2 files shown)


classes/class-p4-context.php
```diff
@@ -59,10 +59,32 @@ public static function set_og_meta_fields( &$context, $post ) {
...
+	 * - If Global Project matches any other value (apart from Local Campaign) then Scope is Global
+	 *
+	 * @param string $global_project The Global Project value.
+	 * @param string $local_project The Local Project value.
+	 * @return string The campaign scope.
+	 */
+	private static function get_campaign_scope( $global_project, $local_project ) {
+		if ( 'not set' === $global_project ) {
+			$campaign_scope = 'not set';
+		} elseif ( $global_project === $local_project ) {
```
Inwerpsel, line 82: I think this should check if the global project is the string "Local Campaign", rather than if it is equal to the current local project.

classes/class-p4-context.php
```diff
@@ -59,10 +59,32 @@ public static function set_og_meta_fields( &$context, $post ) {
...
+	 * - If Global Project equals Local Campaign then Scope is Local.
+	 * - If Global Project equals none then Scope is not set
+	 * - If Global Project matches any other value (apart from Local Campaign) then Scope is Global
+	 *
+	 * @param string $global_project The Global Project value.
+	 * @param string $local_project The Local Project value.
+	 * @return string The campaign scope.
+	 */
+	private static function get_campaign_scope( $global_project, $local_project ) {
+		if ( 'not set' === $global_project ) {
```
Inwerpsel, line 80: Maybe this can be a switch as both if's are checking the same variable. You could also directly return for each case, so you don't nee the `$campaign_scope` variable or break statement.

PR #1082, opened 2020-05-13
PLANET-4998 Use docker-compose with coverage for tests

Inwerpsel:


Inwerpsel:
This was just to test out, the necessary things for collecting coverage were added in 
https://github.com/greenpeace/planet4-base-fork/pull/69
and https://github.com/greenpeace/planet4-docker-compose/pull/47
and require no changes in master theme or plugin repos.

PR #1083, opened 2020-05-13
PLANET-4774 Campaign Import - Skip duplicate post_meta import

sagarsdeshmukh:
[JIRA 4774](https://jira.greenpeace.org/browse/PLANET-4774)

- The [WP-Importer](https://wordpress.org/plugins/wordpress-importer/) plugin has a bug, if post already exists, it skip the post from import but [imports post_meta data](https://github.com/WordPress/wordpress-importer/blob/master/src/class-wp-import.php#L898).

Fix - 
- Added a `wp_import_post_meta` filter to skip duplicate post_meta from import.

### DIFF
--- a/classes/class-p4-campaign-importer.php
+++ b/classes/class-p4-campaign-importer.php
@@ -27,6 +27,7 @@ if ( ! class_exists( 'P4_Campaign_Importer' ) ) {
 			add_filter( 'wp_import_post_meta', [ $this, 'process_campaign_metas' ] );
 			add_filter( 'wp_import_post_data_processed', [ $this, 'set_imported_campaigns_as_drafts' ], 10, 2 );
 			add_action( 'import_end', [ $this, 'action_import_end' ], 10, 0 );
+			add_filter( 'wp_import_post_meta', [ $this, 'skip_duplicate_postmeta_import' ], 10, 2 );
 		}
 
 		/**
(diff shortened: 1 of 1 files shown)


classes/class-p4-campaign-importer.php
```diff
@@ -281,5 +282,26 @@ public function process_campaign_metas( $post_meta ) {
...
+		 * Skip already existing postmeta data from import.
+		 *
+		 * @param array   $post_meta The to be imported post meta fields.
+		 * @param integer $post_id Post ID.
+		 *
+		 * @return array The cleaned post meta fields.
+		 */
+		public function skip_duplicate_postmeta_import( $post_meta, $post_id ) {
+			$campaign_postmeta = get_post_meta( $post_id );
+			if ( $post_meta && $campaign_postmeta ) {
```
Inwerpsel, line 296: Both of these are variables meta of a campaign, the difference is one exists in db already. Maybe you can rename `$campaign_postmeta` to `$existing_postmeta`.

classes/class-p4-campaign-importer.php
```diff
@@ -281,5 +282,26 @@ public function process_campaign_metas( $post_meta ) {
...
+
+		/**
+		 * Skip already existing postmeta data from import.
+		 *
+		 * @param array   $campaign_postmeta The to be imported post meta fields.
+		 * @param integer $post_id Post ID.
+		 *
+		 * @return array The cleaned post meta fields.
+		 */
+		public function skip_duplicate_postmeta_import( $campaign_postmeta, $post_id ) {
```
Inwerpsel, line 294: You can make it just `$postmeta`, as this same function is called for the meta of the attachments in the import file too.

classes/class-p4-campaign-importer.php
```diff
@@ -281,5 +282,26 @@ public function process_campaign_metas( $post_meta ) {
...
+		 * Skip already existing postmeta data from import.
+		 *
+		 * @param array   $postmeta The to be imported post meta fields.
+		 * @param integer $post_id Post ID.
+		 *
+		 * @return array The cleaned post meta fields.
+		 */
+		public function skip_duplicate_postmeta_import( $postmeta, $post_id ) {
+			$existing_postmeta = get_post_meta( $post_id );
+			if ( $postmeta && $existing_postmeta ) {
```
Inwerpsel, line 296: You can use `array_filter` on `$postmeta` with function returning `!array_key_exists( $meta['key'], $existing_postmeta )`.

sagarsdeshmukh: Hi @Inwerpsel , The callback function of `array_filter` won't get access to `$existing_postmeta` var. I tried with with `global` declaration, but it's not working.

PR #1084, opened 2020-05-13
PLANET-4987: Prevent sync of non-pdf attachments to ElasticSearch

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-4987

This is a PoC of applying [a filter](https://github.com/10up/ElasticPress/blob/develop/includes/classes/Elasticsearch.php#L903) to hijack the "bulk index" requests sent to ElasticSearch for the attachments.

Attachments should be handled by the [Documents class](https://github.com/10up/ElasticPress/blob/develop/includes/classes/Feature/Documents/Documents.php#L226) in ElasticPress, which seems to handle the extensions supported by default by EP (pdf, xls, docx... some others).

But when you enable the Documents feature, "bulk indexing" requests are being sent to ElasticSearch to actually index everything, ignoring any MIME type filtering, this sounds like it's being overlooked by ElasticPress, the Documents feature itself seems to be poorly maintained:  the UI to enable the feature doesn't even work properly.

## How to test this

1. Go to the ElasticPress page on WP's Dashboard, and open the "Documents" section. **But wait!** Don't click **Enable** and **Save** yet:
image
2. Clicking "Enable" and "Save" won't work, first you have to run this line (gently provided by PieterV): `document.querySelector('label[for="feature_active_documents_enabled"] input').setAttribute('data-field-name', 'active')` to add that "field name" attribute to the input.
3. Then click **Save**, the Documents Feature should be enabled now. It should show a circle filled with yellow.
4 _(Feel free to refresh the page to make sure it worked)_
5. Open ElasticHQ by going to: http://localhost:5000/ 
6. Connect to the cluster at http://elasticsearch:9200 (remember the hostname should be `elasticsearch` here, not `localhost`. 
image
7. Open the `planet4test-post-1` index. You should see 29 documents indexed (provided you are using the default content):
image
8. Go back to the ElasticPress admin page and click on the Sync icon (next to the cog wheel on the top-right corner):
image
9. Once it's finished, go back to ElasticHQ. As there is only one PDF file in the default content, you should see there are now 30 documents indexed.

If you follow these steps in the `master` branch, you'll end up with 262 documents indexed, because it will include all the Media Library files.

## Notes

* A caveat: this is currently only working using the Dashboard Sync, not the CLI. So you can actually test the indexing without the filter running `make elastic` in the `docker-compose` folder. You'll end up with 262 docs synced.
* I added a few `error_log` statements because debugging interactively with Xdebug was impractical, I'll remove those after review.
* Probably needs a few comments about the body format.

## How it works

The body of a bulk index request is separated by new lines and looks like this:
```
{ "index": ...id, (etc.) }
{ "post_id": 123, ...(the full content of the object to be indexed) }
...(an empty line)
{ "index": ... } ... and so on
``` 

The filter looks for attachment and removes the ones that are not in the  `P4_Search::DOCUMENT_TYPES` list from the request's body.

### DIFF
--- a/classes/class-p4-search.php
+++ b/classes/class-p4-search.php
@@ -232,6 +232,20 @@ if ( ! class_exists( 'P4_Search' ) ) {
 				[ Features::factory()->get_registered_feature( 'documents' ), 'setup_document_search' ],
 				10
 			);
+
+			add_filter(
+				'ep_intercept_remote_request',
+				function() {
+					return true;
+				},
+				10
+			);
+			add_filter(
+				'ep_do_intercept_request',
+				[ self::class, 'intercept_ep_request' ],
+				10,
+				4
+			);
 		}
 
 
@@ -1018,6 +1032,59 @@ if ( ! class_exists( 'P4_Search' ) ) {
 				$where     .= ' AND ' . $wpdb->posts . '.post_mime_type IN("' . $mime_types . '","") ';
 			}
 			return $where;
+		}
+
+		/**
+		 * Intercept ElasticPress request filter.
+		 * All these parameters are passed from EP.
+		 *
+		 * @param object $request Request object from ElasticPress.
+		 * @param array  $query Query from ElasticPress.
+		 * @param array  $args Arguments for the request.
+		 * @param array  $failures Errors array.
+		 */
+		public static function intercept_ep_request( $request, $query = null, $args = null, $failures = null ) {
+			$total_docs_count    = 0;
+			$indexed_docs_count  = 0;
+			$filtered_body_lines = [];
+
+			// Is this a bulk index task on the attachment pipeline?
+			if ( $query && ! ! preg_match( '/_bulk\?pipeline.*attachment/', $query['url'] ) ) {
+				$body_lines      = explode( "\n", $args['body'] );
+				$body_line_count = count( $body_lines );
+
+				for ( $i = 0; $i < $body_line_count; $i += 3 ) {
+
+					$total_docs_count++;
+
+					// e.g.: { index; ... }.
+					$index_header  = $body_lines[ $i ];
+					$index_content = $body_lines[ $i + 1 ];
+
+					$index_content_data = json_decode( $index_content, true );
+
+					// Not an attachment (no MIME type).
+					if ( 'attachment' !== $index_content_data['post_type'] ) {
+						$filtered_body_lines[] = $index_header;
+						$filtered_body_lines[] = $index_content;
+						$filtered_body_lines[] = '';
+						$indexed_docs_count++;
+					} elseif (
+							! empty( $index_content_data['post_mime_type'] ) &&
+							in_array( $index_content_data['post_mime_type'], self::DOCUMENT_TYPES, true )
+						) {
+						$filtered_body_lines[] = $index_header;
+						$filtered_body_lines[] = $index_content;
+						$filtered_body_lines[] = ''; // The third element ($body_lines[$i + 2]) is just an empty string.
+						$indexed_docs_count++;
+					}
+				}
+
+				$args['body'] = implode( "\n", $filtered_body_lines );
+			}
+
+			$request = wp_remote_request( $query['url'], $args );
+			return $request;
 		}
 
 		/**
--- a/package-lock.json
+++ b/package-lock.json
@@ -9124,7 +9124,8 @@
         },
         "yargs-parser": {
           "version": "13.1.1",
-          "resolved": "",
+          "resolved": "https://registry.npmjs.org/yargs-parser/-/yargs-parser-13.1.1.tgz",
+          "integrity": "sha512-oVAVsHz6uFrg3XQheFII8ESO2ssAf9luWuAd6Wexsu4F3OtIW0o8IribPXYrD4WC24LWtPrJlGy87y5udK+dxQ==",
           "dev": true,
           "requires": {
             "camelcase": "^5.0.0",


.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
...
+			"name": "Listen for XDebug",
+			"type": "php",
+			"request": "launch",
+			"port": 9010,
+			"pathMappings": {
+				"/app/source/public": "${workspaceRoot}/app/public"
+			}
+		}
+	]
+}
```
comzeradd, line 17: I'm a bit skeptical about including editor specific configuration to the code. If you have this locally you can gitignore it at `.git/info/exclude`.

pablocubico: Ha, that's vestigial, I'll remove it. I'll add it to `.gitignore`

pablocubico: File removed from version control!

pablocubico:
Update:

Got some feedback from @Inwerpsel , who raised this concern: ⚠️  posts get synced individually despite of the bulk sync, we need to check that because then the docs would get indexed on a post-by-post basis. I investigated and it the outcome is:

> It looks like the hook is also triggered on post updates so I think we could intercept those. The query has a different URL (elasticsearch:9200/planet4test-post-1/post/_bulk, note that it says "bulk" but it's still a single post), it looks like it only indexes one document, the post, not the attachments (you can only see one "index" entry in the request's body and it is for the post only). I pushed a commit just to add some extra logging to the temp log file I'm using so it prints all requests.
> 
> Anyway, this is all happening in the editor/dashboard, but we still need to find a way to intercept the bulk indexing when creating the index from the CLI...
> 

(I shamelessly posted what I said to Pieter via RocketChat)

So I guess this needs further testing just in case. I'll do more tests tomorrow.

cc @sagarsdeshmukh @comzeradd

pablocubico:
@comzeradd Last update:

Ok, so I removed all the `error_log` debugging and did some more tests.

- About @Inwerpsel 's concern about WP indexing images on single Post updates: it does not index images, so we are good. To test that, just create a post without images, you should see the indexed Documents counter in ElasticHQ increase by 1. Then add a post with many images, the counter should also increase by 1, not more.
- The syncing of all documents (note that "images" are actually "posts" in WP, and "posts" are "Indexables" to ElasticPress), only happened while bulk indexing, either when clicking the sync icon, or when running the ElasticPress `index` CLI command (which is aliased as `make elastic` in our [Makefile](https://github.com/greenpeace/planet4-docker-compose/blob/master/Makefile#L230).
- The filter added solves the problem only when clicking the Sync icon.
- I'm not a fan of creating a fork, but as the ElasticPress CLI command ignores any filters added on the master theme (or the blocks plugin for that matter), I couldn't find a way to apply the same logic using filters, so the remaining option to fix the CLI indexing is forking the plugin. I created a fork but I'm not sure which branch of the *ElasticPress* repo we should be using... I have to double-check that. In the meantime, it's on EP's `develop` branch: https://github.com/greenpeace/ElasticPress/commit/d5a509a897fb272d8774465122c0b22fb6eb3a69
- Another option to avoid the fork would be to remove the CLI indexing from the Makefile, along with any calls to the `elastic` call there, and just run the indexing manually from the Dashboard as needed, for example, when a new P4 site is created or the index has to be rebuilt for some other reason.

comzeradd:
I don't think we currently rely on the cli command. We mostly rely on the asynchronous task WP is running to index new changes and documents. The upstream cli command is indeed useless at its current form, we have a [5075](https://jira.greenpeace.org/browse/PLANET-5075) to try create our own command that emulates the behavior of the asynchronous task and can respect the filters. This will prevent us from forking the plugin, since for the most part we are fine with it.

Another option would be to open an issue upstream, to see if there is an interest to contribute your patch directly to them.

classes/class-p4-search.php
```diff
@@ -1020,6 +1034,59 @@ public static function edit_search_mime_types( $where ) : string {
...
+		 * @param array  $args Arguments for the request.
+		 * @param array  $failures Errors array.
+		 */
+		public static function intercept_ep_request( $request, $query = null, $args = null, $failures = null ) {
+			$total_docs_count    = 0;
+			$indexed_docs_count  = 0;
+			$filtered_body_lines = [];
+
+			// Is this a bulk index task on the attachment pipeline?
+			if ( $query && ! ! preg_match( '/_bulk\?pipeline.*attachment/', $query['url'] ) ) {
```
Inwerpsel, line 1052: This filter would allow achieving the same thing, but in a much easier way: https://github.com/10up/ElasticPress/blob/develop/includes/classes/Indexable/Post/Post.php#L318

It contains the indexed item before it is passed to the bulk request or json encoded. If we check the mime type there then we can avoid it making it into the bulk request in the first place by returning `false`. That means we could achieve the same thing as here with about 4 lines of code, instead of 40, and no regex :)

pablocubico: Hmmm but how would you actually prevent the "post" (for example, an image) from being sent to the index by using this filter? 

According to the blockdoc it seems like it's just there for backwards compatibility, it's being applied inside the Elasticpress code just to do a little transform in the structure of the arguments here:

https://github.com/10up/ElasticPress/blob/3b482e632521310dbb82d9602d54bf5f2b281c9c/includes/classes/Feature/Documents/Documents.php#L213

But there's no way to prevent the sync from actually happening, or I'm not seeing it 🤔

comzeradd:
Should we close this one?

pablocubico:
Closing this in favor of a newer PR @Inwerpsel merged. 💯

PR #1088, opened 2020-05-19
PLANET-5099 Add PSR-4 autoloading

sagarsdeshmukh:
- Added PSR-4 autoloading in composer
- Remove class_exists() checks from classes dir
- For now I added the auto-generated autoload files(vendor/composer/*) in git, but we will automate this step in build pipeline.


lithrel:
One more proposition: a full implementation of PSR4 would also imply changing filenames of classes ([point 3.3 of the specs](https://www.php-fig.org/psr/psr-4/#2-specification)), so something like
`git mv classes/class-p4-post.php classes/P4_Post.php`  
for all classes.

Inwerpsel:
> One more proposition: a full implementation of PSR4 would also imply changing filenames of classes ([point 3.3 of the specs](https://www.php-fig.org/psr/psr-4/#2-specification)), so something like
> `git mv classes/class-p4-post.php classes/P4_Post.php`
> for all classes.

We can also follow the widely used convention of putting our code in a `src` folder. And, since we use namespaces, there is no reason to preserve the `P4_` prefixes in every classname.

We can also consider not using `_`, as the words are already split because of the camel case. That will make the class names a lot less awkward to type.

Something that could make sense too is to create a few sub-namespaces, so that not everything is in the same folder, as in PSR4 a sub-namespace translates to a separate folder.

Inwerpsel:
We created separate PRs for some of the changes in this PR: https://github.com/greenpeace/planet4-master-theme/pull/1120, https://github.com/greenpeace/planet4-master-theme/pull/1124, https://github.com/greenpeace/planet4-master-theme/pull/1123

PR #1090, opened 2020-05-21
PLANET-4891 Enforce Google Login for users with greenpeace.org email

dpivo:
Added filter to check if user has greenpeace email, and warns them to login with google

### DIFF
--- a/classes/class-p4-master-site.php
+++ b/classes/class-p4-master-site.php
@@ -44,6 +44,13 @@ class P4_Master_Site extends TimberSite {
 	protected $child_css = [];
 
 	/**
+	 * Variable that lets us know if the user has or hasn't used google to log in
+	 *
+	 * @var boolean $google_login_error
+	 */
+	protected $google_login_error = false;
+
+	/**
 	 * P4_Master_Site constructor.
 	 */
 	public function __construct() {
@@ -131,6 +138,8 @@ class P4_Master_Site extends TimberSite {
 			]
 		);
 
+		add_filter( 'authenticate', [ $this, 'enforce_google_signon' ], 4, 3 );
+		add_filter( 'authenticate', [ $this, 'check_google_login_error' ], 30, 3 );
 		add_filter( 'login_headerurl', [ $this, 'add_login_logo_url' ] );
 		add_filter( 'login_headertext', [ $this, 'add_login_logo_url_title' ] );
 		add_action( 'login_enqueue_scripts', [ $this, 'add_login_stylesheet' ] );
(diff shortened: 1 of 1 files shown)


classes/class-p4-master-site.php
```diff
@@ -605,6 +606,27 @@ public function view_meta_box_search( $post ) {
...
+	 */
+	function enforce_google_signon( $user, $username, $password ) {
+
+
+		if ( defined( 'WP_DEBUG' ) && is_bool( WP_DEBUG ) && WP_DEBUG ) {
+			return $user;
+		}
+
+		if ( strpos( $user->data->user_email, '@greenpeace' ) !== false ) {
+			return new WP_Error( 'login', __( 'You are trying to login with a Greenpeace email. Please use Google login button instead.' ) );
```
comzeradd, line 624: do we need to add the `planet4-master-theme-backend` text domain here for the translations?

classes/class-p4-master-site.php
```diff
@@ -605,6 +606,27 @@ public function view_meta_box_search( $post ) {
...
+	 * Forces a user to login using Google Auth if they have a greenpeace email
+	 * 
+	 * @param WP_User $user The current user logging in.
+	 * @param String  $username The username of the user.
+	 * @param String  $password The password of the user.
+	 */
+	function enforce_google_signon( $user, $username, $password ) {
+
+
+		if ( defined( 'WP_DEBUG' ) && is_bool( WP_DEBUG ) && WP_DEBUG ) {
```
Inwerpsel, line 619: `WP_DEBUG === true` is equivalent to `is_bool( WP_DEBUG ) && WP_DEBUG`

classes/class-p4-master-site.php
```diff
@@ -605,6 +606,27 @@ public function view_meta_box_search( $post ) {
...
+			return $user;
+		}
+
+		if ( strpos( $user->data->user_email, '@greenpeace' ) !== false ) {
+			return new WP_Error( 'login', __( 'You are trying to login with a Greenpeace email. Please use Google login button instead.' ) );
+		} else {
```
Inwerpsel, line 625: You don't need the else as the if returns early.

classes/class-p4-master-site.php
```diff
@@ -605,6 +606,27 @@ public function view_meta_box_search( $post ) {
...
+	 * @param String  $password The password of the user.
+	 */
+	function enforce_google_signon( $user, $username, $password ) {
+
+
+		if ( defined( 'WP_DEBUG' ) && is_bool( WP_DEBUG ) && WP_DEBUG ) {
+			return $user;
+		}
+
+		if ( strpos( $user->data->user_email, '@greenpeace' ) !== false ) {
```
Inwerpsel, line 623: Maybe less likely to ever occur, but this will also match unintended addresses like `someone@greenpeace_is_cool.com`. Since the ticket says "Prohibit login with username/password for user with greenpeace.org email", I suppose we shouldn't enforce this for other greenpeace domains (not sure if they're used in emails). So we could have this check if the email address ends with `@greenpeace.com`.

comzeradd: We should limit that to `greenpeace.org`. This is the global id for all GP staff. Other TLDs may not even use g-suite.

comzeradd:
I deployed this on [pluto](https://k8s.p4.greenpeace.org/test-pluto/admin) but I get the warning even when I try to login with the Google button. So I can't login at all.

classes/class-p4-master-site.php
```diff
@@ -605,6 +606,30 @@ public function view_meta_box_search( $post ) {
...
+
+		if ( defined( 'WP_DEBUG' ) && WP_DEBUG === true ) {
+			return $user;
+		}
+
+		if ( is_wp_error( $user ) ) {
+			return $user;
+		}
+
+		if ( strpos( $user->data->user_email, '@greenpeace.org' ) !== false ) {
```
lithrel, line 626: `=== false` :)

If you want to be exact (in case somebody ever uses a `greenpeace.org` subdomain like me@greenpeace.org.example.com, which will probably never happen, but eh 🤷 ), you can check from the end of the string with [`substr`](https://www.php.net/substr):
```php
$domain = '@greenpeace.org';
if ( substr( $user->data->user_email, -strlen($domain) ) !== $domain ) {
    // error
}
```
maybe even use [`mb_substr`](https://www.php.net/manual/en/function.mb-substr.php) (same syntax as substr) ? I'm not sure of what is allowed in email addresses, it might be safer.

comzeradd:
I re-deployed on pluto and it seems to work as expected. @lithrel @Inwerpsel can you have another pass in the code?

classes/class-p4-master-site.php
```diff
@@ -605,6 +614,59 @@ public function view_meta_box_search( $post ) {
...
 	}
 
+	/**
+	 * Forces a user to login using Google Auth if they have a greenpeace.org email
+	 *
+	 * @param WP_User $user The current user logging in.
+	 * @param String  $username The username of the user.
+	 * @param String  $password The password of the user.
+	 */
+	public function enforce_google_signon( $user, $username = null, $password = null ) {
```
lithrel, line 624: Nitpicking about type hints, but it helps to catch error early, for both functions
```php
	 * @param null|WP_User|WP_Error $user The current user logging in.
	 * @param null|string           $username The username of the user.
	 * @param null|string           $password The password of the user.
	 * @return null|WP_User|WP_Error
	 */
	public function ... ( $user, ?string $username = null, ?string $password = null ) {
```

classes/class-p4-master-site.php
```diff
@@ -605,6 +614,59 @@ public function view_meta_box_search( $post ) {
...
+		if ( defined( 'WP_DEBUG' ) && WP_DEBUG === true ) {
+			return $user;
+		}
+
+		if ( is_wp_error( $user ) ) {
+			return $user;
+		}
+
+		$domain = '@greenpeace.org';
+		if ( mb_substr( $user_data->data->user_email, -strlen( $domain ) ) === $domain ) {
```
lithrel, line 647: I reach this point if I try to log in with a user that does not exist, which triggers some notices because `$user_data` is `false`, so to not have notices and because `get_user_by()` returns `WP_User|false` I would change the first `if` like this
```php
if ( empty( $username ) || empty( $password ) ) {
	return $user;
} 

if ( strpos( $username, '@' ) ) {
	$user_data = get_user_by( 'email', trim( wp_unslash( $username ) ) );
} else {
	$login     = trim( $username );
	$user_data = get_user_by( 'login', $login );
}

if ( false === $user_data ) {
	return $user;
}
```

classes/class-p4-master-site.php
dpivo: Nice!

classes/class-p4-master-site.php
dpivo: Good catch!

PR #1093, opened 2020-05-22
PLANET-5058 Update styleguide for table colors

mleray:
See https://jira.greenpeace.org/browse/PLANET-5058
We want to offer the same color options for the Table block as we do for the Spreadsheet block. The only difference is that tables also have a footer, which requires an additional color.
Related PRs: 
- Styleguide: https://github.com/greenpeace/planet4-styleguide/pull/56
- Gutenberg blocks: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/278


mleray:
@pablocubico Maybe those extra changes you see in the submodule are because I've had to do a rebase?

PR #1094, opened 2020-05-25
PLANET-5032 Remove duplicate ES filter on post type

Inwerpsel:
see https://jira.greenpeace.org/browse/PLANET-5032
* There is already a WP filter in `class-p4-search.php` that adds the
same ES filter. Moreover, the removed one added it on `post_type`
instead of `post_type.raw`. This caused page filter to break if the
ElasticPress language is something else than English. No idea why it did work for English.

### DIFF
--- a/classes/class-p4-elasticsearch.php
+++ b/classes/class-p4-elasticsearch.php
@@ -48,12 +48,6 @@ if ( ! class_exists( 'P4_ElasticSearch' ) ) {
 														],
 													];
 												}
-												// Make sure it is a Page.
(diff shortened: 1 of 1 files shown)


Inwerpsel:
> Looks good  There is no test instance for this?

Thanks! There's one on the ticket but I still needed to update it to the latest version. Done now: https://k8s.p4.greenpeace.org/test-deimos/?s=spread+sheet&orderby=_score&f%5Bctype%5D%5BPage%5D=2

PR #1097, opened 2020-05-27
PLANET-2209 Add success notification on P4 settings saved

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-2209

Added a success notification message after saving P4 settings.

Screenshot 2020-05-27 at 09 38 47

## Testing

Go to `/wp-admin` Settings > Planet4, and save the form.  
If no problem occurs, a message "Settings saved." appears, as it does in Settings > General.

## Notes

I followed the same method used in [General Settings code](https://github.com/WordPress/WordPress/blob/57fb3c6cf016678ab38d7a636b8df41fa2d955f1/wp-admin/options.php#L313), and put it in an [after_save](https://github.com/CMB2/CMB2/blob/af56b9eb227a3db9694510ca21bf9e9058e640f0/includes/CMB2.php#L854) hook of CMB2. Also parameterized the ID of the metabox, so that the hook doesn't break if we change it. I need somebody to tell me how wrong I am 😨 

I have used the domain `planet4-master-theme-backend` for the string `Settings saved.` although it's a string used by wordpress itself. I did it because a sniff `WordPress.WP.I18n.TextDomainMismatch` insulted me, but I don't know if it's the right thing to do here.

### DIFF
--- a/classes/class-p4-settings.php
+++ b/classes/class-p4-settings.php
@@ -12,6 +12,11 @@ if ( ! class_exists( 'P4_Settings' ) ) {
 	class P4_Settings {
 
 		/**
+		 * ID of the Metabox
+		 */
+		private const METABOX_ID = 'option_metabox';
+
+		/**
 		 * Option key, and option page slug
 		 *
 		 * @var string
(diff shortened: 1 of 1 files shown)


classes/class-p4-settings.php
```diff
@@ -18,6 +18,13 @@ class P4_Settings {
 		 */
 		private $key = 'planet4_options';
 
+		/**
+		 * ID of the Metabox
+		 *
+		 * @var string
+		 */
+		private $metabox_id = 'option_metabox';
```
Inwerpsel, line 26: You can make this a `private const`

Inwerpsel:
> I followed the same method used in [General Settings code](https://github.com/WordPress/WordPress/blob/57fb3c6cf016678ab38d7a636b8df41fa2d955f1/wp-admin/options.php#L313), and put it in an [after_save](https://github.com/CMB2/CMB2/blob/af56b9eb227a3db9694510ca21bf9e9058e640f0/includes/CMB2.php#L854) hook of CMB2. Also parameterized the ID of the metabox, so that the hook doesn't break if we change it. I need somebody to tell me how wrong I am 

I guess you're not wrong, seems the proper way to do it in WP. Well, proper as in there is no better way they offer :)

> I have used the domain `planet4-master-theme-backend` for the string `Settings saved.` although it's a string used by wordpress itself. I did it because a sniff `WordPress.WP.I18n.TextDomainMismatch` insulted me, but I don't know if it's the right thing to do here.

Looks like that is indeed the recommended way, as the core texts might change without notice, which would result in untranslated text, so plugins or themes should always use their own domain.

PR #1103, opened 2020-06-03
PLANET-5060 Remove external links features for Take Action card links

mleray:
For external links we usually show an svg icon, but for the Take Action card (https://jira.greenpeace.org/browse/PLANET-5060) and for the Take Action boxout (https://jira.greenpeace.org/browse/PLANET-5126) it doesn't fit the design.

Also, we usually open external links in a new tab, but Take Action cards always open in the same page, and for the Take Action boxouts it's a setting (open in new tab or not) that is taken care of in this file.

### DIFF
--- a/assets/src/js/external_links.js
+++ b/assets/src/js/external_links.js
@@ -10,7 +10,7 @@ export const setupExternalLinks = function($) {
       </g>
   </svg>`;
 
-  $('div.page-template a:not(.btn), article a:not(.btn)').each(function () {
+  $('div.page-template a:not(.btn):not(.cover-card-heading), article a:not(.btn)').each(function () {
     let href = undefined === $(this).attr('href') ? '' : $(this).attr('href');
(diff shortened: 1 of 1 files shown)


assets/src/js/external_links.js
```diff
@@ -10,7 +10,7 @@ export const setupExternalLinks = function($) {
       </g>
   </svg>`;
 
-  $('div.page-template a:not(.btn), article a:not(.btn)').each(function () {
+  $('div.page-template a:not(.btn):not[class^="cover-card"], article a:not(.btn)').each(function () {
```
comzeradd, line 13: I think I found why tests fail :bulb: the selection should be enclosed inside `()`, similar to `(.btn)`. Now the script fails, causing some follow-up js code not to run at all.

You can also make the selector more explicit, since we only need to catch the header.

eg. `:not([class="cover-card-heading"])`

mleray: In that case I'll just change it to `:not(.cover-card-heading)` :)

PR #1107, opened 2020-06-08
PLANET-5145 Fix deleted terms crashing search

Inwerpsel:
* If a term was deleted but ElasticSearch still had the old tag, then
this would cause a fatal error because the term was fetched and then
assumed it existed, causing a WP_Error to be used as a string.
* Return early if there are no posts.

### DIFF
--- a/classes/class-p4-search.php
+++ b/classes/class-p4-search.php
@@ -351,45 +351,41 @@ if ( ! class_exists( 'P4_Search' ) ) {
 
 			$posts = $this->query_posts( $paged );
 
-			if ( $posts ) {
-				foreach ( $posts as $post ) {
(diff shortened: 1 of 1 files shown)


classes/class-p4-search.php
```diff
@@ -351,45 +351,41 @@ protected function get_posts( $paged = 1 ) : array {
...
-
-						// @todo Ensure the term link is synced to ElasticSearch so we don't have to fetch it here.
-						$template_post->tags = array_map(
-							function ( $tag ) {
-								$tag['link'] = get_term_link( $tag['term_id'] );
-								return $tag;
-							},
-							$template_post->tags
-						);
+			if ( ! $posts ) {
```
lithrel, line 354: `::query_posts` only returns `array`, so maybe `if ( empty( $posts ) )` ?

PR #1108, opened 2020-06-08
PLANET-5128: Expand images lazy-loading on Image block

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-5128

### Note
Wordpress 5.5 will add the `loading` attribute for all images by default. We will have to check if this does not break our own implementation on browsers that support it (and see if we want to migrate to it ? Browser support might not be good enough).
- Implementation in browsers: https://caniuse.com/#feat=loading-lazy-attr
- Announcement: https://make.wordpress.org/core/2020/01/29/lazy-loading-images-in-wordpress-core/
- Changes on `core`: https://core.trac.wordpress.org/changeset/47554

## Changes

The approach I took follows function [`wp_make_content_images_responsive`](https://github.com/WordPress/WordPress/blob/5.4-branch/wp-includes/media.php#L1490) that adds `srcset` and `sizes` attributes to all wp images in content. This `scrset` attribute triggers a network call, so it has to be edited before display.  
The modification currently impacts all images in content that are not already lazy-loaded, so it also impacts Columns block for example, but it can be limited to images with "wp-image-{}" class (image block) if you think it's better.

## Tests

Images added by the Image block should have attributes `data-src`, `data-srcset` and `data-sizes`, and should not trigger a network call if they are out of the viewport.  
Other images like Columns block in the Explore page should also be lazy-loaded.  
Images already marked as `lazyload` or `preload` should not change.  
Test instance available on [Tavros](https://k8s.p4.greenpeace.org/test-tavros/)

_I have unit tests there https://gist.github.com/lithrel/2ab39f07391390e66d13b5fcd0121cbb but I don't know yet how to integrate them properly_

### DIFF
--- a/classes/class-p4-master-site.php
+++ b/classes/class-p4-master-site.php
@@ -168,6 +168,7 @@ class P4_Master_Site extends TimberSite {
 			10,
 			2
 		);
+		add_filter( 'the_content', [ $this, 'make_content_images_lazyload' ] );
 
 		add_action( 'init', [ $this, 'login_redirect' ], 1 );
 		add_filter( 'attachment_fields_to_edit', [ $this, 'add_image_attachment_fields_to_edit' ], 10, 2 );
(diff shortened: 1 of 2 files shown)


sagarsdeshmukh:
Tested the updated PHPUnit test code on local. looks good 👍

PR #1115, opened 2020-06-11
PLANET-4818: PoC for a dashboard P4 message

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-4818

PoC for a single informative message from Planet4 team on the dashboard.

Screenshot 2020-06-12 at 15 31 56

## Functionalities

The ticket didn't say too much about specific functionalities, so this was extrapolated.

The message:
- is in HTML format
- is integrated by a developer  
- appears only on the dashboard
- is dismissible, at a user level
- once dismissed, doesn't reappear until a new release is pushed

and:
- only one Planet4 message can appear at the same time

### Alternatives

- Fetching a message from a document
  We could think about fetching the message from a distant document, it raises the problems of formatting and control of that document.
- Adding a permanent widget, with a Planet4 newsfeed
  This would require a newsfeed; I don't know if the widget would always appear on top, I think the user can reorder those at will and could put us at the bottom, which I guess would defeat the goal of the ticket
- Insert the message in database through a patch (`post-deploy`)
  Approximately equivalent, but it could also allow us to keep in the instance a history of those messages, that the user could read/search in

### Changes

A function plugged on `admin_notices` is in charge of displaying the message,  a function on `wp_ajax_*` handles the dismiss. No existing code is changed.  

## Tests

A message should appear on the dashboard. Once dismissed by clicking on the little cross on the top right of the message box, it should disappear and not reappear on refresh. You can make it reappear by modifying the user meta `last_p4_notice` to a lower version.
```console
$ wp user meta get 1 last_p4_notice
2.32.3
$ wp user meta update 1 last_p4_notice '2.32.0'
Success: Updated custom field 'last_p4_notice'.
```

### DIFF
--- a/classes/class-p4-master-site.php
+++ b/classes/class-p4-master-site.php
@@ -16,6 +16,20 @@ use Timber\Menu as TimberMenu;
 class P4_Master_Site extends TimberSite {
 
 	/**
+	 * Key of notice seen by user
+	 *
+	 * @var string
+	 */
+	private const DASHBOARD_MESSAGE_KEY = 'last_p4_notice';
+
+	/**
+	 * Version of notice
+	 *
+	 * @var string
+	 */
+	private const DASHBOARD_MESSAGE_VERSION = '0.1';
+
+	/**
 	 * Theme directory
 	 *
 	 * @var string $theme_dir
@@ -174,6 +188,8 @@ class P4_Master_Site extends TimberSite {
 		add_filter( 'attachment_fields_to_edit', [ $this, 'add_image_attachment_fields_to_edit' ], 10, 2 );
 		add_filter( 'attachment_fields_to_save', [ $this, 'add_image_attachment_fields_to_save' ], 10, 2 );
 		add_action( 'init', [ $this, 'p4_register_core_image_block' ] );
+		add_action( 'admin_notices', [ $this, 'show_dashboard_notice' ] );
+		add_action( 'wp_ajax_dismiss_dashboard_notice', [ $this, 'dismiss_dashboard_notice' ] );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


Inwerpsel:
Looks great so far :100: 

> We could think about fetching the message from a distant document, it raises the problems of formatting and control of that document.
Maybe we can pull this from Github in a way? Could be that they have something in place specifically for this purpose. Formatting and control doesn't seem an issue in that case, as I assume the APIs Github offers for that are stable and rarely change.

> The ticket didn't say too much about specific functionalities, so this was extrapolated.
I think when we discussed this earlier the idea also included the ability to send more types of notifications, maybe even to specific users (e.g. "A new post was submitted for review" sent to only the users that can perform review / a specific user). If we would want to set up something like that it would be interesting to already create a notifications table / post type to store the read status for each user.

We could also use a meta field that can have multiple values for a key, where we can store all unread messages (or all read messages, not sure which approach would work best).

lithrel:
> I think when we discussed this earlier the idea also included the ability to send more types of notifications, maybe even to specific users (e.g. "A new post was submitted for review" sent to only the users that can perform review / a specific user). If we would want to set up something like that it would be interesting to already create a notifications table / post type to store the read status for each user.

So far neither @suzi-greenpeace nor Luca expressed a need for a message history or a more versatile message system, and Luca asked if we could test this this week, so I'm a bit conflicted about this; I'd be happy to code a bigger thing, but I don't have more demand or features descriptions <sub>and I'm known for adding unnecessary features everywhere</sub> 🤷 

I've hardcoded the version number, but do we have that string updated every release somewhere in a constant or in db ?

comzeradd:
Looks nice! :+1: 

I wouldn't go so far as keeping the history of all the messages. Regarding the versioning, the only place I can think of is that we save all packages current version to display them in the [Dev Report](https://github.com/greenpeace/planet4-base-fork/blob/develop/tasks/post-deploy/02-get-gp-packages.sh). If you think that's usable, we can pick the master-theme version, or amend that script to include the base-fork version. Otherwise we could also use just an arbitrary version number.

lithrel:
Right, master-theme version would make sense. The fact that this is a dev thing only makes me a bit scared it could disappear/change format, so I'll stick with an arbitrary static version for now, we'll see if we find a use for that version somewhere else.
I'll keep that bit of code here, and if we end up using that version number, I'll try to unify the gathering of information between dev report and this in a specific class.
```php
	private function set_theme_version(): void
	{
		$packages = get_option( 'greenpeace_packages', [] );
		if ( empty( $packages ) ) {
			return;
		}

		$mt_key = array_search( 'greenpeace/planet4-master-theme', array_column( $packages, 0 ) );
		if ( false === $mt_key ) {
			return;
		}
		
		$this->theme_version = isset( $packages[ $mt_key ][ 1 ] )
			? ltrim( $packages[ $mt_key ][ 1 ], 'v' )
			: null;
	}
```

PR #1123, opened 2020-06-18
PLANET-5099 Use psr-4

Inwerpsel:

All PHP classes in `classes` directory converted to PSR-4 format in namespace `P4\MasterTheme` (we can still change the naming).

* The `src` dir already existed and contained an unused css file, which I removed.
* I split it in mostly 1 commit per class, though there are a few with more than one.
* Because I used PHPStorm to move the classes into the namespace, this introduced some changes in these commits which are overwritten by later commits. This might seem a bit of a detour but it helped in ensuring the code kept working with each change. I noticed that I could use PHPStorm "move to namespace" functionality to handle both moving to src and adapting the internal namespaces only after a few commits (I believe during [this one](https://github.com/greenpeace/planet4-master-theme/pull/1123/commits/a8dc821039691f88fd8f074dfa5148cf5b70e96e)), so the earlier ones might look a bit different.
* Besides removing the prefix and underscores, I kept the naming so that we can adapt the old autoloading files to include the new one, which contain an alias to the old name at the end. I verified that this works as there is a test that depends on it, which was failing without it (see https://app.circleci.com/pipelines/github/greenpeace/planet4-master-theme/1308/workflows/0ebfa803-54e3-4465-b54f-f97890dd4e79/jobs/8334 and reproduced error on test-uranus ([sentry-error](https://sentry.greenpeace.org/greenpeace/planet4/issues/1593/?query=is%3Aunresolved)).
* At some points I couldn't prevent PHPStorm from applying CS rules while doing the move. I manually removed most of them again, but kept some that are easy to review.
* I also included a commit at the end to fix the last CS errors flagged in the pipeline. It didn't seem worth the effort to amend each respective commit and there's not that much of them. But maybe for reviewing it can be easier to check that commit first, so that you can recognize these changes when going through the full PR file changes.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,6 +25,11 @@
     "wiki": "https://github.com/greenpeace/planet4-master-theme/wiki",
     "source": "https://github.com/greenpeace/planet4-master-theme"
   },
+  "autoload": {
+    "psr-4": {
+      "P4\\MasterTheme\\": "src/"
+    }
+  },
   "extra": {
     "installer-paths": {
       "vendor/plugins/{$name}/": ["type:wordpress-plugin"]
--- a/functions.php
+++ b/functions.php
@@ -25,6 +25,7 @@ function simple_value_filter( string $filter_name, $value, $priority = null ): v
 	);
 }
 
+use P4\MasterTheme\Loader;
 use Timber\Timber;
 
 if ( ! class_exists( 'Timber' ) ) {
@@ -55,5 +56,5 @@ if ( ! class_exists( 'Timber' ) ) {
 	}
 }
 
-require_once __DIR__ . '/classes/class-p4-loader.php';
-P4_Loader::get_instance();
+require_once __DIR__ . '/src/Loader.php';
+Loader::get_instance();
--- a/functions.php
+++ b/functions.php
@@ -69,7 +69,6 @@ if ( ! class_exists( 'Timber' ) ) {
 	}
 }
 
-require_once __DIR__ . '/src/Loader.php';
 Loader::get_instance();
 
 require_once 'load-class-aliases.php';
--- a/functions.php
+++ b/functions.php
@@ -5,6 +5,19 @@
  * @package P4MT
  */
 
+// Composer install might have happened inside the master-theme directory, instead of the app root. Specifically for
+// php lint and unit test job this is the case.
+// Also this helps with local development environments that pulled in changes for master-theme but did not rerun
+// base-fork's composer installation, which is currently tricky to do as it puts the fetched versions of master theme
+// and the plugins into the theme and plugin folders, which might be messy if you have changes there.
+// With this fallback for tests in place, you can just run composer dump autoload in master-theme.
+// Probably there's a better way to handle this, but for now let's try load both.
+if ( file_exists( __DIR__ . '/vendor/autoload.php' ) ) {
+	require_once __DIR__ . '/vendor/autoload.php';
+} else {
+	require_once __DIR__ . '/../../../../vendor/autoload.php';
+}
+
 /**
  * A simpler way to add a filter that only returns a static value regardless of the input.
  *
@@ -58,3 +71,5 @@ if ( ! class_exists( 'Timber' ) ) {
 
 require_once __DIR__ . '/src/Loader.php';
 Loader::get_instance();
+
+require_once 'load-class-aliases.php';
--- a/src/AnalyticsValues.php
+++ b/src/AnalyticsValues.php
@@ -324,5 +324,3 @@ final class AnalyticsValues {
 	}
 }
 
-class_alias( AnalyticsValues::class, 'P4_Analytics_Values' );
-
--- a/src/AnalyticsValues.php
+++ b/src/AnalyticsValues.php
@@ -7,7 +7,7 @@
 
 namespace P4\MasterTheme;
 
-use P4_Smartsheet;
+use P4\MasterTheme\Smartsheet;
 use P4_Smartsheet_Client;
 
 /**
@@ -40,13 +40,14 @@ final class AnalyticsValues {
 	/**
 	 * Extract values from smartsheets.
 	 *
-	 * @param P4_Smartsheet      $global_smartsheet Smartsheet containing global projects data.
-	 * @param P4_Smartsheet|null $local_smartsheet Smartsheet containing local projects data.
+	 * @param Smartsheet      $global_smartsheet Smartsheet containing global projects data.
+	 * @param Smartsheet|null $local_smartsheet Smartsheet containing local projects data.
+	 *
 	 * @return static
 	 */
 	public static function from_smartsheets(
-		P4_Smartsheet $global_smartsheet,
-		P4_Smartsheet $local_smartsheet = null
+		Smartsheet $global_smartsheet,
+		Smartsheet $local_smartsheet = null
 	): self {
 		$project_name_column = $global_smartsheet->get_column_index( 'Global Project Standard' );
 		$approved_column     = $global_smartsheet->get_column_index( 'Standard Approved' );
--- a/src/AnalyticsValues.php
+++ b/src/AnalyticsValues.php
@@ -7,9 +7,6 @@
 
 namespace P4\MasterTheme;
 
-use P4\MasterTheme\Smartsheet;
-use P4\MasterTheme\SmartsheetClient;
-
 /**
  * Data class for post editor analytics values.
  */
--- a/src/AnalyticsValues.php
+++ b/src/AnalyticsValues.php
@@ -5,10 +5,15 @@
  * @package P4MT
  */
 
+namespace P4\MasterTheme;
+
+use P4_Smartsheet;
+use P4_Smartsheet_Client;
+
 /**
  * Data class for post editor analytics values.
  */
-final class P4_Analytics_Values {
+final class AnalyticsValues {
 	private const CACHE_KEY = 'analytics_global_projects';
 
 	/**
@@ -320,3 +325,6 @@ final class P4_Analytics_Values {
 		return $matching_project['tracking_id'];
 	}
 }
+
+class_alias( AnalyticsValues::class, 'P4_Analytics_Values' );
+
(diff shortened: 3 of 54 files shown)


composer.json
```diff
@@ -11,10 +11,6 @@
   ],
   "license": "MIT",
   "repositories": [
-    {
-      "type": "composer",
-      "url": "https://wpackagist.org"
-    },
```
Inwerpsel, line 17: Remove this as it's identical to the other entry.

Inwerpsel:
> I think you only missed the `P4_Loader::theme_file_ver` recently merged
> 
> ```
> $ rg "P4_[a-zA-Z0-9_]*(::|\()" planet4-master-theme
> persistence/app/public/wp-content/themes/planet4-master-theme/src/PostReportController.php
> 105:			P4_Loader::theme_file_ver( 'admin/js/posts_report.js' ),
> 
> persistence/app/public/wp-content/themes/planet4-master-theme/src/PostCampaign.php
> 211:			P4_Loader::theme_file_ver( 'admin/css/campaign.css' )
> 
> persistence/app/public/wp-content/themes/planet4-master-theme/src/ControlPanel.php
> 296:			P4_Loader::theme_file_ver( 'admin/css/dashboard.css' )
> 302:			P4_Loader::theme_file_ver( 'admin/js/dashboard.js' ),
> 
> persistence/app/public/wp-content/themes/planet4-master-theme/src/Campaigns.php
> 430:			P4_Loader::theme_file_ver( "admin/js/$this->taxonomy.js" ),
> 
> persistence/app/public/wp-content/themes/planet4-master-theme/src/MasterSite.php
> 251:			P4_Loader::theme_file_ver( 'admin/css/login.css' )
> 590:			P4_Loader::theme_file_ver( 'assets/build/bootstrap.min.css' )
> ```

Yeah noticed too as it made the test fail, thanks for the alertness though! I had too much blind faith in PHPStorm's resolve conflicts functionality (it did the right thing but I forgot to check the errors in the conflict resoution dialog).

Inwerpsel:
After the last rebase [the pipeline failed](https://app.circleci.com/pipelines/github/greenpeace/planet4-master-theme/1315/workflows/1a16980e-96b2-40fb-bf71-94938a89f964/jobs/8353), though looking at [the diff with the previous last commit](https://github.com/greenpeace/planet4-master-theme/compare/30c929b067ed4c1a941c2326a9dd5ec23d002848..8c94a7c70a536b1ee6964207c8f6550612fe5903), which did pass, I couldn't find anything that could have caused it to fail. What failed was the `Build - Build containers` step, which actually runs docker inside of it as part of the build (not sure why it needs to do this), so because of that it didn't contain a stack trace, it only prints the logs of the running container. They contained this:
```
build_proxy | 2020/06/24 12:42:12 [error] 120#120: *150 FastCGI sent in stderr: "PHP message: PHP Warning:  require_once(/app/source/public/wp-content/themes/planet4-master-theme/src/\MasterTheme\Loader.php): failed to open stream: No such file or directory in /app/source/public/wp-content/themes/planet4-master-theme/src/Loader.php on line 89
build_proxy | PHP message: PHP Fatal error:  require_once(): Failed opening required '/app/source/public/wp-content/themes/planet4-master-theme/src/\MasterTheme\Loader.php' (include_path='.:/usr/share/php') in /app/source/public/wp-content/themes/planet4-master-theme/src/Loader.php on line 89" while reading response header from upstream, client: 127.0.0.1, server: www.planet4.test, request: "GET / HTTP/1.1", upstream: "fastcgi://172.21.0.4:9000", host: "localhost"
b
```
After going through the code a couple of times and finding nothing I decided to rerun the job and it did pass :exploding_head: . So I gues it was some kind of fluke of the CI job, so I'll rerun it a couple of times to ensure that it's nothing that can actually occur on production.

src/AnalyticsValues.php
```diff
@@ -5,10 +5,15 @@
  * @package P4MT
  */
 
+namespace P4\MasterTheme;
+
+use P4\MasterTheme\Smartsheet;
```
Inwerpsel, line 10: I might remove these internal imports again as they're not really needed. They were added by PHPStorm as I renamed the classes one by one, and each time it needed to add it first to classes that used the renamed class but were not in the namespace yet.

src/Cookies.php
```diff
@@ -5,17 +5,19 @@
...
  * @since 1.9
  */
-class P4_Cookies {
+class Cookies {
 
 	const COOKIE_NAME = 'greenpeace';
 
 	/**
-	 * P4_Cookies constructor.
+	 * P4\MasterTheme\P4_Cookies constructor.
```
Inwerpsel, line 20: Right still need to update comment references, I'm on it.

src/Loader.php
```diff
@@ -60,12 +72,20 @@ private function __construct( $services ) {
...
+		if ( file_exists( __DIR__ . '/../vendor/autoload.php' ) ) {
+			require_once __DIR__ . '/../vendor/autoload.php';
+		} else {
+			require_once __DIR__ . '/../../../../../vendor/autoload.php';
+		}
 		try {
 			// Class names need to be prefixed with P4 and should use capitalized words separated by underscores. Any acronyms should be all upper case.
 			spl_autoload_register(
 				function ( $class_name ) {
 					if ( strpos( $class_name, 'P4_' ) !== false ) {
```
lithrel, line 86: while we're here `0 === strpos( $class_name, 'P4_' )` would be safer 🙏

src/Loader.php
```diff
@@ -60,12 +72,20 @@ private function __construct( $services ) {
 	 * Load required files.
 	 */
 	private function load_files() {
+		// Composer install might have happened inside the master-theme directory, instead of the app root.
+		// Probably there's a better way to handle this, but for now let's try load both.
+		if ( file_exists( __DIR__ . '/../vendor/autoload.php' ) ) {
+			require_once __DIR__ . '/../vendor/autoload.php';
+		} else {
+			require_once __DIR__ . '/../../../../../vendor/autoload.php';
```
lithrel, line 80: don't know if there's a better way yet 🤷‍♂️ but here's 2 versions with less dots / less based on current dir
```php
// equivalent to __DIR__ . '/../vendor/autoload.php';
dirname(__DIR__, 1) . '/vendor/autoload.php';
\get_template_directory() . '/vendor/autoload.php';

// equivalent to __DIR__ . '/../../../../../vendor/autoload.php';
dirname(__DIR__, 5) . '/vendor/autoload.php';
dirname(\ABSPATH) . '/vendor/autoload.php';
```
for whatever reason, [`get_home_path()`](https://developer.wordpress.org/reference/functions/get_home_path/) is not available there :|

src/PostReportController.php
```diff
@@ -5,10 +5,12 @@
...
  */
 
+namespace P4\MasterTheme;
+
 /**
- * Class P4_Post_Report_Controller
+ * Class P4\MasterTheme\P4_Post_Report_Controller
  */
-class P4_Post_Report_Controller {
+class PostReportController {
```
sagarsdeshmukh, line 13: Can we make the classname to `PostReport` only?

src/Loader.php
sagarsdeshmukh: My +1 for moving a autoloading statement in functions.php and dropping a `load_files()` method from Loader class.

src/PostReportController.php
Inwerpsel: We would need to check if something still uses the old class name, as currently the fallback transforms the old class name into the new one to find the file. But maybe let's do renames separately after this is merged? There are a couple more that could use more accurate naming (e.g. `ImageCompression` which does more than compression), so could make sense to look at these together, as we might also introduce some folders.

Inwerpsel: But definitely well spotted that we should change this :100: Controller is a term that is used for so many things that it doesn't add any meaning but rather adds confusion.

src/AnalyticsValues.php
Inwerpsel: done

src/Cookies.php
Inwerpsel: done

src/Loader.php
Inwerpsel: It would be nice to be able to remove the custom function already, one concern that I have is that this changes the loading order, which might cause some unexpected surprises. Like you encountered with `ImageCompression`. There might be more of these cases so we would need to check all usages of WP files if we would need to add the `require_once` calls to our files.

Inwerpsel: > I've tried a different approach for the autoload, the diff is in this gist https://gist.github.com/lithrel/78e1a7824292d162bb4d01a6b0c2afd8
Yes that's definitely better. I included the changes from the gist but just moved it in a separate file. For the `ImageCompression` class it actually makes sense to require the files as the class doesn't work without them anyway. :+1:

Inwerpsel:
I didn't rename the test classes yet, as these are not autoloaded so we can do that separately.

functions.php
```diff
@@ -55,5 +69,7 @@ function( $template ) {
 	}
 }
 
-require_once __DIR__ . '/classes/class-p4-loader.php';
-P4_Loader::get_instance();
+require_once __DIR__ . '/src/Loader.php';
```
lithrel, line 72: I guess we don't need it since the autoloader is included on top of the page ?

Inwerpsel:
> You kept the require dotted path `../../../` on purpose ? there was a thumb up on my comment, wondering if you forgot about it or concluded it was better this way

I first overlooked applying that change, but checking it now, it looks like PHPStorm is not recognizing this syntax and can't point to the correct file like it does when we use `../../..`. So maybe let's keep it like it is so that it doesn't keep reporting this as a warning.
Screenshot from 2020-06-26 13-33-30


> This PR makes me happy

lithrel:
> PHPStorm is not recognizing this syntax

Completely right, just went to upvote this with all the thumbs I have https://youtrack.jetbrains.com/issue/WI-35143

Inwerpsel:
Did a few sanity checks for some things that have low coverage on https://k8s.p4.greenpeace.org/test-uranus/ (campaign import/export, settings, metaboxes). All seems well, so let's merge this :crossed_fingers:

PR #1124, opened 2020-06-18
PLANET-5099 Use ::class references instead of strings

Inwerpsel:

So that we don't miss any when changing file names/structure.

Tested here: https://github.com/greenpeace/planet4-test-uranus/blob/6e76e4316e8f34f751a1b2013c6b518348fedaf8/composer-local.json

### DIFF
--- a/classes/class-p4-master-site.php
+++ b/classes/class-p4-master-site.php
@@ -37,13 +37,6 @@ class P4_Master_Site extends TimberSite {
 	protected $sort_options;
 
 	/**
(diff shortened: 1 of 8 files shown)


classes/class-p4-master-site.php
```diff
@@ -36,13 +36,6 @@ class P4_Master_Site extends TimberSite {
 	 */
 	protected $sort_options;
 
-	/**
```
Inwerpsel, line 39: This was unused.

PR #1125, opened 2020-06-18
PLANET-5099 Add a command as alternative for eval

Inwerpsel:
* The usage of `wp eval` in planet4-docker-compose makes it hard to
provide backwards compatibility for P4_Activator when we move to psr-4,
so provide a WP CLI command instead.

PR for the post deploy task in base-fork: https://github.com/greenpeace/planet4-master-theme/pull/1125

Also cleaned up role registration a bit: 
* There were 2 separate calls outside of the functions per role which also added some capabilities, while there was no reason for them to be separate so I added them with the rest.
* Create a map with all roles and their capabilities, which is less noisy to watch than a `add_cap` call inside of separate functions + mandatory docblock ...
* Made functions static so we don't need to create pointless instances.

Deployed here https://github.com/greenpeace/planet4-test-uranus/commit/0695728f02053e478561b2de976c09ece7917ea6

### DIFF
--- a/classes/class-p4-activator.php
+++ b/classes/class-p4-activator.php
@@ -29,21 +29,6 @@ class P4_Activator {
 	 * Run activation functions.
 	 */
 	public static function run(): void {
-		self::add_custom_roles_and_capabilities();
-	}
-
-	/**
-	 * Add campaigner role and its capabilities.
-	 */
-	private static function add_custom_roles_and_capabilities(): void {
-
-		$campaigner = new P4_Campaigner();
-		$campaigner->register_role_and_add_capabilities();
-
-		// Needed to allow the editor rule to change the author of a post in the document sidebar. The users data for that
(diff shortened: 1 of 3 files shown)


Inwerpsel:
> > Made functions static so we don't need to create pointless instances.
> 
> On a side-note, pointless instances are sometimes better than static usage, as it is can become [quite difficult to unit test](http://misko.hevery.com/2008/12/15/static-methods-are-death-to-testability/) a static method using other static methods (you end up not being able to inject a mock of your dependency)

Probably in those cases this indicates there is a structural problem with the code being tested and mocking might not be the right solution. Not sure if mocking is ever a good solution, really. Instead of creating a mock in tests (which is kinda an "invasive procedure" as you're basically changing the code only during tests by tricking the runtime to resolve a class to something else), it's probably better to invest the same effort in creating an interface in the source code, so that you can easily replace the dependency with another class implementing the interface.

PR #1127, opened 2020-06-22
PLANET-5186: Switch script version to filectime

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-5186
Linked: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/302

Script version is not automatically updated when we make changes, which leads to cache problems if we forget it.

## Fix

Switch enqueue/register scripts from hardcoded version to `filectime`.

Versions were found with regex `[0-9]{1,2}\.[0-9]{1,2}`, and by looking at `wp_enqueue_style` and `wp_enqueue_script` usage.
_I didn't write a specific larger function for this, because at this point I would prefer to create a new class that could be used in other repos and would have the paths needed and version function for each equivalent wordpress enqueue/register functions; and I'm not doing that because we currently are in a namespace refactoring + another large refactoring. This modification seems like enough at the moment._

## Test

On the editor side while editing a campaing, in the network calls or in html source, calls to `planet4-master-theme/` assets should be versioned with a timestamp as version number. You can filter by `ver=159` or `planet4-master-theme` to see those calls.  
No network call related to this theme should be a 404, on the public and the admin sides.

### DIFF
--- a/classes/class-p4-loader.php
+++ b/classes/class-p4-loader.php
@@ -201,4 +201,18 @@ final class P4_Loader {
 
 		return false;
 	}
+
+	/**
+	 * @param string $rel_path Relative path to the file.
+	 * @return int timestamp of file creation
+	 */
+	public static function theme_file_ver( string $rel_path ): int {
+		$filepath = trailingslashit( get_template_directory() ) . $rel_path;
+		$ctime    = filectime( $filepath );
+		if ( $ctime ) {
+			return $ctime;
+		}
+
+		return time();
+	}
 }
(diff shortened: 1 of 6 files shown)


classes/class-p4-loader.php
```diff
@@ -201,4 +201,18 @@ public function do_not_delete_autosave( $delete = null, $post = null, $force_del
...
 		return false;
 	}
+
+	/**
+	 * @param string $rel_path Relative path to the file.
+	 * @return int timestamp of file creation
+	 */
+	public static function theme_file_ver( string $rel_path ): int {
+		$filepath = trailingslashit( get_template_directory() ) . $rel_path;
+		$ctime = filectime( $filepath );
```
comzeradd, line 211: > Equals sign not aligned with surrounding assignments; expected 4 spaces but found 1 space

:grin:

lithrel: _starts to write a 50 pages ADR to explain how wrong this is 🔨_

PR #1130, opened 2020-06-25
PLANET-4987 Exclude unwanted attachments from EP

Inwerpsel:
While I was debugging why my previous suggested solution didn't work, I stumbled on this other filter that allows us to exclude unwanted attachments from the list of ids that ElasticPress will sync, before it starts looping through these. The benefit is that way it will also show correct numbers in the UI and can prevent issues caused by the number not corresponding to the actual posts being synced.

Unfortunately this was not solvable by adding more args, as WP query args don't support the kind of complex or condition we need (either not post_type attachment or having one of the wanted mime types), so instead I had to first fetch all unwanted ids. It needs to be done with raw SQL, as regular query with args was not returning attachments with no post_parent.

The additional query shouldn't be an issue as this function is only called in non-performance sensitive contexts (CLI and dashboard sync).

Still todo: write the `IN` query with parameters to make it support multiple wanted doc types.

### DIFF
--- a/exporter-helper.php
+++ b/exporter-helper.php
@@ -6,22 +6,6 @@
  */
 
 /**
- * Generate a bunch of placeholders for use in an IN query.
- * Such a shame WordPress doesn't offer a decent way to do bind IN statement params and we have to do this ourselves.
- *
- * @param array $items The items to generate placeholders for.
- * @param int   $start_index The start index to use for creating the placeholders.
- * @return string The generated placeholders string.
- */
-function generate_list_placeholders( array $items, int $start_index ): string {
-	$placeholders = [];
-	foreach ( range( $start_index, count( $items ) + $start_index - 1 ) as $i ) {
-		$placeholders[] = "%$i\$d";
-	}
-	return implode( ',', $placeholders );
-}
-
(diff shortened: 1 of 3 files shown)


pablocubico:
So what about the ugly UI bug in the settings page? We could use something like:

```
if ( 'toplevel_page_elasticpress' == $hook ) { ... }
```

...inside `enqueue_admin_assets` to add a script for that specific page and fix the UI temporarily, but maybe they fixed it already in the plugin, haven't checked what the latest version is. 🤔

Inwerpsel:
> So what about the ugly UI bug in the settings page? We could use something like:
> 
> ```
> if ( 'toplevel_page_elasticpress' == $hook ) { ... }
> ```
> 
> ...inside `enqueue_admin_assets` to add a script for that specific page and fix the UI temporarily, but maybe they fixed it already in the plugin, haven't checked what the latest version is.

@pablocubico Yeah let's check first if they fixed it already. There's already a ticket to upgrade the plugin from 3.2 to 3.4, I can't image they didn't fix it in the meanwhile, so we can look at it in the context of that ticket.

PR #1135, opened 2020-07-01
PLANET-5189 ImageArchive

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-5189
Deployed on https://k8s.p4.greenpeace.org/pieterdev/wp-admin/admin.php?page=gpi-image-picker 

A first iteration of a rewrite of the Orange Logic Media API integration (specifically Greenpeace's instance of it). This also includes a small class with some functions to handle toggling features, as we'll intially merge the code but disable it as a feature on all live instances. When the code is stable enough we can enable this feature on some NROs to test it before enabling it for everyone.

ImageArchive (name subject to change):
* Frontend javascript and twig templates are entirely replaced by a React component.
* Add a `ImagePicker` component which uses render props. That way this component is only concerned with picker behavior and doesn't care about how the list items and sidebar look like.
* Some code of the wp-ajax endpoints was preserved, but it uses the WordPress REST API instead.
* The `Image` class resembles the `MediaImage` class, but it also contains some domain logic that acts on the instance (instead of in a separate `MediaHelper` and `MediaImageMapper` class), to provide more terse semantics. Setters are removed, as well as any getter that isn't used (much less public getters required as things happen in instance methods).
* The new picker can be accessed on a dedicated page https://www.planet4.test/wp-admin/admin.php?page=gpi-image-picker . 
* All new functionality should be hidden as long as the feature isn't enabled.

Todo (can be done after this PR as long as feature isn't enabled):
* Write tests.
* Integrate the archive picker with the post editor, so that an image can be directly included in a post without having to manually move it into the WP media library. Should be able to reuse the `ArchivePicker` component, but with different buttons.
* Provide decent scss. I just added a plain css file for now to avoid having to determine the right scss structure already. In this case, since the picker component is intended to be reused with different controls, it might make sense to add this scss to the same folder as the components js files. Similar to how gutenberg components are organized. Maybe we can consider doing this for existing components too(mainly blocks)?
* Evaluate the data that is shown in the sidebar for a selected image. We already fetch a lot of information from the Media API that we don't display yet. There might also be some other interesting fields
* The Media API has a [query language](https://docs.google.com/document/d/1qM0H2rbSuUqA66Yvd4UwYwP6M2T51lWm95r1dEWOPNA/edit#heading=h.3rdcrjn) that can do a lot more than the things we currently use. We can use it to implement some advanced filters which can really help editors find the content they need. Main filtering by date, orientation and keywords (i.e. tags) look interesting.

Feature management:
* This is just a class with some static functions and a constructor that adds hooks (the instance is thrown away, but that's how other things already work).
* "Register" a feature by providing a metabox option for it, then `Features::active( $name )` will check this option.

### DIFF
--- a/assets/src/js/Components/ArchivePicker.js
+++ b/assets/src/js/Components/ArchivePicker.js
@@ -1,8 +1,11 @@
 import { Component, Fragment } from '@wordpress/element';
+import { TextControl, QueryControls } from '@wordpress/components';
 import { ImagePicker, toSrcSet } from './ImagePicker';
 import { SingleSidebar } from './archivePicker/SingleSidebar';
 import { MultiSidebar } from './archivePicker/MultiSidebar';
 import classNames from 'classnames';
+import debounce from 'debounce';
+// import debounce from 'debounce-promise';
 
 const { apiFetch } = wp;
 const { addQueryArgs } = wp.url;
@@ -15,26 +18,19 @@ class ArchivePicker extends Component {
       loading: true,
       error: false,
       images: [],
-      currentPage: 1,
+      currentPage: 0,
       filters: {},
       searchText: null,
     };
     this.loadNextPage = this.loadNextPage.bind( this );
     this.updateFromUploadedResponse = this.updateFromUploadedResponse.bind( this );
     this.renderSidebar = this.renderSidebar.bind( this );
+    this.search = this.search.bind( this );
+    this.includeInWp = this.includeInWp.bind( this );
   }
 
   async componentDidMount() {
-    try {
-      const images = await this.fetchImages( {} );
-      this.setState( {
-        loading: false,
-        images: images,
-      } );
-    } catch ( error ) {
-      console.log( error );
-      this.setState( { error } );
-    }
+    await this.loadNextPage( {} );
   }
 
   async fetchImages( args ) {
@@ -44,16 +40,18 @@ class ArchivePicker extends Component {
   }
 
   async loadNextPage() {
-    const nextPage = this.state.currentPage + 1;
+    const pageIndex = this.state.currentPage + 1;
 
     try {
       this.setState( { loading: true } );
       const nextImages = await this.fetchImages( {
-        page: nextPage
+        page: pageIndex,
+        search_text: this.state.searchText,
       } );
+      const withPageLabel = nextImages.map(image => ({...image, title: `${pageIndex} -- ${image.title}`}));
       this.setState( {
-        currentPage: nextPage,
-        images: [ ...this.state.images, ...nextImages ]
+        currentPage: pageIndex,
+        images: [ ...this.state.images, ...withPageLabel ]
       } );
     } catch ( e ) {
       this.setState( { error: true } );
@@ -95,6 +93,11 @@ class ArchivePicker extends Component {
     } );
   }
 
+  async search( text ) {
+    this.setState( { images: [], searchText: text, currentPage: 0 } );
+    await this.loadNextPage();
+  }
+
   renderList( parent ) {
     const { props, isSelected, toggleMultiSelection, toggleSingleSelection } = parent;
 
@@ -115,11 +118,11 @@ class ArchivePicker extends Component {
         data-wordpress-id={ wordpress_id }
         className={ classNames( { 'picker-selected': isSelected( image ) } ) }>
         <img
-          srcSet={ toSrcSet( sizes ) }
-          title={ title }
+          srcSet={ toSrcSet( sizes, { maxWidth: 900 } ) }
+          title={ `${title}` }
           alt={ alt }
-          width={ 200 }
-          height={ 200 * ( original.height / original.width ) }
+          width={ 200 * ( original.width / original.height ) }
+          height={ 200  }
           onClick={ ( event ) =>
             event.ctrlKey
               ? toggleMultiSelection( image )
@@ -137,6 +140,8 @@ class ArchivePicker extends Component {
       return <SingleSidebar
         parent={ parent }
         includeInWp={ this.includeInWp }
+        processingError={ this.state.processingError }
+        processingImages={ this.state.processingImages }
       />;
     }
 
@@ -156,13 +161,17 @@ class ArchivePicker extends Component {
     } = this.state;
 
     return <Fragment>
+      <TextControl
+        onChange={ debounce( this.search, 500 ) }
+        disabled={ this.state.loading }
+      />
       { loading && (
         <div className={ 'archive-picker-loading' }> loading...</div>
       ) }
       { !!error && (
         <div>
           <h3>API error:</h3>
-          <p> error.message </p>
+          <p> { error.message } </p>
         </div>
       ) }
       <ImagePicker
--- a/assets/src/js/Components/ArchivePicker.js
+++ b/assets/src/js/Components/ArchivePicker.js
@@ -1,0 +1,119 @@
+import { Component, Fragment } from '@wordpress/element';
+import { ImagePicker } from './ImagePicker';
+import { SingleSidebar } from './archivePicker/SingleSidebar';
+
+const { apiFetch } = wp;
+const { addQueryArgs } = wp.url;
+
+class ArchivePicker extends Component {
+
+  constructor( props ) {
+    super( props );
+    this.state = {
+      loading: true,
+      error: false,
+      images: [],
+      selectedImage: null,
+      currentPage: 1,
+      filters: {},
+      searchText: null,
+    };
+    this.loadNextPage = this.loadNextPage.bind( this );
+    this.updateFromUploadedResponse = this.updateFromUploadedResponse.bind( this );
+    this.renderSidebar = this.renderSidebar.bind( this );
+  }
+
+  async componentDidMount() {
+    try {
+      const images = await this.fetchImages( {} );
+      this.setState( {
+        loading: false,
+        images: images,
+      } );
+    } catch (error) {
+      console.log( error );
+      this.setState( { error } );
+    }
+  }
+
+  async fetchImages( args ) {
+    return apiFetch( {
+      path: addQueryArgs( '/planet4/v1/image-archive/fetch', args ),
+    } );
+  }
+
+  async loadNextPage() {
+    const nextPage = this.state.currentPage + 1;
+
+    try {
+      this.setState( { loading: true } );
+      const nextImages = await this.fetchImages( {
+        page: nextPage
+      } );
+      this.setState( {
+        currentPage: nextPage,
+        images: [...this.state.images, ...nextImages]
+      } );
+    } catch (e) {
+      this.setState( {
+        error: true,
+      } );
+    } finally {
+      this.setState( {
+        loading: false,
+      })
+    }
+  }
+
+  updateFromUploadedResponse( updatedImages ) {
+    const newImages = this.state.images.map( stateImage => {
+      const updated = updatedImages.find( updatedImage => updatedImage.id === stateImage.id );
+      if ( updated ) {
+        return updated;
+      }
+      return stateImage;
+    } );
+    this.setState( {
+      images: newImages,
+    } );
+  }
+
+  renderSidebar( { selectedImages } ) {
+    if ( selectedImages.length === 1 ) {
+      return <SingleSidebar
+        image={ selectedImages[ 0 ] }
+        // todo: clean up state management.
+        onIncludeInWP={ this.updateFromUploadedResponse }
+      />;
+    }
+  }
+
+  render() {
+
+    return <Fragment>
+      { this.state.loading && (
+        <div className={ 'archive-picker-loading' }> loading...</div>
+      ) }
+      { !!this.state.error && (
+        <Fragment>
+          <h3>API error:</h3>
+          <p> error.message </p>
+        </Fragment>
+      ) }
+      { this.state.images.length > 0 && (
+        <ImagePicker
+          images={ this.state.images }
+          renderSidebar={ this.renderSidebar }
+          onNearListBottom={ async () => {
+            if ( !this.state.loading ) {
+              await this.loadNextPage();
+            }
+          } }
+        />
+      ) }
+    </Fragment>;
+  }
+}
+
+export default ArchivePicker;
+
--- a/assets/src/js/Components/ArchivePicker.js
+++ b/assets/src/js/Components/ArchivePicker.js
@@ -14,7 +14,7 @@ class ArchivePicker extends Component {
     super( props );
     this.state = {
       loading: true,
-      error: false,
+      error: null,
       images: [],
       currentPage: 0,
       filters: {},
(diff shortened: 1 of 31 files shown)


assets/src/js/Components/archivePicker/archivePickerSidebar.js
```diff
@@ -0,0 +1,22 @@
+import { SingleSidebar } from './SingleSidebar';
+import { MultiSidebar } from './MultiSidebar';
+
+export const archivePickerSidebar = ( archivePicker ) => ( imagePicker ) => {
```
Inwerpsel, line 4: I initially added the `archivePicker` argument as a quick fix after moving this render prop function outside of it, with the intention of replacing it later by passing on only certain props. But actually this way might be more clear about the intention of this function, i.e. it should provide the `imagePicker` sidebar for the `ArchivePicker` component, and the signature is also recognizable.

src/ImageArchive/ApiClient.php
```diff
@@ -0,0 +1,235 @@
...
+		);
+		// Authentication failure.
+		if ( is_wp_error( $response ) ) {
+			$response = $response->get_error_message() . ' ' . $response->get_error_code();
+
+		} elseif ( WP_Http::ACCEPTED !== $response['response']['code'] ) {
+			$response = $response['response']['message'] . ' ' . $response['response']['code'];
+		}
+
+		if ( ! is_array( $response ) || ! ( $response['body'] ) ) {
```
lithrel, line 124: `wp_safe_remote_post` says that it will return an array `containing 'headers', 'body', 'response', 'cookies', and 'filename' elements` but I can't find any guarantee about that in wp code, when returning things from `apply_filter` hooks. So maybe `if ( ! is_array( $response ) || empty( $response['body'] ) ) {` would be safer (to no get a notice) ?

src/ImageArchive/Image.php
```diff
@@ -0,0 +1,240 @@
...
+
+	/**
+	 * @param array $data The data returned by Orange Logic API.
+	 * @param array $images_in_wordpress Array with all images that are already in WP.
+	 *
+	 * @return static The created instance.
+	 */
+	public static function from_api_response( array $data, array $images_in_wordpress ): self {
+		$image               = new self();
+		$image->archive_id   = $data['SystemIdentifier'];
```
lithrel, line 99: any idea of those types ? we could typecast and document the types in property declaration

src/ImageArchive/Image.php
```diff
@@ -0,0 +1,240 @@
...
+			'post_status'    => 'inherit',
+		];
+
+		// Check title has full stop at the end, if not then add it.
+		$alt_text = rtrim( $title, '.' ) . '.';
+
+		$attachment_id = wp_insert_attachment( $attachment, $upload_file['file'], 0, true );
+
+		if ( is_wp_error( $attachment_id ) ) {
+			throw new RuntimeException( __( 'Error while inserting attachment...!', 'planet4-master-theme-backend' ) );
```
lithrel, line 215: Can we add wp error message here ? or is it logged elsewhere ?

src/ImageArchive/Image.php
```diff
@@ -0,0 +1,240 @@
...
+
+		require_once ABSPATH . 'wp-admin/includes/image.php';
+
+		// Generate the metadata for the attachment, and update the database record.
+		$attachment_data = wp_generate_attachment_metadata( $attachment_id, $upload_file['file'] );
+
+		wp_update_attachment_metadata( $attachment_id, $attachment_data );
+
+		// Add credit to alt field.
+		if ( '' !== $this->credit ) {
```
lithrel, line 226: `!empty(trim())` in case api gives us something strange ? (or maybe do this in property assignment)

src/ImageArchive/Rest.php
```diff
@@ -0,0 +1,109 @@
...
+class Rest {
+	/**
+	 * @var string The API path.
+	 */
+	private const REST_NAMESPACE = 'planet4/v1';
+
+	/**
+	 * Add some REST API endpoints if feature is active.
+	 */
+	public static function endpoints(): void {
```
lithrel, line 32: register_endpoints() ? I feel like endpoints() would give me endpoints

src/ImageArchive/UiIntegration.php
```diff
@@ -0,0 +1,111 @@
...
+	 */
+	public static function media_library_post_upload_ui() {
+		global $pagenow;
+		$classes = 'button media-button button-primary button-large add_media switchtoml';
+
+		// Add the insert media class only when not in the editor, i.e. when on the "Media > Add New" page.
+		if ( ( 'post.php' !== $pagenow ) && ( ! in_array( get_post_type(), [ 'post', 'page', 'campaign' ], true ) ) ) {
+			$classes .= ' insert-media';
+		}
+		print '<button id="db-upload-btn" class="' . $classes . '">' . esc_html__( //phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
lithrel, line 51: In `Features::admin_page_display` we're using closing/opening php tags to display something, here it is print, and I'm sure I've used echo somewhere else; I don't know what we should use (clearly not found of closing/opening php tags), but we could opt for only one way to do this

src/ImageArchive/UiIntegration.php
```diff
@@ -0,0 +1,111 @@
...
+		) . '</button>'; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+	}
+
+	/**
+	 * @param string[] $tabs Existing tabs passed by the filter.
+	 *
+	 * @return string[] Same tabs with image archive tab added to them.
+	 */
+	public static function image_archive_tab( $tabs ): array {
+		$tabs['image_archive'] = 'GPI Image Archive';
```
lithrel, line 63: will this one be translatable ?

src/Features.php
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	/**
+	 * Check whether a feature is active.
+	 *
+	 * @param string $name The name of the feature we're checking.
+	 *
+	 * @return bool Whether the feature is active.
+	 */
+	public static function active( string $name ): bool {
```
lithrel, line 48: `is_active()` might be more clear ? (might also be my french, active means activate 😬 )

src/Loader.php
```diff
@@ -192,11 +196,66 @@ public function do_not_delete_autosave( $delete = null, $post = null, $force_del
...
+	 * @param string $path The path of the file.
+	 *
+	 * @throws RuntimeException If the file doesn't exist, or filectime fails in some other way.
+	 * @return int Timestamp of last file change.
+	 */
+	private static function get_timestamp( string $path ): int {
+		$ctime = filectime( $path );
+
+		if ( ! $ctime ) {
+			throw new RuntimeException( "Tried to get file change time of {$path} but failed to." );
```
lithrel, line 215: You could have included the runtime exception in `theme_file_ver` directly, what made you extract this part ? (I feel like the function name is more explicit about what it does, but in the context of the class `get_file_timestamp()` might be better, and in the context of `wp_enqueue_style` `theme_file_ver()` seemed more versatile if a change about versioning process was needed)

src/Exception/RemoteCallFailed.php
```diff
@@ -0,0 +1,17 @@
+<?php
+/**
+ * Exception RemoteCallFailed.
+ *
+ * @package P4\MasterTheme
```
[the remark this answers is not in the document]
Inwerpsel: Yeah I just added it for now so the linter doesn't trip over it :stuck_out_tongue: . I'm not sure if something is still using these `@package` annotations, but they're definitely redundant now we have namespaces, so we can probably drop the rule that it should be there. Similarly we could also drop the required file comment, at least for classes, as they also require a comment on the class itself.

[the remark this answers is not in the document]
Inwerpsel: Removed them here: https://github.com/greenpeace/planet4-master-theme/pull/1138/files

src/ImageArchive/ApiClient.php
Inwerpsel: Good point, in fact `$response` could be anything as it's passed through a filter before it's returned https://github.com/WordPress/WordPress/blob/master/wp-includes/class-http.php#L446

comzeradd:
Since we have `PLANET-5222`, I'd place the Feature toggle inside the Planet4 settings to avoid a new menu entry for now.

src/ImageArchive/UiIntegration.php
```diff
@@ -0,0 +1,111 @@
...
+	public static function media_library_post_upload_ui() {
+		global $pagenow;
+		$classes = 'button media-button button-primary button-large add_media switchtoml';
+
+		// Add the insert media class only when not in the editor, i.e. when on the "Media > Add New" page.
+		if ( ( 'post.php' !== $pagenow ) && ( ! in_array( get_post_type(), [ 'post', 'page', 'campaign' ], true ) ) ) {
+			$classes .= ' insert-media';
+		}
+		print '<button id="db-upload-btn" class="' . $classes . '">' . esc_html__( //phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+			'Upload From GPI Media Library',
```
comzeradd, line 52: That should be renamed, right?

assets/src/js/Components/ArchivePicker.js
```diff
@@ -0,0 +1,151 @@
...
+        onChange={ event => this.setState( { enteredSearch: event.target.value } ) }
+      >
+        <input
+          type={ 'text' }
+          disabled={ this.state.loading }
+        />
+        <input type={ 'submit' } value={ __( 'Search' ) }/>
+      </form>
+      { loading && (
+        <div className={ 'archive-picker-loading' }> loading...</div>
```
sagarsdeshmukh, line 128: need string translation

src/ImageArchive/Image.php
Inwerpsel: The API documentation doesn't explicitly state the types, but we can figure them out and document them here :+1: maybe something we can do in a follow up PR?

assets/src/js/Components/ArchivePicker.js
```diff
@@ -0,0 +1,151 @@
+const __ = ( str ) => wp.i18n.__( str, 'planet4-master-theme-backend' );
```
sagarsdeshmukh, line 1: The Loco translate plugin string extractor not able to evaluate variables/constants. 
We need to use - `__('string to translate', 'text-domain')` format of translation function.
Even we are not able to use textdomain variable here.

src/ImageArchive/UiIntegration.php
Inwerpsel: Yeah just preserved that from the original code. I'll change it to `echo`, as the closing/opening php tags in the middle of the file indeed looks weird :+1:

src/ImageArchive/UiIntegration.php
Inwerpsel: Good catch, I did this similarly to how it was done in the media library plugin, but indeed it looks like the translation needs to be performed here :+1:

src/Loader.php
Inwerpsel: > what made you extract this part ? 

I included the exception in that function just to avoid duplicating it, it's a private function that is used by 3 functions, for all of which it makes sense to throw the same exception.

> get_file_timestamp() might be better

The function argument already makes it clear that this acts on a file. In a lot of cases it can really help reduce redundancy if you regard the parameter as part of the function name, as it will be present both in the declaration and the usages (assuming the variable you pass to it is accurately named). Especially if the argument is a class this helps reducing repetition e.g. `get_file_timestamp(File $file)` vs `get_timestamp(File $file)`.

> and in the context of wp_enqueue_style theme_file_ver() seemed more versatile if a change about versioning process was needed

Not sure if I understand this correctly, can you elaborate?

src/ImageArchive/UiIntegration.php
Inwerpsel: Indeed, but maybe we can remove this tab.

I intially started setting it up in a tab of the media modal like in the original media library plugin. But since the frontend is a self-contained app, we can easily include it on a separate page, which I already did as it's easier to develop in.

assets/src/js/Components/ArchivePicker.js
Inwerpsel: Right, good catch, didn't think of that :+1: Too bad, I thought I found a way to only specify the domain once per file, but probably would be hard to make this work with Loco, so I'll remove it.

src/Loader.php
lithrel: > The function argument already makes it clear that this acts on a file.  

Ah that's interesting, usually I'm reasoning more in term of methods for which the main subject is the class itself (like `File::get_timestamp()` for example, and the rest of the arguments are more like options), my guess is that you're displaying argument names in phpstorm ? (I don't at the moment, but I understand your point in that context)

> Not sure if I understand this correctly, can you elaborate?

I just meant that wp expects this parameter to be a file version, so naming the function for what it gives made sens to me (whatever it returns that will act as is, timestamp, version, anything else) as we wouldn't have to change its name if we want it to return a version number for example, but again, if you display arguments names I see your point.

I was saying this somewhere else, once we have PSR-4 everywhere we could extract those utilities in specific classes and reuse them in gutenberg-blocks too.

src/ImageArchive/Image.php
lithrel: follow-up PR sounds good to me :)

src/Exception/RemoteCallFailed.php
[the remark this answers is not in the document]
Inwerpsel: thanks for the review, merged it now :+1:

src/Loader.php
Inwerpsel: > so naming the function for what it gives made sense to me

I did name the function for what it gives, as it gives the timestamp :D But you're right that without having the word "version" in there somewhere it's not that clear. It also doesn't help that WP abbreviates it to `$ver`... So I instead assigned  the return value to a variable.

src/ImageArchive/UiIntegration.php
Inwerpsel: I removed the tab as we'll probably want to integrate this in another way so that's in not in the iframe modal.

assets/src/js/Components/ArchivePicker.js
Inwerpsel: Maybe for now we should wait with the string translation until we have the final text? Since the feature is not intended to be used live yet, we might make changes to these texts before they are ever used. So seems to make more sense to only add strings to the translations when we know they will be used, to avoid unnecessary translation work.

assets/src/js/Components/ImagePicker.js
```diff
@@ -0,0 +1,95 @@
+import { Component } from '@wordpress/element';
+import classNames from 'classnames';
+
+export const getSizesUnder = ( sizes, maxWidth) => sizes.filter( size => {
+  console.log( sizes );
```
lithrel, line 5: Keeping this one ?

Inwerpsel: Good catch removed it :ok_hand:

PR #1139, opened 2020-07-09
PLANET-5142 Add Cloudflare's Image Optimization url twig filter(cf_img_url)

sagarsdeshmukh:
Ref.
[JIRA 5142](https://jira.greenpeace.org/browse/PLANET-5142)
https://developers.cloudflare.com/images/responsive

- Add a twig filter which returns cloudflare ready(cloudflare optimized) image url
- Add P4 setting to toggle the cloudflare image url functionality
- Option added in P4 setting to try different `options` values from [cloudflare](https://developers.cloudflare.com/images/about) (`https://zone/cdn-cgi/image/options/source-image`) [This field is intended for testing purpose only, we will remove it once finish with testing]

Related PR: [#321](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/321)

### DIFF
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -1170,54 +1170,72 @@ class MasterSite extends TimberSite {
 		$twig->addFilter(
 			new Twig_SimpleFilter(
 				'cf_img_url',
-				function ( $source, $srcset = '' ) use ( $cf_img_optimization, $cf_options_txt ) {
-					if ( ! $cf_img_optimization ) {
+				$cf_img_optimization
+					? function ( string $source, string $src_set = '' ) use ( $cf_options_txt ) {
+						return $this->img_to_cloudflare( $source, $src_set, $cf_options_txt );
+					}
+					: function ( string $source ) {
 						return $source;
 					}
-
-					// Cloudflare url format - https://zone/cdn-cgi/image/options/source-image .
-					// eg. <img src="/cdn-cgi/image/width=80,quality=75/uploads/avatar1.jpg">.
-					// More info. https://developers.cloudflare.com/images/about .
-					$zone      = '';
-					$prefix    = '/cdn-cgi/image/';
-					$options[] = 'format=auto';
-
-					if ( $cf_options_txt ) {
-						$options[] = esc_attr( $cf_options_txt );
-					}
-
-					// Case 2: Prepare a srcset cloudflare images url.
-					if ( empty( $srcset ) ) {
-						$srcset_array = explode( ',', $source );
-						$cf_srcset    = $source;
-						$options[]    = 'fit=contain';
-
-						foreach ( $srcset_array as $img_data ) {
-							list( $img_url, $img_width ) = explode( ' ', trim( $img_data ) );
-
-							// Update image width in options array.
-							$srcset_options = array_merge( $options, [ 'width=' . (int) $img_width ] );
-							$cf_img_url     = $zone . $prefix . implode( ',', $srcset_options ) . '/' . $img_url;
-
-							$cf_srcset = str_replace( $img_url, $cf_img_url, $cf_srcset );
-						}
-						return $cf_srcset;
-					}
-
-
-					// Case 1: Prepare a src cloudflare image url.
-					preg_match( '/' . preg_quote( $source, '/' ) . '\s(\d+)/', $srcset, $img_width );
-
-					if ( isset( $img_width[1] ) && $img_width[1] ) {
-						$options[] = 'width=' . $img_width[1];
-						return $zone . $prefix . implode( ',', $options ) . '/' . $source;
-					}
-
-					return $source;
-				}
 			)
 		);
 
 		return $twig;
+	}
+
+	/**
+	 * Converts Image urls in src/srcset content to cloudflare optimized image url.
+	 *
+	 * @see More info. https://developers.cloudflare.com/images/about
+	 * @see eg. <img src="/cdn-cgi/image/width=80,quality=75/uploads/avatar1.jpg">.
+	 *
+	 * @param string $source The raw img src/srcset to be filtered.
+	 * @param string $srcset The raw img srcset.
+	 * @param string $cf_options_txt The options value set from P4 settings.
+	 * @return string Converted img src/srcset content.
+	 */
+	public function img_to_cloudflare(
+		string $source,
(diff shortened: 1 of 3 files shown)


src/MasterSite.php
```diff
@@ -1153,4 +1154,69 @@ public function dismiss_dashboard_notice(): void {
...
+		$options             = get_option( 'planet4_options' );
+		$cf_img_optimization = $options['cloudflare_img_opt'] ?? false;
+		$cf_options_txt      = $options['cloudflare_options_txt'] ?? '';
+
+		// Add Twig filter(cf_img_url) which convert image url/s into cloudflare optimized image url/s.
+		$twig->addFilter(
+			new Twig_SimpleFilter(
+				'cf_img_url',
+				function ( $source, $srcset = '' ) use ( $cf_img_optimization, $cf_options_txt ) {
+					if ( ! $cf_img_optimization ) {
```
lithrel, line 1174: could we move this outside of this call, maybe before the `$twig->addFilter` ? It seems like if no optimization is selected in global options, we don't have to add this filter at all

src/MasterSite.php
```diff
@@ -1153,4 +1154,69 @@ public function dismiss_dashboard_notice(): void {
...
+					// More info. https://developers.cloudflare.com/images/about .
+					$zone      = '';
+					$prefix    = '/cdn-cgi/image/';
+					$options[] = 'format=auto';
+
+					if ( $cf_options_txt ) {
+						$options[] = $cf_options_txt;
+					}
+
+					if ( '' !== $srcset ) {
```
lithrel, line 1189: if you invert the condition you can drop an `else` I think
```php
	// Case 2: Prepare a srcset cloudflare images url.
	if ( empty( $srcset ) ) {
		$srcset_array = explode( ',', $source );
		$cf_srcset    = $source;
		$options[]    = 'fit=contain';

		foreach ( $srcset_array as $img_data ) {
			list( $img_url, $img_width ) = explode( ' ', trim( $img_data ) );

			// Update image width in options array.
			$srcset_options = array_merge( $options, [ 'width=' . (int) $img_width ] );
			$cf_img_url     = $zone . $prefix . implode( ',', $srcset_options ) . '/' . $img_url;

			$cf_srcset = str_replace( $img_url, $cf_img_url, $cf_srcset );
		}
		return $cf_srcset;
	}


	// Case 1: Prepare a src cloudflare image url.
	preg_match( '/' . preg_quote( $source, '/' ) . '\s(\d+)/', $srcset, $img_width );

	if ( isset( $img_width[1] ) && $img_width[1] ) {
		$options[] = 'width=' . $img_width[1];
		return $zone . $prefix . implode( ',', $options ) . '/' . $source;
	}

	return $source;
```

src/MasterSite.php
sagarsdeshmukh: if we skip the `$twig->addFilter` part, the frontend where we use `cf_img_url` would be invalid and through's error.

sagarsdeshmukh:
> I'd be tempted to add an allowList for the cloudflare options, as it can be defined by users and ends up on the frontend; maybe just checking the keys used (explode on commas, then on equals), I guess the url is escaped by twig so there's not too much risk of injection of another url or xss. Might be a better idea to use an allowList in the settings, to display an error message when people save a wrong option/format

The idea of adding a `Cloudflare Image Optimization Options` field in setting is temporary/for testing purposes only, we will remove this field after we are done with testing the cloudflare image optimization functionality.
I also checked in CMB2 field for adding a validation but it will be tricky and time consuming.

I added a example of options text string in help info.

lithrel:
This function reminded me of the lazyload I made in another one, so I though it was easily testable through unit test; I made a PoC there https://github.com/greenpeace/planet4-master-theme/compare/planet-5142...planet-5142-testable. 
As you can see I extracted the anonymous function so that it takes only the necessary parameters; by doing this, testing it is just a matter of [writing various parameters](https://github.com/greenpeace/planet4-master-theme/compare/planet-5142...planet-5142-testable#diff-439b0effd602fc73a5df26e5ba1d84cbR67) and [the expected result](https://github.com/greenpeace/planet4-master-theme/compare/planet-5142...planet-5142-testable#diff-439b0effd602fc73a5df26e5ba1d84cbR68), and it makes it more easy to see what the function is actually doing.  
My view is that your tests can even function as documentation for various options, if we have a bug we just have to add it to the test list, and if I have a doubt about something in the code it's very easy to test.

sagarsdeshmukh:
> This function reminded me of the lazyload I made in another one, so I though it was easily testable through unit test; I made a PoC there [planet-5142...planet-5142-testable](https://github.com/greenpeace/planet4-master-theme/compare/planet-5142...planet-5142-testable).
> As you can see I extracted the anonymous function so that it takes only the necessary parameters; by doing this, testing it is just a matter of [writing various parameters](https://github.com/greenpeace/planet4-master-theme/compare/planet-5142...planet-5142-testable#diff-439b0effd602fc73a5df26e5ba1d84cbR67) and [the expected result](https://github.com/greenpeace/planet4-master-theme/compare/planet-5142...planet-5142-testable#diff-439b0effd602fc73a5df26e5ba1d84cbR68), and it makes it more easy to see what the function is actually doing.
> My view is that your tests can even function as documentation for various options, if we have a bug we just have to add it to the test list, and if I have a doubt about something in the code it's very easy to test.

Thanks @lithrel for the unit test branch. It's absolutely make sense to add it 💯 . I added few more variations of inputs in test code. please have a look.

sagarsdeshmukh:
> Excellent 👍 🤸
> Do you know if it's possible to have an image url without a width and srcset ? I wonder what would be displayed if width gets to 0 there https://github.com/greenpeace/planet4-master-theme/pull/1139/files#diff-5bc221d166fcb4619227bb8679dc1aefR1220

I am not sure about this usecase, but I here is a example what cloudflare return in case of width=0
https://www.greenpeace.org/cdn-cgi/image/format=auto,fit=contain,width=0/https:/storage.googleapis.com/planet4-cidev-stateless/2018/10/6aa365fe-oceans-gp0stom6c-1024x683.jpg

lithrel:
That's tiny 😁 If you think it can ever happen, maybe add a condition to use `width=` only if `$img_width` is not null. If not, fine for me 👍

PR #1144, opened 2020-07-14
PLANET-5272 Add hreflang metadata tags

comzeradd:


src/MasterSite.php
```diff
@@ -279,6 +279,37 @@ public function set_featured_image( $post_id, $post, $update ) {
...
+			if ( 0 === $key ) {
+				$metadata .= "<link rel='alternate' hreflang='x-default' href='" . $letter[ $key ]['url'] . "'>\n";
+				continue;
+			}
+			foreach ( $letter as $country ) {
+				$lang = $country['lang'];
+				foreach ( $lang as $item ) {
+					if ( isset( $item['locale'] ) ) {
+						foreach ( $item['locale'] as $code ) {
+							$metadata .= "<link rel='alternate' hreflang='" . $code . "' href='" . $item['url'] . "'>\n";
```
Inwerpsel, line 303: Since we're constructing html here, maybe it's easier to manage if we create a small twig template that loops over an array of data given to it? Then this function would be simpler as it would only need to filter the ones with a locale.

src/MasterSite.php
```diff
@@ -279,6 +279,37 @@ public function set_featured_image( $post_id, $post, $update ) {
 		}
 	}
 
+	/**
+	 * Generates hreflang metadata from countries.json template.
+	 */
+	public function generate_hreflang_meta() {
```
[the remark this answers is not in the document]
comzeradd: If I turn this to `static` then I can't access `$this` object that I'm using to find the `theme_dir`.

src/MasterSite.php
```diff
@@ -279,6 +279,37 @@ public function set_featured_image( $post_id, $post, $update ) {
...
+		$countries = json_decode( $body, true );
+
+		$metadata = '';
+
+		foreach ( $countries as $key => $letter ) {
+			if ( 0 === $key ) {
+				$metadata .= "<link rel='alternate' hreflang='x-default' href='" . $letter[ $key ]['url'] . "'>\n";
+				continue;
+			}
+			foreach ( $letter as $country ) {
```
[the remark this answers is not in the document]
comzeradd: It is unfortunately :) Because then we would have to refactor the js code on both the country selector and the landing page. Maybe after this is done we could re-think whether it makes sense to render the country selector in the backend.

src/MasterSite.php
```diff
@@ -279,6 +279,37 @@ public function set_featured_image( $post_id, $post, $update ) {
 		}
 	}
 
+	/**
+	 * Generates hreflang metadata from countries.json template.
+	 */
+	public function generate_hreflang_meta() {
+
+		$request   = wp_remote_get( $this->theme_dir . '/templates/countries.json' );
```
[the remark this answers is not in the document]
comzeradd: I added a cache check.

src/MasterSite.php
comzeradd: Moved the html part in the template and just left the array construction here.

src/MasterSite.php
```diff
@@ -275,6 +275,41 @@ public function set_featured_image( $post_id, $post, $update ) {
...
 
+	/**
+	 * Generates hreflang metadata from countries.json template.
+	 */
+	public function generate_hreflang_meta() {
+
+		$countries = wp_cache_get( 'countries' );
+
+		if ( false === $countries ) {
+			$request   = wp_remote_get( $this->theme_dir . '/templates/countries.json' );
```
lithrel, line 286: I'm confused about this, do we have to do an http request for what looks like a local file ? Does `$json = file_get_contents( get_template_directory() .  '/templates/countries.json' )` works ?

comzeradd: Indeed! `file_get_contents` was also my first thought, but `phpcs` was unhappy suggesting `wp_remote_get`. That was because I used `$this->theme_dir`. Using `get_template_directory` instead made `phpcs` happy again.

src/MasterSite.php
```diff
@@ -275,6 +275,40 @@ public function set_featured_image( $post_id, $post, $update ) {
...
+	 * Generates hreflang metadata from countries.json template.
+	 */
+	public static function generate_hreflang_meta() {
+
+		$countries = wp_cache_get( 'countries' );
+
+		if ( false === $countries ) {
+			$body      = file_get_contents( get_template_directory() . '/templates/countries.json' );
+			$countries = json_decode( $body, true );
+			wp_cache_set( 'countries', $countries );
```
lithrel, line 288: To be safesafesafe and not cache a bad value I'd add
```php
$body      = file_get_contents( get_template_directory() . '/templates/countries.json' );
$countries = json_decode( $body, true );
if ( empty( $countries ) ) {
    return null;
}
wp_cache_set( 'countries', $countries );
```

and change method signature to `public static function generate_hreflang_meta(): ?array {`

comzeradd: Thanks! commit amended

templates/countries.json
```diff
@@ -8,14 +8,17 @@
   "A": [
     {
       "name": "Africa",
+      "codes": ["ZA", "CM", "CD", "SN", "KE", "GH"],
```
lithrel, line 11: I think you don't use those anymore ?

comzeradd: We want those for the GeoIP detection to provide the proper button, like we do for other NROs. For some reason these were missing from Africa, so I just squeezed them in now.

PR #1148, opened 2020-07-16
PLANET-5320 Explicitly set SameSite and Secure attributes on cookie

oekeur:
Just explicitly setting the attributes in the setCookie method. :)

### DIFF
--- a/assets/src/js/cookies.js
+++ b/assets/src/js/cookies.js
@@ -3,7 +3,7 @@ export const setupCookies = function($) {
   window.createCookie = function(name, value, days) {
     let date = new Date();
     date.setTime(date.getTime() + (days * 24 * 60 * 60 * 1000));
-    document.cookie = encodeURI(name) + '=' + encodeURI(value) + ';domain=.' + document.domain + ';path=/;' + '; expires=' + date.toGMTString();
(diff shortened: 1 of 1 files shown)


oekeur:
Hi Florent,

That's strange that you don't have https on the local env.. Because LetsEncrypt is included and setup on the local environment.
(it indeed throws some warnings because of the traefik that's in between)

I do agree with you proposed solution for supporting http, but I doubt if it should..

Or do you refer to the minimal implementation of the local env (I haven't used that (yet))

lithrel:
Oh you're right, it runs on `https` too, I've stayed on http because of the message at startup, thanks ! But our acceptance tests [still run on `http`](https://github.com/greenpeace/planet4-base-fork/blob/develop/.env.codeception#L6), so maybe a condition on `document.location.protocol` would be ok, until we run those tests on https ? (or wait until we run those on https ?) I don't think our current tests will fail, but if we write some cookie sensitive tests in the meantime, it might be hard to understand.

comzeradd:
The local dev environments does have `https` configuration, but I think LetsEncrypt stopped supporting `.test` TLD in the meantime so you get a security warning if you open `www.planet4.test` through `https`.

oekeur:
Closing because of inactivity

PR #1151, opened 2020-07-17
PLANET-5221 Use CircleCI checkout code and remove gcloud activation

Inwerpsel:
* In order to run the same pipeline as internal PRs on fork PRs too, I needed to get around some authentication difficulties in fetching that repository with composer. The simplest way seemed to be to use CircleCI's "checkout code" step, which works regardless of whether it's internal or a fork. In https://github.com/greenpeace/planet4-builder/pull/45/files#diff-54216e5f4228fa168f9fa147150a842bR75 some logic was added so that it uses this checked out code instead by adding a composer repository pointing to the folder.
* Remove the `activate-gcloud-account.sh` script as it depends on secrets (not available for forks), and is not used anyway.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -114,7 +114,7 @@ jobs:
             else
               MASTER_THEME_BRANCH="${CIRCLE_BRANCH}"
             fi
-            MASTER_THEME_BRANCH=dev-${MASTER_THEME_BRANCH} \
+            MASTER_THEME_BRANCH=dev-${MASTER_THEME_BRANCH}#${CIRCLE_SHA1} \
             MERGE_SOURCE=git@github.com:greenpeace/planet4-base-fork.git \
             MERGE_REF=develop \
             FORK_USER=$CIRCLE_PR_USERNAME \
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,7 +81,7 @@ jobs:
 
   acceptance-tests:
     docker:
-      - image: greenpeaceinternational/p4-builder:latest
+      - image: greenpeaceinternational/p4-builder:planet-5221
     working_directory:  /home/circleci/
     environment:
       APP_HOSTNAME: www.planet4.test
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -99,7 +99,6 @@ jobs:
       - run:
           name: Build - Configure
           command: |
-            activate-gcloud-account.sh
             mkdir -p /tmp/workspace/var
             mkdir -p /tmp/workspace/src
             echo "${CIRCLE_BUILD_NUM}" > /tmp/workspace/var/circle-build-num
```
comzeradd, line 104: Probably this will also fail. I think `CIRCLE_BUILD_NUM` is considered a environmental variable and also not shared with forks.

Inwerpsel: [Looks like it's working](https://app.circleci.com/pipelines/github/greenpeace/planet4-master-theme/1480/workflows/05ebeeb9-6a40-45c2-81b7-8cbc253c44f5/jobs/8848/steps), I think the built in variables are always available. https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables.

Inwerpsel: But running into another issue now. `CIRCLE_BRANCH` has the value `pull/1152`, so that gets passed to the composer file and it tries to load it as a branch, which fails. So we would need to pass in the forking repository, so that we can add it to the composer file, like here https://stackoverflow.com/questions/13498519/how-to-require-a-fork-with-composer.

.circleci/config.yml
```diff
@@ -108,7 +109,15 @@ jobs:
           working_directory: /home/circleci
           command: |
             echo "Master theme branch is ${CIRCLE_BRANCH}"
-            MASTER_THEME_BRANCH=dev-${CIRCLE_BRANCH} MERGE_SOURCE=git@github.com:greenpeace/planet4-base-fork.git MERGE_REF=develop make
+            if [ -n "$CIRCLE_PR_NUMBER" ]; then
+              BRANCH="$(./get-fork-pr-branch.sh)"
+            else
+              BRANCH="${CIRCLE_BRANCH}"
+            fi
+            MASTER_THEME_BRANCH=dev-${BRANCH}#${CIRCLE_SHA1} \
```
Inwerpsel, line 117: I added the `$CIRCLE_SHA1` to make the pipeline use the exact commit for which it started, regardless of what the head of the branch is. I don't remember if I it was actually required to make it work, but I guess it won't hurt?

.circleci/config.yml
```diff
@@ -108,7 +109,15 @@ jobs:
...
+            MERGE_SOURCE=git@github.com:greenpeace/planet4-base-fork.git \
+            MERGE_REF=develop \
+            make ci
```
comzeradd, line 120: Is it intentional that this goes from `make` to `make ci`?

Inwerpsel: Yes, it's needed to run the tests without secrets. `make` has the `push` step at the end (https://github.com/greenpeace/planet4-builder/blob/846861ee8d3dfb2a984997082d5677e73a10b395/src/Makefile#L162-L162), whereas `make ci` doesn't (https://github.com/greenpeace/planet4-builder/blob/846861ee8d3dfb2a984997082d5677e73a10b395/src/Makefile#L166-L166)

Inwerpsel: In the blocks plugin we actually already use `make ci` in the same place https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/c9b268862b8bb2a62b68893751e5721353fc785c/.circleci/config.yml#L121-L121

comzeradd (approved):
:+1:

Inwerpsel:
Started anew here https://github.com/greenpeace/planet4-master-theme/pull/1208

PR #1158, opened 2020-07-28
PLANET-5296 Flush TAB posts cache if act page status changes

sagarsdeshmukh:
Ref. [JIRA #5296](https://jira.greenpeace.org/browse/PLANET-5296)

**Issue :**
- The TAB is no longer visible only when the viewer is logged in to P4.
- When not logged in to P4 and viewing the Post incognito the TAB is still present.
- It isn't until the cache is purged that the TAB disappears.

**Fix :**
- On change of act page status purge the cache of related posts which uses act page in TAB block

**Testing :**
- Add a [new TA page](https://k8s.p4.greenpeace.org/test-venus/wp-admin/post-new.php?post_type=page)
- Add/edit post and select newly added TA page from "Take Action Page Selector" dropdown at bottom of post or add a "Take Action Boxout" block from P4 blocks with new TA page
- Check the new post on frontend, the TAB is visible as intended
- Now change the status of TA page to draft/pending review
- Check the post in private window or different browser, the TAB should no longer visible on post

e.g. https://k8s.p4.greenpeace.org/test-venus/story/28509/fukushima-and-the-2020-olympics/

Related PR [#336](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/336)

### DIFF
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -325,7 +325,59 @@ class MasterSite extends TimberSite {
 		if ( defined( 'DOING_AUTOSAVE' ) && DOING_AUTOSAVE ) {
 			return;
 		}
+
+		$this->clean_boxout_posts_cache( $post_id, $post_after, $post_before );
+
 		clean_post_cache( $post_id );
+	}
+
+	/**
+	 * Flush Take Action Boxout(TAB) posts cache, if the TAB act page status changes.
+	 *
+	 * @param int     $post_id The ID of the current Post.
+	 * @param WP_Post $post_after The current Post.
+	 * @param WP_Post $post_before Whether this is an existing post being updated or not.
+	 */
+	private function clean_boxout_posts_cache( $post_id, $post_after, $post_before ): void {
+		$parent_act_id = (int) planet4_get_option( 'act_page' );
+		if ( 'page' !== $post_after->post_type || $parent_act_id !== $post_after->post_parent ) {
+			return;
+		}
+
+		// Flush cache only when a page status changes from publish to any non-public status & vice versa.
+		if (
+			( $post_before->post_status === $post_after->post_status ) ||
+			( 'publish' !== $post_before->post_status && 'publish' !== $post_after->post_status )
+		) {
+			return;
+		}
+
+		global $wpdb, $nginx_purger;
+
+		// Search for those posts, who use TAB($post_id) from "Take Action Page Selector" dropdown.
+		// phpcs:disable
+		$sql          = 'SELECT post_id FROM %1$s WHERE meta_key = \'p4_take_action_page\' AND meta_value = %2$d';
+		$prepared_sql = $wpdb->prepare( $sql, $wpdb->postmeta, $post_id );
+		$boxout_posts = $wpdb->get_col( $prepared_sql );
+		// phpcs:enable
+
+		// Search for those posts, who use TAB($post_id) as a block inside block editor.
+		$take_action_boxout_block = '%<!-- wp:planet4-blocks/take-action-boxout {"take_action_page":' . $post_id . '} /-->%';
+		// phpcs:disable
+		$sql          = 'SELECT ID FROM %1$s WHERE post_type = \'post\' AND post_status = \'publish\' AND post_content LIKE \'%2$s\'';
+		$prepared_sql = $wpdb->prepare( $sql, $wpdb->posts, $take_action_boxout_block );
+		$result       = $wpdb->get_col( $prepared_sql );
+		// phpcs:enable
+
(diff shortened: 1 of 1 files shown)


src/MasterSite.php
```diff
@@ -325,6 +325,52 @@ public function clean_post_cache( $post_id, $post_after, $post_before ) {
 		if ( defined( 'DOING_AUTOSAVE' ) && DOING_AUTOSAVE ) {
 			return;
 		}
+
+		// Flush TAB posts cache, if the selected act page status change.
```
Inwerpsel, line 329: Maybe it's easier if we put this in a separate function. That way you can invert the 2 if-statements inside to return from that function, so we don't need to wrap the statements inside of them.

src/MasterSite.php
```diff
@@ -325,6 +325,52 @@ public function clean_post_cache( $post_id, $post_after, $post_before ) {
...
+				$tab_posts = [];
+				// Search for those posts, who use TAB($post_id) from "Take Action Page Selector" dropdown.
+				// phpcs:disable
+				$sql          = 'SELECT post_id FROM %1$s WHERE meta_key = \'p4_take_action_page\' AND meta_value = %2$d';
+				$prepared_sql = $wpdb->prepare( $sql, $wpdb->postmeta, $post_id );
+				$result       = $wpdb->get_results( $prepared_sql );
+				// phpcs:enable
+
+				foreach ( $result as $tab_post ) {
+					$tab_posts[] = $tab_post->post_id;
```
Inwerpsel, line 349: If we use `$wpdb->get_col( $prepared_sql );`, then we don't need this foreach.

src/MasterSite.php
```diff
@@ -325,6 +325,52 @@ public function clean_post_cache( $post_id, $post_after, $post_before ) {
...
+		if ( 'page' === $post_after->post_type && $parent_act_id === $post_after->post_parent ) {
+
+			// Flush cache only when a page status changes from publish to any non-public status & vice versa.
+			if (
+				( 'publish' === $post_before->post_status && 'publish' !== $post_after->post_status ) ||
+				( 'publish' === $post_after->post_status && 'publish' !== $post_before->post_status )
+			) {
+				global $wpdb, $nginx_purger;
+
+				$tab_posts = [];
```
Inwerpsel, line 340: We probably use it in other places in the code, but it's kinda confusing that it abbreviates to an existing and common word `tab`. Could be less ambiguous if we refer to it as `$boxout_posts`?

src/MasterSite.php
```diff
@@ -325,6 +325,52 @@ public function clean_post_cache( $post_id, $post_after, $post_before ) {
...
+
+				// Search for those posts, who use TAB($post_id) as a block inside block editor.
+				$take_action_boxout_block = '%<!-- wp:planet4-blocks/take-action-boxout {"take_action_page":' . $post_id . '} /-->%';
+				// phpcs:disable
+				$sql          = 'SELECT ID FROM %1$s WHERE post_type = \'post\' AND post_status = \'publish\' AND post_content LIKE \'%2$s\'';
+				$prepared_sql = $wpdb->prepare( $sql, $wpdb->posts, $take_action_boxout_block );
+				$result       = $wpdb->get_results( $prepared_sql );
+				// phpcs:enable
+
+				foreach ( $result as $tab_post ) {
```
Inwerpsel, line 360: Here we an use `$wpdb->get_col` too.

src/MasterSite.php
Inwerpsel: Then we're good until we have a second kind of boxout :smile:

src/MasterSite.php
Inwerpsel: So would be something like
```
private function check_act_page_change ($post_id) {
  if ( [[not the act page]] ) {
    // nothing to do
    return;
  }
  if ( [[status didn't change]] ) {
    // nothing to do
    return;
  }
}
```

src/MasterSite.php
```diff
@@ -325,6 +325,52 @@ public function clean_post_cache( $post_id, $post_after, $post_before ) {
...
+					$tab_posts[] = $tab_post->ID;
+				}
+
+				// Flush TAB posts cache.
+				$tab_posts = array_unique( $tab_posts );
```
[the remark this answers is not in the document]
sagarsdeshmukh: If for same post, editor has selected dropdown value as well as added TAB block, the 2 query give same post id which won't skip by `SELECT DISTINCT ID` condition.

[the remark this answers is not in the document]
lithrel: Haaa right my bad, missed a query 🤦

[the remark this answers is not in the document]
sagarsdeshmukh: np :)

PR #1159, opened 2020-07-30
PLANET-5232 Add a visited state to blue links

dpivo:
Added a function that cycles through a tags on a page and adds a class if the link color is blue. That class contains the :visited styles

### DIFF
--- a/assets/src/js/app.js
+++ b/assets/src/js/app.js
@@ -2,7 +2,7 @@ import { setupCookies } from './cookies';
 import { setupAuthor } from './author';
 import { setupCommentsAnchor } from './comments_anchor';
 import { setupCountrySelect } from './country_select';
-import { setImageTitlesFromAltText } from './global';
+import { setImageTitlesFromAltText, setBlueLinkStyles } from './global';
 import { setupHeader } from './header';
 import { setupImageAlign } from './img_align';
 import { setupLoadMore } from './load_more';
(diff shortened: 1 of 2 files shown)


comzeradd:
Why not do this in css? The "blue links" are the ones using `$link-color` and it seems we only use that [in a couple of places](https://github.com/greenpeace/planet4-styleguide/search?q=link-color&unscoped_q=link-color). We could add a `$link-color-visited` variable and adjust the css code.

mleray:
> in a couple of places

mleray:
> Why not do this in css? The "blue links" are the ones using `$link-color` and it seems we only use that [in a couple of places](https://github.com/greenpeace/planet4-styleguide/search?q=link-color&unscoped_q=link-color). We could add a `$link-color-visited` variable and adjust the css code.

Made a mess with the PR, sorry 🤦 Anyway, at first it was using css but the `$link-color` is set in `src/base/_typography.scss` and other links (for example buttons) don't have a `:visited` state, they were picking up the changes too and we don't want that to happen... So in css we would either need a really long selector (i.e. `:not(.btn)` etc) or we would need to override the `:visited` selector for all of them. With the risk of course to forget some of them, or to forget to add this selector in case we add new links :/

comzeradd:
Yes, good point. I left a comment in the ticket for Will to review. This seems like a hack to me. If we want visited color state we should code one for all of our link variations and classes (which is a good practice anyway).

comzeradd:
Per latest ticket comment, we are not doing this after all.

PR #1162, opened 2020-08-12
PLANET-4808: Display of cmb2 fields in rest api response

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-4808
Related to: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/358
Doc: https://github.com/CMB2/CMB2/wiki/REST-API#registered-rest-fields

In order to get full functionalities for the [WYSIWYG version of the Split Two Columns block](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/358), I need those fields to appear in the rest api response giving a list of pages.  
The loading of MetaboxRegister is based on `wp_is_json_request()`, which needs a header `Http-Accept` or `Content-Type` with `application/json` if you want to test this. 
I also tried to base this solution on the constant `REST_REQUEST`, but it is defined far too late in the process.

## Test:
Request a list of pages from the rest api.
```console
> curl -v -H "Content-Type: application/json" "http://www.planet4.test/wp-json/wp/v2/pages?per_page=100&sort_order=asc&sort_column=post_title&parent=11&post_status=publish&context=view&_locale=user" | jq
```
Each page returned should have a `cmb2` object, containing a `p4_metabox` object with various fields.

Screenshot 2020-08-12 at 08 14 14

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -93,7 +93,7 @@ final class Loader {
 				);
 			}
 
-			// Load P4 Metaboxes only when adding/editing a new Page/Post/Campaign.
+			// Load `Campaigns` class only when adding/editing a new tag.
 			if ( 'edit-tags.php' === $pagenow || 'term.php' === $pagenow ) {
 				$this->default_services[] = Campaigns::class;
 			}
@@ -114,6 +114,10 @@ final class Loader {
 		// Run Activator after theme switched to planet4-master-theme or a planet4 child theme.
 		if ( get_option( 'theme_switched' ) ) {
 			$this->default_services[] = Activator::class;
+		}
+
+		if ( wp_is_json_request() ) {
+			$this->default_services[] = MetaboxRegister::class;
 		}
 
 		$services = array_merge( $services, $this->default_services );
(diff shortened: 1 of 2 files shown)


src/Loader.php
```diff
@@ -93,7 +93,7 @@ private function load_services( $services ) {
 				);
 			}
 
-			// Load P4 Metaboxes only when adding/editing a new Page/Post/Campaign.
+			// Load Campaigns only when adding/editing a new Tag.
```
sagarsdeshmukh, line 96: Just to clarify the reference of Campaigns::class is not related to the P4CG or "Campaign Pages" here, At very beginning of P4 (when we don't have P4CG as a post type) we were refer the tags to campaign ( [eg](https://github.com/greenpeace/planet4-master-theme/blob/master/src/Search.php#L747) )
and [Category to Issue](https://github.com/greenpeace/planet4-master-theme/blob/master/src/Search.php#L734)

lithrel: Oups, thanks for that, I'll fix the comment 👍

PR #1165, opened 2020-08-18
PLANET-5275: Running accessibility report with pa11y

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-5275
Ref: https://github.com/greenpeace/planet4/issues/73

Requires: https://github.com/greenpeace/planet4-base-fork/pull/91 (pa11y configuration)  
Requires: https://github.com/greenpeace/planet4-docker-compose/pull/52 (targets for pa11y installation and run)

A report generated by circle-ci is visible there https://9211-80401279-gh.circle-artifacts.com/0/planet4-docker-compose/artifacts/pa11y/index.html

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,6 +8,8 @@ workflows:
       - php73-tests
       - acceptance-tests:
           context: org-global
+      - a11y-tests:
+          context: org-global
       - rips-test:
           context: rips-scans
           filters:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -109,18 +109,22 @@ jobs:
...
           name: Test - Run tests
           command: |
             export BUILD_TAG="build-$(cat /tmp/workspace/var/circle-build-num)"
             export APP_IMAGE=gcr.io/planet-4-151612/planet4-base-app:${BUILD_TAG}
             export OPENRESTY_IMAGE=gcr.io/planet-4-151612/planet4-base-openresty:${BUILD_TAG}
             make -C planet4-docker-compose ci test
+      - run:
+          name: Test - Run accessibility report
+          command: |
+            make -C planet4-docker-compose test-pa11y
```
comzeradd, line 127: Do you think, this can be broken in into a separate job? (eg. a11y-tests). That way it can run in parallel, but most importantly it would allow us to merge this and have the tests run without blocking PRs until we fix all the issues reported.

lithrel: So I've made it run in another job, and [it seems to work](https://app.circleci.com/pipelines/github/greenpeace/planet4-master-theme/1606/workflows/f19ad42a-2edb-44fc-b852-ae4797c260a5), but because I still want to have a report of warnings (and maybe notices), I'm doing something not pretty with jq and an exit code, don't know if there's a better way.  
What I'm wondering is, how can I configure this job so that it fails without blocking a PR merge ?

comzeradd: I just adjusted repo settings, so instead of requiring the whole `main` workflow to pass it would require only the 3 jobs we want. We will still see a11y job check, but it won't block the merge button.

comzeradd (approved):
Looks nice! Extra :pancakes: for making abstracting re-usable config parts.

comzeradd:
Oh, and don't forget to update branches before merge :exclamation:

PR #1166, opened 2020-08-24
PLANET-4811 Add backtop markup to base template

Inwerpsel:
* Similarly to how .skip-links was already set up, we can add the
backtop markup here and hide it. That way it's already directly under
the body, and we don't need to dynamically append it. Since the click
handler is very short we can directly include it here for simplicity.

### DIFF
--- a/templates/base.twig
+++ b/templates/base.twig
@@ -11,7 +11,8 @@
 		<li><a href="#content">{{ __( 'Skip to Content', 'planet4-master-theme' ) }}</a></li>
 		<li><a href="#footer">{{ __( 'Skip to Footer', 'planet4-master-theme' ) }}</a></li>
 	</ul>
-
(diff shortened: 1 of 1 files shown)


templates/base.twig
```diff
@@ -11,7 +11,8 @@
 		<li><a href="#content">{{ __( 'Skip to Content', 'planet4-master-theme' ) }}</a></li>
 		<li><a href="#footer">{{ __( 'Skip to Footer', 'planet4-master-theme' ) }}</a></li>
 	</ul>
-
+	<a style="display: none;" title="{{ __('Go to the top of the page.', 'planet4-master-theme') }}" class="back-top"
```
comzeradd, line 14: Instead of the inline `style` here you can probably use the `d-none` class [from bootstrap](https://getbootstrap.com/docs/4.5/utilities/display/#notation).

mleray (approved):
Looks good to me 👍

PR #1170, opened 2020-08-28
Planet 5221

Inwerpsel:
Check if the behavior is the same as for master branch.


Inwerpsel:
Just for reference the last commit (`will this cancel pipeline in main PR?`) did in fact cancel the workflow of a non-fork PR which had the previous commit as its head (5e1ce87) and was still running. Probably a bug (or at least weird behavior) in CircleCI, though unlikely to happen in practice.

PR #1172, opened 2020-08-31
PLANET-5292 create record, move to separate class

Inwerpsel:
* MigrationRecord holds the data of a particular run of a migration
script.
* Move migration logic into it's own class.
* Throw a specific exception.

### DIFF
--- a/src/MigrationLog.php
+++ b/src/MigrationLog.php
@@ -2,6 +2,8 @@
 
 namespace P4\MasterTheme;
 
+use P4\MasterTheme\Migrations\MigrationRecord;
+
 /**
  * A log of all migrations that have run, saved as a WP option.
  */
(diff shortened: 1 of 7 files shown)


src/MigrationLog.php
```diff
@@ -40,7 +42,7 @@ public static function from_wp_options(): self {
 	 */
 	public function already_ran( string $migration_id ): bool {
 		foreach ( $this->done_migrations as $migration ) {
-			if ( $migration['id'] === $migration_id ) {
+			if ( $migration['id'] === $migration_id && $migration['success'] ) {
```
Inwerpsel, line 45: We might want to look into allowing (maybe enforcing) some kind of success query, so that we can actually determine whether the script needs to run for any given data.

Inwerpsel: Though that might be a bit overkill for this PR, all I wanted to do initially was make an option default to true :smile:

PR #1173, opened 2020-09-02
Collect vars and add beta toggles for theme editor

Inwerpsel:
This adds the build script to make the CSS variables file used for the theme editor, as well as a toggle which will be used by https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/495

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "lint:css": "wp-scripts lint-style"
   },
   "devDependencies": {
-    "@greenpeace/dashdash": "^1.0.1",
+    "@greenpeace/dashdash": "^1.0.2",
     "@wordpress/components": "^8.3.2",
     "@wordpress/scripts": "3.3.0",
     "autoprefixer": "^9.6.1",
--- a/package.json
+++ b/package.json
@@ -26,7 +26,8 @@
     "svg-sprite-loader": "^4.1.6",
     "terser-webpack-plugin": "^2.2.1",
     "webpack": "^4.40.2",
-    "webpack-fix-style-only-entries": "^0.4.0"
+    "webpack-fix-style-only-entries": "^0.4.0",
+    "postcss-css-variables-extract": "^0.1.3"
   },
   "browserslist": [
     "last 2 version",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -36,6 +36,7 @@
     "bootstrap": "^4.3.1",
     "classnames": "^2.2.6",
     "jquery": "^3.5.1",
-    "popper.js": "^1.15.0"
+    "popper.js": "^1.15.0",
+    "postcss-css-variables-extract": "^0.1.2"
```
Inwerpsel, line 40: Probably I should move this to the dev dependencies if we only need it during build

src/Features.php
```diff
@@ -66,6 +70,24 @@ private static function get_fields(): array {
...
+				'name' => __( 'Theme editor (experimental)', 'planet4-master-theme-backend' ),
+				'desc' => __(
+					'Enable CSS variables based theme editor for logged in users. Do not use in production yet.',
+					'planet4-master-theme-backend'
+				),
+				'id'   => self::THEME_EDITOR,
+				'type' => 'checkbox',
+			],
+			[
+				'name' => __( 'Theme editor non-logged in(experimental)', 'planet4-master-theme-backend' ),
```
Inwerpsel, line 83: I made this a separate checkbox to be sure people see the warning about enabling it for all users.

comzeradd: I left a comment on the other PR. I think it would make more sense not to enable that for non-logged-in users.

comzeradd:
Since you are on the Features page, can you make a small css tweak to make sure the option text stays on the same line?

fearures

Looks like this line is causing the problem:

```css
.cmb-th+.cmb-td {
    float: left;
}
```

Inwerpsel:
> Since you are on the Features page, can you make a small css tweak to make sure the option text stays on the same line?
> 
> fearures
> 
> Looks like this line is causing the problem:
> 
> ```css
> .cmb-th+.cmb-td {
>     float: left;
> }
> ```

Added this fix https://github.com/greenpeace/planet4-master-theme/blob/de89e006b98c7ad97b55b855e3c3223479d0fe0a/admin/css/campaign.css#L37-L39

comzeradd (approved):

PR #1175, opened 2020-09-03
Class to facilitate parameterized queries

Inwerpsel:
* Sql query parameters need to be be passed separately to $wpdb->prepare
and the values need to be in the same order as the placeholders in the
query. This makes the code hard to read and prone to bugs if a mistake
in the position is made. The class introduced here solves this problem
by allowing to specify these values in the query, where it will echo
back a placeholder and keep track of all values in the order they were
passed in. This ensure that the order will be the same, and that the
values can be seen in the position they will be used.

### DIFF
--- a/src/Exception/SqlInIsEmpty.php
+++ b/src/Exception/SqlInIsEmpty.php
@@ -1,0 +1,12 @@
+<?php
+
+namespace P4\MasterTheme\Exception;
+
+use Exception;
+
+/**
+ * An empty array was given to be used in a SQL IN query, but that doesn't work.
+ */
+class SqlInIsEmpty extends Exception {
+
+}
--- a/src/Search.php
+++ b/src/Search.php
@@ -1114,12 +1114,14 @@ abstract class Search {
 	public static function exclude_unwanted_attachments( $args ) {
 		global $wpdb;
 
-		$in_placeholders = generate_list_placeholders( self::DOCUMENT_TYPES, 2, 's' );
+		$params = new SqlParameters();
 
-		$sql = 'SELECT id FROM %1$s WHERE post_type = "attachment" AND post_mime_type NOT IN (' . $in_placeholders . ')';
+		$sql = 'SELECT id FROM ' . $params->object( $wpdb->posts )
+			. ' WHERE post_type = "attachment"'
+			. ' AND post_mime_type NOT IN (' . $params->string_array( self::DOCUMENT_TYPES ) . ')';
 
 		$unwanted_attachment_ids = $wpdb->get_col(
-			$wpdb->prepare( $sql, array_merge( [ $wpdb->posts ], self::DOCUMENT_TYPES ) ) // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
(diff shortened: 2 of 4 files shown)


lithrel (approved):
:+1: 
I've added a test there if you want https://github.com/greenpeace/planet4-master-theme/compare/sql-params-wrapper...sql-params-wrapper-test because it seemed like a good candidate for that (a bit critical, well isolated).  

Side-notes: <sub>because the less code, the more comments of course</sub>  
- I feel like the `*_array` methods might be better named `*_list` according to what they return.
- You could add a `float()` method, wordpress `prepare()` [apparently](https://github.com/WordPress/WordPress/blob/master/wp-includes/wp-db.php#L1258) only handles string, int and float placeholders.
- Because parameters are typed, you could have a method `list` that replaces `*_array` methods for any list of arguments (or any array, what you prefer :man_shrugging: )
```php
	public function list(...$args): string {
		$params = [];
		foreach ( $args as $arg ) {
			if ( is_string( $arg ) ) {
				$params[] = $this->string( $arg );
			} elseif ( is_int( $arg ) ) {
				$params[] = $this->int( $arg );
			} else {
				throw new \Exception( 'Wrong type buddy.' );
			}
		}
		return implode( ',', $params );
	}
```
(not that we need a list of disparate arguments in an sql query I realize, it's just one less method)

Inwerpsel:
> I've added a test there if you want [sql-params-wrapper...sql-params-wrapper-test](https://github.com/greenpeace/planet4-master-theme/compare/sql-params-wrapper...sql-params-wrapper-test) because it seemed like a good candidate for that (a bit critical, well isolated).

Nice, I'll include it :+1: 

> * I feel like the `*_array` methods might be better named `*_list` according to what they return.

Makes sense to use list as it's a more generic term :+1: 

> * You could add a `float()` method, wordpress `prepare()` [apparently](https://github.com/WordPress/WordPress/blob/master/wp-includes/wp-db.php#L1258) only handles string, int and float placeholders.

Indeed, though maybe it's better to only add it when we need to use it, don't want to rush it to include it in this PR and end up writing a bug :slightly_smiling_face: 

> * Because parameters are typed, you could have a method `list` that replaces `*_array` methods for any list of arguments (or any array, what you prefer man_shrugging )
> 
> ```
> 	public function list(...$args): string {
> 		$params = [];
> 		foreach ( $args as $arg ) {
> 			if ( is_string( $arg ) ) {
> 				$params[] = $this->string( $arg );
> 			} elseif ( is_int( $arg ) ) {
> 				$params[] = $this->int( $arg );
> 			} else {
> 				throw new \Exception( 'Wrong type buddy.' );
> 			}
> 		}
> 		return implode( ',', $params );
> 	}
> ```
> 
> (not that we need a list of disparate arguments in an sql query I realize, it's just one less method)

Don't we actually want all elements passed to an IN query to be of the expected type?

If we want that, then that single function would also need a parameter for which type each element should have. But then it would get quite complex to avoid a small amount of duplication in 2 otherwise simple functions.

lithrel (approved):
good changes, and having different typed list methods makes sense :+1:

PR #1176, opened 2020-09-04
PLANET-5297 Remove custom lazy-loading implementation

comzeradd:


mleray (commented):
I just have one questions about this! I found [this function](https://github.com/greenpeace/planet4-master-theme/blob/master/src/Search.php#L286) that still specifies "Lazy-loading" in the description, it's probably good but just wanted to double check 🙂

comzeradd:
Yes, I also noticed that but doesn't seem to have anything related to lazy loading. I removed the comment to avoid confusion.

mleray (approved):
Looks good to me 👍

PR #1177, opened 2020-09-09
PLANET-5363 Revert "PLANET-4536: Perf: Reduce unused dependencies from bootstrap in code"

Inwerpsel:
Reverts greenpeace/planet4-master-theme#1116

Ref: https://jira.greenpeace.org/browse/PLANET-5363
Ref reverted ticket: https://jira.greenpeace.org/browse/PLANET-4536

### DIFF
--- a/assets/src/scss/bootstrap-build.scss
+++ b/assets/src/scss/bootstrap-build.scss
@@ -15,8 +15,8 @@ Text Domain: planet4-master-theme
 // Full bootstrap:
 // @import "~bootstrap/scss/bootstrap";
 
-@import "~bootstrap/scss/bootstrap-grid.scss";
-@import "~bootstrap/scss/bootstrap-reboot.scss";
+// @import "~bootstrap/scss/bootstrap-grid.scss";
+// @import "~bootstrap/scss/bootstrap-reboot.scss";
 
 // Required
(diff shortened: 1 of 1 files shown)


comzeradd:
Not sure if we want to bring back everything. Mostly concerned about things that might have changed in our css code in the meantime. For the collapsible elements the relevant code seems to be in `_cards.scss` and `_transitions.scss` (I don't see the first one on the list, but [here it is](https://github.com/twbs/bootstrap/blob/main/scss/_card.scss)).

@pablocubico thoughts?

Inwerpsel:
@comzeradd Indeed, probably best to start by re-adding these parts we currently know are missing, which should be pretty safe. Maybe we can follow it up with a ticket to check if there are other parts we need to add again. For example if we missed usages in the content or in child themes that depend on it.

Inwerpsel:
Made a PR to add only `_cards.scss` and `_transitions.scss` https://github.com/greenpeace/planet4-master-theme/pull/1179, which solves the currently known issues. I'll close this one for now.

pablocubico:
@comzeradd Agreed, I think only `_transitions` is needed for the accordion, but we chatted about this with Pieter, I remembered those were re-added for the Bulgaria site, so I guess let's go ahead and add both `cards` and `transitions`.

PR #1179, opened 2020-09-10
PLANET-5363 Re-add bootstrap card and transitions

Inwerpsel:
* These were removed, but turns out that they're needed anyway in a few
places.

We may want to check the other parts to in a later PR, this one just re-adds the ones which we know already are needed.
Ref: https://jira.greenpeace.org/browse/PLANET-5363, also fixes https://jira.greenpeace.org/browse/PLANET-5475

@comzeradd `_cards.scss` is indeed used by the columns block, though at first glance it looks like it's barely actually using it. Most of the rules of those classes are overridden. Probably best for now to re-add it, but maybe afterwards we can add those few rules it does use to our css and drop the dependency.

Test instance:
- search page: https://k8s.p4.greenpeace.org/test-mars/?s=test&orderby=_score
- page with a columns task style block (accordion only on mobile): https://k8s.p4.greenpeace.org/test-mars/act/join-the-movement-for-clean-air/

### DIFF
--- a/assets/src/scss/bootstrap-build.scss
+++ b/assets/src/scss/bootstrap-build.scss
@@ -37,7 +37,7 @@ Text Domain: planet4-master-theme
 // @import "~bootstrap/scss/_breadcrumb.scss";
 // @import "~bootstrap/scss/_button-group.scss";
 // @import "~bootstrap/scss/_buttons.scss";
-// @import "~bootstrap/scss/_card.scss";
+@import "~bootstrap/scss/_card.scss";
 @import "~bootstrap/scss/_carousel.scss";
 // @import "~bootstrap/scss/_close.scss";
 // @import "~bootstrap/scss/_code.scss";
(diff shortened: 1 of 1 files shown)


comzeradd (approved):

pablocubico:
@comzeradd @Inwerpsel Looks good, I think we need to talk about Bootstrap in sync with the design team, probably having Magali & Matt on board, because we are currently using only a few bits of it, but not following the Bootstrap rules (e.g.: the grid) as its intended, or overriding a lots of styles in some cases. I think we can actually benefit from it by using it a little more, following the Bootstrap patterns and examples more closely, and having some overrides, it could even be the basis of out design system. For example, using `.btn-primary` and `.btn-secondary`, or the Navbar styles, or Cards, Alerts. etc., adapting the styles for those classes with clear, straight-forward overrides. The site won't look like a "Bootstrap site".

I did this for other projects and it has a few benefits, your code would then be consistent with the Bootstrap docs, so you can practically refer to their examples for a lot of dev-time situations (for example, when we are building a new block).

I already made some progress with the Grid component, in order to use an 8px grid, but to do that, we actually have to fix the way we are building the layout for the pages, despite we use an 8px grid or not.

PR #1181, opened 2020-09-22
PLANET-5484 Add asset build job

Inwerpsel:


Inwerpsel:
I'll close this for now as we went for [building the assets on on tag pushes only as a first step](https://github.com/greenpeace/planet4-master-theme/pull/1182), which already solves removes the asset building from a production deploy. We'll get back to this later to do move the building job earlier so that tests also don't need to build.

PR #1183, opened 2020-09-24
PLANET-5515 When 0 search results set 204 http header

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-5515

Deployed on my test instance. eg. https://k8s.p4.greenpeace.org/nikosdev/?s=blahblah&orderby=_score

The above page returns 0 zero results, with the normal search template. But the http status header is `404`.

This can be tested putting that url at [httpstatus](https://httpstatus.io/) or looking at web console.

http

### DIFF
--- a/src/Search.php
+++ b/src/Search.php
@@ -355,6 +355,7 @@ abstract class Search {
 		$posts = $this->query_posts( $paged );
 
 		if ( empty( $posts ) ) {
+			status_header( 204 );
 			return [];
 		}
 


sagarsdeshmukh (approved):
LGTM 👍

comzeradd:
Did a minor change to `204` (No Content) instead of `404`, which feels less aggressive and accurate.

We can merge this to quickly mitigate spam bots, but I keep investigating the idea @Inwerpsel suggested to add a custom http header.

PR #1185, opened 2020-10-01
PLANET - 5538 Updated search.twig for Search Results title

powersaudrey25:


powersaudrey25:
I am unsure as to why the checks failed, could someone explain why?

Inwerpsel:
> I am unsure as to why the checks failed, could someone explain why?

Some jobs were set up to require secrets (such as credentials) which are set up on our CircleCI instance for this GitHub repository. This is the case for the acceptance tests. We're currently working on fixing that actually, so we may be able to try out the fix on this PR. see https://jira.greenpeace.org/browse/PLANET-5221

PR #1188, opened 2020-10-01
PLANET-5538 Use h1 for search title

powersaudrey25:
PLANET-5521 Fix syntax

Fixes greenpeace/planet4#104

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,10 +6,8 @@ workflows:
     jobs:
       - php72-tests
       - php73-tests
-      - acceptance-tests:
-          context: org-global
-      - a11y-tests:
-          context: org-global
+      - acceptance-tests
+      - a11y-tests
       - create-release-zip:
           context: org-global
           filters:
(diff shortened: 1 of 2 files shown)


Inwerpsel:
Related issue: https://github.com/greenpeace/planet4/issues/104

Inwerpsel (commented):
The code change looks good, but could you provide a commit message describing what the commit changes? We follow the approach described in this article: https://chris.beams.io/posts/git-commit/ . It mainly involves using imperative form, in this case something like "Use h1 for search title". You can also prefix it with the Jira ticket number PLANET-5538, so "PLANET-5538 Use h1 for search title". Then we can see for each commit what the story behind it is when looking at Git history.

Inwerpsel:
A bit of context here, as some of this was discussed in the chat of Open Source Day 2020. There is [a pending PR](https://github.com/greenpeace/planet4-master-theme/pull/1151/files#diff-1d37e48f9ceff6d8030570cd36286a61L111-R121) (Jira [PLANET-5221](https://jira.greenpeace.org/browse/PLANET-5221)) for this repository which will allow to run the acceptance and accessibility tests on PRs from external forks. That PR still needs to be updated because in the meanwhile other changes to the same configuration were merged.

We thought it would be nice to get the test result for this PR, as there's likely a test that will fail on the markup change. So I pointed to the pending PR and gave some context about the changes that are needed, and @powersaudrey25 made the config work in a couple of tries :100: The tests are running, though some of them are failing unexpectedly. But it's probably a small issue as most are passing :crossed_fingers:

comzeradd:
Thank you so much for the contribution @powersaudrey25! Change has been merged to the `master` branch.

PR #1195, opened 2020-10-06
Load ponyfill only for ie

Inwerpsel:
The ponyfill is only needed for ie11 and Opera mini (ignoring some browsers with virtually no usage).
https://caniuse.com/css-variables
We can add a conditional if we only want to support ie11, though the ponyfill then wouldn't apply to Opera Mini.

I still need to do some tests on ie if it always loads the script.

### DIFF
--- a/assets/src/js/cssvarsponyfill.js
+++ b/assets/src/js/cssvarsponyfill.js
@@ -1,4 +1,6 @@
 /* global cssVars */
 export const setupCSSVarsPonyfill = () => {
-  cssVars();
+  if ('undefined' !== typeof cssVars) {
+    cssVars();
+  }
 };
--- a/assets/src/js/cssvarsponyfill.js
+++ b/assets/src/js/cssvarsponyfill.js
@@ -1,5 +1,17 @@
 /* global cssVars */
 export const setupCSSVarsPonyfill = () => {
+  // Taken from https://stackoverflow.com/questions/36653217/opera-mini-browser-detection-using-javascript.
+  const isOperaMini = Object.prototype.toString.call(window.operamini) === '[object OperaMini]';
+
+  if (isOperaMini) {
+    const script = document.createElement('script');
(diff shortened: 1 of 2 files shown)


comzeradd:
We can also check the library to see what exactly it does. For instance from [its description](https://jhildenbiddle.github.io/css-vars-ponyfill/):

> Limitations: Custom property declaration support is limited to `:root` and `:host` rulesets.

Does this mean that it only supports variables inside `:root`? :thinking: 

Checking user agent feels like a dirty hack :hammer:

Inwerpsel:
> Does this mean that it only supports variables inside `:root`? 
> 

The limitation is only for property declarations, so the following is not supported:
`.mydiv { --my-var: green }`. But we don't make use of that afaik. `var()` calls can be used anywhere.

> Checking user agent feels like a dirty hack

Yep, just also realized that we can't do both that and cache the page :facepalm: so I'll drop that again.

Inwerpsel:
I tested doing the UA check client side, which should reliably detect it. However the result seems inconsistent, maybe loading the script only at that point is too late and the variables are not applied most of the time.

Opera Mini turns out to be an interesting story. If you install the the app and just run a site that doesn't have the ponyfill, you'll find that css variables do actually work (unless you're on a really slow connection, explained below why). It's only when you enable "extreme savings mode" that they stop working.

The reason behind this is that the extreme mode will actually make your phone connect with Opera servers that "preload" the page (inc all static resources), then transforms it into a lightweight representation with less static assets. Those servers are running an older version of Opera which doesn't support css variables yet (though will work with the ponyfill).

The default mode of the app is "Automatic", which will only use the servers if the connection is slow.

see https://dev.opera.com/articles/opera-mini-and-javascript/

comzeradd:
One more test we could, is see how the site looks in these browsers without this library (thus without css variables support).
If we can define some sane fallbacks that make the site at least look descent and readable, it's a fair compromise in order to completely remove that library.

Inwerpsel:
Closing in favor of https://github.com/greenpeace/planet4-master-theme/compare/css-vars-feature-detection

PR #1196, opened 2020-10-07
Planet 5485 Rewrite ImageArchive using hooks

mleray:
See https://jira.greenpeace.org/browse/PLANET-5485

I've also added a small fix for the multi-selection to work with Mac keyboards 🙂


assets/src/js/Components/ArchivePicker.js
```diff
@@ -1,151 +1,115 @@
...
+      onChange={event => setEnteredSearch(event.target.value)}
+    >
+      <input
+        type={'text'}
+        disabled={loading}
+      />
+      <input type={'submit'} value={__('Search', 'planet4-master-theme-backend')} />
+    </form>
+    {loading && (
+      <div className={'archive-picker-loading'}> loading...</div>
```
lithrel, line 69: There's a [Spinner component](https://github.com/WordPress/gutenberg/tree/162bc50a487f987d1c3250d769f07a28598fe3ae/packages/components/src/spinner) in gutenberg if we want :)

assets/src/js/Components/archivePicker/SingleSidebar.js
```diff
@@ -1,135 +1,114 @@
...
+  const original = largestSize(image) || {};
+  const aspectRatio = original.height / original.width;
 
-  async componentDidUpdate( prevProps ) {
-    if ( prevProps.image !== this.props.image ) {
-      this.preloadImage();
+  const preloadImage = async () => {
+    if (image) {
+      setImageLoaded(false);
+      await fetch(largestSize(getSizesUnder(image.sizes, PREVIEW_MAX_SIZE)));
```
lithrel, line 28: `getSizesUnder()` returns an array of `image` objects, which does not seem to fit `largestSize()`; fetch here doesn't fetch anything ? I noticed this because there's an ajax call to `/undefined` on every click to select an image

lithrel (commented):
The button `Include in WP` doesn't appear if multiple images are selected, I don't know if it's intentional ?
Looking good, doing another pass tomorrow :ok_hand:

assets/src/js/Components/archivePicker/SingleSidebar.js
Inwerpsel: The reason I had added this was because otherwise the sidebar would render with the new data but still the old image. Then only when the image is fully loaded it would replace the previous image. This can take a long time if it's a big image.

I had first added the preloading, then later also added `getSizesUnder`, because the original image was extremely large in some cases. But looks like I broke it here as `largestSize` expects the image itself.

I didn't notice because even with the wrong fetch it still causes the previous image to be replaced with the loading placeholder.

Inwerpsel: You can see the original issue if you initialize `imageLoaded` as true and remove the preloading effect. Click on a few images and you'll see some of the time the previous image is displayed with the data of the new one for multiple seconds.

Inwerpsel: But it was more a quick fix at the end of the PR, I'm sure there's a better way to solve it.

Inwerpsel: Ah yes there is indeed https://github.com/facebook/react-native/issues/9195#issuecomment-238190537

Inwerpsel: Checked locally and this seems to work
```
            <img
              key={image.original.url}
              srcSet={toSrcSet(image.sizes, { maxWidth: PREVIEW_MAX_SIZE })}
              title={image.title}
              alt={image.title}
            />
```

assets/src/js/Components/ArchivePicker.js
```diff
@@ -1,151 +1,115 @@
...
+        <p> {error.message} </p>
+      </div>
+    )}
+    <div className={'image-picker'}>
+      <ul
+        className={'picker-list'}
+        onScroll={async (event) => {
+          if (isNearScrollEnd(event)) {
+            await loadPage(currentPage + 1, searchText);
+            setCurrentPage(currentPage + 1);
```
Inwerpsel, line 83: If we only use the `setCurrentPage` call here, and pass `currentPage` as a dependency [here](https://github.com/greenpeace/planet4-master-theme/pull/1196/files?file-filters%5B%5D=.js&file-filters%5B%5D=No+extension&hide-deleted-files=true&w=1#diff-867d658bf27f0e63a343817562d12adeR47), it should do the same thing.

Inwerpsel: We also would need to add the search text there, but that would make sense regardless. We might later keep track of the search text in the url and apply the search filter on the initial load.

assets/src/js/Components/archivePicker/SingleSidebar.js
mleray: Updated! 😄

assets/src/js/Components/ArchivePicker.js
mleray: Updated! 🙂

mleray:
> The button `Include in WP` doesn't appear if multiple images are selected, I don't know if it's intentional ?

@Inwerpsel This was intentional right? I'm pretty sure it was like that before refactor

Inwerpsel:
> > The button `Include in WP` doesn't appear if multiple images are selected, I don't know if it's intentional ?
> 
> @Inwerpsel This was intentional right? I'm pretty sure it was like that before refactor

Indeed. Well, I intended to include it but we wanted to wrap up and it's behind a feature flag anyway.

Inwerpsel (commented):
Posting these comments already, will continue later but I think we're almost there :+1:

assets/src/js/Components/ArchivePicker.js
mleray: I looked into it, we are missing some dependencies to make it work in master-theme... Since this feature is hidden for now and UX needs to be reviewed anyway, we can keep it in mind for then! 🙂

assets/src/js/Components/archivePicker/useImages.js
```diff
@@ -0,0 +1,85 @@
...
+    setLoading(true);
+
+    try {
+      const nextImages = await apiFetch({
+        path: addQueryArgs('/planet4/v1/image-archive/fetch', {
+          page: pageIndex,
+          search_text: searchedText,
+        })
+      });
+      const withPageLabel = nextImages.map(image => ({
```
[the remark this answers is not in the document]
mleray: Removed the mapping 🙂

Inwerpsel (approved):
Looks a lot better with hooks :+1:  There are some small leftover `async`s that we can clean up now, so would be nice to have that out of the way. Maybe also make it recognize the rename of `archivePickerList` to preserve the history of that file if you have time.

assets/src/js/Components/ArchivePicker.js
```diff
@@ -1,151 +1,115 @@
...
-        className={ 'archive-picker-search' }
-        onSubmit={ event => {
-          event.preventDefault();
-          this.search();
-        } }
-        onChange={ event => this.setState( { enteredSearch: event.target.value } ) }
+  return <Fragment>
+    <form
+      className={'archive-picker-search'}
+      onSubmit={async event => {
```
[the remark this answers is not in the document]
mleray: Indeed!

assets/src/js/Components/ArchivePicker.js
```diff
@@ -1,151 +1,115 @@
...
+    {!!error && (
+      <div>
+        <h3>API error:</h3>
+        <p> {error.message} </p>
+      </div>
+    )}
+    <div className={'image-picker'}>
+      <ul
+        className={'picker-list'}
+        onScroll={async (event) => {
```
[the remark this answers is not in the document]
mleray: Indeed!

assets/src/js/Components/archivePicker/useImages.js
```diff
@@ -0,0 +1,81 @@
...
+const { addQueryArgs } = wp.url;
+
+export const useImages = () => {
+  const [images, setImages] = useState([]);
+  const [loading, setLoading] = useState(false);
+  const [processingImages, setProcessingImages] = useState(false);
+  const [processingError, setProcessingError] = useState(null);
+  const [error, setError] = useState(null);
+
+  const loadPage = async (pageIndex, searchedText) => {
```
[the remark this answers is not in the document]
Inwerpsel: Nvm, it needs to be async as it uses await, not sure what I was thinking :facepalm:

assets/src/js/Components/ArchivePicker.js
```diff
@@ -1,151 +1,115 @@
...
+      <div>
+        <h3>API error:</h3>
+        <p> {error.message} </p>
+      </div>
+    )}
+    <div className={'image-picker'}>
+      <ul
+        className={'picker-list'}
+        onScroll={event => {
+          if (isNearScrollEnd(event) && !loading) {
```
Inwerpsel, line 82: Can you swap these so that it doesn't need to calculate the position if we're scrolling while loading?

PR #1202, opened 2020-10-13
Include dash dash plugin in webpack

Inwerpsel:
Plugin to transform CSS variables shorthand syntax. Same changes as in https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/410/files


lithrel (approved):
:+1:

PR #1206, opened 2020-10-19
Always add apps login filter

Inwerpsel:
* This filter is needed to prevent the cookie from being set anywhere
but the login page. If you don't add the filter then the cookie is
always set.
* The checks were in place to prevent other things from setting cookies.
Those things are not there anymore, so we don't need the checks.

Ref: https://jira.greenpeace.org/browse/PLANET-5605
deployed on https://k8s.p4.greenpeace.org/test-iocaste/

### DIFF
--- a/src/Cookies.php
+++ b/src/Cookies.php
@@ -6,32 +6,11 @@ namespace P4\MasterTheme;
  * Class Cookies
  */
 class Cookies {
-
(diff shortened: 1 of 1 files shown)


sagarsdeshmukh (approved):
Totally make sense 👍

PR #1215, opened 2020-10-22
Planet 5481 - Improve Image Archive UI

mleray:
See https://jira.greenpeace.org/browse/PLANET-5481

- [x] Use a dashicon (eg. https://developer.wordpress.org/resource/dashicons/#format-image)
- [x] Push the menu right after the Media (maybe remove "GPI" to make it shorter)
- [x] Rename the first submenu item to "Library"
- [x] Use same UI elements for the search bar and the buttons as the rest of wp-admin
- [x] Use grid instead of flexbox for displaying the image, to have them use the same dimensions (similar to the current medialibrary plugin)
- [x] Image link after importing should open on the same window

You can see the improved UI version [here](https://www-dev.greenpeace.org/test-iocaste/wp-admin/admin.php?page=gpi-image-picker)!

### DIFF
--- a/admin/css/picker.css
+++ b/admin/css/picker.css
@@ -1,22 +1,32 @@
 /* Todo: Change to scss, separate styles that are essential to the ImagePicker component to work (if there are any)
      from ArchivePicker styles*/
 
-.archive-picker-search input[type=submit] {
-  display: inline-block;
+.archive-picker-search {
+  margin-top: 20px;
+}
+
+.archive-picker-search input {
+  margin-right: 10px;
 }
 
 .picker-list {
   max-width: 60%;
   max-height: 75vh;
   overflow-y: scroll;
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  grid-template-columns: repeat(2, 1fr);
+  gap: 0 10px;
+}
+
+@media (min-width: 768px) {
+  .picker-list {
+    grid-template-columns: repeat(3, 1fr);
+  }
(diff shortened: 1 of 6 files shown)


lithrel (approved):
Code looks good to me :+1: 

Side-notes: compared to the Image gallery, the archive layout behaves differently on mobile and stretches images, but the image gallery has a popup for selected image instead of a sidebar so that's probably more difficult to handle here. Maybe it should go on top of the image list instead of staying on the side ? (in another PR :p)
The border color for selected image is also more visible in Image gallery (blue instead of yellow).

admin/css/picker.css
```diff
@@ -1,22 +1,26 @@
...
 }
 
 .picker-list {
   max-width: 60%;
   max-height: 75vh;
   overflow-y: scroll;
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  grid-template-columns: calc(33% - 10px) calc(33% - 10px) calc(33% - 10px);
```
comzeradd, line 17: Minor tweak here, to make it look better on mobile, we could use something similar to the [Gallery Block](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/styles/blocks/Gallery_grid.scss#L32-L46).

For small screens:
```
grid-template-columns: repeat(2, 1fr);
```

For bigger screens:
```
grid-template-columns: repeat(3, 1fr);
```

PR #1217, opened 2020-10-26
PLANET-5556 Hide Cloudflare plugin options

comzeradd:
Since we have multiple site behind Cloudflare, changing an option in one site affects all of them. This removes all settings options and menus of the Cloudflare plugin to prevent an admin accidentally changing something essential.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -114,6 +114,28 @@ add_action(
 	}
 );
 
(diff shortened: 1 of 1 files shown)


comzeradd:
Closing this for now, at least until we conclude [PLANET-5619](https://jira.greenpeace.org/browse/PLANET-5619)

PR #1221, opened 2020-11-02
PLANET-5460 Disable Wordpress(wp5.5) Blocks Directory feature

sagarsdeshmukh:
[JIRA 5460](https://jira.greenpeace.org/browse/PLANET-5460)

The branch is pin on [planet-test-nix](https://www-dev.greenpeace.org/test-nix/wp-admin/post-new.php?post_type=page) test instance for testing.

### DIFF
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -195,6 +195,9 @@ class MasterSite extends TimberSite {
 		add_action( 'admin_notices', [ $this, 'show_dashboard_notice' ] );
 		add_action( 'wp_ajax_dismiss_dashboard_notice', [ $this, 'dismiss_dashboard_notice' ] );
 		add_filter( 'timber/twig', [ $this, 'p4_optimize_img_url' ] );
+
+		// Disable WordPress(WP5.5) Block Directory.
+		$this->disable_block_directory();
 	}
 
 	/**
@@ -1278,5 +1281,25 @@ class MasterSite extends TimberSite {
 		}
 
 		return $source;
+	}
+
+	/**
+	 * Remove block directory assets and rest endpoint.
+	 */
+	public function disable_block_directory(): void {
+		remove_action( 'enqueue_block_editor_assets', 'wp_enqueue_editor_block_directory_assets' );
+		add_filter( 'rest_endpoints', [ $this, 'disable_block_directory_endpoint' ] );
+	}
+
+	/**
+	 * Remove block directory endpoint.
+	 *
+	 * @param array $endpoints The available endpoints.
+	 *
+	 * @return array
+	 */
+	public function disable_block_directory_endpoint( array $endpoints ): array {
+		unset( $endpoints['/wp/v2/block-directory/search'] );
+		return $endpoints;
 	}
 }


src/MasterSite.php
```diff
@@ -195,6 +195,18 @@ function ( $block_content, $block ) {
 		add_action( 'admin_notices', [ $this, 'show_dashboard_notice' ] );
 		add_action( 'wp_ajax_dismiss_dashboard_notice', [ $this, 'dismiss_dashboard_notice' ] );
 		add_filter( 'timber/twig', [ $this, 'p4_optimize_img_url' ] );
+
+		// Disable WordPress(WP5.5) Block Directory.
+		global $wp_version;
+		if ( version_compare( $wp_version, '5.5', '>=' ) ) {
```
comzeradd, line 201: Just to avoid the check, I'm thinking that we could create a new `wp55` branch out of master and merge this there. Because I think we may have other similar code changes that need to happen before we upgrade to 5.5

sagarsdeshmukh: The `wp55` branch is already exist, but far behind of `master`. On merging of this PR, all `master` branch changes will automatically add in `wp55` branch.

comzeradd: Probably created that branch a long time :) I just rebased it and commits got cleared.

src/MasterSite.php
```diff
@@ -195,6 +195,15 @@ function ( $block_content, $block ) {
...
 		add_filter( 'timber/twig', [ $this, 'p4_optimize_img_url' ] );
+
+		// Disable WordPress(WP5.5) Block Directory.
+		add_action(
+			'plugins_loaded',
+			function() {
+				remove_action( 'enqueue_block_editor_assets', 'wp_enqueue_editor_block_directory_assets' );
+				remove_action( 'enqueue_block_editor_assets', 'gutenberg_enqueue_block_editor_assets_block_directory' );
+			}
+		);
```
lithrel, line 206: OK, I can't make the block directory work for some reason (it makes an OPTION request to its endpoint when I type in the box and that's all, no search after that); but I can see the script `block-directory.js` still loaded in the network tab.
From what I've read and tested, you don't need a hook to [remove an action from a theme](https://github.com/WordPress/gutenberg/issues/23961#issuecomment-688957626), and I can't see where the `gutenberg_enqueue_block_editor_assets_block_directory` is used, I can only assume it is specific to the Gutenberg repo and transformed at some point, but not included in the Wordpress install. In the end using only this one (without `add_action()`) seems to work:
```php
remove_action( 'enqueue_block_editor_assets', 'wp_enqueue_editor_block_directory_assets' );
```
Which version are you on ? I just installed 5.5.3 so maybe we are on a different one (5.5.2 was just a few days before)

sagarsdeshmukh: Hi @lithrel - Thanks for the review :) . I am using the same WPv5.5.3 version. Could you try commenting the `define('DISALLOW_FILE_MODS', true );` line in `wp-config.php` file.

I updated the PR with above changes. please have a look.

lithrel: > Could you try commenting the `define('DISALLOW_FILE_MODS', true );` line in `wp-config.php` file.

That was it, thanks !

lithrel (approved):
:+1:

PR #1223, opened 2020-11-04
PLANET-5325 Take Action boxout A/B test

mleray:
### Description

See https://jira.greenpeace.org/browse/PLANET-5325

- **Remove external link markup**: we don't want to show the external link markup in Take Action covers, including Take Action boxouts. In order to achieve this, we need to update the selector as Take Action boxouts are only in posts (that don't have the `page-template` class).
- **Update post content z-index value**: this is to prevent the Take Action boxout going under the author block and comment form on mobile (scrolling version)

Related gutenberg-blocks PR: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/364

### DIFF
--- a/assets/src/js/external_links.js
+++ b/assets/src/js/external_links.js
@@ -10,7 +10,7 @@ export const setupExternalLinks = function($) {
       </g>
   </svg>`;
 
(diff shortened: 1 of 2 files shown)


lithrel (approved):
:+1:

PR #1224, opened 2020-11-04
PLANET-5644 Remove text transform on buttons

mleray:
### Description

See https://jira.greenpeace.org/browse/PLANET-5644
We also need to update the button default texts to make sure they follow the new format!

#### Testing
See the fixed version here: https://www-dev.greenpeace.org/test-jupiter/

#### Related PRs:
- styleguide: https://github.com/greenpeace/planet4-styleguide/pull/81
- gutenberg-blocks: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/426


templates/tease-search.twig
```diff
@@ -121,7 +121,7 @@
 							data-ga-category="Search Results"
 							data-ga-action="Call to Action"
 							data-ga-label="{{ ga_page_type }}">
-								{{ link_text ?: __( 'take action', 'planet4-master-theme' ) }}
+								{{ link_text ?: __( 'Take action', 'planet4-master-theme' ) }}
```
Inwerpsel, line 124: If we change the case here, these won't be translated anymore, since the translation files still have the old case.
https://github.com/greenpeace/planet4-master-theme/blob/master/languages/planet4-master-theme-nl_NL.po#L305-L306

mleray: Yep, I know. I guess we need to update the translations files then? Not sure how that works 🤔

Inwerpsel: Indeed, you could include those changes in this PR. I guess this shouldn't cause issues for the loco plugin, but maybe we should double check this to be sure? cc @comzeradd

Inwerpsel: We probably also need to update all translations too, as they all seem to be completely in lowercase (inc first letter) too. https://github.com/search?p=8&q=org%3Agreenpeace+%22msgid+take+action%22&type=Code

comzeradd: I can add a commit here to update the translation files

Inwerpsel: I set up the test instance with some languages so we can test others than just English. Also added WPML https://github.com/greenpeace/planet4-test-jupiter/blob/develop/composer-local.json

comzeradd: I pushed a new commit with translation files updated. @mleray you have to pull it locally too, if you plan in making any further changes to this branch

sagarsdeshmukh: Hi @Inwerpsel - You could also [sync data](https://app.gitbook.com/@greenpeace/s/planet4/ci-cd/test-environments#data-synchronization) from NRO site.

mleray: We also released these changes in all dev instances, so different languages can be tested these 🙂

comzeradd (approved):

PR #1230, opened 2020-11-10
PLANET-5503 Remove hammer and slick CSS and JS resources

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-5503

* Hammer is only used by a block which also performs this enqueue. (We do need to check if none of the NRO code relies on it being there.)
* Slick is added as a dependency by the blocks plugin, but the only usage it has is currently not working. So we can remove both CSS and JS files of it until we have fixed the usage. We need to merge https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/434 first for that, to remove the dependency. (currently causes failing tests anyway)

### DIFF
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -691,7 +691,6 @@ class MasterSite extends TimberSite {
 		wp_enqueue_script( 'main' );
 
 		wp_enqueue_script( 'slick', 'https://cdnjs.cloudflare.com/ajax/libs/slick-carousel/1.9.0/slick.min.js', [], '1.9.0', true );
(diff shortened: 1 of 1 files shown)


pablocubico (dismissed):
LGTM. I noticed this on friday while working on the Carousel Header, we can probably remove slick too, since it's only used (as far as I recall) in the Covers block, via the `pubslider.js` script. 

Maybe we could enqueue it conditionally as GPNL does:

https://github.com/greenpeace/planet4-gpnl-plugin-gutenberg-blocks/blob/develop/classes/blocks/class-heroimage.php#L74

pablocubico (approved):
Looks good! Less libs 🙌

PR #1232, opened 2020-11-11
PLANET-5055 Pinned the ElasticPress Search algorithm to v3.4

sagarsdeshmukh:
[JIRA 5055](https://jira.greenpeace.org/browse/PLANET-5055)

The [ElasticPress v3.5](https://github.com/10up/ElasticPress/releases) version includes a revamp of the search algorithm, which probably breaks the backwards compatibility. 

After discussion with @Inwerpsel , we agree on to pin the EP search algorithm to older version (v3.4) and we will test & evaluate the new algorithm changes in separate ticket.

Related PR # [102](https://github.com/greenpeace/planet4-base-fork/pull/102)

### DIFF
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -195,6 +195,9 @@ class MasterSite extends TimberSite {
 		add_action( 'admin_notices', [ $this, 'show_dashboard_notice' ] );
 		add_action( 'wp_ajax_dismiss_dashboard_notice', [ $this, 'dismiss_dashboard_notice' ] );
 		add_filter( 'timber/twig', [ $this, 'p4_optimize_img_url' ] );
+
+		// Pin the ElasticPress to v3.4 search algorithm.
+		simple_value_filter( 'ep_search_algorithm_version', '3.4' );
 	}
 
 	/**


sagarsdeshmukh:
The phpunit tests fails, because of the `is_plugin_active ()` function executes before loading of `wp-admin/includes/plugin.php` file.

Here we have 2 options -
1. Add `include_once( ABSPATH . 'wp-admin/includes/plugin.php' );` line before the function call
2. Move this filter code to blocks plugin

Any suggestions 🤔

src/MasterSite.php
```diff
@@ -195,6 +195,16 @@ function ( $block_content, $block ) {
 		add_action( 'admin_notices', [ $this, 'show_dashboard_notice' ] );
 		add_action( 'wp_ajax_dismiss_dashboard_notice', [ $this, 'dismiss_dashboard_notice' ] );
 		add_filter( 'timber/twig', [ $this, 'p4_optimize_img_url' ] );
+
+		// Pin the ElasticPress to v3.4 search algorithm.
+		if ( \is_plugin_active( 'elasticpress/elasticpress.php' ) ) {
```
Inwerpsel, line 200: I think we can get around the issue by simply not doing the `is_plugin_active` check. The `ep_search_algorithm_version` filter is exclusive to the plugin, adding it without the plugin loaded shouldn't cause any issues.

src/MasterSite.php
```diff
@@ -195,6 +195,16 @@ function ( $block_content, $block ) {
...
+			add_filter(
+				'ep_search_algorithm_version',
+				function() {
+					return '3.4';
```
Inwerpsel, line 204: We have a wrapper function around `add_filter` for in case the function only returns a static value, without needing to write the function.

```
simple_value_filter( 'ep_search_algorithm_version', '3.4');
```

Inwerpsel (approved):
Looks good :+1:

PR #1235, opened 2020-11-12
PLANET-5619 Restrict CF plugin

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-5619

* Remove any actions that would trigger a full purge.
* Remove settings page. 1) from the menu 2) because removing from menu
still leaves the page accessible, add a function to the same action with
a higher priority.

Deploying it now on a test instances without the CF plugin first as we'll probably do that for prod too.

I'm currently verifying if the plugin will function correctly by just putting the right credentials into `wp_options` (likely it will).

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -114,6 +114,23 @@ add_action(
 	}
 );
 
+// Ensure no actions trigger a purge everything.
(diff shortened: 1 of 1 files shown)


comzeradd (approved):
:+1:

PR #1236, opened 2020-11-12
PLANET-5585: Normalize action search results appearance with other results

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-5585
Ref: https://jira.greenpeace.org/browse/PLANET-4374

> Background images for Take Action pages on search results are pixelated and make the text hard to read

This PR makes the actions in a search result look like the other results.  
It also fixes general search results layout.

## Fix 

- Suppression of some exception rules and markup specific to action results.  
- Added a margin-top to the button so it's not stuck under the text.
- Fixed some margins, paddings, width and markup so that it fits the layout given in https://jira.greenpeace.org/browse/PLANET-4374

fix-search-results

## Test

- Run a search on Content type `Action` : https://www-dev.greenpeace.org/test-uranus/?s=&orderby=_score&f[ctype][Action]=0
- Results should fit [the mockup](https://jira.greenpeace.org/browse/PLANET-5585); no background image, a thumbnail if available, and an action button under it.

### DIFF
--- a/assets/src/scss/pages/search/_search-results.scss
+++ b/assets/src/scss/pages/search/_search-results.scss
@@ -23,26 +23,11 @@
   }
 
   ul {
-    border-top: 1px solid $grey-20;
-
-    @include small-and-up {
-      border-top: 0;
-    }
-
     li {
-      padding: 15px 0;
-      margin: 0 $n15;
+      padding: 0 24px 32px 24px;
+      margin: 0 auto 24px;
       border-bottom: 1px solid $grey-20;
       position: relative;
-
-      @include small-and-up {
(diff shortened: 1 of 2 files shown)


mleray (approved):
Looks good! 👍 One thing I noticed on the test instance is that the links just reload the page instead of going to the post, which is strange but it seems unrelated since it also happens in other test instances 🤷‍♀️

lithrel:
Ah right, got it locally too, I think it's when Elasticsearch is not ready or hasn't indexed anything yet, we don't handle this very well :grimacing:  I'll open another ticket for that

PR #1240, opened 2020-11-23
PLANET-5619 Command to update CF key from config

Inwerpsel:
The Cloudflare plugin uses this key from the db. We'll put it in wp-config during deploy (https://github.com/greenpeace/planet4-builder/pull/62 https://github.com/greenpeace/planet4-helm-wordpress/pull/13/files), so that it can be put into the db in a post deploy script (https://github.com/greenpeace/planet4-base-fork/pull/104/files) which uses the command introduced in this PR.

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -154,11 +154,32 @@ final class Loader {
 			return;
 		}
 
-		$command = static function ( $args, $assoc_args ) {
+		$activator_command = static function ( $args, $assoc_args ) {
 			Activator::run();
 		};
+		WP_CLI::add_command( 'p4-run-activator', $activator_command );
 
-		WP_CLI::add_command( 'p4-run-activator', $command );
+		/**
+		 * Put the CF API key into the options table, where the CF plugin uses it from.
+		 */
+		$put_cf_key_in_db = static function ( $args ) {
+			$hostname = $args[0];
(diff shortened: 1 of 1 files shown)


src/Loader.php
```diff
@@ -154,11 +154,28 @@ public function load_commands() {
...
+
+		/**
+		 * Put the CF API key into the options table, where the CF plugin uses it from.
+		 */
+		$put_cf_key_in_db = static function ( $args ) {
+			$hostname = $args[0];
+			if ( empty( $hostname ) ) {
+				WP_CLI::error( 'Please specify the hostname.' );
+			}
+
```
lithrel, line 170: We could add an error if `CLOUDFLARE_API_KEY` is undefined (assuming activating the following options would cause some problems with a key missing)

lithrel (approved):
:+1:

src/Loader.php
Inwerpsel: @lithrel Added the check :+1: A missing/wrong/expired key should only result in the purge request failing.

Inwerpsel: https://github.com/cloudflare/Cloudflare-WordPress/blob/91ba8ddd4c9478d1615e5067d0759e4af6e8849e/src/WordPress/Hooks.php#L168-L174

PR #1242, opened 2020-11-25
PLANET-5286 Hide all details on password protected post/page/campaigns

sagarsdeshmukh:
[JIRA 5286](https://jira.greenpeace.org/browse/PLANET-5286)

**Task -**
- Hide all details on password protected post/page/campaigns

**Testing -**
- Create new page, post, Campaign
- Make it password protected and check on frontend, the page should only display the add password form and submit button
- Check the page source and page meta tags (page title, open graph fields etc)
- Test with adding page header fields and changing the page template

### DIFF
--- a/page.php
+++ b/page.php
@@ -69,6 +69,10 @@ $context['post_tags']           = implode( ', ', $post->tags() );
 $context['custom_body_classes'] = 'brown-bg ';
 
 if ( post_password_required( $post->ID ) ) {
+
+	// Hide the page title from links to the extra feeds.
+	remove_action( 'wp_head', 'feed_links_extra', 3 );
+
 	$context['login_url'] = wp_login_url();
 
 	Timber::render( 'single-page.twig', $context );
--- a/single-campaign.php
+++ b/single-campaign.php
@@ -116,6 +116,10 @@ foreach ( range( 1, 5 ) as $i ) {
 }
 
 if ( post_password_required( $post->ID ) ) {
+
+	// Hide the campaign title from links to the extra feeds.
+	remove_action( 'wp_head', 'feed_links_extra', 3 );
+
 	$context['login_url'] = wp_login_url();
 
 	Timber::render( 'single-password.twig', $context );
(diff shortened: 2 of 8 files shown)


lithrel (commented):
Testing with post, I can see the post in 3 other places:
- The title appears in comments link `<link rel="alternate" type="application/rss+xml"` in the html header; I think we have to alter args of `feed_links_extra()` with action `wp_head`
- The post appears in our Articles block `Recent articles` with title and excerpt; we probably need to add filters to `Articles::get_posts()`
- I can also get the title with an ajax request to `https://www.planet4.test/wp-json/wp/v2/posts/{postId}`; not sure how we should modify this response, there are other sensitive informations like created time, last modified datetime, author id; instead of finding what to filter we could send a response WP_Error with status 401 (not sure where to put this one, `WP_REST_Posts_Controller` has functions like `check_read_permission()` but no filter or action in there :/)

sagarsdeshmukh:
> Testing with post, I can see the post in 3 other places:
> 
> * The title appears in comments link `<link rel="alternate" type="application/rss+xml"` in the html header; I think we have to alter args of `feed_links_extra()` with action `wp_head`
> * The post appears in our Articles block `Recent articles` with title and excerpt; we probably need to add filters to `Articles::get_posts()`
> * I can also get the title with an ajax request to `https://www.planet4.test/wp-json/wp/v2/posts/{postId}`; not sure how we should modify this response, there are other sensitive informations like created time, last modified datetime, author id; instead of finding what to filter we could send a response WP_Error with status 401 (not sure where to put this one, `WP_REST_Posts_Controller` has functions like `check_read_permission()` but no filter or action in there :/)

Thank @lithrel for the review 👍 
Good catch about 1 & 3rd point, I added a filter which hide post title in comments link meta.
The articles block listing a password protected post issue is addressed in [separate ticket ](https://jira.greenpeace.org/browse/PLANET-5726) and PR at [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/447).  
About the REST API, it will need a hack into default functionality, so we could go with as it is.

lithrel (approved):
> About the REST API, it will need a hack into default functionality, so we could go with as it is.

Good to me :+1:  as I said before, it's all a matter of what the person proposing the feature has in mind. In the meantime, this is still better than what we have at the moment so let's go with this :)

PR #1243, opened 2020-12-01
Remove preload and line break in href

Inwerpsel:
* This line break caused an issue with Cloudflare's APO, which inlines
Google Fonts for better performance. They clearly didn't expect a
newline in there and did a bad request to Google resulting in a 400.
Cloudflare then proceeded to include that 400 response in the page, as
though it was the expected styles. This meant a html document was inside
of a style tag in the head, breaking the site very badly.
* Remove preload as APO doesn't remove it when inlining the google fonts
response.

### DIFF
--- a/templates/google-fonts.twig
+++ b/templates/google-fonts.twig
@@ -1,1 +1,0 @@
-https://fonts.googleapis.com/css?family=Roboto:300,400,500,700,900|Lora:400,400i,700&display=swap&subset=latin-ext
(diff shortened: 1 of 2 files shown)


pablocubico (approved):
Looks good. Couldn't test the APO setup but at least the URL seems fine. :)

comzeradd (approved):
Nice catch 🍻

PR #1244, opened 2020-12-01
PLANET-5685 Add auto test instance booking and deploy, use test instance for lighthouse check

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-5685

To get reliable performance test results we need to run Lighthouse we need to deploy the code to a test instance, since these have the same setup (Cloudflare) as production. Automatically deploying each commit to a branch of an open PR allows this, and additionally helps ensuring our test instances are always on the latest version of the code.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -242,9 +242,15 @@ jobs:
               exit 1
             fi
       - run:
-          name: "Call webhook to get free instance (mock for now). If an instance is already on the branch, use that one."
+          name: "Python script deps"
           command: |
-            echo oberon > /tmp/workspace/test-instance
+            pip install requests requests_oauthlib
+      - run:
+          name: "Book instance"
+          command: |
+            JIRA_USERNAME=planet4 python /home/circleci/book-test-instance.py \
+            --pr-url $CIRCLE_PULL_REQUEST \
+            --results /tmp/workspace/booking.json >/tmp/workspace/test-instance
       - run: activate-gcloud-account.sh
       - run:
           name: "Put zip in cloud storage"
@@ -260,6 +266,7 @@ jobs:
           root: /tmp/workspace
           paths:
             - test-instance
+            - booking.json
 
   lighthouse:
     executor: lighthouse-check/default
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,31 +1,29 @@
 ---
 version: 2.1
 
+orbs:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -4,10 +4,20 @@ version: 2.1
 workflows:
   main:
     jobs:
-      - php72-tests
-      - php73-tests
-      - acceptance-tests
-      - a11y-tests
+      # Get new instance if none was used for the PR, else figure out which instance.
```
Inwerpsel, line 7: I replaced the other jobs with stubs that finish immediately, to not bogart concurrent jobs.

.circleci/config.yml
```diff
@@ -205,12 +225,65 @@ jobs:
...
     steps:
-      - run: apk add curl zip git openssh-client jq python2 make g++
+      - attach_workspace:
+          at: /tmp/workspace
+      - run:
+          name: "Check if this is for an open PR."
+          command: |
+            if [ -z "$CIRCLE_PULL_REQUEST" ];then
+              echo echo "No PR, skipping instance deploy"
+              exit 1
```
Inwerpsel, line 242: Still need to determine what we do for this job when there is no open PR. Probably best to make it fail?

lithrel: can we do a `step halt` or something that would stop this process without failing the whole workflow ?

Inwerpsel: Wouldn't `step halt` make the job pass and start the next jobs too?

lithrel: Absolutely no idea, didn't check the docs yet :D

Inwerpsel: But the pipeline cannot really pass without these tests, so I guess failure is the right status? If you push a branch without a PR it will fail on that job, but you can create a PR and re-run from fail and it should work.

Inwerpsel: Found more info on `halt` on the forum than in the docs btw https://discuss.circleci.com/t/anyway-to-terminate-circleci-entirely/27486

Inwerpsel: We could try allowing `request-instance` to fail :thinking:

Inwerpsel: Though it would probably still show the workflow as failed in GitHub

lithrel (approved):
:+1:

PR #1245, opened 2020-12-02
PLANET-5726 Hide password protected content on post listing & search page

sagarsdeshmukh:
[JIRA 5726](https://jira.greenpeace.org/browse/PLANET-5726)

**Task** -
- Hide password protected content from post listing pages like taxonomy, author, p4 post type pages etc
- Hide password protected content from search result page

**Blocker** -
 - ~~The elasticPress plugin won’t provide option/support to hide password protected content from search result(quite surprising)~~

I found some discussion and open PR which offer the solution , but it will take time to get approval and merging -
- https://github.com/10up/ElasticPress/pull/1789
- https://github.com/10up/ElasticPress/pull/1913

~~Till the time we could consider some temporary solution-~~
- ~~We already have a meta field check which skip the posts from search result ( `p4_do_not_index` [meta field](https://github.com/greenpeace/planet4-master-theme/blob/master/src/Search.php#L463))~~

~~Any suggestions or if I miss something~~

The solution suggested by @Inwerpsel works like a charm. Thanks 🚀

### DIFF
--- a/src/Search.php
+++ b/src/Search.php
@@ -450,6 +450,7 @@ abstract class Search {
 			'no_found_rows'  => true,
 			'post_type'      => self::get_post_types(),
 			'post_status'    => [ 'publish', 'inherit' ],
+			'has_password'   => false,  // Skip password protected content.
 		];
 
 		if ( $paged > 1 ) {
--- a/src/Search.php
+++ b/src/Search.php
@@ -236,6 +236,12 @@ abstract class Search {
 			10,
 			1
 		);
+		add_filter(
+			'ep_post_query_db_args',
+			[ self::class, 'hide_password_protected_content' ],
+			10,
+			1
+		);
 	}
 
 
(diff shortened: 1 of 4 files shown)


src/Search.php
```diff
@@ -1127,8 +1143,27 @@ public static function exclude_unwanted_attachments( $args ) {
...
+		return $unwanted_attachment_ids;
+	}
 
-		return $args;
+	/**
+	 * Fetch all password protected posts that we don't want to include in search.
+	 *
+	 * @return array The password protected post ids.
+	 */
+	public static function get_password_protected_post_ids():array {
```
Inwerpsel, line 1154: For attachments we needed to execute another query, because wp_query didn't have a way to specify the post_mime_type in the args, so we had to add the first query. However for post password we probably don't need to do that and can add it directly to the query.
```
public static function exclude_unwanted_posts( $args ) {
   $args['has_password'] => false,
}
```
see https://stackoverflow.com/questions/7538959/how-to-exclude-password-protected-posts-in-wordpress-loop

Inwerpsel (changes requested):
We don't need the additional query, which could affect performance, but can directly add post_password = false to the arguments.

sagarsdeshmukh:
> We don't need the additional query, which could affect performance, but can directly add post_password = false to the arguments.

Yes, make sense. 💯 I updated the PR, please have a look.

Inwerpsel (approved):
Looks good :+1:

PR #1250, opened 2020-12-04
PLANET-5753: Replace jQuery usage from author.js

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-5753
Ref: https://jira.greenpeace.org/browse/PLANET-5663

`author.js` script is used to change the appearance of a link when hovering a related image, by toggling a class name.
We can replace this by a CSS rule.

Good to note that the effect on hovering the image is different from the effect hovering the title itself, I didn't change this, I don't know if it's on purpose.

## Fix

- Remove `author.js` script and duplicated code in `search.js`
- Add a CSS rule to reach sibling element

## Test

- Run a search
- On the results, when hovering an image, the title of the result should be underlined


assets/src/scss/pages/search/_search-results.scss
```diff
@@ -61,6 +57,10 @@
           width: max-content;
         }
       }
+
+      .search-result-item-image:hover + * .search-result-item-headline {
```
Inwerpsel, line 61: One tiny drawback is that `+` selectors depend on the elements being next to each other, which could be easy to overlook and break when changes to the markup are made. I guess using `~` is a little more robust. But then again changing the markup should involve checking dependent CSS, so feel free to ignore :shrug:

Inwerpsel (approved):
Nice fix :+1:

assets/src/scss/pages/search/_search-results.scss
lithrel: Makes sense, specially for a feature easy to overlook :ok_hand:

PR #1254, opened 2020-12-09
PLANET-5781: Update the P4 message board with the Holiday party

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-5781

> change the message in the P4 message board to advertise the holiday party

Screenshot from 2020-12-09 15-55-55

## Code

Update of message and increment of message version to trigger reappearance of the dashboard message for every user.

## Note

The revert of this message for [5782](https://jira.greenpeace.org/browse/PLANET-5782) won't need a version increment, as it should stay hidden for people who discarded it before.

### DIFF
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -30,7 +30,7 @@ class MasterSite extends TimberSite {
 	 *
 	 * @var string
 	 */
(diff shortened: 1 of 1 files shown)


comzeradd (approved):
That was quick! :partying_face: :balloon:

PR #1255, opened 2020-12-10
PLANET-5162 Fix author override value in RSS feed

sagarsdeshmukh:
Ref: [JIRA 5162](https://jira.greenpeace.org/browse/PLANET-5162)

---
**Issue :**
- The RSS feed not showing the author override value in place of author name (`dc:creator`) in RSS feed.

**Testing:**
- Add post with author override value
- After checkout of this branch goto RSS feed link(https://www.planet4.test/feed/)
- The post feed will show the author override value at author name


mleray (approved):
Looks good to me 👍

PR #1256, opened 2020-12-10
PLANET-5395: Some ElasticPress settings can not be saved

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-5395

> ~Due to a bug in the ElasticPress plugin~ we're not able to save EP settings. There seems to be a missing html attribute that is used to send the update request.

Our html tag filter overwrites the elasticpress plugin own filters (and any other :grimacing:).

The tag attribute filter is too widely applied, and should only concern `post` content.
`MasterSite::set_custom_allowed_attributes_filter` only concerns `post` context. It is mainly used with the `wp_kses_post()` function.

## Fix 

- Passing `context` parameter to our filter
- Limiting filter to `post` context (ElasticPress uses its own context `ep-html`)

## Test

Test ElasticPress dashboard before and after fix:
- enable/disable radio boxes are mutually exclusive (can't check both at the same time)
- saving features status work as expected


sagarsdeshmukh (approved):
Nice work @lithrel , Tested locally. PR LGTM 👍

PR #1257, opened 2020-12-10
PLANET-5387 Campaigner User Role missing from Author dropdown field

dpivo:
Ref: https://jira.greenpeace.org/browse/PLANET-5387

---

Added author filter to ensure the users with campaigner role are included in dropdown

### DIFF
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -118,6 +118,7 @@ class MasterSite extends TimberSite {
 		add_action( 'admin_menu', [ $this, 'add_restricted_tags_box' ] );
 		add_action( 'do_meta_boxes', [ $this, 'remove_default_tags_box' ] );
 		add_action( 'pre_insert_term', [ $this, 'disallow_insert_term' ], 1, 2 );
+		add_filter( 'wp_dropdown_users_args', [ $this, 'filter_authors' ], 10, 1 );
 		add_filter( 'wp_image_editors', [ $this, 'allowedEditors' ] );
 		add_filter(
 			'jpeg_quality',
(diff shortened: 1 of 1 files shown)


src/MasterSite.php
```diff
@@ -118,6 +118,7 @@ protected function hooks() {
 		add_action( 'admin_menu', [ $this, 'add_restricted_tags_box' ] );
 		add_action( 'do_meta_boxes', [ $this, 'remove_default_tags_box' ] );
 		add_action( 'pre_insert_term', [ $this, 'disallow_insert_term' ], 1, 2 );
+		add_action( 'wp_dropdown_users_args', [ $this, 'filter_authors' ] );
```
sagarsdeshmukh, line 121: The `wp_dropdown_users_args` is a [filter](https://developer.wordpress.org/reference/hooks/wp_dropdown_users_args/), so it will be use with `apply_filters ()` function.

src/MasterSite.php
```diff
@@ -733,6 +734,19 @@ public function view_meta_box_search( $post ) {
...
 	}
 
+	/**
+	 * Applies filters for list of users in dropdown
+	 *
+	 * @param null|Array $args The filter options and values.
+	 */
+	public function filter_authors( $args ) {
+		if ( isset( $args['who'] ) ) {
+			$args['role__in'] = [ 'administrator', 'author', 'campaigner', 'contributor', 'editor', 'subscriber' ];
```
sagarsdeshmukh, line 744: The JIRA ticket won't have a details about showing the `subscriber` & `contributor` user in dropdown. not sure do we need confirmation on it? 🤷🏽‍♂️

comzeradd: We can safely skip `subscriber`, since we don't actively use that role and they don't have any backend permissions anyway.

comzeradd (approved):
Looks good :+1:

PR #1261, opened 2020-12-11
PLANET-5654: WP_Block_Type_Registry::register was called incorrectly

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-5654

## Problem

`register_block_type('core/image')` is called twice, first by WP, then by us to add a render callback. This triggers an error WP >= 5.5.

## Fix

Fix by hooking to the `register_block_type_args` which allows to modify block type args at register, including `render_callback`.
~The fix also works for WP < 5.5~ <sub>no it does not</sub> 
Added a version compare to use this fix only on versions >= 5.5

## Test

The render callback we add is used to add credits to images caption.
- Add an image in a post. The image has to have something in the _Credit_ field of the Attachment details.
- The credit appears at the end of the image caption on the frontend.

Screenshot from 2020-12-10 20-11-15

### DIFF
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -195,7 +195,7 @@ class MasterSite extends TimberSite {
 		add_action( 'init', [ $this, 'login_redirect' ], 1 );
 		add_filter( 'attachment_fields_to_edit', [ $this, 'add_image_attachment_fields_to_edit' ], 10, 2 );
 		add_filter( 'attachment_fields_to_save', [ $this, 'add_image_attachment_fields_to_save' ], 10, 2 );
-		add_action( 'init', [ $this, 'p4_register_core_image_block' ] );
+		add_filter( 'register_block_type_args', [ $this, 'register_core_blocks_callback' ] );
 		add_action( 'admin_notices', [ $this, 'show_dashboard_notice' ] );
 		add_action( 'wp_ajax_dismiss_dashboard_notice', [ $this, 'dismiss_dashboard_notice' ] );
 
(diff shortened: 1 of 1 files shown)


src/MasterSite.php
```diff
@@ -195,7 +195,7 @@ function ( $block_content, $block ) {
 		add_action( 'init', [ $this, 'login_redirect' ], 1 );
 		add_filter( 'attachment_fields_to_edit', [ $this, 'add_image_attachment_fields_to_edit' ], 10, 2 );
 		add_filter( 'attachment_fields_to_save', [ $this, 'add_image_attachment_fields_to_save' ], 10, 2 );
-		add_action( 'init', [ $this, 'p4_register_core_image_block' ] );
+		add_filter( 'register_block_type_args', [ $this, 'register_core_blocks_callback' ] );
```
Inwerpsel, line 198: How can this filter work for WP < 5.5 if it was [only added in 5.5](https://developer.wordpress.org/reference/hooks/register_block_type_args/)? :thinking:

lithrel: Oh, because I'm extremely dumb and tested on the wrong version :facepalm:

comzeradd: We can merge it to the `wp66` branch if that helps

lithrel: There's a fix for this on `wp55` https://github.com/greenpeace/planet4-master-theme/blob/wp55/src/MasterSite.php#L1099 that is not on `wp56` https://github.com/greenpeace/planet4-master-theme/blob/wp56/src/MasterSite.php#L1099, we might be missing some stuff or rebase ? @Inwerpsel do you have preferences on the fix ? I feel unregister/register is more cumbersome, but it might also be more efficient because it's probably executed less times in a process than `register_block_type_args`

Inwerpsel: @lithrel your fix makes a lot more sense than the one on `wp55`, I wasn't aware of that new filter when I made that.

I think the easiest for us would be to keep the current `register_block_type` filter, but wrap it in a WP version check (WP < 5.5). We can already add the new filter, since it won't be called in WP < 5.5.

That way we can merge to master, don't have this change pending, and can switch WP versions locally without needing to check out a different branch of one of the repos.

As for efficiency, it gets called just once per request, and doesn't do anything expensive. If you want to be sure you can measure it, but I'm pretty sure it's negligible.

comzeradd: Just to avoid confusion, I rebased [wp56](https://github.com/greenpeace/planet4-master-theme/tree/wp56) with all the relevant commits, and deleted `wp55`.

lithrel:
Added a version compare to trigger one or the other register hook; to mitigate for my dumbness I also deployed on https://www-dev.greenpeace.org/test-titan/story/28961/test-image/ which is on 5.4

Inwerpsel (approved):

PR #1263, opened 2020-12-14
PLANET-5759 Remove unused img_align script

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-5759

---

This one seems unused. `wp-caption` was probably a default WP class in the pro-Gutenberg era.


lithrel (approved):
:+1:

PR #1265, opened 2020-12-16
PLANET-5782: Restore the P4 message board after the Holiday party announcement

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-5782

> Once the Holiday party is over (December 18th), we should put back the original message  (well, slightly adjusted)

## Fix 

Made the message time dependent, as we might not deploy anything during the holidays.

## Test

You can play with the `over_date` locally to see the dashboard message change, if `over_date` is in the past.

- Before `over_date`: Party message
- Starting on `over_date`: Regular slack message

### DIFF
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -2,6 +2,7 @@
 
 namespace P4\MasterTheme;
 
+use DateTimeImmutable;
 use Timber\Timber;
 use Timber\Site as TimberSite;
 use Timber\Menu as TimberMenu;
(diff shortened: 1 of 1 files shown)


src/MasterSite.php
```diff
@@ -1195,21 +1196,48 @@ public function show_dashboard_notice(): void {
...
-				<li>🎶 <a href="https://www.youtube.com/playlist?list=PLRWtdTW6NaMUAxoBmV1baj7CStIIhuXtc&jct=DoU-zKSovfpKY5InMOOQTOiYwM6p7g" target="_blank">Here\'s a YouTube playlist for everyone to add their favourite songs</a></li>
-			</ul>
-		</p>
-		<p>☮️  🥂 💚</p>';
+		$now           = new \DateTimeImmutable();
+		$over_date     = new \DateTimeImmutable( '2020-12-19' );
+		$party_is_over = $now >= $over_date;
+
+		return $party_is_over
+			? '<h2>Welcome to the new P4 message board!</h2>
```
Inwerpsel, line 1209: Maybe this is not the best place to use a ternary :sweat_smile: We could instead have 2 return statements
```
if (!$party_is_over) { return 'party string'; }

return 'normal string';
```
Then the next deploy after we just need to remove the party part, and don't need to modify the lines with the normal return value.

lithrel: Good point :+1:

Inwerpsel (approved):
:birthday:

PR #1266, opened 2020-12-16
Split title into a new template block

comzeradd:
That would make it easier for child themes that just want to override that part only.


sagarsdeshmukh (approved):
👍

Inwerpsel (approved):
:+1: Looks cleaner too. Though we should probably discourage NROs from overriding these templates unless there's no other way, since each change we make will need to be propagated to all overriding templates, and it's easy to miss doing so.

PR #1267, opened 2020-12-24
PLANET-5680 fix static call to search function

dpivo:
Ref: https://jira.greenpeace.org/browse/PLANET-5680

---

Changed function to static function to fix calling non static function error


lithrel (approved):
:+1:

PR #1269, opened 2021-01-04
PLANET-5782: Restore the P4 message board

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-5782

Removing holiday party message.
Message version is not incremented so that people who dismissed it previously won't see it again.

Screenshot from 2021-01-04 12-46-26


sagarsdeshmukh (approved):
👍

PR #1270, opened 2021-01-07
PLANET-5434 Remove php-7.2

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-5434

php-8 job is not added yet, but since we already switched to 7.3 no need to test for a previous version.


lithrel (approved):
:+1:

PR #1272, opened 2021-01-07
PLANET-5820 Anti flicker selector insteadof toggle

Inwerpsel:
* The toggle controlled whether to hide everything before the test is
applied, which harmed performance.
* Instead provide an option for a selector that will be hidden before
the test is ready.
* Always run the snippet since we're not hiding anything unless
specifically configured.
* Rename `.async-hide` class to something more accurate.
* Provide `.ab-hide` class which will be hidden while loading.

Ref: https://jira.greenpeace.org/browse/PLANET-5820

Tested and UAT approved on https://www-dev.greenpeace.org/test-deimos . It's set up there with the take action boxout selector, and Julia confirmed that the tests ran succesfully.

### DIFF
--- a/templates/html-header.twig
+++ b/templates/html-header.twig
@@ -7,6 +7,13 @@
 	{% include 'blocks/meta_fields.twig' %}
 
 	{% if google_tag_value %}
+		{% if ab_hide_selector %}
+		<style>
+			.google-optimize-loading {{ ab_hide_selector }} {
+				opacity: 0 !important;
+			}
+		</style>
+		{% endif %}
 		<script>
 			var google_tag_value = '{{ google_tag_value }}';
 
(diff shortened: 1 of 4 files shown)


templates/html-header.twig
```diff
@@ -7,6 +7,13 @@
 	{% include 'blocks/meta_fields.twig' %}
 
 	{% if google_tag_value %}
+		{% if ab_hide_selector %}
+		<style>
+			.google-optimize-loading {{ ab_hide_selector }} {
```
Inwerpsel, line 12: This does allow entering things that mess up the HTML, but that is also the case for other existing fields (like google_tag_value).

mleray (approved):
Looks good to me 👍

PR #1278, opened 2021-01-14
Remove anti-flicker snippet setting

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-5820

This should have been removed in the previous PR, but was overlooked.


mleray (approved):
Looks good to me 👍

PR #1285, opened 2021-01-21
PLANET-5897: Add HTTP Header class that sends CSP and X-Frame headers

stduerre:
Ref: https://github.com/greenpeace/planet4/issues/132

Adds an HTTP Header class that sends two headers by default:
`Content-Security-Policy: frame-ancestors 'self'`
`X-Frame-Options: SAMEORIGIN`

In addition, it contains a filter hook to allow modification of the allowed frame ancestors list in the `Content-Security-Policy` header. In practice, the filter can be used to whitelist domains of 3rd party websites to embed Planet4 content.
When the filter is used to add allowed frame ancestors, the `X-Frame-Options` header is not sent. 

This PR is intended to replace sending of the `X-Frame-Options: SAMEORIGIN` header [directly from NGINX](https://github.com/greenpeace/planet4-docker/blob/develop/src/planet-4-151612/openresty/etc/nginx/conf.d/security.conf#L8). The config for NGINX has to be altered separately.

### DIFF
--- a/src/HttpHeaders.php
+++ b/src/HttpHeaders.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace P4\MasterTheme;
+
+/**
+ * Class HttpHeaders
+ */
+class HttpHeaders {
+	/**
+	 * Headers constructor.
+	 */
+	public function __construct() {
+		add_action( 'wp_headers', [ $this, 'send_content_security_policy_header' ], 10, 1 );
+	}
+
+
+	/**
+	 * Send Content Security Policy (CSP) HTTP headers.
+	 */
+	public function send_content_security_policy_header() {
+		$allowed_frame_ancestors = array( '\'self\'' );
+
+		// Filter hook to allow modification of trusted frame ancestors.
+		$modified_allowed_frame_ancestors = apply_filters( 'planet4_csp_allowed_frame_ancestors', $allowed_frame_ancestors );
+
+		if ( ! is_array( $modified_allowed_frame_ancestors ) ) {
+			$modified_allowed_frame_ancestors = $allowed_frame_ancestors;
+		}
+
+		$directives = array(
+			'frame-ancestors' => 'frame-ancestors ' . implode( ' ', $modified_allowed_frame_ancestors )
+		);
+
+		$csp_header = 'Content-Security-Policy: ' . implode( '; ', $directives );
+		$csp_header = preg_replace( "/\r|\n/", "", $csp_header );
+
+		header( $csp_header );
+
+		// In addition, send the "X-Frame-Options" header when no other trusted frame ancestors were added through the filter.
+		if ( $modified_allowed_frame_ancestors === $allowed_frame_ancestors ) {
+			header( 'X-Frame-Options: SAMEORIGIN' );
+		}
+	}
+}
--- a/src/HttpHeaders.php
+++ b/src/HttpHeaders.php
@@ -17,7 +17,7 @@ class HttpHeaders {
 	/**
 	 * Send Content Security Policy (CSP) HTTP headers.
 	 */
-	public function send_content_security_policy_header() {
+	public function send_content_security_policy_header( $headers ): array {
 		$default_allowed_frame_ancestors = [ '\'self\'' ];
 
 		/**
@@ -36,14 +36,16 @@ class HttpHeaders {
 			'frame-ancestors ' . implode( ' ', $allowed_frame_ancestors ),
 		];
 
-		$csp_header = 'Content-Security-Policy: ' . implode( '; ', $directives );
+		$csp_header = implode( '; ', $directives );
 		$csp_header = preg_replace( "/\r|\n/", '', $csp_header );
 
-		header( $csp_header );
+		$headers[ 'Content-Security-Policy' ] = $csp_header;
 
 		// In addition, send the "X-Frame-Options" header when no other trusted frame ancestors were added through the filter.
 		if ( $allowed_frame_ancestors === $default_allowed_frame_ancestors ) {
-			header( 'X-Frame-Options: SAMEORIGIN' );
+			$headers[ 'X-Frame-Options' ] = 'SAMEORIGIN';
 		}
+
+		return $headers;
 	}
 }
--- a/src/HttpHeaders.php
+++ b/src/HttpHeaders.php
@@ -23,7 +23,7 @@ class HttpHeaders {
 		// Filter hook to allow adding trusted frame ancestors.
 		$additional_allowed_frame_ancestors = apply_filters( 'planet4_csp_allowed_frame_ancestors', [] );
 
-		$allowed_frame_ancestors = array_merge($default_allowed_frame_ancestors, (array)$additional_allowed_frame_ancestors);
+		$allowed_frame_ancestors = array_merge( $default_allowed_frame_ancestors, (array) $additional_allowed_frame_ancestors );
 
 		$directives = [
 			'default-src *',
--- a/src/HttpHeaders.php
+++ b/src/HttpHeaders.php
@@ -23,7 +23,9 @@ class HttpHeaders {
 		// Filter hook to allow adding trusted frame ancestors.
 		$additional_allowed_frame_ancestors = apply_filters( 'planet4_csp_allowed_frame_ancestors', [] );
 
-		$allowed_frame_ancestors = array_merge( $default_allowed_frame_ancestors, (array) $additional_allowed_frame_ancestors );
+		if ( is_array( $additional_allowed_frame_ancestors ) ) {
+			$allowed_frame_ancestors = array_merge( $default_allowed_frame_ancestors, $additional_allowed_frame_ancestors );
+		}
 
 		$directives = [
 			'default-src * \'self\' data: \'unsafe-inline\' \'unsafe-eval\'',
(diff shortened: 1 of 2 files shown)


src/HttpHeaders.php
```diff
@@ -0,0 +1,44 @@
...
+	 * Send Content Security Policy (CSP) HTTP headers.
+	 */
+	public function send_content_security_policy_header() {
+		$allowed_frame_ancestors = [ '\'self\'' ];
+
+		// Filter hook to allow modification of trusted frame ancestors.
+		$modified_allowed_frame_ancestors = apply_filters( 'planet4_csp_allowed_frame_ancestors', $allowed_frame_ancestors );
+
+		if ( ! is_array( $modified_allowed_frame_ancestors ) ) {
+			$modified_allowed_frame_ancestors = $allowed_frame_ancestors;
```
Inwerpsel, line 27: Since this variable is used further in the code, it's more convenient to just call it `$allowed_frame_ancestors`, and also more accurate, since it might not be modified. We can give a lengthy name to the base config instead.

Then catching the case of the filter returning an invalid value becomes more readable
```php
if ( ! is_array( $allowed_frame_ancestors ) ) {
			$allowed_frame_ancestors = $default_allowed_frame_ancestors;

File: src/HttpHeaders.php [hunk 2]
```diff
@@ -0,0 +1,44 @@
...
+	public function __construct() {
+		add_action( 'wp_headers', [ $this, 'send_content_security_policy_header' ], 10, 1 );
+	}
+
+
+	/**
+	 * Send Content Security Policy (CSP) HTTP headers.
+	 */
+	public function send_content_security_policy_header() {
+		$allowed_frame_ancestors = [ '\'self\'' ];
```
Inwerpsel, line 21: Would we add additional ancestors here, or will it always be `\self\`?

I guess there's also no reason to disallow `\self\', so maybe we shouldn't pass this to the filter, and can instead make the filter ask only for "extra" ancestors? That would in turn make the logic a bit simpler

stduerre:
The CSP header currently generated looks like this by default:
`Content-Security-Policy: default-src *; frame-ancestors 'self'`

It can be modified through the filter to look something like this:
`Content-Security-Policy: default-src *; frame-ancestors 'self' *.greenpeace.ch`

It's important to mention that using `default-src: *;` is considered an unsafe CSP directive because it allows loading sources from anywhere. I added it like that because it most closely resembles what we've had before: no CSP at all. I think it would make sense to look into developing this into a more refined CSP in the future, but it's out of scope for this issue. Extensive testing would be required to make sure not to break any functionality on P4 websites.

src/HttpHeaders.php
stduerre: I agree, there's no reason to disallow `'self'`. I modified the filter so it only allows adding to the list of ancestors.

stduerre:
Updated default CSP to:
`content-security-policy: default-src * 'self' data: 'unsafe-inline' 'unsafe-eval'; frame-ancestors 'self'`

Upon further testing I found that the last version would block some resources that P4 site might use currently and the change is needed to not cause any issues at the moment.

src/HttpHeaders.php
```diff
@@ -0,0 +1,43 @@
...
+	/**
+	 * Send Content Security Policy (CSP) HTTP headers.
+	 */
+	public function send_content_security_policy_header() {
+		$default_allowed_frame_ancestors = [ '\'self\'' ];
+
+		// Filter hook to allow adding trusted frame ancestors.
+		$additional_allowed_frame_ancestors = apply_filters( 'planet4_csp_allowed_frame_ancestors', [] );
+
+		$allowed_frame_ancestors = array_merge( $default_allowed_frame_ancestors, (array) $additional_allowed_frame_ancestors );
```
Inwerpsel, line 26: Maybe best to not cast it to an array and just document the filter needs to return an array of strings (and maybe check if it actually is one before doing the merge). Casting to an array will do some unwanted conversions. For example if the filter by accident returns an object (like a `WP_ERROR`), then it will be converted to an array where the keys are the keys of the object (including private properties). See https://www.php.net/manual/en/language.types.array.php#language.types.array.casting

stduerre: Removed the cast and instead checked for `$additional_allowed_frame_ancestors` to be an array. 

Not sure what to do with the documentation. I used a one line comment for consistency with the rest of the master theme. Considering the limited space I didn't mention the return type, every filter function is expected to return the first argument.

I could change it to how the [Wordpress Handbook](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-hooks-actions-and-filters) suggests documenting a hook. It also only specifies the parameter and assumes the return value is the same. But then it's inconsistent with the rest of the code. What's better?

Example:
```
/**
 * Filter hook to allow adding trusted frame ancestors to the Content Security Policy.
 * 
 * @param array $additional_allowed_frame_ancestors Array of domains to whitelist as frame ancestors.
 */
```

Inwerpsel: Well, we don't enforce how phpdoc should look for filters, so I guess we can indeed follow how WordPress does it.

Inwerpsel:
@stduerre The unit tests are failing because of a warning

> OpenGraphTest::test_post_open_graph_data with data set #0 (array('post', 'The name of the place is Babylon', 'test-social-url', 'test content', 'post excerpt'), 'single.php')
Cannot modify header information - headers already sent by (output started at /tmp/wordpress-tests-lib/includes/bootstrap.php:151)

Not sure if you have access to this in CircleCI? There's 9 tests failing with the same error. I assume it's only an issue in the unit tests, since we use the `wp_headers` action.

Inwerpsel:
I created this draft PR of the pulled branch so that it can pick up the PR title from there, which it needs to do a test instance deploy. https://github.com/greenpeace/planet4-master-theme/pull/1294

stduerre:
> 
> 
> @stduerre The unit tests are failing because of a warning
> 
> > OpenGraphTest::test_post_open_graph_data with data set #0 (array('post', 'The name of the place is Babylon', 'test-social-url', 'test content', 'post excerpt'), 'single.php')
> > Cannot modify header information - headers already sent by (output started at /tmp/wordpress-tests-lib/includes/bootstrap.php:151)
> 
> Not sure if you have access to this in CircleCI? There's 9 tests failing with the same error. I assume it's only an issue in the unit tests, since we use the `wp_headers` action.

I can see the error message, but can only guess what output is sent. OFten times, this error message is due to PHP errors or warnings being sent. I agree this is probably an issue with how the tests are run, not this particular PR. The HttpHeaders class in this PR uses the `wp_headers` hook, which should be the correct way to do it.

I'm not sure how these tests are run, but is there a chance that once this PR is merged, the errors show up every time any of these tests is run? In that case, they'd need to be investigated. But that's out of scope for this ticket.

oekeur:
What's the status on this? :)

lithrel:
@stefan I got a look at the failing test; I think we're using `wp_headers` hook the wrong way, it should [return a list of headers](https://github.com/WordPress/wordpress-develop/blob/a45c8715410b1f362218da80eda9bd0248fd5503/src/wp-includes/class-wp.php#L495) instead of sending them right away - wordpress [handles the `headers_sent()` verification](https://github.com/WordPress/wordpress-develop/blob/a45c8715410b1f362218da80eda9bd0248fd5503/src/wp-includes/class-wp.php#L510) process which protects the tests from this kind of error.

So a modification like this seems to work (on my side) and doesn't break tests:
```diff
diff --git a/src/HttpHeaders.php b/src/HttpHeaders.php
index b82dc552..6bd6a7ec 100644
--- a/src/HttpHeaders.php
+++ b/src/HttpHeaders.php
@@ -17,7 +17,8 @@ class HttpHeaders {
 	/**
 	 * Send Content Security Policy (CSP) HTTP headers.
 	 */
-	public function send_content_security_policy_header() {
+	public function send_content_security_policy_header( $headers ): array {
+
 		$default_allowed_frame_ancestors = [ '\'self\'' ];
 
 		/**
@@ -35,15 +36,15 @@ class HttpHeaders {
 			'default-src * \'self\' data: \'unsafe-inline\' \'unsafe-eval\'',
 			'frame-ancestors ' . implode( ' ', $allowed_frame_ancestors ),
 		];
-
-		$csp_header = 'Content-Security-Policy: ' . implode( '; ', $directives );
+		$csp_header = implode( '; ', $directives );
 		$csp_header = preg_replace( "/\r|\n/", '', $csp_header );
-
-		header( $csp_header );
+		$headers['Content-Security-Policy'] = $csp_header;
 
 		// In addition, send the "X-Frame-Options" header when no other trusted frame ancestors were added through the filter.
 		if ( $allowed_frame_ancestors === $default_allowed_frame_ancestors ) {
-			header( 'X-Frame-Options: SAMEORIGIN' );
+			$headers['X-Frame-Options'] = 'SAMEORIGIN';
 		}
+
+		return $headers;
 	}
 }
```

Can you try this and check on your side if it still work as expected ?

stduerre:
@lithrel You're right about that, it's supposed to only return the headers, not send them. My mistake, thanks for spotting it! 

> Can you try this and check on your side if it still works as expected ?

Yes, perfectly. The headers are exactly the same as before.

lithrel (approved):
LGTM, thanks @stduerre  :+1: 

As mentioned, this currently results in 2 similar headers `X-Frame-Options` (the other one coming from [nginx security conf](https://github.com/greenpeace/planet4-docker/blob/a8b070482b456d43479a2b22f66fbac018582819/src/planet-4-151612/openresty/etc/nginx/conf.d/security.conf#L8)), we'll need to adapt the conf for that.

PR #1286, opened 2021-01-27
PLANET-5870: Cache clear command

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-5870

Adding this as a command first so that we can make it a post-deploy step for test instances. We can start that way and observe if purging all html on each deployment causes any issues. Then we can set it up for a few production NROs and see how it goes. If it doesn't cause issues (e.g. Cloudflare deciding to evict us from cache more frequently once they pick up these entries usually don't live long), we could do this for all sites on each deploy.

I added a feature flag for this, so that we're not blocked to use it on production by needing more code changes. Tested off and on behavior on https://app.circleci.com/pipelines/github/greenpeace/planet4-test-janus/37/workflows/a3b88549-3fb8-4d73-926b-2e42e61b34ad/jobs/199

### DIFF
--- a/src/Commands.php
+++ b/src/Commands.php
@@ -33,7 +33,7 @@ class Commands {
 		 * Put the CF API key into the options table, where the CF plugin uses it from.
 		 */
 		$put_cf_key_in_db = static function ( $args ) {
-			$hostname = $args[0];
+			$hostname = $args[0] ?? null;
 			if ( empty( $hostname ) ) {
 				WP_CLI::error( 'Please specify the hostname.' );
 			}
@@ -65,7 +65,7 @@ class Commands {
 
 			if ( isset( $assoc_args['urls'] ) ) {
 				$urls = explode( ',', $assoc_args['urls'] );
-			} elseif ( $assoc_args['all'] ) {
+			} elseif ( isset( $assoc_args['all'] ) ) {
 				$post_types = isset( $assoc_args['post-types'] )
 					? explode( ',', $assoc_args['post-types'] )
 					: [
--- a/src/Commands.php
+++ b/src/Commands.php
@@ -1,10 +1,25 @@
 <?php
+/**
+ * Commands.
+ */
 
 namespace P4\MasterTheme;
 
+use CF\Integration\DefaultConfig;
+use CF\Integration\DefaultIntegration;
+use CF\Integration\DefaultLogger;
+use CF\WordPress\DataStore;
+use CF\WordPress\WordPressAPI;
+use CF\WordPress\WordPressClientAPI;
 use WP_CLI;
 
+/**
+ * Class with a static function just because PHP can't autoload functions.
+ */
 class Commands {
+	/**
+	 * Add some WP_CLI commands if we're in CLI.
+	 */
 	public static function load() {
 		if ( ! defined( 'WP_CLI' ) || ! WP_CLI ) {
 			return;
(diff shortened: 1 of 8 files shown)


src/Commands.php
```diff
@@ -0,0 +1,116 @@
...
+					$joined = implode( $chunk, "\n" );
+					WP_CLI::warning( "Chunk $i failed, one or more of these didn't work out: \n$joined" );
+				}
+				$progress->tick();
+			}
+		};
+
+		$activator_command = static function ( $args, $assoc_args ) use ( $purge_urls ) {
+			Activator::run();
+			$purge_urls( [], [ 'all' => true ] );
```
Inwerpsel, line 110: This is just temporary to keep the changes in the same PR

src/Commands.php
```diff
@@ -0,0 +1,118 @@
...
+			if ( ! Features::is_active( Features::CLOUDFLARE_DEPLOY_PURGE ) ) {
+				WP_CLI::warning( 'Purge on deploy is not enabled, not purging.' );
+
+				return;
+			}
+
+			if ( ! defined( 'CLOUDFLARE_PLUGIN_DIR' ) ) {
+				define( 'CLOUDFLARE_PLUGIN_DIR', WP_PLUGIN_DIR . '/cloudflare/' );
+			}
+			require_once CLOUDFLARE_PLUGIN_DIR . 'vendor/autoload.php';
```
lithrel, line 64: regular autoload doesn't work for this ? (it works locally without it at least)

src/Commands.php
```diff
@@ -0,0 +1,118 @@
...
+
+			if ( isset( $assoc_args['urls'] ) ) {
+				$urls = explode( ',', $assoc_args['urls'] );
+			} elseif ( $assoc_args['all'] ) {
```
lithrel, line 68: `isset()` ?

lithrel (approved):
Also check Sonar for code smell and complexity https://sonarcloud.io/project/issues?id=greenpeace_planet4-master-theme&pullRequest=1286&resolved=false&types=CODE_SMELL (not always fixable but interesting)

src/Commands.php
Inwerpsel: Not sure if it's really "autoloading" it, could be the classes happen to be loaded already at the point of execution. But if we don't add this then there's no explicit way of saying where these classes are coming from. If the plugin is not present this should also make it fail as early as possible.

src/Commands.php
```diff
@@ -0,0 +1,118 @@
...
+				WP_CLI::error( 'Please specify the hostname.' );
+			}
+
+			if ( ! defined( 'CLOUDFLARE_API_KEY' ) || empty( CLOUDFLARE_API_KEY ) ) {
+				WP_CLI::error( 'CLOUDFLARE_API_KEY constant is not set.' );
+			}
+
+			$domain_parts = explode( '.', $hostname );
+
+			$root_domain = implode( '.', array_slice( $domain_parts, - 2 ) );
```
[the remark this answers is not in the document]
Inwerpsel: Agree it's a shaky way to get the root domain, but I only moved this code from `Loader` without changes, so I rather not blow up this PR by adding a new dependency :sweat_smile: We can look at it in a separate PR :ok_hand:

Inwerpsel:
> Also check Sonar for code smell and complexity https://sonarcloud.io/project/issues?id=greenpeace_planet4-master-theme&pullRequest=1286&resolved=false&types=CODE_SMELL (not always fixable but interesting)

I followed Sonar and actually glad I did, it's much nicer to have each command in  its own file. It's still tripping over a too long line and the runtime exception, but I don't think the latter is an issue for a command. Also Squiz doesn't understand `@inheritDoc` correctly, but seems they fixed it recently, maybe we can bump it instead of working around. https://github.com/squizlabs/PHP_CodeSniffer/issues/2770#issuecomment-762532979

PR #1288, opened 2021-01-28
PLANET-5449 Fix coverage path

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-5449

---

Also did a small fix for `@covers` otherwise it fails


sagarsdeshmukh (approved):
LGTM 👍

PR #1292, opened 2021-02-08
PLANET-5869 Protect posts with "locked" meta

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-5869

---

If env var `LOCK_TEST_CONTENT` is true, add a filter that disallows changing any post that has the `test-content` meta.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -99,14 +99,49 @@ if ( ! class_exists( 'Timber' ) ) {
 	);
 
 	return;
-} else {
-	// Enable Timber template cache unless this is a debug environment.
-	if ( defined( 'WP_DEBUG' ) && is_bool( WP_DEBUG ) ) {
-		Timber::$cache = ! WP_DEBUG;
-	} else {
-		Timber::$cache = true;
-	}
 }
+
+// Enable Timber template cache unless this is a debug environment.
+Timber::$cache = defined( 'WP_DEBUG' ) && is_bool( WP_DEBUG ) ? ! WP_DEBUG : true;
+
+// Prevent posts from being saved if they have the "locked" meta key. There are plugins for this too, but usually they
+// still allow admins to make changes.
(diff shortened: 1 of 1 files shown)


comzeradd:
Can you add a more detailed commit summary on the logic behind this? The code looks good, but it's 100% clear to me why we need this if we import defaultcontent sql on each run.

Inwerpsel:
> Can you add a more detailed commit summary on the logic behind this? The code looks good, but it's 100% clear to me why we need this if we import defaultcontent sql on each run.

Added it :ok_hand: The main reason is to avoid having to import the whole defaultcontent sql.

functions.php
```diff
@@ -99,14 +99,49 @@ function( $template ) {
...
-		Timber::$cache = ! WP_DEBUG;
-	} else {
-		Timber::$cache = true;
-	}
 }
+
+// Enable Timber template cache unless this is a debug environment.
+Timber::$cache = defined( 'WP_DEBUG' ) && is_bool( WP_DEBUG ) ? ! WP_DEBUG : true;
+
+// Prevent posts from being saved if they have the "locked" meta key. There are plugins for this too, but usually they
```
lithrel, line 107: "test-content" meta key ?

lithrel (approved):
:+1: tested by using wp cli `wp post meta add 427 test-content true`

Since you talk about creating an export of test content, we could move this code in its own class at some point, having the meta key in a constant and adding simple methods to call, `TestContent::export`, `TestContent::initUserCapLock` and so on, just to not put too much in this file

Inwerpsel:
Coming back to this old PR, I'm not sure if it's still relevant. We could instead add the files to Git and import them before every test.

If we do that there might still be a place for "protecting" this content, although simply overwriting it in case of changes before running a test might be better.

PR #1293, opened 2021-02-09
PLANET-5939: Campaign text hidden by navbar

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-5939

On a campaign page with title hidden and no subtitle or description, the text is on top of the page and gets hidden by the navbar.

Screenshot from 2021-02-09 16-57-13

The div `.page-header` has a padding that largely compensate the navbar height, but it is not always visible. When the title is hidden (option "Hide page title" in Campaign and Page) and there's no subtitle or description, it shrinks as `.page-header-hidden` and the main container has no padding-top.  
The padding has to be precise, so that when we add an EN form as first block, it doesn't look weird or has no white empty bar at the top.

The problem also exists on Pages. Solution has to be tested on:

| | Page type | Title visible* | First block |
| --- | --- | --- | --- |
| Minimal navigation |
| | Campaign | no title | <ul><li>[x] paragraph</li><li>[x] EN form</li></ul> |
| | Campaign | title | <ul><li>[x] paragraph</li><li>[x] EN form</li></ul> |
| Main navigation |
| | Campaign | no title | <ul><li>[x] paragraph</li><li>[x] EN form</li></ul> |
| | Campaign | title | <ul><li>[x] paragraph</li><li>[x] EN form</li></ul> |
| | Page | no title | <ul><li>[x] paragraph</li><li>[x] EN form</li></ul> |
| | Page | title | <ul><li>[x] paragraph</li><li>[x] EN form</li></ul> |

_***title visible** can be header title, subtitle or description, each of those will make the div `page-header` appear, and this div has its own padding._

## Fix 
Assigning a padding-top the height of the current navbar, in this context only (header hidden, no title).  
Padding is different according to main nav or minimal nav.

## Test

Any of those situations should work.  
Test instance has 3 campaigns pages to check: 
- https://www-dev.greenpeace.org/test-pluto/campaign/campaign-with-title/
- https://www-dev.greenpeace.org/test-pluto/campaign/campaign-no-title/
- https://www-dev.greenpeace.org/test-pluto/campaign/campaign-no-title-en-form/
- https://www-dev.greenpeace.org/test-pluto/page-no-title/

### DIFF
--- a/assets/src/scss/pages/_page.scss
+++ b/assets/src/scss/pages/_page.scss
@@ -33,6 +33,38 @@ div.page-template {
     max-width: 100%;
     height: auto;
   }
+
+  .page-header-hidden + &.no-page-title {
+    padding-top: $menu-height-small;
+
+    @include medium-and-up {
+      padding-top: $menu-height-large;
+    }
+
+    & > p:first-child {
+      padding-top: $space-xs;
+
+      @include small-and-up {
+        padding-top: $space-sm;
+      }
+
+      @include medium-and-up {
+        padding-top: $space-md;
+      }
+
+      @include large-and-up {
+        padding-top: $space-lg;
+      }
+
+      @include x-large-and-up {
+        padding-top: $space-xl;
+      }
+    }
+
+    .navigation-bar_min ~ & {
+      padding-top: $min-height;
+    }
+  }
 }
 
 .single-campaign {
--- a/templates/page.twig
+++ b/templates/page.twig
@@ -1,7 +1,7 @@
 {% extends "base.twig" %}
 
 {% block content %}
-	<div class="page-template">
+	<div class="page-template {% if hide_page_title_checkbox == 'on' %}no-page-title{% endif %}">
 		{{ post.content|raw }}
 	</div>
 {% endblock %}
--- a/templates/single-page.twig
+++ b/templates/single-page.twig
@@ -5,7 +5,7 @@
 		<div class="page-header-background"></div>
 	</div>
 
-	<div class="page-template">
+	<div class="page-template {% if hide_page_title_checkbox == 'on' %}no-page-title{% endif %}">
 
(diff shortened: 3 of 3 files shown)


comzeradd:
One thing to check is how does this change affects Campaigns that use an EN form side style as the first block on that page.

eg. https://www.greenpeace.org/international/campaign/climate-emergency/

mleray:
I would add more padding at the top for the campaign with no title (https://www-dev.greenpeace.org/test-pluto/campaign/campaign-no-title/), I think it looks better if the top padding is similar to the bottom one 🙂 But maybe something to double-check with design!

lithrel:
@mleray I've asked Mag, who pointed to the [vertical-spacing](https://zeroheight.com/05f6e9516/p/34a6e6-layout) of the new design system. These values are not yet implemented, but we currently have some [spacing variables](https://github.com/greenpeace/planet4-styleguide/blob/53490a5d6faeea333ec16da166828ad94b6d0443/src/base/_variables.scss#L40) so I've used those thinking that it will be updated when we switch to the new system. Does it make sense ?
The logic for page without title now is : move content down the height of the navbar, and if the first block is a paragraph add generic vertical spacing to it.

mleray:
> @mleray I've asked Mag, who pointed to the [vertical-spacing](https://zeroheight.com/05f6e9516/p/34a6e6-layout) of the new design system. These values are not yet implemented, but we currently have some [spacing variables](https://github.com/greenpeace/planet4-styleguide/blob/53490a5d6faeea333ec16da166828ad94b6d0443/src/base/_variables.scss#L40) so I've used those thinking that it will be updated when we switch to the new system. Does it make sense ?
> The logic for page without title now is : move content down the height of the navbar, and if the first block is a paragraph add generic vertical spacing to it.

Makes sense to me 🙂 Thanks for checking!

mleray (approved):
Looks good to me 👍

PR #1295, opened 2021-02-10
PLANET-5756 Remove jQuery from external_link.js and pdf_icon.js

mleray:
### Description
See https://jira.greenpeace.org/browse/PLANET-5756

After discussion with Magali it was decided to not show the external link icon in headings/titles (as we already do for the PDF icon)

### Related PRs
- styleguide: https://github.com/greenpeace/planet4-styleguide/pull/96


assets/src/js/pdf_icon.js
```diff
@@ -1,12 +1,26 @@
...
+      return;
+    } else {
+      const text = link.textContent || link.innerText;
+      if (text.trim().length === 0) {
+        return;
+      }
 
-    if (!(link.parent('h1, h2, h3, h4, h5, h6').length || link.has('img').length)) {
-      link.addClass('icon-link pdf-link');
+      link.classList.add('pdf-link');
```
Inwerpsel, line 23: If all we're doing in the end is adding a class, maybe we could replicate the whole script in pure CSS? :thinking:

assets/src/js/external_links.js
```diff
@@ -1,27 +1,39 @@
...
+  const article = document.getElementsByTagName('article')[0];
 
-  $('div.page-template a:not(.btn):not(.cover-card-heading), article a:not(.btn):not(.cover-card-heading)').each(function () {
-    let href = undefined === $(this).attr('href') ? '' : $(this).attr('href');
-    if (href != '' && href.indexOf(siteURL) <= -1 && href.length > 0) {
-      if ($(this).text().trim().length == 0) {
+  let links = [];
+
+  if (page) {
+    links = links.concat([...page.querySelectorAll('a:not(.btn):not(.cover-card-heading)')]);
```
Inwerpsel, line 10: I guess the previous selector should still work in vanilla JS? `div.page-template a:not(.btn):not(.cover-card-heading), article a:not(.btn):not(.cover-card-heading)`

assets/src/js/pdf_icon.js
mleray: I don't think that would be possible because we need to check that the link has no image as child, and that's not possible in pure CSS 🤔

assets/src/js/pdf_icon.js
```diff
@@ -1,12 +1,26 @@
...
-  'use strict';
+export const setupPDFIcon = () => {
+  const links = [...document.querySelectorAll('a[href*=".pdf"]')];
 
-  $('a[href*=".pdf"]').each(function() {
-    const link = $(this);
+  links.forEach(link => {
+    // We don't want to show the icon in headings/titles
+    const linkParent = link.parentElement.nodeName;
+    if (['H1', 'H2', 'H3', 'H4', 'H5', 'H6'].includes(linkParent)) {
```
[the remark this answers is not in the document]
mleray: This may be shorter, but I feel like the "deconstructed" way is more readable 😬

assets/src/js/pdf_icon.js
```diff
@@ -1,12 +1,20 @@
...
-  $('a[href*=".pdf"]').each(function() {
-    const link = $(this);
-
-    if (!(link.parent('h1, h2, h3, h4, h5, h6').length || link.has('img').length)) {
-      link.addClass('icon-link pdf-link');
+  links.forEach(link => {
+    // We don't want to show the icon in headings/titles,
+    // in empty links or in links that are images
+    const linkParent = link.parentElement.nodeName;
+    const text = link.textContent || link.innerText;
```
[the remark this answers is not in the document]
mleray: Yeah I guess we can remove it, I probably added it at some point because it's there for external links 🤔

assets/src/js/external_links.js
```diff
@@ -1,27 +1,25 @@
...
-      }
+  links.forEach(link => {
+    // We don't want to show the icon in headings/titles,
+    // or in links that are images
+    const href = link.href || '';
+    const text = link.textContent || link.innerText;
+    if (
+      ['H1', 'H2', 'H3', 'H4', 'H5', 'H6'].includes(link.parentElement.nodeName)
+      || !href
+      || href.includes(siteURL)
```
Inwerpsel, line 14: For this it makes sense to filter more in the initial query selector, since currently it will match all links on the page and loop them. Which is usually a lot more than the ones we're targeting. I think we can use an attribute selector 
```css
:not([href*=${siteURL}])
```
 Then we also don't need to check for the href being empty.

mleray: Updated 🙂

Inwerpsel (approved):

PR #1296, opened 2021-02-11
PLANET-5752 Remove jQuery from enhancedDonateButton.js

sagarsdeshmukh:
Ref: [JIRA 5752](https://jira.greenpeace.org/browse/PLANET-5752)

---

**Testing steps:**
- Go to WP admin and enable the "Show donate button below mobile navigation bar" setting from **Planet 4 >> Donate button**
- Test the feature in mobile and desktop screen
- The functionality should work as it is after removing the jQuery

### DIFF
--- a/assets/src/js/enhancedDonateButton.js
+++ b/assets/src/js/enhancedDonateButton.js
@@ -1,34 +1,58 @@
-export const setupEnhancedDonateButton = function() {
-  function isMobile() {
-    return window.matchMedia('(max-width: 576px)').matches;
-  }
+export const setupEnhancedDonateButton = () => {
+  const isMobile = () => window.matchMedia('(max-width: 576px)').matches;
 
-  function setupDonateButton() {
+  const setupDonateButton = () => {
+    const donateButton = document.querySelector('.btn-donate-top');
     if ( isMobile() ) {
-      if (!$('.btn-enhanced-donate.btn-donate-top').length) {
-        $('.btn-enhanced-donate').clone().appendTo('body').addClass('btn-donate-top');
-        $('.btn-donate-top').parent('body').addClass('with-donate-on-top');
+      if ( ! donateButton ) {
+        const enhancedDonateBtn = document.querySelector('.btn-enhanced-donate').cloneNode(true);
+        enhancedDonateBtn.classList.add('btn-donate-top');
+
+        document.querySelector('body').appendChild(enhancedDonateBtn);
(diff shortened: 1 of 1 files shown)


assets/src/js/enhancedDonateButton.js
```diff
@@ -1,34 +1,58 @@
...
-        $('.btn-donate-top').parent('body').addClass('with-donate-on-top');
+      if ( ! document.querySelector('.btn-enhanced-donate.btn-donate-top') ) {
+        const enhancedDonateBtn = document.querySelector('.btn-enhanced-donate').cloneNode(true);
+        enhancedDonateBtn.classList.add('btn-donate-top');
+
+        document.querySelector('body').appendChild(enhancedDonateBtn);
+        document.querySelector('body').classList.add('with-donate-on-top');
+      }
+    } else {
+      const buttonClass = document.querySelector('.btn-donate-top');
```
mleray, line 14: I would rename this variable (maybe to just `button`?) since the selector returns the whole element and not just the class 🙂

assets/src/js/enhancedDonateButton.js
```diff
@@ -1,34 +1,58 @@
...
+  const isMobile = () => window.matchMedia('(max-width: 576px)').matches;
 
-  function setupDonateButton() {
+  const setupDonateButton = () => {
     if ( isMobile() ) {
-      if (!$('.btn-enhanced-donate.btn-donate-top').length) {
-        $('.btn-enhanced-donate').clone().appendTo('body').addClass('btn-donate-top');
-        $('.btn-donate-top').parent('body').addClass('with-donate-on-top');
+      if ( ! document.querySelector('.btn-enhanced-donate.btn-donate-top') ) {
+        const enhancedDonateBtn = document.querySelector('.btn-enhanced-donate').cloneNode(true);
```
mleray, line 7: Shouldn't we first check that `document.querySelector('.btn-enhanced-donate')` exists? 🤔

sagarsdeshmukh: yes, here we need to check `document.querySelector('.btn-donate-top')`, PR updated :)

mleray: Oh no that's not what I meant, sorry 🙂 I meant we should check that `document.querySelector('.btn-enhanced-donate')` exists before cloning it, but now I see that you already check that in `ready` function so it's fine ✌️

mleray (approved):

PR #1298, opened 2021-02-12
PLANET-5850 Remove CSS vars ponyfill

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-5850

After doing some testing I noticed that the ponyfill has a large impact on performance. I don't know yet exactly what causes this impact, it could be the additional domain for a blocking resource, or the ponyfill might interfere with render by accessing style attributes. Normally it should not have to do anything for browsers that do support CSS variables, but maybe just checking for support affects page load.

I [temporarily put a commit on this branch](https://github.com/greenpeace/planet4-master-theme/commit/77faaca03c024f9f80b747d8fca8453efe0b8d12) that has a toggle for the removal so I could test both scenarios. The difference for the test instance was quite massive, without the ponyfill it was getting just under 70, with the ponyfill this dropped to under 50. It seems that this increase is due to total blocking time.

### DIFF
--- a/assets/src/js/app.js
+++ b/assets/src/js/app.js
@@ -6,7 +6,6 @@ import { setupLoadMore } from './load_more';
 import { setupPDFIcon } from './pdf_icon';
 import { setupSearch } from './search';
 import { setupExternalLinks } from './external_links';
-import { setupCSSVarsPonyfill } from './cssvarsponyfill';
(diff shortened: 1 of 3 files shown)


comzeradd (approved):

comzeradd:
Looks good. I checked latest page views for browsers that don't support css variables. Opera Mini hits are so low that are not even registered. IE11 is around 0.05%.

PR #1303, opened 2021-02-16
Use python3 for book-instance

comzeradd:
Removing the python call, since the script is executable and already configured to use python3


lithrel (approved):
:+1:

PR #1305, opened 2021-02-16
PLANET-5906: Migrate to Bootstrap 5

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-5906

Migrate to Bootstrap 5.

Changes:
* [_media.scss](https://github.com/twbs/bootstrap/blob/v4.6.0/scss/_media.scss) no longer exists, but it can be replaced with the `d-flex` utility.
* `sr-only` is now `visually-hidden`

Also check the [styleguide PR](https://github.com/greenpeace/planet4-styleguide/pull/98) and the [blocks PR](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/505).

---

### DIFF
--- a/assets/src/scss/bootstrap-build.scss
+++ b/assets/src/scss/bootstrap-build.scss
@@ -20,9 +20,9 @@ Text Domain: planet4-master-theme
 
 // Required
 // See: https://getbootstrap.com/docs/4.3/getting-started/theming/#importing
-@import "~bootstrap/scss/functions";
-@import "~bootstrap/scss/variables";
-@import "~bootstrap/scss/mixins";
+// @import "~bootstrap/scss/functions";
+// @import "~bootstrap/scss/variables";
+// @import "~bootstrap/scss/mixins";
 
 // TODO: Add variable overrides for the 16px grid here.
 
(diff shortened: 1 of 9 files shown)


assets/src/scss/bootstrap-build.scss
```diff
@@ -12,57 +12,50 @@ Tags: light, accessibility-ready
...
-// Required
-// See: https://getbootstrap.com/docs/4.3/getting-started/theming/#importing
 @import "~bootstrap/scss/functions";
 @import "~bootstrap/scss/variables";
 @import "~bootstrap/scss/mixins";
+@import "~bootstrap/scss/utilities";
 
-// TODO: Add variable overrides for the 16px grid here.
+// Layout & components
+@import "~bootstrap/scss/root";
```
mleray, line 27: We had these first files commented out before, do we now want to use them or is this a mistake? 🤔

pablocubico: Yes, we want! These were included before as part of `bootstrap-grid` and `bootstrap-reboot`, which are partials including a subset of things. This was looks closer to the original file and makes it easier to see what's being loaded with more granularity.

mleray (approved):
Can you please fix the [stylelint issue](https://app.circleci.com/pipelines/github/greenpeace/planet4-master-theme/2705/workflows/e9ec6448-4dcc-483f-bbfb-16a4f1498d25/jobs/14004) and the conflict? Otherwise, tested on local, looks good to me 👍

PR #1307, opened 2021-02-17
PLANET-5844: Investigate using Google Optimize with optimize.js

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-5844

WIP.
* We should conditionally load optimize only if the cookies are accepted.
* Maybe we could also try loading it only when we know we are running a test... 🤔 

---

### DIFF
--- a/templates/blocks/google_tag_manager.twig
+++ b/templates/blocks/google_tag_manager.twig
@@ -1,10 +1,10 @@
 {% if google_tag_value %}
   {% if ab_hide_selector %}
-  <style>
-    .google-optimize-loading {{ ab_hide_selector }} {
-      opacity: 0 !important;
-    }
-  </style>
+    <style>
+      .google-optimize-loading {{ ab_hide_selector }} {
+        opacity: 0 !important;
+      }
+    </style>
   {% endif %}
   <script>
     var google_tag_value = '{{ google_tag_value }}';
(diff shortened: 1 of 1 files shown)


comzeradd:
> We should conditionally load optimize only if the cookies are accepted.

Do we only run the a/b test experiments on people who have accepted cookies? 🤔

pablocubico:
> > We should conditionally load optimize only if the cookies are accepted.
> 
> Do we only run the a/b test experiments on people who have accepted cookies? 🤔

@comzeradd this is the current condition: https://github.com/greenpeace/planet4-master-theme/blob/master/templates/blocks/google_tag_manager.twig#L52

comzeradd:
@pablocubico in general it looks good, but I think we should also add an option (Planet 4 > Analytics) to load this conditionally when needed.

Can you also adjust [this css rule](https://github.com/greenpeace/planet4-master-theme/blob/master/admin/css/campaign.css#L37) (and maybe move it in another file) to make it more generic to match all settings options? Something like that should work:

```
.planet4_options .cmb-th + .cmb-td {
  float: none;
}

File: templates/blocks/google_tag_manager.twig [hunk 1]
```diff
@@ -54,18 +54,11 @@
...
-          'gtm.start':
-          new Date().getTime(), event: 'gtm.js'
-        });
-        var f = d.getElementsByTagName(s)[0],
-            j = d.createElement(s), dl = l != 'dataLayer' ? '&l=' + l : '';
-        j.async = true;
-        j.src = 'https://www.googletagmanager.com/gtm.js?id=' + i + dl;
-        f.parentNode.insertBefore(j, f);
-      })(window, document, 'script', 'dataLayer', google_tag_value);
+      var script = document.createElement('script');
```
Inwerpsel, line 57: Do we need to replace the code on the left with the optimize loading?

PR #1309, opened 2021-02-18
PLANET-5959 Add lazy youtube loading

Inwerpsel:
* Add a way to force feature settings using query string. Can be used to
alternate lighthouse tests between both versions so that both are
affected about equally by the run conditions.
* Use srcdoc with an image using YouTube's url schema to avoid loading
the massive player on page load.

Ref: https://jira.greenpeace.org/browse/PLANET-5959
---

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -65,7 +65,7 @@
     "selector-pseudo-element-case": "lower",
     "selector-type-case": "lower",
     "selector-type-no-unknown": [true, {
-       "ignoreTypes": ["/^--[a-z]\\w*(--?[a-z0-9]+)*--$/"]
+       "ignoreTypes": ["/^--[a-z]\\w*(--?[a-z0-9]+)*--$/", "lite-youtube"]
     }],
     "selector-max-empty-lines": 0,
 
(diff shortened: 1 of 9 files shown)


assets/src/js/vendors/lite-youtube.js
```diff
@@ -0,0 +1,124 @@
+/* eslint-disable */
```
comzeradd, line 1: Since we are going to maintain this code from now on, maybe it makes sense to lint it? Same for the css part.

Inwerpsel: Sure, though maybe we can just build the script by taking it from node_modules instead :thinking: Are you ok with handling this in a follow up PR so we can take time to set it up properly and sites can already benefit from it in the meanwhile?

comzeradd: We can fix the linting issues now, since it's many lines of code and probably it's mostly trivial changes, to avoid adding some technical debt here. And we iterate later on how is the best way to load the script.

Inwerpsel: I made it use the npm package instead, so that we don't have to re-apply our code style each time we update.

comzeradd: Ah much better 🙂

comzeradd (approved):

PR #1310, opened 2021-02-19
Fix stylelint recursive path

comzeradd:
stylelint needs a double-quoted path, otherwise it doesn't make a recursive lookup.

Counterpart of greenpeace/planet4-plugin-gutenberg-blocks#510. In this case no scss code needed fixing.


mleray (approved):
Looks good to me 👍

PR #1317, opened 2021-02-25
PLANET-5971 Fix blocks js loading when logged in

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-5971

---

* By not making jquery depend on it when logged in, since in that case
WordPress enqueues jquery in the head.


mleray (approved):
Looks good to me 👍

PR #1320, opened 2021-03-01
Bump dashdash, whitelist shorthand

Inwerpsel:
Ref: no ticket, just a bump to match the same bump in https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/495

---

This bumps dashdash to the latest version, which supports a shorthand to generate variable names from the selector.
```
.foo-bar {
  _-- {
    color: green;
    background-color: yellow;
  }
}
```
produces the same result as
```
.foo-bar {
  --foo-bar-- {
    color: green;
    background-color: yellow;
  }
}
```
which is the same as
```
.foo-bar {
  color: var(--foo-bar--color, green);
  background-color: var(--foo-bar--background-color, yellow);
}
```

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -65,7 +65,7 @@
     "selector-pseudo-element-case": "lower",
     "selector-type-case": "lower",
     "selector-type-no-unknown": [true, {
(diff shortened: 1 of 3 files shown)


comzeradd (approved):
Can you adjust also Sonacloud stylelint config if needed?

PR #1321, opened 2021-03-03
Move over svg icons from styleguide

comzeradd:
Since icons are part of master-theme build system, it makes more sense to keep them here.


lithrel (approved):
:+1:

PR #1323, opened 2021-03-03
PLANET-5981 Updates for Bootstrap 5

mleray:
### Description

See https://jira.greenpeace.org/browse/PLANET-5981

Changes made to comply with the [migration doc](https://getbootstrap.com/docs/5.0/migration/):
- Rename all Bootstrap attributes' prefixes from `data-` to `data-bs-`
- Rename `float-right` to `float-end`
- Add `form-select` to select components instead of `form-control` (which now has the `appearance: none;` hiding the select's arrow)
- Remove `mr-sm-2` class
- Rename `text-right` to `text-end`
- Update page-template width for xxl screens (new breakpoint added as of Bootstrap 5)
- Remove `screen-reader-text` class to use Bootstrap's `visually-hidden` instead
- Add `align-items-start` class to search results
- Fix search for rtl direction

### Testing

You can check the changes on [this test instance](https://www-dev.greenpeace.org/test-janus/), basically everything should look/behave the same as it does in the current sites 🙂 

### Related PRs

- [styleguide](https://github.com/greenpeace/planet4-styleguide/pull/104)
- [gutenberg-blocks](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/521)

### DIFF
--- a/assets/src/js/header.js
+++ b/assets/src/js/header.js
@@ -38,7 +38,7 @@ export const setupHeader = function($) {
     const clickedElement = evt.target;
     $('button[aria-expanded="true"]').each(function(i, button) {
       const $button = $(button);
-      const buttonTarget = $($button.data('target')).get( 0 );
+      const buttonTarget = $($button.data('bs-target')).get( 0 );
       if (buttonTarget && ! $.contains(buttonTarget, clickedElement)) {
         // Spoof a click on the open menu's toggle to close that menu.
         $button.trigger('click');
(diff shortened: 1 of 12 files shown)


lithrel:
`.form-select` arrow doesn't adapt to rtl (stays on the right side), but I tested by adding `dir="rtl"` to `<html>` tag, I don't know if my test process is wrong, if we build specific css for rtl or something else.

mleray:
> `.form-select` arrow doesn't adapt to rtl (stays on the right side), but I tested by adding `dir="rtl"` to `<html>` tag, I don't know if my test process is wrong, if we build specific css for rtl or something else.

Good catch! I've just pushed a fix and the test instance is updating 🙂

lithrel (approved):
:+1:

PR #1324, opened 2021-03-03
PLANET-5035 Make sure author image is full height

mleray:
### Description

See [PLANET-5035](https://jira.greenpeace.org/browse/PLANET-5035): this is an old ticket that was closed but apparently the issue comes back from time to time in very specific cases (probably due to the author image dimensions?) This fix should solve it for good. 😬


comzeradd (approved):
:+1:

PR #1325, opened 2021-03-04
PLANET-5983: Add variables

Inwerpsel:

---

Easiest to review with whitespace changes hidden: https://github.com/greenpeace/planet4-master-theme/pull/1325/files?diff=split&w=1

Current proposed migration path:
* Add all variables needed to recreate the campaign themes (CSS overrides living under https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/tree/d90243cf861c5e5909db59c546516bc6f58315cb/assets/src/styles/campaigns, as well as the sidebar options.
* Recreate the themes (either by adding a JSON/CSS file for each one in our code, or by allowing themes to be created and saved in the database without code change)
* Switch all existing campaigns to use these instead of the CSS overrides.
* Remove the old override based code.

In this PR (might split up further):
* Add variables unless adding them has a significant risk for regressions (e.g. have to change our CSS logic to make the variable work, only had to make minor changes which seem safe).
* I removed some declarations that are not used or not needed, when they also prevented variables from working. Also some minor modifications. I added explanatory comments for most of these in the PR.
* In some cases I needed to add hover properties that weren't present because they are the same as the non-hover property in our theme, but were being customized in campaign themes. e.g. https://github.com/greenpeace/planet4-master-theme/pull/1325/files#diff-2bc1c6e089a7a01997bb250bca7b661343c35a1512e49d22a11b1f00ef03b783R183

The test instance has data from GPI, so we can test all current campaigns if they still look the same with these changes. https://www-dev.greenpeace.org/test-nix/?s=&orderby=_score&f%5Bctype%5D%5BCampaign%5D=4

Note that if a campaign uses one of the campaign themes, the experimental theme editor will not work. This is not a problem since we intend to replace those themes altogether.

### DIFF
--- a/assets/src/scss/components/_buttons.scss
+++ b/assets/src/scss/components/_buttons.scss
@@ -12,23 +12,25 @@
 .btn,
 .wp-block-button a,
 .wp-block-file .wp-block-file__button {
+  --button-- {
+    font-family: $roboto;
+    text-align: center;
+    font-weight: bold;
+    border-radius: 4px;
+    padding: 0 30px;
+    transition-property: color, background-color, border-color;
+    transition-duration: 150ms;
+    transition-timing-function: linear;
+  }
   display: inline-block;
   position: relative;
-  font-family: $roboto;
   font-size: $font-size-sm;
-  text-align: center;
   text-decoration: none;
   color: $white;
-  font-weight: bold;
-  border-radius: 4px;
   border: 1px solid transparent;
   cursor: pointer;
   line-height: 3;
-  padding: 0 $n30;
   appearance: none;
-  transition-property: color, background-color, border-color;
-  transition-duration: 150ms;
-  transition-timing-function: linear;
 
   &:hover,
   &:focus,
@@ -70,23 +72,32 @@
   line-height: 3.1;
 }
 
+$primary-button-box-shadow: 0 2px 5px rgba(0, 0, 0, .25);
+
 .btn-primary,
 .wp-block-button.is-style-cta a {
-  background-color: $orange;
-  color: var(--btn-primary-color, $white);
-  box-shadow: 0 2px 5px rgba(0, 0, 0, .25);
+  --button-primary-- {
+    background: $orange;
+    color: $white;
+    // Adding this here so that the variable exists, though technically not needed.
+    // This allows for the hover variable to be previewed.
+    border-color: transparent;
+    box-shadow: $primary-button-box-shadow;
 
-  &:hover,
-  &:focus,
-  &:active {
-    background-color: $orange-hover;
-    border-color: $orange-hover;
-  }
+    &:hover,
+    &:focus,
+    &:active {
+      background: $orange-hover;
+      color: $white;
+      border-color: $orange-hover;
+      box-shadow: $primary-button-box-shadow;
+    }
 
-  &:disabled,
-  &.disabled,
-  &[disabled] {
-    background-color: $orange;
+    &:disabled,
+    &.disabled,
+    &[disabled] {
+      background: $orange;
+    }
   }
 }
 
@@ -94,36 +105,39 @@
 .wp-block-button.is-style-secondary a,
 [class="wp-block-button"] a,
 .wp-block-file .wp-block-file__button {
-  background-color: transparentize($white, .7);
-  border-color: $dark-blue;
-  color: $dark-blue;
-  box-shadow: none;
+  --button-secondary-- {
+    background: transparentize($white, .7);
+    border-color: $dark-blue;
+    color: $dark-blue;
+    box-shadow: none;
+
+    &:visited {
+      color: $dark-blue;
+    }
+
+    &:hover,
+    &:focus {
+      background: $dark-blue;
+      border-color: $dark-blue;
+      color: $white;
+    }
+
+    &:active {
+      background: $active-blue;
+      border-color: $active-blue;
+    }
+
+    &:disabled,
+    &.disabled,
+    &[disabled] {
+      background: $white;
+      color: $dark-blue;
+    }
+  }
+
   white-space: nowrap;
   overflow: hidden;
   text-overflow: ellipsis;
-
-  &:visited {
-    color: $dark-blue;
-  }
-
-  &:hover,
-  &:focus {
-    background-color: $dark-blue;
-    border-color: $dark-blue;
-    color: $white;
-  }
-
-  &:active {
-    background-color: $active-blue;
-    border-color: $active-blue;
-  }
-
-  &:disabled,
-  &.disabled,
-  &[disabled] {
-    background-color: $white;
-    color: $dark-blue;
-  }
 }
 
 // The new core buttons block wrap in "wp-block-buttons" class which apply
@@ -151,21 +165,29 @@
   }
 }
 
+$donate-button-box-shadow: 0 2px 5px rgba(0, 0, 0, .25);
+
 .btn-donate,
 .wp-block-button.is-style-donate a {
-  background-color: var(--btn-donate-background-color, $aquamarine);
-  color: var(--btn-donate-color, $grey-80);
-  line-height: 1.65;
-  min-width: 180px;
-  margin: 0;
-  padding: 2px $n30;
-  box-shadow: 0 2px 5px rgba(0, 0, 0, .25);
-  text-transform: none;
+  --button-donate-- {
+    background: $aquamarine;
+    color: $grey-80;
+    min-width: 180px;
+    padding: 2px 30px;
+    box-shadow: $donate-button-box-shadow;
+    border-width: 1px;
+    border-color: transparent;
 
-  &:hover,
-  &:focus {
-    color: transparentize($grey-80, 0.2);
+    &:hover,
+    &:focus {
+      background: $aquamarine;
+      color: transparentize($grey-80, 0.2);
+      box-shadow: $donate-button-box-shadow;
+    }
   }
+  line-height: 1.65;
+  margin: 0;
+  text-transform: none;
 
   @include large-and-up {
     font-size: $font-size-md;
--- a/assets/src/scss/layout/_breadcrumbs.scss
+++ b/assets/src/scss/layout/_breadcrumbs.scss
@@ -31,7 +31,6 @@
   .tag-wrap {
     display: inline;
     position: relative;
-    color: $link-color;
   }
 
   .tag-item--main + .tag-wrap,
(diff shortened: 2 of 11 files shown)


src/PostCampaign.php
```diff
@@ -456,34 +490,6 @@ public static function replace_font_aliases( array $css_vars ): array {
 		return $css_vars;
 	}
 
-	/**
```
Inwerpsel, line 459: I couldn't find any usages of this, I assume it's not used anymore. @pablocubico can you confirm?

pablocubico: No, I think the only use was in the line you deleted in this same PR.
If I recall correctly, this was here to satisfy a weird requirement, we had a spec for the hover colors (the "campaign_primary_color" was actually the hover color), but not for the idle state. 🤷

templates/html-header.twig
```diff
@@ -37,6 +37,12 @@
 	{% if css_vars %}
 		{% include 'css-variables.twig' %}
 	{% endif %}
+	<script>
```
Inwerpsel, line 40: This is just a quick fix for now to ensure the theme's properties are set early enough. It does the same thing as https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/d90243cf861c5e5909db59c546516bc6f58315cb/assets/src/theme/initializeThemeEditor.js#L37-L40 but that runs too late so it causes a flicker effect.

assets/src/scss/styleguide/src/components/_buttons.scss
```diff
@@ -12,23 +12,25 @@
 .btn,
 .wp-block-button a,
 .wp-block-file .wp-block-file__button {
+  --button-- {
+    font-family: $roboto;
+    text-align: center;
+    font-weight: bold;
+    border-radius: 4px;
+    padding: 0 30px;
+    transition-property: color, background-color, border-color;
```
Inwerpsel, line 21: These transition related properties are mostly made variable to be able to disable the transition, and not to fine tune it. The transition causes a weird effect for the climate theme, which uses gradient buttons. 
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/3cfd56585552a4e3df8e43ad3e442d2ddc7d5278/assets/src/styles/campaigns/themes/_theme_climate.scss#L22-L25
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/3cfd56585552a4e3df8e43ad3e442d2ddc7d5278/assets/src/styles/campaigns/themes/_theme_climate.scss#L35-L39

Inwerpsel: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/3cfd56585552a4e3df8e43ad3e442d2ddc7d5278/assets/src/styles/campaigns/themes/_theme_climate.scss#L22-L25

Inwerpsel: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/3cfd56585552a4e3df8e43ad3e442d2ddc7d5278/assets/src/styles/campaigns/themes/_theme_climate.scss#L35-L39

assets/src/scss/layout/_breadcrumbs.scss
```diff
@@ -31,7 +31,6 @@
   .tag-wrap {
     display: inline;
     position: relative;
-    color: $link-color;
```
Inwerpsel, line 34: I checked which things have the `tag-wrap` class, and all of them only have `tag-item` children, which also set the same color. So we can safely remove this one.

assets/src/scss/layout/_footer.scss
```diff
@@ -60,15 +62,11 @@
     padding: 55px 0 $n25;
   }
 
-  p {
-    font-family: $roboto;
-  }
-
-  a {
-    color: var(--footer_links_color, inherit);
+  a --footer--links-- {
```
Inwerpsel, line 65: Note that there is both this one which targets any `a` tag in the footer, but there's also the `.footer-links` class, which is a menu. Could be confusing, maybe we need to rename one of the two?

assets/src/scss/layout/_footer.scss
```diff
@@ -60,15 +62,11 @@
     padding: 55px 0 $n25;
   }
 
-  p {
-    font-family: $roboto;
```
Inwerpsel, line 64: Also not needed because it already has this font family.

assets/src/scss/layout/_footer.scss
```diff
@@ -78,7 +76,9 @@
 }
 
 .footer-social-media {
-  color: var(--footer_links_color, $grey-20);
+  _-- {
+    color: $grey-20;
```
Inwerpsel, line 80: Technically putting this together with the `a` selector below, which sets the hover color, would be more straightforward. But I preserved it so we don't change the CSS specificity.

assets/src/scss/layout/_footer.scss
```diff
@@ -235,7 +245,6 @@
   line-height: 1;
 
   .footer-social-media {
-    color: var(--footer_links_color, $grey-20);
```
Inwerpsel, line 238: Same property and value already set outside media query.

assets/src/scss/layout/_navbar.scss
```diff
@@ -70,6 +70,14 @@ $min-height: 40px;
 $navbar-default-height: 60px;
 
 .top-navigation {
+  _-- {
+    background: transparentize($dark-blue, 0.2);
+    border-bottom-width: 0;
```
Inwerpsel, line 75: Had to add these because the climate theme has a white border. We could also check if we can remove it from that theme. If I remember correctly it was added because the color of the fixed navbar didn't contrast enough with the content.

assets/src/scss/layout/_navbar.scss
```diff
@@ -709,14 +727,15 @@ $navbar-default-height: 60px;
...
+      left: 10%;
+    }
+    max-height: calc(100vh - #{$menu-height-large});
     position: absolute;
-    height: 344px;
-    width: 80%;
-    left: 10%;
     overflow-x: hidden;
     overflow-y: hidden !important;
-    max-height: 344px;
```
Inwerpsel, line 718: I added `height` since it seems useful to try different values for this, especially if other properties are changed that affect the height needed. I removed `max-height` because otherwise you'd have to change both of them, and we set them to the same value anyway.

webpack.config.js
```diff
@@ -44,9 +52,9 @@ module.exports = {
             options: {
               ident: 'postcss',
               plugins: () => [
-                dashDash(),
-                 cssVariables( { preserve: true, exportVarUsagesTo: allCssVars } ),
-                 require('autoprefixer'),
+                dashDash({ mediaQueryAliases, mediaQueryAtStart: true }),
```
Inwerpsel, line 55: I first implemented the media query prefix at the start of the variable, but then realized it also makes sense to put it right before the CSS property name. Couldn't make up my mind, so I made it configurable.

assets/src/scss/layout/_footer.scss
mleray: Good point, maybe the class could be renamed `.footer-menu`?

Inwerpsel: Makes sense :ok_hand: Probably better to rename the class separately, but we can already give it the right name instead of generating it from the class (just pushed this change) https://github.com/greenpeace/planet4-master-theme/pull/1325/files#diff-8b7c0326c01b36e050f9be000f365759c8fd1ba4653b49ddd09eec4bba3f3f11R152

assets/src/scss/components/_buttons.scss
```diff
@@ -12,23 +12,25 @@
...
+    border-radius: 4px;
+    padding: 0 30px;
+    transition-property: color, background-color, border-color;
+    transition-duration: 150ms;
+    transition-timing-function: linear;
+  }
   display: inline-block;
   position: relative;
-  font-family: $roboto;
   font-size: $font-size-sm;
```
mleray, line 27: Why is this font-size not a variable? 🤔

assets/src/scss/components/_buttons.scss
```diff
@@ -12,23 +12,25 @@
...
-  text-align: center;
   text-decoration: none;
   color: $white;
```
mleray, line 29: Also this color maybe? (and later in the file, L32, the line-height?)

assets/src/scss/components/_buttons.scss
Inwerpsel: Because there's also the `button-small` and `button-large` classes. And the donate button also sets a custom size, but only for `large-and-up`. So I'm not sure if it makes sense to also add it to the "base" button class.

assets/src/scss/components/_buttons.scss
```diff
@@ -151,22 +165,30 @@
...
+    border-color: transparent;
+
+    &:hover,
+    &:focus {
+      background: $aquamarine;
+      color: transparentize($grey-80, 0.2);
+      box-shadow: $donate-button-box-shadow;
+    }
+  }
   line-height: 1.65;
```
mleray, line 188: This line-height should also maybe be a variable?

assets/src/scss/components/_buttons.scss
Inwerpsel: This one I didn't make variable because all other classes override it. Maybe we can remove this declaration too, but probably best separately.

assets/src/scss/components/_buttons.scss
Inwerpsel: We probably want to look at how to configure button spacing more widely first, which might involve changing the CSS a bit to make it more consistent. Then we probably need to add less variables, or we can add more appropriate / easier to work with ones.

assets/src/scss/layout/_navbar.scss
```diff
@@ -616,10 +632,12 @@ $navbar-default-height: 60px;
 }
 
 .country-list {
+  _-- {
+    background: $x-dark-blue;
```
mleray, line 636: Maybe we should use the `background-color` property here instead of just `background`? It's more specific

Inwerpsel: With background all values you'd put in `background-color` would work the same way, but you can do more things like gradients, images.

Inwerpsel: In this case it already used `background`, in some other cases (like buttons) I changed it.

assets/src/scss/components/_buttons.scss
mleray: There is actually one use case where I think it's used, it's for the Button block when users manually select the white text color... But indeed it doesn't need to be investigated in this PR 🙂

mleray (approved):
Looks good to me 👍

comzeradd (approved):

PR #1326, opened 2021-03-04
PLANET-5990 Restrict max-width on floated elements

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-5990

---

On images this has no effect, since they already have `max-width`,
but other elements can expand outside of the post container when floating.

> _[Test post](https://www-dev.greenpeace.org/test-proteus/story/1007/social-media-test/)_

Required for greenpeace/planet4-plugin-gutenberg-blocks#524


mleray (approved):
Looks good to me 👍

PR #1327, opened 2021-03-05
PLANET-5995: Load icons from less deep level

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-5995

---

`require.context` has some kind of problem with going up 3 levels. It causes it to swap a closing bracket's position resulting in invalid syntax of the output. `src` makes sense to have the icons in anyway since they're only used in the build.


lithrel (approved):
:+1:

PR #1328, opened 2021-03-09
PLANET-5501 Update clear search filters button styles

mleray:
### Description

When updating the button styles for [PLANET-5501](https://jira.greenpeace.org/browse/PLANET-5501) we forgot the one to clear search filters 😬 We probably missed it because it didn't have the `btn` class so I added it, which also avoids some duplicate code

Screenshot 2021-03-09 at 09 42 24

### Testing

You can see the updated button [here](https://www-dev.greenpeace.org/test-umbriel/?s=&orderby=_score&f%5Bcat%5D%5BNature%5D=3) for instance


sagarsdeshmukh (approved):
LGTM 👍

PR #1331, opened 2021-03-10
Prevent unneeded prefixes, we autoprefix

Inwerpsel:
Ref: no ticket for now

---

Since we use autoprefixer, we can use a stylelintrule to detect any prefixes in our source that would be added during build anyway. This adds the rule and fixes violations, except `-webkit-box-orient` which is buggy.


mleray (approved):
Looks good to me 👍 We should check at some point whether or not that `webkit-box-orient` is really needed in the author block 😁

PR #1332, opened 2021-03-11
test against version we use

Inwerpsel:
Ref: 

---

Our tests are running against the latest version of WP, not against the one we use in production. This came up by a new issue in 5.7 that broke a test.

Could need more changes than just the WP version since it's still failing.


lithrel (approved):
:+1:

PR #1333, opened 2021-03-11
PLANET-5748 Update password form UI as per design system

sagarsdeshmukh:
Ref: [JIRA 5748](https://jira.greenpeace.org/browse/PLANET-5748)

---
eg.
Page: https://www-dev.greenpeace.org/sagardev/test-password-protected-page/
Post: https://www-dev.greenpeace.org/sagardev/test-password-protected-post/
Campaign: https://www-dev.greenpeace.org/sagardev/campaign/test-password-protected-campaign/

### DIFF
--- a/assets/src/scss/pages/_page.scss
+++ b/assets/src/scss/pages/_page.scss
@@ -89,12 +89,72 @@ div.page-template {
   }
 }
 
-.error-message {
-  color: $orange-hover;
+.password-form {
+  .input-text{
+    font-size:16px;
+    line-height:1.6;
+    font-weight:400;
+    margin-bottom:16px;
+    position:relative;
+  }
+
+  .input-text input{
+    border:1px solid $grey-20;
+    border-radius:4px;
+    display:inline-block;
+    font-size:16px;
+    padding:15px 14px 8px;
+    min-width:400px;
+
+    &.is-invalid {
+      border-color: $dark-orange;
+      background: $white;
+
+      &:focus {
+        border-color: $dark-orange;
+      }
+    }
+  }
+
+  .input-text input:hover{
+    border:1px solid $grey-40;
+  }
+
+  .input-text input:focus{
+    border-color:rgba(0,109,253,0.75);
+    outline:0 none;
+  }
+
+  .invalid-pw-feedback {
+    font-family: $roboto;
+    background: $dark-orange;
+    color: $white;
+    border-radius: 4px;
+    font-size: 12px;
+    font-weight: 500;
+    padding: 6px 12px;
+    width: fit-content;
+    position: relative;
+    pointer-events: none;
+    margin-top: 6px;
+
+    &:after {
+      bottom: 100%;
+      border: solid transparent;
+      content: "";
+      height: 0;
+      width: 0;
+      position: absolute;
+      border-bottom-color: $dark-orange;
+      border-width: 5px;
+      right: auto;
+      left: 16px;
+
+      html[dir="rtl"] & {
+        left: auto;
+        right: 16px;
+      }
+    }
+  }
 }
 
-.password-field {
-  padding-top: $space-xs;
-  padding-bottom: $space-xs;
-  margin-bottom: $space-xs;
-}
--- a/templates/single-page.twig
+++ b/templates/single-page.twig
@@ -7,13 +7,13 @@
 
 	<div class="page-template">
 
-		<form id="password-form" class="" action="{{login_url}}?action=postpass" method="post">
+		<form id="password-form" class="password-form" action="{{login_url}}?action=postpass" method="post">
 			<div class="mb-3">
 				<label for="pwbox-{{post.ID}}" class="form-label"><h6>{{ __( 'To see the content of this page, please enter your password  below', 'planet4-master-theme' ) }}</h6></label>
-				<div class="col-sm-8">
-					<input class="form-control password-field" name="post_password" id="pwbox-{{post.ID}}" type="password">
+				<div class="input-text">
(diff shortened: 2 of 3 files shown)


assets/src/scss/pages/_page.scss
```diff
@@ -89,12 +89,71 @@ div.page-template {
...
+  .input-text{
+    font-size:16px;
+    line-height:1.6;
+    font-weight:400;
+    margin-bottom:16px;
+    position:relative;
+  }
+
+  .input-text input{
+    border:1.5px solid $grey-20;
```
mleray, line 102: The border width should also be 1px here, at the moment it looks a bit weird on hover

assets/src/scss/pages/_page.scss
```diff
@@ -89,12 +89,71 @@ div.page-template {
...
+  .input-text input:hover{
+    border:1px solid $grey-40;
+  }
+
+  .input-text input:focus{
+    border-color:rgba(0,109,253,0.75);
+    outline:0 none;
+  }
+
+  .invalid-pw-feedback {
```
mleray, line 128: I think the font family for the error message should be Roboto, at the moment it's Lora

templates/single-page.twig
```diff
@@ -7,13 +7,13 @@
...
 
-		<form id="password-form" class="" action="{{login_url}}?action=postpass" method="post">
+		<form id="password-form" class="password-form" action="{{login_url}}?action=postpass" method="post">
 			<div class="mb-3">
 				<label for="pwbox-{{post.ID}}" class="form-label"><h6>{{ __( 'To see the content of this page, please enter your password  below', 'planet4-master-theme' ) }}</h6></label>
-				<div class="col-sm-8">
-					<input class="form-control password-field" name="post_password" id="pwbox-{{post.ID}}" type="password">
+				<div class="input-text">
+					<input name="post_password" id="pwbox-{{post.ID}}" type="password" {% if not is_password_valid %}class="is-invalid"{% endif %}>
+					{% if not is_password_valid %} <div class="invalid-pw-feedback">{{ __( 'This field is required', 'planet4-master-theme' ) }}</div>{% endif %}
```
mleray, line 15: I find this message a bit confusing, shouldn't it say something more along the lines of "Invalid password, please try again"? 🤔 Unless the user actually didn't enter anything

mleray (changes requested):
Added a couple of comments but actually for the CSS, since we already have it for the error messages in the [EN form styles](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/styles/blocks/ENForm/components/_enform.scss#L103): maybe it would be better to move this code to the styleguide in `_forms.scss`? Then we avoid some duplication. The only thing is that we need to make sure that it still works for other input types too (checkboxes, radios, etc), so maybe it would be better to do this as part of [PLANET-5984](https://jira.greenpeace.org/browse/PLANET-5984) where I actually add the error messages to these 🤔

assets/src/scss/pages/_page.scss
magalifatome: Yes, I agree with Maud

assets/src/scss/pages/_page.scss
magalifatome: Yep, that's correct, thanks @mleray

templates/single-page.twig
magalifatome: Yes, we can apply this message @mleray and @sagarsdeshmukh

sagarsdeshmukh:
> Added a couple of comments but actually for the CSS, since we already have it for the error messages in the [EN form styles](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/styles/blocks/ENForm/components/_enform.scss#L103): maybe it would be better to move this code to the styleguide in `_forms.scss`? Then we avoid some duplication. The only thing is that we need to make sure that it still works for other input types too (checkboxes, radios, etc), so maybe it would be better to do this as part of [PLANET-5984](https://jira.greenpeace.org/browse/PLANET-5984) where I actually add the error messages to these 🤔

Hi @mleray , Thanks for the review 🙂  yes indeed I was initially tried to reuse the  [EN form error message styles](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/styles/blocks/ENForm/components/_enform.scss#L103) but that was closely attached with EN forms elements, then I copied the required part from their, but it would be a nice idea to move this reusable style in styleguide.
We will optimise it in Planet-5984. I updated the PR with review fixes, please have a look.

mleray (approved):
Looks good to me 👍 It worked well for me on all 3 test pages (post, page & campaign) but for Magali apparently the error didn't show on pages 🤷‍♀️

sagarsdeshmukh:
> Looks good to me 👍 It worked well for me on all 3 test pages (post, page & campaign) but for Magali apparently the error didn't show on pages 🤷‍♀️

Thanks @mleray , I got the [issue](https://jira.greenpeace.org/browse/PLANET-5748?focusedCommentId=240928&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-240928) but won’t have any solution on it 😕

mleray:
> > Looks good to me 👍 It worked well for me on all 3 test pages (post, page & campaign) but for Magali apparently the error didn't show on pages 🤷‍♀️
> 
> Thanks @mleray , I got the [issue](https://jira.greenpeace.org/browse/PLANET-5748?focusedCommentId=240928&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-240928) but won’t have any solution on it 😕

Just saw that on the ticket, we can discuss it during DS, maybe someone will have an idea of how to solve this!

PR #1334, opened 2021-03-12
PLANET-5848 Remove blue caption from classic editor images

mleray:
### Description

See https://jira.greenpeace.org/browse/PLANET-5848

### Testing
- [Old version](https://www-dev.greenpeace.org/test-jupiter/explore/ships/arctic-sunrise/) with the blue caption overlay
- [New version](https://www-dev.greenpeace.org/test-tavros/explore/ships/arctic-sunrise/) with the regular caption styles

### DIFF
--- a/assets/src/scss/pages/post/_article-content.scss
+++ b/assets/src/scss/pages/post/_article-content.scss
@@ -285,7 +285,7 @@
 .wp-caption {
   position: relative;
   display: inline-block;
-  width: auto !important;
+  width: auto;
   max-width: 100%;
 
   &.alignnone {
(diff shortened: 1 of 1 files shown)


assets/src/scss/pages/post/_article-content.scss
```diff
@@ -299,7 +299,7 @@
   }
 
   img {
-    max-width: 100%;
+    width: 100%;
```
lithrel, line 302: This ends up adapting the image width to the caption width, which might generate unexpected layout changes. But I'm not sure how to handle that, there are several cases:

Testing with an image 150x150 to see the stretch

What I've inserted, disabling `width:100%` (it falls back to a `max-width: 100%;` from `div.page-template img` on a page; appears the same in the editor):
Screenshot from 2021-03-12 10-12-08

Current PR:
On insertion
Screenshot from 2021-03-12 10-06-37
When adding an alignement
Screenshot from 2021-03-12 10-10-00

(Weirdly, on insertion there is no alignement, and once you add one you can only change it, not remove it (but it's not part of this ticket))

My guess would be, we should probably not change anything on how images are currently streched but maybe there's a design choice there ? No idea of the impact it would have on current content

mleray: That's a good point, it's not the scope of this ticket to fix this and it might cause issues. I'll revert the changes related to the width 👍

lithrel (approved):
:+1:

PR #1335, opened 2021-03-15
PLANET-6004 Fix pdf link icon (hover) colors

mleray:
### Description

See https://jira.greenpeace.org/browse/PLANET-6004

### Testing

You can see the fixed links & buttons on [this page](https://www-dev.greenpeace.org/test-leda/links-and-buttons-with-icons/) for example.

Note: the issue with the width of the secondary buttons will be solved as part of [PLANET-5980](https://jira.greenpeace.org/browse/PLANET-5980).

### Related PRs
- [styleguide](https://github.com/greenpeace/planet4-styleguide/pull/107)
- [blocks](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/531)

### DIFF
--- a/assets/src/js/external_links.js
+++ b/assets/src/js/external_links.js
@@ -1,7 +1,7 @@
 export const setupExternalLinks = () => {
   const siteURL = window.location.host;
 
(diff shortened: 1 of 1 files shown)


assets/src/js/pdf_icon.js
```diff
@@ -13,5 +13,7 @@ export const setupPDFIcon = () => {
     }
 
     link.classList.add('pdf-link');
+    const color = window.getComputedStyle(link).getPropertyValue('color');
```
Inwerpsel, line 16: Pretty sure we don't want to be calling `getComputedStyle` during page load since it causes reflow.

Inwerpsel (approved):

PR #1337, opened 2021-03-16
PLANET-5983: Move in editorStyle from blocks plugin

Inwerpsel:

---

Needs to be merged together with https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/533

This moves in all parts of the editor styles that are not relevant to the blocks in the blocks plugin. After this there will be more parts of the styleguide that can be moved into master theme.


lithrel (approved):
:+1:

PR #1338, opened 2021-03-16
PLANET-5983: Move styleguide parts to theme

Inwerpsel:

---
This contains the following changes:
- files from the styleguide, with their history (see https://github.com/greenpeace/planet4-styleguide/pull/109/files)
- update `style.scss` and `editorStyle.scss` to use the new location
- remove some vendor prefixes (the styleguide doesn't have the rule)

No testing tip except it should behave exactly the same as before.

### DIFF
--- a/assets/src/scss/base/_body.scss
+++ b/assets/src/scss/base/_body.scss
@@ -1,0 +1,27 @@
+html,
+body {
+  overflow-x: hidden;
+  scroll-behavior: smooth;
+  scroll-padding-top: 4rem;
+}
+
+body {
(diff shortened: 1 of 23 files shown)


comzeradd:
I think components should remain in the styleguide. There are being used in other repos too.

Inwerpsel:
> I think components should remain in the styleguide. There are being used in other repos too.

Right :facepalm: I knew I was forgetting something. Luckily I kept the branch from before I did that :sweat_smile: https://github.com/greenpeace/planet4-master-theme/pull/1339

PR #1339, opened 2021-03-16
Move but keep components

Inwerpsel:
Ref: 

---


Inwerpsel:
We ended up moving in the entire styleguide https://github.com/greenpeace/planet4-master-theme/pull/1344

PR #1341, opened 2021-03-22
PLANET-6021 Hide core upgrade notice

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6021

---

We already watch for new releases, so this notice only causes confusion to users.
To test: confirm that the notice is not shown at the top of every admin page.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -132,6 +132,16 @@ add_action(
 	1
 );
 
+/**
+ * Hide core updates notification in the dashboard, to avoid confusion while an upgrade is already in progress.
+ */
+function hide_wp_update_nag() {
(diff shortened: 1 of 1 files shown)


functions.php
```diff
@@ -132,6 +132,14 @@ function () {
 	1
 );
 
+// Hide core updates notification in the dashboard, to avoid confusion while an upgrade is already in progress.
+function hide_wp_update_nag() {
```
Inwerpsel, line 136: `update_nag` are WP's words, not mine :grin:

comzeradd (approved):

PR #1342, opened 2021-03-25
PLANET-6028 explicitly define bootstrap as dep of parent-style

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6028

---

We didn't define this earlier but it didn't cause any issues. However since last deploy the Japanese child theme is now loading bootstrap after the master theme CSS, which changes the order of CSS and so causes most `a` tags to be underlined (and likely more issues). Could be it was already loading like this before but it only started causing issues with Bootstrap 5 upgrade.


lithrel (approved):
:+1:

PR #1350, opened 2021-03-31
PLANET-6034 Disable Bootstrap 5 XXL breakpoint

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-6034

---

There should be no visual change above `1400`. The last valid breakpoint should be at `1200`.


mleray (commented):
I had added a small change for the new breakpoint in [_page.scss](https://github.com/greenpeace/planet4-master-theme/blob/master/assets/src/scss/pages/_page.scss#L22), I guess we should remove that then? 🤔

comzeradd:
> I had added a small change for the new breakpoint in [_page.scss](https://github.com/greenpeace/planet4-master-theme/blob/master/assets/src/scss/pages/_page.scss#L22), I guess we should remove that then?

Thanks! I would have missed that. I'll add it here.

mleray (approved):
Looks good to me 👍

Inwerpsel (approved):

PR #1351, opened 2021-03-31
Fix Finland localized name

comzeradd:
This came as a request from GP Nordic. It turns out we had it wrong all along.

### DIFF
--- a/templates/countries.json
+++ b/templates/countries.json
@@ -231,7 +231,7 @@
       "codes": ["FI"],
       "lang": [
         {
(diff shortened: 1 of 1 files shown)


lithrel (approved):
:+1: 
We're gonna have to ping the few NROs overriding the country selector too https://github.com/search?l=JSON&q=org%3Agreenpeace+Suomea&type=Code

comzeradd:
Good catch! Not sure why they override it, but I'll send them a notice.

PR #1353, opened 2021-04-02
PLANET-6043 Fix search pages image aspect ratios

mleray:
### Description 

See https://jira.greenpeace.org/browse/PLANET-6043

We added this fix to the [search page results](https://github.com/greenpeace/planet4-master-theme/blob/master/templates/tease-search.twig#L19) when upgrading to Bootstrap 5, but forgot to update the other templates that feature search results.
This PR proposes a fix using the new BS 5 classes, but there is [another PR](https://github.com/greenpeace/planet4-master-theme/pull/1354) without the BS 5 classes, which avoids some duplication 🙂

### DIFF
--- a/templates/tease-author.twig
+++ b/templates/tease-author.twig
@@ -1,4 +1,4 @@
-<li id="result-row-{{ post.ID }}" class="d-flex search-result-list-item">
+<li id="result-row-{{ post.ID }}" class="d-flex align-items-start search-result-list-item">
 	<img
 		class="d-flex search-result-item-image"
 		src="{{ post.thumbnail.src('thumbnail') }}"
--- a/templates/tease-taxonomy-post.twig
+++ b/templates/tease-taxonomy-post.twig
@@ -1,4 +1,4 @@
-<li id="result-row-{{ post.ID }}" class="d-flex search-result-list-item">
+<li id="result-row-{{ post.ID }}" class="d-flex align-items-start search-result-list-item">
 	<img
 		class="d-flex search-result-item-image"
 		src="{{ post.thumbnail.src('thumbnail') ?? dummy_thumbnail }}"


templates/tease-author.twig
```diff
@@ -1,4 +1,4 @@
-<li id="result-row-{{ post.ID }}" class="d-flex search-result-list-item">
+<li id="result-row-{{ post.ID }}" class="d-flex align-items-start search-result-list-item">
```
Inwerpsel, line 1: Not sure if these bootstrap single declaration classes are really an improvement :thinking: Doesn't it scatter the styling between the templates and our own `search-result-list-item` class? That makes it easier for issues like this one to go unnoticed.

mleray: I thought about that, but our css code for this `search-result-list-item` class is actually also split between 3 css files, so I don't know if it would be much better 🤔

mleray: That being said, it's true that 2 of these css files are basically the same so there's room for improvement 😬 
- [_category.scss](https://github.com/greenpeace/planet4-master-theme/blob/master/assets/src/scss/pages/_category.scss)
- [_page_type.scss](https://github.com/greenpeace/planet4-master-theme/blob/master/assets/src/scss/pages/_page_type.scss)

comzeradd: I'm generally in favor of using more Bootstrap classes (especially for things like alignment and spacing). This can potentially make it easier for child themes too.

mleray: Nvm, found a solution and opened [another PR](https://github.com/greenpeace/planet4-master-theme/pull/1354) 😄

Inwerpsel: @comzeradd Can you explain how it makes things easier? Not saying there is no benefit to it, but I don't see it atm, and I do see some downsides.

comzeradd: The main benefit is that NROs will know what this class is doing and that we can't change its behavior.

Inwerpsel: Doesn't that mean we need to change the markup each time we configure a flex property? It feels a lot easier to work with if these things are directly in CSS and you don't need to know how all of these classes are named, they're not exactly the same name as the property represents.
```html
<li 
  id="result-row-{{ post.ID }}"
  class="d-flex align-items-start justify-content-around flew-grow-1 flex-fill-md search-result-list-item"
>

```

comzeradd: Bootstrap classes follow the css name attribute they refer to (at least for the most part). But yes there is a trade off. Doing changes in markup vs maintaining less css code.

comzeradd:
@mleray we can close this one, right?

mleray:
> @mleray we can close this one, right?

If we decide to go for the [other solution](https://github.com/greenpeace/planet4-master-theme/pull/1354), then yes! 😄

templates/tease-author.twig
Inwerpsel: Just because in absolute numbers we have less lines of CSS, doesn't necessarily mean we will have an easier time maintaining our CSS. There's the extra level of complexity (move a part of the declarations to another file, written as a class which you need to know how what it translates to). It's also a bit unclear in which cases we should use the bootstrap class technique, and in which cases we still use properties, so we'll end up with a mix of both.

comzeradd: We already do that. We don't have our own grid css code, we rely on Bootstrap for that. Same for most of the forms UI (eg. `form-control`). It would be more consistent to utilize Bootstrap even more. In then end, it's easier to understand what `align-items-start` does that trying to figure out where and how `search-result-list-item` is being used.

mleray:
Closed in favor of https://github.com/greenpeace/planet4-master-theme/pull/1354

templates/tease-author.twig
Inwerpsel: Well for the grid and "semantical" classes like `form-control`, these result in properties that make sense on each screen size (in case of the grid because it's designed for that). Other properties are much more likely to change on different screen sizes, like `align-items` and `justify-content` could be different if we show 4 items in a row on big screens and only 1 on mobile.

PR #1354, opened 2021-04-02
PLANET-6043 Fix search pages image aspect ratios

mleray:
### Description 

See https://jira.greenpeace.org/browse/PLANET-6043

We added this fix to the [search page results](https://github.com/greenpeace/planet4-master-theme/blob/master/templates/tease-search.twig#L19) when upgrading to Bootstrap 5, but forgot to update the other templates that feature search results. 
There is already a [PR open](https://github.com/greenpeace/planet4-master-theme/pull/1353) for that issue using the new BS 5 classes, but this PR gives another solution without using the BS 5 classes, which avoids some duplication 🙂

### DIFF
--- a/assets/src/scss/pages/search/_search-results.scss
+++ b/assets/src/scss/pages/search/_search-results.scss
@@ -28,6 +28,7 @@
       margin: 0 auto 24px;
       border-bottom: 1px solid $grey-20;
       position: relative;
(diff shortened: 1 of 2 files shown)


comzeradd (approved):
This looks like the simplest fix for now :+1:

PR #1355, opened 2021-04-05
PLANET-6045 Add relevant Acceptance tests

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-6045

---

First step in moving tests to app repos. These are the tests that test theme functionality.


lithrel (approved):
:+1:

lithrel:
I realize now that `tests/data/test-campaign.lorem-ipsum.xml` was copied here, but not removed in `planet4-base`, and there's no command to copy ` tests/data` folders to `tests/_data`, probably a step missing or just a test ?

comzeradd:
Good catch! I have the `_data` folder locally already so I missed that. I removed it from base and added an extra `rsync` command.

PR #1358, opened 2021-04-07
PLANET-6053 Remove inline-block from icon links to properly wrap

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-6053

---

`inline-block` forces links with long text to try to fit on the same line causing weird white space on paragraphs.

Not sure why we had that initially. It looks these were added when we removed `jQuery` from the js part of these links.


mleray (approved):
Tested on local, looks good to me 👍 This was indeed added during the jQuery removal but right now I can't remember why 🤔

PR #1360, opened 2021-04-13
PLANET-5983 Ensure theme fonts are loaded, add page section header variables

Inwerpsel:

---

While adding variables to the blocks CSS I find that some need more variables in master theme.

We also need to include the custom fonts for the default theme as well.

Also included a commit to use an existing variable in the twig template instead of a constant. These don't throw errors when undefined.

### DIFF
--- a/assets/src/scss/layout/_page-section.scss
+++ b/assets/src/scss/layout/_page-section.scss
@@ -1,4 +1,8 @@
-.page-section-header {
+.page-section-header _-- {
+  text-transform: none;
+  color: inherit;
+  font-weight: var(--headings--font-weight, bold);
+  font-size: 2.5rem;
   margin-bottom: $space-md;
 }
 
--- a/single-campaign.php
+++ b/single-campaign.php
@@ -100,15 +100,13 @@ $context['post_tags']       = implode( ', ', $post->tags() );
 $context['custom_styles']   = $custom_styles;
 $context['css_vars']        = PostCampaign::css_vars( $campaign_meta );
 
-if ( $theme_name && 'default' !== $theme_name ) {
-	$context['custom_font_families'] = [
-		'Montserrat:300,400,500,600,700,800',
-		'Kanit:400,500,600,800',
-		'Open+Sans:400,500,600,800',
-		'Anton:400,500,600,800',
-		'Amiko:400,500,600,800',
-	];
-}
+$context['custom_font_families'] = [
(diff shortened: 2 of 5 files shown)


planet-4:

:new_moon: [nix](https://www-dev.greenpeace.org/test-nix)

:watch: 2021.04.30 15:21:24

templates/html-header.twig
```diff
@@ -38,7 +38,7 @@
 	{% endif %}
 
 	{% if custom_styles %}
-		<style type="text/css">
+		<style>
```
Inwerpsel, line 41: This attribute is now deprecated

templates/head/google_fonts.twig
```diff
@@ -1,5 +1,51 @@
 <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,700,900&display=swap&subset=latin-ext"/>
 <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Lora:400,400i,700&display=swap&subset=latin-ext"/>
+{% if  custom_font_families %}
```
Inwerpsel, line 3: I suppose this won't affect performance, since only the needed fonts are loaded by the browser. The few additional bytes in the HTML should have no measurable impact.

single-campaign.php
```diff
@@ -100,15 +100,13 @@ static function ( $page ) {
...
-if ( $theme_name && 'default' !== $theme_name ) {
-	$context['custom_font_families'] = [
-		'Montserrat:300,400,500,600,700,800',
-		'Kanit:400,500,600,800',
-		'Open+Sans:400,500,600,800',
-		'Anton:400,500,600,800',
-		'Amiko:400,500,600,800',
-	];
-}
+$context['custom_font_families'] = [
```
Inwerpsel, line 103: I removed the check because some of these can be used for the default theme as well.

assets/src/scss/layout/_page-section.scss
```diff
@@ -1,4 +1,8 @@
-.page-section-header {
+.page-section-header _-- {
+  text-transform: none;
```
Inwerpsel, line 2: This is customized by current themes on this class, but base didn't have the corresponding styles. Not sure if it really makes sense as a separate option :thinking: I'll check some live content to see how it looks like without.

mleray (approved):
Looks good to me 👍

PR #1361, opened 2021-04-13
PLANET-6019: Verify Elasticpress activation before adding filters

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-6019
Related to: https://github.com/greenpeace/planet4-docker-compose/pull/85

On local instance, Elasticsearch container is disabled by default.
Elasticpress throws some exceptions when it is activated (no server to connect to), and once deactivated the Search functionality throws some more.

This fixes exceptions thrown when Elasticpress plugin is disabled, and keeps some links working in the result page.

## Test

- Disable Elasticpress 
  `docker-compose exec php-fpm wp plugin deactivate elasticpress`
- Go to a search page
  - on `master` branch, you should get an error 
  - on this branch, the page appears and result links are valid

### DIFF
--- a/src/Search.php
+++ b/src/Search.php
@@ -112,6 +112,10 @@ abstract class Search {
 	 * Add filters that are needed by both the initial page load and subsequent ajax page loads.
 	 */
 	public static function add_general_filters(): void {
+		if ( ! is_plugin_active( 'elasticpress' ) ) {
+			return;
+		}
+
 		// Call apply filters to catch issue in WPML's ElasticPress integration, which uses the wrong filter name.
 		add_filter(
 			'ep_formatted_args',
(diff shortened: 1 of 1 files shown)


Inwerpsel (approved):
Looks good :+1: Maybe we should also consider always running ElasticPress locally? It always runs in production, so devs might end up missing things. And disabling it makes the search act quite differently locally, which can cause confusion.

lithrel:
Yeah, I'm mostly disabling it because it throws periodic fatal errors in the background when it doesn't find an elasticsearch server (which is disabled by default), and multiple ones during a request, which pollutes logs a lot and is quite inconvenient during a debugging sessions, but I'll also look into how to fix that problem.

PR #1364, opened 2021-04-15
PLANET-6080 Add fonts in HTML head for campaigns

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6080

---

These fonts were loaded with an `import` in the stylesheet, which is being removed in https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/556. This PR adds them to the HTML so that Cloudflare APO can optimize them.

Both should be merged at the same time. This one should go first so we don't end up in a state where there are no fonts.

I moved the new code and existing fonts into a new template part.


lithrel (approved):
:+1: 
Tested with https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/556

PR #1365, opened 2021-04-15
PLANET-6082 Fix incorrect old name migration

Inwerpsel:

---

I forgot to update the name fallbacks for campaigns before merging. Also made it a bit more manageable.

How to test: Create a campaign and use the sidebar options. Each of them should result in the expected property being applied.


mleray (approved):

PR #1366, opened 2021-04-16
PLANET-5523 Add Beta Blocks option

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-5523

---

This adds the feature toggle for beta blocks used in by the plugin.

Also makes `tag.php` use a string literal instead of the class constant from the other repo, to avoid hassle when we change the old version name to `OldCovers`.

### DIFF
--- a/src/Features.php
+++ b/src/Features.php
@@ -22,6 +22,8 @@ class Features {
 
 	public const THEME_EDITOR_NON_LOGGED_IN = 'theme_editor_non_logged_in';
 
+	public const BETA_BLOCKS = 'beta_blocks';
+
 	/**
 	 * Get the features options page settings.
 	 *
@@ -88,6 +90,15 @@ class Features {
 					'planet4-master-theme-backend'
 				),
 				'id'   => self::THEME_EDITOR,
+				'type' => 'checkbox',
+			],
+			[
+				'name' => __( 'Allow beta blocks in post editor.', 'planet4-master-theme-backend' ),
+				'desc' => __(
+					'If enabled, beta blocks defined in <a href="https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/planet4-gutenberg-blocks.php">this file</a>. Not for production use.',
+					'planet4-master-theme-backend'
+				),
+				'id'   => self::BETA_BLOCKS,
 				'type' => 'checkbox',
 			],
 		];
--- a/tag.php
+++ b/tag.php
@@ -11,7 +11,6 @@
 
 use P4\MasterTheme\TaxonomyCampaign;
 use Timber\Timber;
-use P4GBKS\Blocks\Covers;
 use P4GBKS\Blocks\Articles;
 use P4GBKS\Blocks\HappyPoint;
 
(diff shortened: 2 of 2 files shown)


comzeradd:
We should also move the current beta blocks in this PR, so they don't disappear once we merge this (Accordion, Spreadsheet, Sub-pages).

Inwerpsel:
> We should also move the current beta blocks in this PR, so they don't disappear once we merge this (Accordion, Spreadsheet, Sub-pages).

They won't disappear, since the setting only affects whether a list of block names is allowed in the plugin:
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/4cad017c7d8d95614a1f4f166e8c6513dab86cb5/planet4-gutenberg-blocks.php#L139-L142
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/4cad017c7d8d95614a1f4f166e8c6513dab86cb5/planet4-gutenberg-blocks.php#L163-L165
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/4cad017c7d8d95614a1f4f166e8c6513dab86cb5/planet4-gutenberg-blocks.php#L248-L258

src/Features.php
```diff
@@ -90,6 +92,15 @@ private static function get_fields(): array {
 				'id'   => self::THEME_EDITOR,
 				'type' => 'checkbox',
 			],
+			[
+				'name' => __( 'Use beta blocks in posts', 'planet4-master-theme-backend' ),
```
mleray, line 96: Isn't this formulation a bit confusing since we have posts, pages and campaigns? 🤔

mleray (approved):
Looks good to me 👍

Inwerpsel:
> > We should also move the current beta blocks in this PR, so they don't disappear once we merge this (Accordion, Spreadsheet, Sub-pages).
> 
> They won't disappear, since the setting only affects whether a list of block names is allowed in the plugin:

Though I did forget to update the setting description. I referred to our source file for now, then we can do a separate PR to move (some of or all) the existing blocks out of the beta category.

src/Features.php
```diff
@@ -90,6 +92,15 @@ private static function get_fields(): array {
 				'id'   => self::THEME_EDITOR,
 				'type' => 'checkbox',
 			],
+			[
+				'name' => __( 'Allow beta blocks in post editor.', 'planet4-master-theme-backend' ),
+				'desc' => __(
+					'If enabled, beta blocks defined in <a href="https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/planet4-gutenberg-blocks.php">this file</a>. Not for production use.',
+					'planet4-master-theme-backend'
```
comzeradd, line 99: This is a bit confusing for editors/admins. The setting should completely hide the whole Beta category. We shouldn't link to the code for that. A more simple text would be:

```
'name' => 'Allow Beta Blocks in the editor'
'desc' => If enabled, blocks that are still not stable enough will be available.
```

comzeradd (changes requested):
Since the setting will disable Beta blocks, I think we still to move the 3 existing ones as part of the Covers PR and merge the two PRs together.

comzeradd (approved):
Ok, let's merge this for now and we can keep the suggested description for the next iteration.

Inwerpsel:
> Ok, let's merge this for now and we can keep the suggested description for the next iteration.

I'll update it as soon as https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/486 is merged :ok_hand:

PR #1367, opened 2021-04-16
remove bullet from breadcrumbs styling

DanceParty:
Ref: 1/2 of https://github.com/greenpeace/planet4/issues/114

Sibling PR: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/560

## Description
This will remove the bullet applied via css `content: "\2022";` as there is not a clean way to hide the bullet from the screenreader when the bullet is added through css like this.

## Potential Issues
After taking a quick look, and noticing that we are updating a file that seems more general than the component we are touching (`_breadcrumbs.scss`) I am a little worried that this might break other functionality by hiding bullets in other places of the website. I would definitely like some 👀 on this PR with that in mind. Thank you ❤️ 

---

### DIFF
--- a/assets/src/scss/layout/_breadcrumbs.scss
+++ b/assets/src/scss/layout/_breadcrumbs.scss
@@ -23,7 +23,6 @@
   .tag-wrap-bullet {
     pointer-events: none;
     display: inline-block;
-    color: $link-color;
     font-weight: 400;
     margin-left: -12px;
     margin-right: 8px;
--- a/assets/src/scss/layout/_breadcrumbs.scss
+++ b/assets/src/scss/layout/_breadcrumbs.scss
@@ -20,19 +20,17 @@
     position: relative;
   }
 
-  .tag-item--main + .tag-wrap,
-  .tag-wrap + .tag-wrap {
-    &:before {
-      content: "\2022";
-      pointer-events: none;
-      display: inline-block;
-      margin-left: -12px;
-      margin-right: 8px;
+  .tag-wrap-bullet {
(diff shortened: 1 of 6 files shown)


DanceParty:
Sorry for the delay, I just incorporated the changes that you suggested! Please let me know how it looks 🙇

assets/src/scss/layout/_breadcrumbs.scss
```diff
@@ -20,19 +20,17 @@
...
-    &:before {
-      content: "\2022";
-      pointer-events: none;
-      display: inline-block;
-      margin-left: -12px;
-      margin-right: 8px;
+  .tag-wrap-bullet {
+    pointer-events: none;
+    display: inline-block;
+    color: $link-color;
```
lithrel, line 26: We didn't use link color for bullets before, no need for it here

lithrel (commented):
@DanceParty  Just a comment on color, and you missed one bullet in `single.twig` (the one between issues and tags, it's the one between "People" and "About us" on this page https://www-dev.greenpeace.org/test-tavros/story/29490/covid-19-compassion-cooperation-courage-choice/)
After that we're good :partying_face:  :+1:

assets/src/scss/layout/_breadcrumbs.scss
DanceParty: Thank you for catching this, still getting used to testing these changes! ❤️

DanceParty:
@lithrel thanks for the patience, the color has been removed <3

PR #1369, opened 2021-04-19
PLANET-6087 Refer to beta category in help

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6087

---

Now that we moved out blocks that were already used in production from the beta category, it will be hidden behind this toggle from now on. So we can refer to it in the help text.

https://www-dev.greenpeace.org/test-pluto/wp-admin/admin.php?page=planet4_settings_features


comzeradd (approved):
🍻

PR #1373, opened 2021-04-30
PLANET-6129 Remove variable collect script

Inwerpsel:
* It's not needed anymore. The same info can be extracted at runtime
fast enough, which is a lot easier to work with. Besides the plugin had
multiple serious issues. It always added a fallback with no option to
disable that. It also for some reason split combined selectors, which
did quite increase the size of the CSS.

Ref: https://jira.greenpeace.org/browse/PLANET-6129

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1510,61 +1510,6 @@
         }
       }
     },
-    "@greenpeace/planet4-postcss-css-variables": {
-      "version": "0.2.0",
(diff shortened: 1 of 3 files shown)


planet-4:

:new_moon: [sinope](https://www-dev.greenpeace.org/test-sinope)

:watch: 2021.04.30 12:13:18

lithrel (approved):
:+1:

PR #1377, opened 2021-05-06
PLANET-6037 Add color transition to button PDF icons

mleray:
### Description

See https://jira.greenpeace.org/browse/PLANET-6037

### Testing

Add a Button block with secondary or donate style and a PDF url (for example [GP annual report](https://www.greenpeace.org/static/planet4-international-stateless/2018/11/f4b36653-greenpeace_international_annualreport2006.pdf)) and make sure it looks good on hover.

### DIFF
--- a/assets/src/scss/components/_buttons.scss
+++ b/assets/src/scss/components/_buttons.scss
@@ -11,6 +11,13 @@
     transition-duration: 150ms;
     transition-timing-function: linear;
   }
+
+  *, &::before, &:after {
+    transition-property: inherit;
+    transition-duration: inherit;
+    transition-timing-function: inherit;
+  }
+
   display: inline-block;
   position: relative;
   font-size: $font-size-sm;


assets/src/scss/base/_icons.scss
```diff
@@ -1,3 +1,11 @@
+// Secondary and donate buttons have a transition on the color
+// Other button styles don't need it since they don't change color on hover
+@mixin with-transition() {
```
Inwerpsel, line 3: It seems more straightforward if we create a separate rule for the transition properties, with all selectors that need these same properties. We could use the same `transition-property` for all of them, since it includes background-color and the other 2 don't cause an issue to add to the icons.

```css
.btn,
.wp-block-button a,
.wp-block-file .wp-block-file__button,
.btn a.pdf-link,
.btn a.external-link {
    transition-property: background-color;
  transition-duration: 150ms;
  transition-timing-function: linear;
}

mleray: External links don't need it, since only the PDF icons are applied to buttons 🤔

Inwerpsel: Ah could be I have the wrong selector, it was just to give an example of what I meant :smile:

Inwerpsel: I think we can just put it on `.btn` that way and don't need to do anything for the button type classes.

mleray: Well the Button block actually doesn't have the `.btn` class but `.wp-block-button` so we'd need at least these 2

Inwerpsel: Btw, these are variable now (one theme depends on changing these to prevent a flicker bug). You can create a second selector and also use the same `--button--` shorthand, should work fine :crossed_fingers:

Inwerpsel (approved):

planet-4:

:new_moon: [umbriel](https://www-dev.greenpeace.org/test-umbriel)

:watch: 2021.05.07 11:53:39

PR #1381, opened 2021-05-14
PLANET-6136 Prepare test instance for themes review

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6136

---

This branch ended up with various small things. I didn't split them up yet because for now it seems manageable to review. But I can still split it if that works better for the reviewer.

- Add the theme to the campaigns listing and allow sorting on it. See screenshot
- Fix fallbacks in editor styles
- Add a "data-base" attribute so that we can use it to create a namespace in localstorage.
- If a campaign has a new theme selected, it will grab it from the options. Not intended to be used in production, but it shouldn't affect any live content so we can merge it. On production you're not able to select a new theme for a campaign.
- If the stripping `-new` from the theme name results in one of the legacy themes, it will still use the sidebar settings corresponding to that theme. This is because when we switch the new themes will effectively use the same settings.

Screenshot from 2021-05-25 10-58-11

### DIFF
--- a/assets/src/scss/editorOverrides.scss
+++ b/assets/src/scss/editorOverrides.scss
@@ -1,6 +1,10 @@
 // Moved from another repo, for history please see https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blame/f5c532ed5704738136224cde305e9a0ffe614ceb/assets/src/styles/editorOverrides.scss
 .editor-post-title__block textarea.editor-post-title__input {
-  font-family: var(--headings--font-family, var(--header-primary-font, $roboto));
+  --headings-- {
+    // Not sure if fallback still needed, preserving for now.
+    font-family: var(--header-primary-font, $roboto);
+  }
+
   font-size: $font-size-xl;
   line-height: 1.225;
   margin-bottom: 30px;
@@ -74,14 +78,17 @@
 
 .wp-block-button.is-style-secondary .wp-block-button__link[role="textbox"],
 [class="wp-block-button"] .wp-block-button__link[role="textbox"] {
-  background-color: var(--btn-secondary-background-color, transparentize($white, .7));
-  border-color: var(--btn-secondary-border-color, $dark-blue);
-  color: var(--btn-secondary-color, $dark-blue);
+  --button-secondary-- {
+    background: transparentize($white, .7);
+    border-color: $dark-blue;
+    color: $dark-blue;
+    padding: 2px $n30;
+  }
+
   box-shadow: none;
   white-space: nowrap;
   overflow: hidden;
   text-overflow: ellipsis;
-  padding: 2px $n30;
 
   html[dir="rtl"] & {
     line-height: 2.5;
@@ -89,20 +96,25 @@
 }
 
 .wp-block-button.is-style-cta .wp-block-button__link[role="textbox"] {
-  background-color: var(--btn-primary-background-color, $orange);
-  border-color: var(--btn-secondary-border-color, $orange);
-  color: var(--btn-primary-color, $white);
-  box-shadow: 0 2px 5px rgba(0, 0, 0, .25);
+  --button-primary-- {
+    background: $orange;
+    border-color: $orange;
+    color: $white;
+    box-shadow: 0 2px 5px rgba(0, 0, 0, .25);
+  }
 }
 
 .wp-block-button.is-style-donate .wp-block-button__link[role="textbox"] {
-  background-color: var(--btn-donate-background-color, $aquamarine);
-  color: var(--btn-donate-color, $grey);
+  --button-donate-- {
+    background: $aquamarine;
+    color: $grey;
+    box-shadow: 0 2px 5px rgba(0, 0, 0, .25);
+  }
+
   line-height: 1.65;
   min-width: 180px;
   margin: 0;
   padding: 2px $n30;
-  box-shadow: 0 2px 5px rgba(0, 0, 0, .25);
   transition: color background-color font-weight 100ms;
 
   .editor-styles-wrapper & {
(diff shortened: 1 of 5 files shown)


planet-4:

:new_moon: [pluto](https://www-dev.greenpeace.org/test-pluto)

:watch: 2021.05.14 09:42:14

templates/html-header.twig
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html {{ fn('language_attributes', 'html') }}>
+<html {{ fn('language_attributes', 'html') }} data-base="{{ constant('WP_HOME') }}">
```
Inwerpsel, line 2: The base data attribute is used by the theme editor to put a local storage namespace, otherwise it loads the same settings on each instance on the same domain. This is just the best fix I came up with, maybe I'm overlooking an existing way to get this value on the front end.

templates/css-variables.twig
```diff
@@ -1,7 +1,7 @@
-<style>
+<style id="test">
 	:root {
 		{% for key, value in css_vars %}
-			--{{ key }}: {{ value }};
+			--{{ key|trim('--', 'left') }}: {{ value|raw }};
```
Inwerpsel, line 4: The existing implementation in the sidebar stored the variables without `--` at the front. For now, in the newer one `--` is included. Trimming here to make both work.

src/PostCampaign.php
```diff
@@ -378,13 +416,34 @@ private static function get_theme_defaults( $theme_json ) {
...
 	 * Determine the css variables for a certain post.
 	 *
 	 * @param array $meta The meta containing the variable values.
+	 *
 	 * @return array The values that will be used for the css variables.
 	 */
 	public static function css_vars( array $meta ): array {
+		$is_new_theme = $meta['theme'] && ! in_array( $meta['theme'], self::get_legacy_themes(), true );
+		if ( $is_new_theme ) {
+			$themes = json_decode( get_option( 'planet4_themes', '[]' ), true );
```
Inwerpsel, line 442: This is probably temporary to allow us to make changes to the theme on the instance. After we confirmed the new version works, we can for now include it in our code base, so that it effectively works the same as the legacy themes.

templates/css-variables.twig
Inwerpsel: The `|raw` filter actually solves an issue production currently has, if there are characters in the value that Twig escapes. Could be there are no such values in the old theme options, but in the new ones I already ran into it.

templates/html-header.twig
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html {{ fn('language_attributes', 'html') }}>
+<html {{ fn('language_attributes', 'html') }} data-base="{% if constant('WP_HOME') is defined %}{{ constant('WP_HOME') }}{% endif %}">
```
Inwerpsel, line 2: I added this because the theme editor uses local storage to store the editor state. It checks for the `data-base` attribute to create a local storage namespace. All sites on the same domain share the local storage, so if we don't provide a namespace these will unintentionally be shared. See https://github.com/Inwerpsel/use-theme-editor/blob/main/src/useLocalStorage.js#L3-L5

assets/src/scss/editorOverrides.scss
```diff
@@ -112,7 +124,7 @@
 }
 
 .editor-rich-text__editable.wp-block-button__link:not([role="textbox"]) {
```
mleray, line 126: This selector doesn't need variables? 🤔

Inwerpsel: I think I tried adding them but it had some difficulties and required further changes, so I kept them out of this PR.

assets/src/scss/layout/_page-header.scss
```diff
@@ -23,23 +23,27 @@
 }
 
 .page-header {
-  color: $grey-80;
-  padding-top: $n90;
-  margin-bottom: $space-md;
-  position: relative;
+  _-- {
+    color: $grey-80;
+    background: inherit;
```
Inwerpsel, line 28: Added it with `inherit` because it's customized by some themes, but by default has the parent's background.

assets/src/scss/editorOverrides.scss
Inwerpsel: I mostly changed this file to fix the incorrect naming for some of the variables, and add some obviously missing ones without going too much into CSS rabbit holes. Created a separate ticket for those https://jira.greenpeace.org/browse/PLANET-6168

src/PostCampaign.php
```diff
@@ -378,15 +416,48 @@ private static function get_theme_defaults( $theme_json ) {
 		return $defaults;
 	}
 
+	/**
+	 * Get whether the theme of a campaign is one of the legacy themes.
+	 *
+	 * @return string[] Whether it's legacy.
+	 */
+	public static function get_legacy_themes(): array {
```
mleray, line 424: Couldn't we use a constant here instead of a function? 🤔

src/PostCampaign.php
```diff
@@ -558,9 +629,8 @@ public static function get_footer_theme( array $css_vars ): array {
 	 */
 	public static function get_theme( array $meta ): string {
 		$theme = $meta['theme'] ?? $meta['_campaign_page_template'] ?? null;
-		$theme = $theme ? $theme : 'default';
 
-		return $theme;
+		return $theme ? $theme : 'default';
```
mleray, line 633: This can be rewritten to `theme ?? 'default'` I think?

Inwerpsel: I wish it could, but it's not the same thing :smile: `??` only would use `default` if the value is `null`, and not if it's an empty string.

mleray: Aaaah I see, nevermind then 😁

Inwerpsel: It could be rewritten to `theme ?: 'default'`, but WordPress coding standards have a yet unidentified problem with that so we can't do that :stuck_out_tongue:

templates/css-variables.twig
mleray: 👍

PR #1384, opened 2021-05-21
PLANET-6237 Slack notifications

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6237

---

Add our own super lightweight implementation of a Slack message carrier to the notifications plugin.

Stores the webhook in a WP option (maybe we can find a better menu location, though the features section is already a place where we put early stuff).

The code that adds Slack only runs on a hook provided by the plugin, so should be safe to merge before we add the plugin. We'll initially set it up for GPI only as a pilot.

The current code was already quite extensively tested, see https://greenpeace-gpi.slack.com/archives/C022728TVT9 . The latest messages in the channel look a bit off, but that's just because of how the notification is set up. We can add a webhook and a test channel webhook to the GPI dev site, it can be used by Kelli to test a notification before setting up on prod.

### DIFF
--- a/src/Features.php
+++ b/src/Features.php
@@ -24,8 +24,6 @@ class Features {
 
 	public const BETA_BLOCKS = 'beta_blocks';
 
-	public const SLACK_WEBHOOK = 'slack_webhook';
-
 	/**
 	 * Get the features options page settings.
 	 *
@@ -102,15 +100,6 @@ class Features {
 				),
 				'id'   => self::BETA_BLOCKS,
 				'type' => 'checkbox',
-			],
-			[
-				'name' => __( 'Slack webhook URL', 'planet4-master-theme-backend' ),
-				'desc' => __(
-					'TMP',
-					'planet4-master-theme-backend'
-				),
-				'id'   => self::SLACK_WEBHOOK,
-				'type' => 'text',
 			],
 		];
 
--- a/src/Features.php
+++ b/src/Features.php
@@ -24,6 +24,8 @@ class Features {
 
 	public const BETA_BLOCKS = 'beta_blocks';
 
+	public const SLACK_WEBHOOK = 'slack_webhook';
+
 	/**
 	 * Get the features options page settings.
 	 *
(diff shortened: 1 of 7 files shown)


src/Features.php
```diff
@@ -101,6 +103,15 @@ private static function get_fields(): array {
 				'id'   => self::BETA_BLOCKS,
 				'type' => 'checkbox',
 			],
+			[
```
Inwerpsel, line 106: Included it here for now, we can probably come up with a better menu location.

src/Notifications/Slack.php
```diff
@@ -0,0 +1,73 @@
...
+	 * Uses $this->add_form_field();.
+	 *
+	 * @return void
+	 * @throws \Exception Not sure which ones.
+	 */
+	public function form_fields() {
+		$body_field = new MrkDwnField(
+			[
+				'label'       => __( 'Message', 'planet4-master-theme-backend' ),
+				'description' => '<a target="_blank" class="external-link" href="https://api.slack.com/reference/surfaces/formatting#basics">Learn about message formatting options</a>',
```
Inwerpsel, line 36: I intentionally didn't add a translation here for now, as the pilot doesn't require it and we might still change it, so we can avoid translators having to spend time on it for now.

src/Features.php
comzeradd: Let's create a new category for this (Planet 4 > Notifications), so we have Features only for features flags.

comzeradd:
Can you also pass `alek13/slack` through SonarCloud? You can fork the github repo to your personal account and then add it as a new project in SonarCloud. Since the project is open source the generated report will be public.

Inwerpsel:
> Can you also pass `alek13/slack` through SonarCloud? You can fork the github repo to your personal account and then add it as a new project in SonarCloud. Since the project is open source the generated report will be public.

Seems good, just checking how to set up the folder exclusions as it's detecting `http://` URLs in the tests as a security issue. 
Those are the only things it picks up for security. Duplications are also only within tests, as are almost all of the code smells .https://sonarcloud.io/dashboard?id=greenpeace_slack

Inwerpsel:
I also analyzed the Notification plugin in the same way and it seems fine as well. Only security issues it flagged are either false positives or non-issues https://sonarcloud.io/project/security_hotspots?id=greenpeace_Notification

planet-4:

:new_moon: [janus](https://www-dev.greenpeace.org/test-janus)

:watch: 2021.07.09 09:28:44

comzeradd (approved):
💯

PR #1387, opened 2021-05-27
PLANET-6019: Fix errors when ElasticPress is disabled

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-6019
Related to: https://github.com/greenpeace/planet4-master-theme/pull/1361

Verify inclusion of ElasticPress classes before using them.
Aiming to fix a CI error during local dev env building - see https://app.circleci.com/pipelines/github/greenpeace/planet4-docker-compose/526/workflows/34039d78-7e7b-46ea-997f-1283a177025b/jobs/901

## Test

- Disable Elasticpress 
  `docker-compose exec php-fpm wp plugin deactivate elasticpress`
- Go to a search page
  - on `master` branch, you should get an error 
  - on this branch, the page appears and result links are valid

### DIFF
--- a/src/Search.php
+++ b/src/Search.php
@@ -216,20 +216,24 @@ abstract class Search {
 		);
 		// Because of how the search is currently set up (using admin-ajax) this ElasticPress filter was not being
 		// applied for subsequent page loads, only for the initial one.
-		if ( ( ! isset( $_GET['orderby'] ) || '_score' === $_GET['orderby'] ) && wp_doing_ajax() ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
-			add_filter(
-				'ep_formatted_args',
-				[ new \ElasticPress\Feature\Search\Search(), 'weight_recent' ],
-				10,
-				2
+		$ep_exists = class_exists( \ElasticPress\Feature\Search\Search::class )
+			&& class_exists( Features::class );
+		if ( $ep_exists ) {
+			if ( ( ! isset( $_GET['orderby'] ) || '_score' === $_GET['orderby'] ) && wp_doing_ajax() ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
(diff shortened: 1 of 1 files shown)


planet-4:

:new_moon: [tavros](https://www-dev.greenpeace.org/test-tavros)

:watch: 2021.05.27 13:05:03

src/Search.php
```diff
@@ -216,20 +216,24 @@ function ( $args ) {
 		);
 		// Because of how the search is currently set up (using admin-ajax) this ElasticPress filter was not being
 		// applied for subsequent page loads, only for the initial one.
-		if ( ( ! isset( $_GET['orderby'] ) || '_score' === $_GET['orderby'] ) && wp_doing_ajax() ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
-			add_filter(
-				'ep_formatted_args',
-				[ new \ElasticPress\Feature\Search\Search(), 'weight_recent' ],
-				10,
-				2
+		$ep_exists = class_exists( \ElasticPress\Feature\Search\Search::class )
```
Inwerpsel, line 219: On production we never run without ElasticSearch, so I'm not sure if we should be making the search code even more complex just so we can support running it locally without it :thinking:

Inwerpsel: I mean, if the test is failing because it's missing ES, that means it's checking a local environment which is significantly different from what production runs. Wouldn't it make more sense to ensure the ES plugin is included in localdev and during the test?

lithrel: I had a my regular argumentation ready for this discussion, but then I just found out that configuring ElasticPress with an empty url for host settings removes the admin panel error message and generates 0 exception, which were the 2 main things I was fighting against. Thanks for the push :partying_face:  closing this, I'm gonna open another one keeping ElasticPress enabled and just modifying config :ok_hand:

PR #1388, opened 2021-06-01
PLANET-6146 Add Hubspot event callback script to track signups

mleray:
### Description

See https://jira.greenpeace.org/browse/PLANET-6146

### DIFF
--- a/templates/head/hubspot.twig
+++ b/templates/head/hubspot.twig
@@ -7,4 +7,13 @@
 	if (getCookieValue('greenpeace') !== '2') {
 		(window._hsp = window._hsp || []).push(['revokeCookieConsent']);
 	}
+
+	window.addEventListener('message', event => {
+		if (event.data.type === 'hsFormCallback' && event.data.eventName === 'onFormSubmitted') {
+			window.dataLayer.push({
+				'event': 'formSubmission',
+				'hs-form-guid': event.data.id
+			});
+		}
+	});
 </script>


templates/html-header.twig
```diff
@@ -8,6 +8,8 @@
 
 	{% include 'blocks/google_tag_manager.twig' %}
 
+	{% include 'blocks/hubspot_signups.twig' %}
```
Inwerpsel, line 11: I also have [some code in my PR for HubSpot cookie consent](https://github.com/greenpeace/planet4-master-theme/pull/1390/files#diff-8479f762f16d76dd3afe830c872b63bab77ba39e3395d34d1123e65a08e6a7baR4-R12) that can be added to the head. Perhaps we can include a check in the twig template for whether the HubSpot plugin is active on the instance, and only output HS related things if needed?

Inwerpsel: [I added a common `head/hubspot.twig` template](https://github.com/greenpeace/planet4-master-theme/pull/1390/files#diff-8479f762f16d76dd3afe830c872b63bab77ba39e3395d34d1123e65a08e6a7baR11-R14) where we can put both things, and the plugin check in my PR. Probably best to not add any more Twig templates to the `blocks` directory, since the name is misleading. Twig actually has something called a "block" (e.g. `{% block head %}{% endblock %}`), but what we include in that directory is not that, it's just other templates not wrapped in a block.

templates/blocks/hubspot_signups.twig
```diff
@@ -0,0 +1,10 @@
+<script type="text/javascript">
+	window.addEventListener("message", event => {
```
Inwerpsel, line 2: We could check whether the page has a `leadin/hubspot-form-block` block and only include this listener if it does.

templates/html-header.twig
mleray: Updated! 👍

templates/blocks/hubspot_signups.twig
Inwerpsel: Though it's a very small listener, probably not worth the effort :shrug:

Inwerpsel (approved):
LGTM code wise. I see the ticket has a few cases to test, so I'm not excluding that we might need more changes. But we'll definitely need this part so should be fine merging already :+1:

planet-4:

:new_moon: [jupiter](https://www-dev.greenpeace.org/test-jupiter)

:watch: 2021.06.04 13:47:08

PR #1390, opened 2021-06-02
PLANET-6155 Prevent HubSpot cookies being set if user didn't consent

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6155

---

Ensure that the `revokeCookieConsent` is the first thing on the _hsp queue.

Luckily, this seems to work even if requiring consent is not enabled for the policy. Even though HubSpot documentation doesn't explicitly mention that revoking is supported in this case, it seems safe enough to do since it's unlikely that they would alter this behavior.

### DIFF
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -2,7 +2,6 @@
 
 namespace P4\MasterTheme;
 
-use DateTimeImmutable;
 use Timber\Timber;
 use Timber\Site as TimberSite;
 use Timber\Menu as TimberMenu;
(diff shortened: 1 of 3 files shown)


planet-4:

:new_moon: [leda](https://www-dev.greenpeace.org/test-leda)

:watch: 2021.06.03 08:00:11

src/MasterSite.php
```diff
@@ -2,7 +2,6 @@
 
 namespace P4\MasterTheme;
 
-use DateTimeImmutable;
```
Inwerpsel, line 5: was unused

mleray (approved):
Tested on local, looks good 👍

PR #1395, opened 2021-06-07
PLANET-5757 jQuery: remove jQuery from global.js

sagarsdeshmukh:
Removed global.js file and added title attribute in twig file & in PHP file by altering original output

Ref: [JIRA 5757](https://jira.greenpeace.org/browse/PLANET-5757)

---

**Testing steps:**
- Add native image block on page/post/campaign and test the frontend page the image has a title attribute
- The same fix is added for some other pages like author profile page, search, taxonomy page etc.

### DIFF
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -538,6 +538,9 @@ class MasterSite extends TimberSite {
 
 		// HubSpot.
 		$context['hubspot_active'] = is_plugin_active( 'leadin/leadin.php' );
+
+		// Dummy thumbnail.
+		$context['dummy_thumbnail'] = get_template_directory_uri() . '/images/dummy-thumbnail.png';
 		return $context;
 	}
 
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -1248,18 +1248,25 @@ class MasterSite extends TimberSite {
 	}
 
 	/**
-	 * Override the Gutenberg core/image block render method output, To add credit field in it's caption text.
+	 * Override the Gutenberg core/image block render method output, To add credit field in it's caption text & image alt text as title.
 	 *
 	 * @param array  $attributes    Attributes of the Gutenberg core/image block.
 	 * @param string $content The image element HTML.
 	 *
-	 * @return string HTML content of image element with credit field in caption.
+	 * @return string HTML content of image element with credit field in caption and alt text in image title.
 	 */
 	public function p4_core_image_block_render( $attributes, $content ) {
-		$image_id = isset( $attributes['id'] ) ? trim( str_replace( 'attachment_', '', $attributes['id'] ) ) : '';
-		$credit   = $image_id ? get_post_meta( $image_id, self::CREDIT_META_FIELD, true ) : '';
-		if ( empty( $image_id ) || empty( $credit ) ) {
+		$image_id      = isset( $attributes['id'] ) ? trim( str_replace( 'attachment_', '', $attributes['id'] ) ) : '';
+		$img_post_meta = $image_id ? get_post_meta( $image_id ) : [];
+		if ( ! $img_post_meta ) {
 			return $content;
+		}
+
+		$credit   = $img_post_meta[ self::CREDIT_META_FIELD ][0] ?? '';
+		$alt_text = $img_post_meta['_wp_attachment_image_alt'][0] ?? '';
+
+		if ( $alt_text ) {
+			$content = str_replace( ' alt=', ' title="' . esc_attr( $alt_text ) . '" alt=', $content );
 		}
 
 		$image_credit = ' ' . $credit;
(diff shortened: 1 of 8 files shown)


src/MasterSite.php
```diff
@@ -1248,16 +1248,27 @@ public function add_image_attachment_fields_to_save( $post, $attachment ) {
...
+		$img_post_meta = $image_id ? get_post_meta( $image_id ) : [];
+		if ( ! $img_post_meta ) {
+			return $content;
+		}
+
+		$credit   = $img_post_meta[ self::CREDIT_META_FIELD ][0] ?? '';
+		$alt_text = $img_post_meta['_wp_attachment_image_alt'][0] ?? '';
+
+		if ( $alt_text ) {
+			$content = str_replace( 'alt=', 'title="' . esc_attr( $alt_text ) . '" alt=', $content );
```
Inwerpsel, line 1269: I think we can match `' alt='` too (so with preceding space), which would decrease the chance of matching unintended things.

templates/tease-author.twig
```diff
@@ -3,6 +3,7 @@
 		class="d-flex search-result-item-image"
 		src="{{ post.thumbnail.src('thumbnail') }}"
 		alt="{{ fn('esc_attr', post.thumbnail.alt|default( post.title ) )|raw }}"
+		title="{{ fn('esc_attr', post.thumbnail.alt|default( post.title ) )|raw }}"
```
Inwerpsel, line 6: This logic is a bit complex and with this change would occur 6 times in 3 files, maybe it's better to write a function for it, or alternative maybe assign it as a template variable? The only difference is [this other one uses `post.post_title`](https://github.com/greenpeace/planet4-master-theme/pull/1395/files#diff-f721ebbf1a207a8e1119c863660a9ec1cac8e82c0e5dcbccebbef1721b41807bR26), but maybe all of them could use `post.title`.

Inwerpsel (approved):
Looks good, left some suggestions. There could be some rare cases where the previous code was adding the title that are not covered by these changes. For example images in a custom HTML block. But can't think of anything that's frequent enough to be blocking. :+1:

src/MasterSite.php
```diff
@@ -1248,16 +1248,27 @@ public function add_image_attachment_fields_to_save( $post, $attachment ) {
...
-	 * @return string HTML content of image element with credit field in caption.
+	 * @return string HTML content of image element with credit field in caption and alt text in image title.
 	 */
 	public function p4_core_image_block_render( $attributes, $content ) {
-		$image_id = isset( $attributes['id'] ) ? trim( str_replace( 'attachment_', '', $attributes['id'] ) ) : '';
-		$credit   = $image_id ? get_post_meta( $image_id, self::CREDIT_META_FIELD, true ) : '';
+		$image_id      = isset( $attributes['id'] ) ? trim( str_replace( 'attachment_', '', $attributes['id'] ) ) : '';
+		$img_post_meta = $image_id ? get_post_meta( $image_id ) : [];
+		if ( ! $img_post_meta ) {
+			return $content;
```
Inwerpsel, line 1262: Sonarcloud complains about this 4th return statement that's added (it only allows 3). Though in this case the added return does make sense, so not sure how we should please Sonar :stuck_out_tongue:

src/MasterSite.php
```diff
@@ -1248,16 +1248,27 @@ public function add_image_attachment_fields_to_save( $post, $attachment ) {
...
+		}
+
+		$credit   = $img_post_meta[ self::CREDIT_META_FIELD ][0] ?? '';
+		$alt_text = $img_post_meta['_wp_attachment_image_alt'][0] ?? '';
+
+		if ( $alt_text ) {
+			$content = str_replace( 'alt=', 'title="' . esc_attr( $alt_text ) . '" alt=', $content );
+		}
+
 		if ( empty( $image_id ) || empty( $credit ) ) {
```
Inwerpsel, line 1272: With the new check you added earlier, it's now impossible that it gets here and `empty( $image_id )` would be true, so I guess we can remove that part of the condition.

templates/tease-author.twig
sagarsdeshmukh: For now use a template variable :)

src/MasterSite.php
sagarsdeshmukh: Removed the one return statement, hope Sonarcloud is happy now 😄

planet-4:

:new_moon: [venus](https://www-dev.greenpeace.org/test-venus)

:watch: 2021.06.07 12:46:22

sagarsdeshmukh:
Hi @Inwerpsel , I updated the PR after last review, could you please take a look on new changes. 🙂

templates/blocks/site_logo.twig
```diff
@@ -5,6 +5,7 @@
 		<img
 			src="{{ data_nav_bar.images }}gp-logo-{{ logo_color }}.svg"
 			alt="Greenpeace"
+			title="Greenpeace"
```
comzeradd, line 8: I don't think we need a tooltip in the site logo, the author profile image, or even the article thumbnails on search and articles block (the other PR). 

According to the [original ticket](https://jira.greenpeace.org/browse/PLANET-1757) the intention was mostly to add tooltips on images being included on posts and pages so users can easily see credit on hover. I don't think we should expand the scope of the initial code.

sagarsdeshmukh: Yes got the point 👍, It was misunderstood by me as applicable over all images.

I will update both the PR accordingly 🙂

sagarsdeshmukh: Both PRs updated, please have a look.

comzeradd (approved):
:+1:

PR #1411, opened 2021-07-02
PLANET-6124: Change the way of setting the target to links when are externals

GP-Dan-Tovbein:
I tackled this modification to be ensure that external links, the ones that were set by the block, would always have set the defined target value. If not, just keep the default.

Linked PR https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/624

Ref: https://jira.greenpeace.org/browse/PLANET-6124

---

### DIFF
--- a/assets/src/js/external_links.js
+++ b/assets/src/js/external_links.js
@@ -12,7 +12,7 @@ export const setupExternalLinks = () => {
       return;
     }
 
-    link.target = '_blank';
+    link.target = link.target ? link.target : '';
     link.classList.add('external-link');
   });
 };


planet-4:

:new_moon: [titan](https://www-dev.greenpeace.org/test-titan)

:watch: 2021.07.02 17:54:00

mleray (changes requested):
The problem is that this file takes care of all links in the page content, and we still need to add the target and class for links that are not in Columns blocks... I think a solution for this issue would probably be to check if the link has the `call-to-action-link` that is [added](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/1b55e9da077de86c40e70aff03a7d8f2fbf7c8ff/assets/src/blocks/Columns/Columns.js#L78) in the Columns block? From a quick search it seems to be used only there 🤔 I'm just not sure whether we want to keep the external link icon in this case or not, since those links are actually not necessarily external that might be confusing... That should probably be checked with the design team

dantovbein:
@mleray, thanks for your review! Have you seen also [this](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/624) PR? Because it's related to this current fix.

GP-Dan-Tovbein:
@mleray I’ve updated the PR. I’ve not added  :not(.call-to-action-link)  since the external icon is not being shown through this way, so, I tackled another method to set target to links.

mleray (approved):
Looks good to me 👍

PR #1413, opened 2021-07-12
PLANET-6189 Implement the new Footer design

dantovbein:
:warning:
This PR is replaced with [this](https://github.com/greenpeace/planet4-master-theme/pull/1465)

---
**Design**
This is the new version of the Footer site based in these [versions](https://p4-designsystem.greenpeace.org/05f6e9516/p/16a899-footer).

**Affected versions:**
- Footer S, M, L and XL
- Minimal Footer S and L. Because we haven't any mockup for minimal's version, we reported it to the Designer and suggested a potential layout and it was approved by Magali. Here is the [thread of the conversation](https://greenpeace-gpi.slack.com/archives/G015K63081W/p1626186385006500)

**What is done.**
- Added a new Facebook and Linkedin rounded icons styled by adding this class name  `[social-name]-rounded` format. i.e. `facebook-rounded`.
- Improve the css using BEM for class names.

**Related PRs**
- [Minimal Footer design](https://github.com/greenpeace/planet4-master-theme/pull/1452)

:warning: **Attention!** 
- The acceptance test is failing because the buttons of the current Footer are upper cased and we won't need that format anymore.


planet-4:

:new_moon: [pandora](https://www-dev.greenpeace.org/test-pandora)

:watch: 2021.07.13 15:58:48

templates/footer.twig
```diff
@@ -1,74 +1,80 @@
...
-						   data-ga-label="{{ social.title }}">
-							{{ name|svgicon }}
-							<span class="visually-hidden">{{ social.title }}</span>
-						</a>
-					</li>
-				{% endfor %}
-			{% endif %}
-		</ul>
+{% set follow_us = __( 'Follow us', 'planet4-master-theme' ) %}
+{% set total_social_icons = social_overrides is defined and social_overrides|length ? social_overrides|length : footer_social_menu|length %}
```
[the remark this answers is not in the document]
Inwerpsel: There's quite a lot of template and logic involved for the social icons, perhaps better to give this its own template? Would also make for a cleaner diff here as this new variable could be inside that template instead.

templates/footer.twig
```diff
@@ -1,74 +1,80 @@
...
-						   data-ga-action="Social Icons"
-						   data-ga-label="{{ social.title }}">
-							{{ name|svgicon }}
-							<span class="visually-hidden">{{ social.title }}</span>
-						</a>
-					</li>
-				{% endfor %}
-			{% endif %}
-		</ul>
+{% set follow_us = __( 'Follow us', 'planet4-master-theme' ) %}
```
Inwerpsel, line 1: What is the reason for assigning this to a variable? The `__` function is used quite generally, I don't see a reason why this instance should be assigned.

assets/src/scss/layout/_footer.scss
```diff
@@ -1,240 +1,239 @@
 .site-footer {
   --footer-- {
     font-family: $roboto;
-    background: $dark-blue;
+    background-color: $dark-blue;
```
Inwerpsel, line 4: These use `background` intentionally, as it supports the same values as `background-color`, but also some more like gradients and background images.

assets/src/scss/layout/_footer.scss
```diff
@@ -1,240 +1,239 @@
...
-  }
 
   a --footer--links-- {
-    color: inherit;
+    color: $white;
 
-    &:hover {
-      color: $white;
-    }
+    // &:hover {
```
Inwerpsel, line 17: This is here intentionally: `--footer--links--` in the selector turns all properties into variables. So where in the default theme they have the same color, it allows a theme to have a different hover color.

Inwerpsel: However this is also a weird case: there's other color variables for different links in the footer. So probably it's better to remove this one and ensure each part can be set individually.

Inwerpsel (commented):
Are all these changes to `_footer.scss` needed to consolidate the template? Not saying we shouldn't refactor that file, but I feel we would be changing too many things at the same time (pretty much every line of CSS) and it would be easy to miss things.

templates/footer.twig
GP-Dan-Tovbein: I agree

templates/footer.twig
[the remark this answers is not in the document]
GP-Dan-Tovbein: I completely agree with you

templates/base.twig
```diff
@@ -41,11 +41,7 @@
 	{% endif %}
 
 	{% block footer %}
-		{% if custom_styles.nav_type == 'minimal' %}
-			{% include 'footer_min.twig' %}
-		{% else %}
-			{% include 'footer.twig' %}
-		{% endif %}
+		{% include 'footer.twig' with { 'nav_type' : custom_styles.nav_type } %}
```
Inwerpsel, line 44: We can do this refactor in a separate PR to make the design changes easier to evaluate.

PR #1419, opened 2021-07-22
PLANET-6083: Add UTM Parameters to share buttons

GP-Dan-Tovbein:
The Social Media share message should include the current page URL with the additional UTM parameters below (based on the [Global UTM Standards](https://app.gitbook.com/@greenpeace/s/insights/global-property/standards/utm-conventions))

Ref: https://jira.greenpeace.org/browse/PLANET-6083

32549_table

### DIFF
--- a/templates/blocks/share_buttons.twig
+++ b/templates/blocks/share_buttons.twig
@@ -1,27 +1,27 @@
 <div class="share-buttons">
 	<!-- Whatsapp -->
-	<a href="https://wa.me/?text={{ social.link }}"
+	<a href="https://wa.me/?text={{ social.link }}&utm_source=whatsapp&utm_medium={{ utm_medium }}{{ utm_content_param }}{{ utm_campaign_param }}"
 		 onclick="dataLayer.push({'event' : 'uaevent', 'eventCategory' : 'Social Share', 'eventAction': 'Whatsapp', 'eventLabel': '{{ social.link }}'});"
 		 target="_blank" class="share-btn whatsapp">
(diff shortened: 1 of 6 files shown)


planet-4:

:new_moon: [umbriel](https://www-dev.greenpeace.org/test-umbriel)

:watch: 2021.07.23 08:45:55

templates/blocks/share_buttons.twig
```diff
@@ -1,27 +1,27 @@
 <div class="share-buttons">
 	<!-- Whatsapp -->
-	<a href="https://wa.me/?text={{ social.link }}"
+	<a href="https://wa.me/?text={{ social.link }}&utm_source=whatsapp&utm_medium={{ utm_medium }}{{ utm_campaign_param }}"
```
comzeradd, line 3: Shouldn't `utm_content` also be part of these urls?

GP-Dan-Tovbein: As far as I understand, we don't need to include it if the local campaign isn't set.

> The utm_campaign value should be based on the “Local Project” value defined in Planet 4 Analytics settings. All values should be in lowercase and spaces should be transformed to dashes . If no value is selected, we don’t need to include the utm_campaign value. 

Correct me if I'm wrong

GP-Dan-Tovbein: Good catch! I missed to add it.

GP-Dan-Tovbein: @comzeradd added

comzeradd (approved):
👍🏻

PR #1420, opened 2021-07-22
PLANET-5983 Make all headings properties variable

Inwerpsel:
Ref: 

---


planet-4:

:new_moon: [pluto](https://www-dev.greenpeace.org/test-pluto)

:watch: 2021.07.26 12:12:40

templates/head/google_fonts.twig
```diff
@@ -1,45 +1,46 @@
 <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,700,900&display=swap&subset=latin-ext"/>
 <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Lora:400,400i,700&display=swap&subset=latin-ext"/>
 {% if  custom_font_families %}
+	{% set fonts_dir = theme_uri ~ "/../../plugins/planet4-plugin-gutenberg-blocks/public/fonts" %}
```
Inwerpsel, line 4: Just to make it work again. Still need to investigate if the old code worked previously and why it broke.

src/PostCampaign.php
```diff
@@ -552,7 +552,7 @@ private static function migrate_old_vars( array $css_vars ): array {
 				continue;
 			}
 			foreach ( $to as $new_name ) {
-				$css_vars[ $new_name ] = $css_vars[ $from ];
+				$css_vars[ '--' . $new_name ] = $css_vars[ $from ];
```
Inwerpsel, line 555: Quick fix because otherwise it puts some variables in there twice. In case of new themes `$css_vars` will contain the version with `--` in front. Without this fix it would add it a second time without 2 dashes. In the template it's currently set up to work with both variants.

Will look for a cleaner way to handle this.

templates/head/google_fonts.twig
Inwerpsel: Fixed in https://github.com/greenpeace/planet4-master-theme/pull/1427

src/PostCampaign.php
Inwerpsel: Addressing this in https://github.com/greenpeace/planet4-master-theme/pull/1428/files

PR #1424, opened 2021-07-29
Poc plugin included

Inwerpsel:
Ref: 

---

For now this just does the "mechanical" steps of including the other repo files and history. We can use this branch to assess how much work is involved before we can include the plugin inside this repo.

Packagist has a dedicated page about how to work with packages in a monorepo, and what the fundamental limitations of composer are. https://blog.packagist.com/installing-composer-packages-from-monorepos/ It describes the problem well, though they state 
>  When installing packages from a Git repository, for example, you cannot checkout different versions for different subdirectories. 

as though that would be a fundamental limitation of Git. However you can in fact do this, using `git worktree`. https://git-scm.com/docs/git-worktree That would allow to do one checkout for all packages in a single repo.

This still doesn't fix the problem that they need to check out the entire repo, but perhaps it makes it easier to address. But maybe not even needed, the size of the monorepo is really within acceptable boundaries in most cases (as long as you need to fetch it once).

It's also more an optimization issue. It doesn't prevent from using a monorepo, it's just Composer authors think it would be wasteful to support in that form (of requiring one download of the monorepo per package in it). It's not a problem in our case.


Inwerpsel:
I'll close this since any new attempt to consolidate these repos would have to start from a recent checkout.

PR #1428, opened 2021-08-03
Ensure migrated old vars override new themes

Inwerpsel:
* By putting "--" in front. New themes currently define them with "--", previously it was adding it without resulting in double definition.
* Remove "footer_color" as that one doesn't come from meta, set right name directly.
* Add some explanatory comments.

To reproduce the issue this fixes on `master`: create a new version of a theme (say `climate-new`). This allows you to use the same sidebar settings as the old `climate` theme. Set the navigation background color using the sidebar. After saving, you'll notice on the public URL of the page that it includes `--top-navigation--background` twice. Also if you're on Chrome, inspect the top navigation and click on the CSS variable to navigate to the defintion. Chrome points you to the first one, while it's actually using the last one (this confused the heck out of me before I realized what's happening). This should be fixed with these changes.

### DIFF
--- a/src/PostCampaign.php
+++ b/src/PostCampaign.php
@@ -443,8 +443,10 @@ class PostCampaign {
 
 			$new_theme = $themes[ $theme ] ?? [];
 
+			// As a way to make new themes use the same config file, check if removing "-new" gives a legacy theme name.
 			$potential_new_version = str_replace( '-new', '', $theme );
 			if ( ! in_array( $potential_new_version, self::LEGACY_THEMES, true ) ) {
+				// If it doesn't, stop here as all code below handles the config file.
 				return $new_theme;
 			}
 			$theme         = $potential_new_version;
@@ -531,7 +533,6 @@ class PostCampaign {
 	 */
 	private static function migrate_old_vars( array $css_vars ): array {
 		$mappings = [
-			'footer_color'            => [ 'footer--background' ],
 			'footer_links_color'      => [
 				'footer-menu--color',
 				'footer--links--hover--color',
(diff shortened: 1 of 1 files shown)


src/PostCampaign.php
```diff
@@ -591,11 +596,11 @@ public static function get_footer_theme( array $css_vars ): array {
...
 		if ( 'white' === $footer_theme ) {
-			$css_vars['footer_links_color'] = $css_vars['footer_links_color'] ? $css_vars['footer_links_color'] : $default_footer_links_color;
-			$css_vars['footer_color']       = '#FFFFFF';
+			$css_vars['footer_links_color']   = $css_vars['footer_links_color'] ? $css_vars['footer_links_color'] : $default_footer_links_color;
+			$css_vars['--footer--background'] = '#FFFFFF';
 		} elseif ( self::DEFAULT_NAVBAR_THEME === $css_vars['campaign_nav_type'] ) {
-			$css_vars['footer_links_color'] = null;
-			$css_vars['footer_color']       = null;
+			$css_vars['footer_links_color']   = null;
+			$css_vars['--footer--background'] = null;
```
Inwerpsel, line 603: I had to increase Sonarcloud's duplicate string literal detection threshold to 5 (instead of default 3). Using a constant or variable like the rule suggests would further complicate this already complex code.

planet-4:

:watch: 2021.08.03 10:49:11

src/PostCampaign.php
```diff
@@ -531,7 +533,6 @@ public static function replace_font_aliases( array $css_vars ): array {
 	 */
 	private static function migrate_old_vars( array $css_vars ): array {
 		$mappings = [
-			'footer_color'            => [ 'footer--background' ],
```
Inwerpsel, line 534: Whereas the other ones are saved on the post, this one is only set by the preceding code. So instead I made it use the right one in the first place.

PR #1432, opened 2021-08-06
Re-add GP green for plastics and climate nav color

Inwerpsel:
This was added to production content before we removed the option, and that content is still using it. If the option is not there, it's
impossible to put the color back using the UI once it's unset or switched to another color. (You can use the console to set it with `wp.data.dispatch('core/editor').setEditedPostAttribute( {meta: {campaign_nav_color: any_value})`)

This is needed now because we want to test switching content to the new version of Plastics theme, and when a theme is switched, it unsets all values on a post that are not allowed in the theme.

I also added the same fix for the Climate theme.

### DIFF
--- a/campaign_themes/climate.json
+++ b/campaign_themes/climate.json
@@ -80,7 +80,8 @@
 						{ "value": "#ffffff" },
 						{ "value": "#000000" },
 						{ "value": "#ff513c" },
-						{ "value": "#007eff" }
+						{ "value": "#007eff" },
+            { "value": "#66cc00" }
 					],
 					"default": "#ff513c"
 				}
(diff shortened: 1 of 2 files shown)


planet-4:

:new_moon: [pluto](test-pluto) | [admin](https://www-dev.greenpeace.org/test-pluto/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-pluto/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-pluto) | [composer-local.json](https://github.com/greenpeace/planet4-test-pluto/blob/main/composer-local.json)

:watch: 2021.08.06 07:42:46

mleray (approved):
👍

PR #1435, opened 2021-08-20
 PLANET-6361 Add JSON file of new themes

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6361

Add the new version as JSON files. I use this format (instead of defining vars in a CSS file) because we want to inline the styles in the back end, and loop over them in the editor to apply them.

Before trying to load the theme from the options (= themes created and saved on the instance) when serving a page, try to load this JSON file for the current theme, which will take precedence.

Last minute I noticed I overlooked the single page using the **Oceans theme** that doesn't use "Hide Page Title". This theme **customized the page header background image** quite a lot, so I added variables that allow to reproduce the old version. _I added it to this branch because it needs to also be deployed on dev sites. I kept it in because it's a small change, but if preferred I can separate it to another PR._

I opted to add the files **as JSON**, instead of CSS. That way it can use the [existing mechanism](https://github.com/greenpeace/planet4-master-theme/blob/66240f9bd46104a20ef2b68cbf880d48d44717ef/templates/css-variables.twig#L1-L7) that outputted only the style options from the editor sidebar. Having these inline ensures that the values can be set from the moment the document is loaded, and there's no delay loading an external file.

It also makes it possible to load the values in the editor and set them with JS.

For test instructions, see https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/647


planet-4:

:new_moon: [pluto](test-pluto) | [admin](https://www-dev.greenpeace.org/test-pluto/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-pluto/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-pluto) | [composer-local.json](https://github.com/greenpeace/planet4-test-pluto/blob/main/composer-local.json)

:watch: 2021.08.20 16:26:49

assets/src/scss/layout/_page-header.scss
```diff
@@ -79,27 +79,40 @@
...
+      bottom: -.5%;
+      background: linear-gradient(360deg, $white 0%, transparent 100%);
+    }
   }
 
   img {
+    _-- {
+      height: 100%;
+      @include mobile-only {
+        height: 100%;
```
Inwerpsel, line 112: The Oceans theme needs to be able to set the height to `auto` on all sizes except mobile.

assets/src/scss/layout/_page-header.scss
```diff
@@ -79,27 +79,40 @@
...
-    height: 75%;
-    bottom: -.5%;
     left: 0;
     right: 0;
-    @include linear-gradient(360deg, $white 0%, transparent 100%);
+
+    _-- {
+      height: 75%;
+      bottom: -.5%;
+      background: linear-gradient(360deg, $white 0%, transparent 100%);
```
Inwerpsel, line 104: In other places we don't use the mixin that was previously used. I checked browser support and the fallbacks are not needed anymore. https://caniuse.com/?search=linear-gradient

PR #1436, opened 2021-08-26
PLANET-6336 Implement new Cookies banner design

mleray:
### Description

See https://jira.greenpeace.org/browse/PLANET-6336

### Testing

On local you can add `new-design` to the [class in cookies.twig](https://github.com/greenpeace/planet4-master-theme/blob/master/templates/cookies.twig#L2) and replace `btn-secondary` with `btn-primary` in the button classes in the same file. That should be enough to see the new design, if you remove all cookies from storage (or uncheck the necessary cookies in the Privacy page) 🙂 

**Note**: for the A/B test these changes are applied by Optimize.

### DIFF
--- a/assets/src/scss/layout/_cookies-new.scss
+++ b/assets/src/scss/layout/_cookies-new.scss
@@ -1,16 +1,35 @@
+@mixin slide-from($direction, $margin) {
+  $animation-name: unique-id() !global;
+
+  @keyframes #{$animation-name} {
+    0% {
+      #{$direction}: -100%;
+    }
+
+    100% {
+      #{$direction}: $margin;
+    }
+  }
+
+  &.shown {
+    animation: 1s $animation-name;
+  }
+}
+
 .cookie-notice.new-design {
   z-index: 100;
   position: fixed;
-  bottom: 0;
   width: 100vw;
   background: white;
-  padding: 26px 24px 32px 24px;
+  padding: 24px 24px 32px;
   visibility: hidden;
   opacity: 0;
   transition: visibility 0.5s, opacity 0.5s linear;
   border-top-left-radius: 4px;
   border-top-right-radius: 4px;
   box-shadow: 0 0 12px rgba(0, 0, 0, 0.15);
+  bottom: 0;
+  @include slide-from(bottom, 0);
 
   &.shown {
     opacity: 1;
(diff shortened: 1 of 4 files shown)


assets/src/scss/layout/_cookies-new.scss
```diff
@@ -0,0 +1,58 @@
+.cookie-notice.new-design {
```
Inwerpsel, line 1: If this markup is very similar to the old version, and we're pretty sure we'll move to this new version, we could do the same thing as with beta blocks. Copy the old version to a new file to freeze it, and make the modifications to the existing file, so that it's easier to track history.

assets/src/scss/layout/_cookies-new.scss
```diff
@@ -1,16 +1,35 @@
...
+}
+
 .cookie-notice.new-design {
   z-index: 100;
   position: fixed;
-  bottom: 0;
   width: 100vw;
   background: white;
-  padding: 26px 24px 32px 24px;
+  padding: 24px 24px 32px 24px;
```
GP-Dan-Tovbein, line 24: Just a suggestion, you could replace to a padding shorthand `padding: 24px 24px 32px` (top|**right & left|**bottom|)

assets/src/scss/layout/_cookies-new.scss
```diff
@@ -38,21 +57,25 @@
     right: 16px;
     left: auto;
     border-radius: 4px;
-    padding: 26px 32px 32px 32px;
+    padding: 24px 32px 32px 32px;
```
GP-Dan-Tovbein, line 60: Same here

GP-Dan-Tovbein (approved):
Code still looking good to me

assets/src/scss/layout/_cookies-new.scss
mleray: Not sure how that would work with the A/B test in Optimize? 🤔

Inwerpsel: It could still contain the same CSS. I mean, if we freeze the old version by copying the file and adding `-old` to the file name, then the PR can modify the existing file name and it would be easy to see which changes are made in the new version.

Inwerpsel: Just a suggestion, could be in this case the history is not that useful :shrug:

assets/src/scss/layout/_cookies-new.scss
GP-Dan-Tovbein: I agree that is less clear, however and thinking about the performance it'll be improve the bundle size, why? because are at least 5 characters less and it's equal to at least 5 bytes less 🤓

GP-Dan-Tovbein: Of course that it won't affect the performance in this case but if you change just this in the rest of files I assume that it will.

mleray: Okay, got it! That's a fair point, I've updated the code 🙂

PR #1438, opened 2021-08-27
PLANET-6253 Add analytical cookies setting in Planet4 settings

sagarsdeshmukh:
The "Enable Analytical Cookies" checkbox is added under Planet4 > Cookies Text setting

Ref: [JIRA 6253](https://jira.greenpeace.org/browse/PLANET-6253)

---

- The "Enable Analytical Cookies" checkbox is added under Planet4 > Cookies Text setting

### DIFF
--- a/src/Settings.php
+++ b/src/Settings.php
@@ -220,7 +220,7 @@ class Settings {
 				],
 			],
 			'planet4_settings_cookies_text'     => [
-				'title'  => 'Cookies Text',
+				'title'  => 'Cookies',
 				'fields' => [
 					[
 						'name'    => __( 'Cookies Text', 'planet4-master-theme-backend' ),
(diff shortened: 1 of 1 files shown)


src/Settings.php
```diff
@@ -243,6 +243,13 @@ public function __construct() {
 						'id'   => 'enforce_cookies_policy',
 						'type' => 'checkbox',
 					],
+
```
Inwerpsel, line 246: The menu item is still "Cookies Text", even though that applies to only 1 out of 3 options. We could rename it to just "Cookies" to be more accurate.

sagarsdeshmukh: As per the discussion on JIRA comment updated as "Cookies Bar Text".

PR #1439, opened 2021-08-27
Don't add our script as a jquery dep

Inwerpsel:
* Some plugins will make jquery load in the head. They shouldn't do
this, but it happens. And when it does, it results in our front end
render script being loaded in the head, when the body doesn't exist yet.
This breaks everything.

Ref: 

---

### DIFF
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -765,16 +765,12 @@ class MasterSite extends TimberSite {
 			$css_creation
 		);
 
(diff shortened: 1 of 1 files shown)


Inwerpsel:
I created this PR after Greenland experienced an issue with a WPML setting causing the issue mentioned in the description. That issue was resolved because we were able to turn off that setting. Keeping the PR open a bit as we might still want to harden against this scenario.

PR #1447, opened 2021-09-13
PLANET-6387 Allow import/export for all post types

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6387

Remove all checks for the "campaign" post type, as this logic works on other types as well.

Makes following things appear in the UI which were only on campaign pages previously:
- Import button on the listing page. All types use the same import page URL, the type is in the imported XML.
- Export button on the listing page for each post.

Otherwise the import is functionality that didn't need any further changes to work on all post types.

I took the liberty to also add an **edit link to the post import output** if it's successful. I imagine people will often want to edit it after import.
Screenshot from 2022-01-14 16-35-05

### DIFF
--- a/load-class-aliases.php
+++ b/load-class-aliases.php
@@ -8,8 +8,8 @@
 
 class_alias( \P4\MasterTheme\Activator::class, 'P4_Activator' );
 class_alias( \P4\MasterTheme\AnalyticsValues::class, 'P4_Analytics_Values' );
-class_alias( \P4\MasterTheme\CampaignExporter::class, 'P4_Campaign_Exporter' );
-class_alias( \P4\MasterTheme\CampaignImporter::class, 'P4_Campaign_Importer' );
+class_alias( \P4\MasterTheme\Exporter::class, 'P4_Campaign_Exporter' );
+class_alias( \P4\MasterTheme\Importer::class, 'P4_Campaign_Importer' );
 class_alias( \P4\MasterTheme\Campaigner::class, 'P4_Campaigner' );
 class_alias( \P4\MasterTheme\Campaigns::class, 'P4_Campaigns' );
 class_alias( \P4\MasterTheme\Context::class, 'P4_Context' );
--- a/src/CampaignImporter.php
+++ b/src/CampaignImporter.php
@@ -24,6 +24,15 @@ class CampaignImporter {
 		add_filter( 'wp_import_post_data_processed', [ $this, 'set_imported_campaigns_as_drafts' ], 10, 2 );
 		add_action( 'import_end', [ $this, 'action_import_end' ], 10, 0 );
 		add_filter( 'wp_import_post_meta', [ $this, 'skip_duplicate_postmeta_import' ], 10, 2 );
+
+		add_action(
+			'wp_import_insert_post',
+			function ( $post_id ) {
+				$link  = get_edit_post_link( $post_id );
+				$title = esc_html( get_the_title( $post_id ) );
+				echo "successfully imported <a href=\"$link\">$title</a>\n"; //phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+			}
+		);
 	}
 
 	/**
--- a/src/CampaignImporter.php
+++ b/src/CampaignImporter.php
@@ -35,10 +35,6 @@ class CampaignImporter {
 	 * @param array   $post Post array.
 	 */
 	public function update_campaign_attachements( $post_id, $original_post_id, $postdata, $post ) {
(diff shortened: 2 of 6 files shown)


planet-4:

:new_moon: [tavros](test-tavros) | [admin](https://www-dev.greenpeace.org/test-tavros/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-tavros/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-tavros) | [composer-local.json](https://github.com/greenpeace/planet4-test-tavros/blob/main/composer-local.json)

:watch: 2021.09.13 12:34:25

src/Loader.php
```diff
@@ -99,9 +99,7 @@ private function load_services( $services ) {
 			}
 
 			// Load `CampaignExporter` class on admin campaign listing page and campaign export only.
-			if ( 'campaign' === filter_input( INPUT_GET, 'post_type', FILTER_SANITIZE_STRING ) || 'export_data' === filter_input( INPUT_GET, 'action', FILTER_SANITIZE_STRING ) ) {
-				$this->default_services[] = CampaignExporter::class;
-			}
+			$this->default_services[] = CampaignExporter::class;
```
Inwerpsel, line 102: Need to check if it's safe to remove the whole condition.

Inwerpsel: Didn't cause any issues :+1:

sagarsdeshmukh:
Can we also change the files names? 🤔 
CampaignExporter.php
CampaignImporter.php
as its applicable to all post types.

src/CampaignImporter.php
```diff
@@ -228,9 +233,7 @@ public function action_import_end() {
 	 * @return array
 	 */
 	public function set_imported_campaigns_as_drafts( $postdata, $post ) {
```
sagarsdeshmukh, line 235: could you please change the function name here.

src/Loader.php
sagarsdeshmukh: I think we should keep the second condition, 
`'export_data' === filter_input( INPUT_GET, 'action', FILTER_SANITIZE_STRING )`
It will make sure the class is only loaded when admin export the post data.

load-class-aliases.php
```diff
@@ -8,8 +8,8 @@
 
 class_alias( \P4\MasterTheme\Activator::class, 'P4_Activator' );
 class_alias( \P4\MasterTheme\AnalyticsValues::class, 'P4_Analytics_Values' );
-class_alias( \P4\MasterTheme\CampaignExporter::class, 'P4_Campaign_Exporter' );
-class_alias( \P4\MasterTheme\CampaignImporter::class, 'P4_Campaign_Importer' );
+class_alias( \P4\MasterTheme\Exporter::class, 'P4_Campaign_Exporter' );
```
Inwerpsel, line 11: We can probably remove these aliases, might do that in a follow up.

src/Loader.php
Inwerpsel: The `Exporter` class is needed to add buttons to the admin listing view. In that case that condition won't be true. It previously worked on campaign pages because it's an OR and it matched the first part.

sagarsdeshmukh: ok

sagarsdeshmukh:
The post taxonomy import is looks like not working, we don’t have it in campaign post types, so didn’t cause any issue in past -
image

sagarsdeshmukh:
Other than taxonomy import issue PR looks good to me 👍

Inwerpsel:
> The post taxonomy import is looks like not working, we don’t have it in campaign post types, so didn’t cause any issue in past - image

Could be this was never implemented, I don't find any code that looks like it should do this. Maybe it's even intentional, as I imagine usually editors don't want to have the import create new taxonomy items from another site. And it's quite challenging to map the imported taxonomy to the local one.

sagarsdeshmukh:
> > The post taxonomy import is looks like not working, we don’t have it in campaign post types, so didn’t cause any issue in past - image
> 
> Could be this was never implemented, I don't find any code that looks like it should do this. Maybe it's even intentional, as I imagine usually editors don't want to have the import create new taxonomy items from another site. And it's quite challenging to map the imported taxonomy to the local one.

I agree that importing taxonomy, will be quite challenging. 
We are using the [wordpress importer](https://github.com/WordPress/wordpress-importer) plugin to import the data, which support's the Categories, tags and terms (from custom taxonomies and term meta) import.

sagarsdeshmukh (approved):

PR #1453, opened 2021-09-15
Refactor of the current Footer campaign version

GP-Dan-Tovbein:
This PR is about to remove duplicated and unnecessary logic used by the current Footer.
Once this PR is merged, we'll be able to rebase the current version.

---


templates/footer.twig
```diff
@@ -1,4 +1,4 @@
-<footer class="site-footer" id="footer">
+<footer id="footer" class="site-footer {{ nav_type == 'planet4' ? 'site-footer_min' : ''}}">
```
GP-Dan-Tovbein, line 1: We don't use `minimal` anymore for Footer Campaigns.

templates/footer.twig
```diff
@@ -37,35 +37,38 @@
...
-					</a>
-				</li>
-			{% endfor %}
-		</ul>
+		{% if nav_type != 'planet4' %}
+			<ul class="list-unstyled footer-links">
+				{% for primary in footer_primary_menu %}
+					<li>
+						<a href="{{ primary.url }}"
+							data-ga-category="Footer Navigation"
```
GP-Dan-Tovbein, line 45: The current minimal implementation doesn't have the `data-ga` attributes in the social buttons. So, I assume we might add them.

assets/src/scss/layout/_footer.scss
```diff
@@ -147,7 +147,7 @@
   --footer--copyright-- {
     color: $grey-10;
   }
-  font: $font-size-xxs;
+  font-size: $font-size-xxs;
```
GP-Dan-Tovbein, line 150: The `font-size` property has not been set correctly since the beginning of this implementation.

footer-error

Inwerpsel: Good catch :ok_hand: I suppose we can do this change as it looks better. Maybe leave a comment on the ticket :pray:

templates/footer.twig
Inwerpsel: Yep, I also think they are supposed to have it. Maybe let Julia know they were missing here and we're adding them.

Inwerpsel (approved):

Inwerpsel:
The test instance deployments seem broken, I tried with a few instances then noticed other deployments are also failing atm. I'll merge this already as I checked locally :rocket:

PR #1458, opened 2021-09-20
Improve and refactor the current Footer version

GP-Dan-Tovbein:
- Add classnames depending of the `nav_type`.
- If the `nav_type` is not defined, the footer will be rendered as the default one.
- Show or hide the primary and secondary nav through the css and not through the template
- Rename css variables

Demo page: https://www-dev.greenpeace.org/test-rhea/campaign/campaign-for-climate-emergency/

---

### DIFF
--- a/templates/base.twig
+++ b/templates/base.twig
@@ -41,7 +41,7 @@
 	{% endif %}
 
 	{% block footer %}
-		{% include 'footer.twig' with { 'nav_type' : custom_styles.nav_type ? 'minimal' : 'default' } %}
+		{% include 'footer.twig' with { 'nav_type' : custom_styles.nav_type == 'planet4' ? 'minimal' : 'default' } %}
 
 		{% block cookies %}
 			{% include 'cookies.twig' %}
--- a/templates/base.twig
+++ b/templates/base.twig
@@ -41,7 +41,7 @@
 	{% endif %}
 
 	{% block footer %}
-		{% include 'footer.twig' with { 'nav_type' : custom_styles.nav_type } %}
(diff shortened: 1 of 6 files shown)


planet-4:

:new_moon: [rhea](test-rhea) | [admin](https://www-dev.greenpeace.org/test-rhea/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-rhea/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-rhea) | [composer-local.json](https://github.com/greenpeace/planet4-test-rhea/blob/main/composer-local.json)

:watch: 2021.09.20 20:45:42

templates/base.twig
```diff
@@ -41,7 +41,7 @@
 	{% endif %}
 
 	{% block footer %}
-		{% include 'footer.twig' with { 'nav_type' : custom_styles.nav_type } %}
+		{% include 'footer.twig' with { 'nav_type' : custom_styles.nav_type ? 'minimal' : 'default' } %}
```
Inwerpsel, line 44: The contents of `custom_styles.nav_type` come from here https://github.com/greenpeace/planet4-master-theme/blob/65fec738a6dc49f3f657a6d17cae89fde6f8ec48/single-campaign.php#L76-L76, and it's a string which can be either `minimal` or `'default`. Both would evaluate to true so it would use the `minimal` type if a page has `default` as a value.

GP-Dan-Tovbein: Got it! Thanks

Inwerpsel (approved):

PR #1461, opened 2021-09-22
PLANET-6070 Implement new Post designs

mleray:
### Description

See https://jira.greenpeace.org/browse/PLANET-6070
- Center posts and single attachment pages
- Update post meta, share buttons, breadcrumbs, author block, articles block

### Testing
Either on local or on [this post](https://www-dev.greenpeace.org/test-tavros/story/29746/test-new-post-designs-this-includes-post-header-post-meta-the-share-buttons-and-the-author-block/) I created for UAT, you can check that everything looks good according to [the designs](https://www.figma.com/file/tdt8olZ6RS4bTMqbo2cvCU/Post-page-V1?node-id=0%3A1). Also make sure that the new "Show more/Show less" buttons work properly in the author block 🙂 These changes also impacted the single attachment pages, since they share a lot of the same code, you can test an example of that [here](https://www-dev.greenpeace.org/test-tavros/oil-spill-and-burnt-forest-action-in-brazil-2/) for instance.

~~**Note 1:** For this PR I've temporarily removed the Take Action Boxout from posts, since it hasn't been updated yet with the new designs and therefore it looks really bad 😬 All the changes will be released together anyway so this will be reverted later on once we start working on the boxouts.~~ I (Pieter) removed this from the branch again because the boxout is almost ready to be merged.

**Note 2**: I've made small commits because there's quite a lot of changes involved, with different parts of the page so I figured it might be easier to review/understand that way. But before merging we could squash everything in one commit as I don't know if it makes sense to keep them separated.


planet-4:

:new_moon: [tavros](test-tavros) | [admin](https://www-dev.greenpeace.org/test-tavros/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-tavros/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-tavros) | [composer-local.json](https://github.com/greenpeace/planet4-test-tavros/blob/main/composer-local.json)

:watch: 2021.09.22 09:38:01

assets/src/js/postAuthorDescription.js
```diff
@@ -0,0 +1,23 @@
+export const setupPostAuthorDescription = () => {
```
Inwerpsel, line 1: Maybe it's an option to use bootstrap for this?

Inwerpsel: Like we do for the main navigation dropdown https://github.com/greenpeace/planet4-master-theme/blob/1511143660231f541c3de4d975c3627be74f4108/templates/navigation-bar.twig#L8-L12

Inwerpsel: If I check [BS's docs on multiple targets](https://getbootstrap.com/docs/5.1/components/collapse/#multiple-targets), their example looks like it has similar markup but both elements hide/show at the same time. Not sure why in our case it does it differently.

It looks somewhat better if the ellipsis `span` is after the rest of the text. It still shows too long but it's hard to notice.

templates/single.twig
```diff
@@ -3,15 +3,11 @@
 {% block content %}
 
 	<div class="clearfix"></div>
-	<div class="skewed-overlay"></div>
```
Inwerpsel, line 6: :partying_face:

templates/blocks/author_profile.twig
```diff
@@ -1,37 +1,53 @@
...
+							data-bs-toggle="collapse"
+							data-bs-target=".multi-collapse"
+							aria-expanded="false"
+						>
+							<span class="show-more">{{ __('Show more', 'planet4-master-theme') }}</span>
+							<span class="show-less">{{ __('Show less', 'planet4-master-theme') }}</span>
+						</button>
+					{% else %}
+						{{ post.author.description|raw }}
+					{% endif %}
```
Inwerpsel, line 48: Since the `else` contains only a small fragment of code, could be easier to read if it's handled first 
```twig
{% if ( post.author.description|length < author_bio_char_limit ) %}
	{{ post.author.description|raw }}
{% else %}
    ...
{% endif %}

File: assets/src/scss/layout/_page-header.scss [hunk 4]
```diff
@@ -126,26 +116,12 @@
   font-family: $roboto;
   font-weight: bold;
   text-transform: none;
-  max-width: 80%;
-  margin-bottom: 10px;
-
```
Inwerpsel, line 131: Do we intend to also change this for pages and campaign pages? It could also cause issues with background image alignment. Maybe we should still include the removed rules for pages and campaign pages in a separate file? :thinking: That would be easier to do if we had code splitting per post type, so we wouldn't have to scope the CSS.

assets/src/scss/pages/post/_author-block.scss
```diff
@@ -1,86 +1,123 @@
...
+    text-align: center;
+    margin-bottom: 16px;
   }
-}
 
-.author-block-info-title {
-  font-size: $font-size-md;
-  font-weight: bold;
-  text-transform: uppercase;
+  .author-block-image {
```
Inwerpsel, line 27: Do we have to move this selector inside? It already has `.author-block` in the name, so we're not really gaining anything by nesting it. If it's very nested and you get a link like this https://github.com/greenpeace/planet4-master-theme/blob/a150337b5e34805ff2cc00e195c2d253e3314432/assets/src/scss/pages/post/_author-block.scss#L68, it's not trivial to locate all the lines to construct the selector if it goes a few levels deep.

Inwerpsel: Whereas before all these were just at the top level, and it's only a few lines up to know what the line does. This may not be obvious for you now because you just worked on the code, but if you're searching through a lot of code and don't have much context it can make a big difference.

Inwerpsel: Compare it to the current version. This is the deepest nested line I found and it's still easy to see what the selector is without scrolling. https://github.com/greenpeace/planet4-master-theme/blob/master/assets/src/scss/pages/post/_author-block.scss#L75

assets/src/js/postAuthorDescription.js
mleray: In our case the ellipsis has the `show` class at the beginning, so the button toggles both, but depending on their class it either shows or hides. I'll test putting the ellipsis at the end, thanks!

mleray: I've pushed a small change just setting `display: none;` for the `collapsing` class, I feel like it's a bit better...?

assets/src/scss/layout/_page-header.scss
mleray: True! I've just pushed a commit to move these changes to `_post.scss` instead so it won't affect pages and campaigns 👍

assets/src/js/postAuthorDescription.js
Inwerpsel: It's still a bit weird because there's a delay when showing the rest of the text before it pops up. I suppose that delay is because BS would do a sliding transition, but in our case it's not sliding and it pops up at the end of the transition duration.

mleray: I agree. I've pushed a new commit adding `transition: none;` and that seems to solve it!

templates/blocks/author_profile.twig
```diff
@@ -1,37 +1,53 @@
...
+						src="{{ fn('get_avatar_url', post.author.id, {'size' : 294 }) }}"
+						alt="{{ post.author.name }}"
+					/>
+				</figure>
+			{% endif %}
 			<div class="author-block-info">
-				<h2 class="author-block-info-title">{{ __( 'about the author', 'planet4-master-theme' ) }}</h2>
+				<h2 class="author-block-info-title">{{ __( 'About the author', 'planet4-master-theme' ) }}</h2>
+				{% if ( post.author.avatar ) %}
+					<figure class="author-block-image">
```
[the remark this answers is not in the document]
mleray: Pushed!

Inwerpsel:
When rebasing this branch against the target branch I ran into many conflicts, some of which were quite complex to solve. Luckily I used PHPStorm's smart conflict resolution tool, else I would not have managed to resolve them.

Nevertheless I didn't fully inspect some of the conflicts, specifically those where `planet-6070` removes code that the `new-post-design` branch had changed. In those cases I didn't try to adapt those changes to `planet-6070`.

I suggest we merge this PR and resolve any issues on `new-post-design`, before merging that one into `master`.

comzeradd:
We could even switch GPI dev site to `new-post-design` branches to have everything in a live instance.

PR #1466, opened 2021-09-24
Planet 6189 new footer design

GP-Dan-Tovbein:
This is the new version of the Footer site based in these [versions](https://p4-designsystem.greenpeace.org/05f6e9516/p/16a899-footer).

**What is done.**
- Footer S, M, L and XL
- Minimal Footer S and L. Because we haven't any mockup for minimal's version, we reported it to the Designer and suggested a potential layout and it was approved by Magali. Here is the [thread of the conversation](https://greenpeace-gpi.slack.com/archives/G015K63081W/p1626186385006500)
- Added a new Facebook and Linkedin rounded icons styled by adding this class name  `[social-name]-rounded` format. i.e. `facebook-rounded`.
- Improve the css using BEM for class names.
- Modify class names from footer
- Layout changes

---

### DIFF
--- a/assets/src/scss/layout/_footer.scss
+++ b/assets/src/scss/layout/_footer.scss
@@ -4,23 +4,27 @@
     background: $dark-blue;
     color: $white;
   }
-  padding: 45px 0 $n25;
   position: relative;
   z-index: 2;
   text-align: center;
   transition: margin 1s;
+  padding-top: 24px;
   clear: both;
   line-height: 0;
 
-  @include medium-and-up {
-    padding: $n25 0 $n20;
+  a {
+    transition: color 100ms linear;
+
+    --footer--link-- {
+      color: inherit;
+
+      &:hover {
+        color: $white;
+      }
+    }
   }
 
-  @include large-and-up {
-    padding: 55px 0 $n25;
-  }
-
-  a --footer--links-- {
+  .icon --footer--icon-- {
     color: inherit;
 
     &:hover {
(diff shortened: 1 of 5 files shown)


planet-4:

:new_moon: [sinope](test-sinope) | [admin](https://www-dev.greenpeace.org/test-sinope/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-sinope/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-sinope) | [composer-local.json](https://github.com/greenpeace/planet4-test-sinope/blob/main/composer-local.json)

:watch: 2021.09.24 13:28:46

templates/footer_social_media.twig
```diff
@@ -1,36 +1,40 @@
-<ul class="footer-social-media list-unstyled col-md-8 col-lg-5">
-  {% if social_overrides is defined and social_overrides|length %}
-    {% for social in social_overrides %}
-      <li>
-        <a href="{{ social.url }}"
+{% set total_social_icons = social_overrides is defined and social_overrides|length ? social_overrides|length : footer_social_menu|length %}
+
+<div class="site-footer__social-media site-footer__social-media--{{ nav_type }}">
```
Inwerpsel, line 3: I don't think BEM plays very well here.

Inwerpsel: I mean, there's `site-footer` on everything :smile:

GP-Dan-Tovbein: I don't agree with this because this is how BEM works (block__element--modifier).
`site-footer` is the block, then, every element inside the block should be named as `block__element`.

`site-footer__social-media`: In this case, `site-footer` is the block and `social_media` is the element.
`site-footer__social-media--{{ nav_type }}`: In this case `site-footer` is the block, `social_media` is the element and `--{{ nav_type }}` is the modifier.

So, if you use BEM, `site-footer` will be added in all classnames since is the `block`.

Look at [this](https://simple-web.dev/bem-by-example) example.
> It's important to combine the original class, and the modifier class even if it looks a bit ugly.

themes/climate.json
```diff
@@ -19,6 +18,18 @@
   "--footer--copyright--year--color": "#66cc00",
   "--footer--links--color": "#66cc00",
   "--footer--links--hover--color": "#66cc00",
+  "--footer--background": "#ffffff",
```
GP-Dan-Tovbein, line 21: These are the new variables added to new footer design

templates/footer_social_media.twig
Inwerpsel: I understand what the logic behind BEM is, my point is that it doesn't work well in this instance. I'm open to being convinced to using a methodology for naming if you can provide a good argument for it. Just saying 
> I don't agree with this because this is how BEM works 

is not very convincing.

Inwerpsel: Let me go into detail about some things that are not playing well here:
- `site-footer__social-media-list--justify-space-between`: "modifier" in BEM is very open to interpretation and ends up containing all kinds of things. In this case it corresponds to a CSS rule that we happen to put on lists that have 4 to 6 items.
- This uses a `span` with a title, and then uses BEM to make it "semantic". Whereas using an `h2` or other heading would be more semantic, and probably remove the need to add a separate selector, as we could write CSS as `.site-footer__social-media h2`.
- `site-footer__social-media--{{ nav_type }}`: the footer type is already present in [on the `footer` element this is inside](https://github.com/greenpeace/planet4-master-theme/pull/1466/files#diff-9ccc5c1e49803dab74b7d5fb3b391bb54c4521688a0f01b887e1d082a4759e60R1). So we can also write the following CSS `site-footer--minimal .social-media {...}`

comzeradd: Although BEM is a nice standard for css coding, we are not using it anywhere else in our codebase (at least not intentionally). So I'd be very careful of introducing it now, since we also need to see how Design Tokens may affect the way we name css classes.

So although `site-footer__social-media-list--justify-space-between` would probably be correct in this case, if we wanted to follow BEM, it strikes to me as completely different compared to the rest of our codebase. My suggestion would be to keep it simple for now, and try to keep the same css classes logic as the current footer. We can re-visit usage of BEM, once we start working on Design Tokens and css variables more extensively.

GP-Dan-Tovbein: Sorry, my last message was added at the same time as yours @comzeradd.

Thanks both for your comments 😄! I'll remove BEM on my code and hope we should find another way for styling our blocks, maybe `styled-components` 🤞 It could be a good experiment to do 😉

assets/src/scss/base/_mixins.scss
```diff
@@ -205,3 +205,11 @@ $nested-comment-left-margin: 50px;
     }
   }
 }
+
+@mixin justifyContent($contentDistribution) {
```
GP-Dan-Tovbein, line 209: I've added this `mixin` to avoid duplicating code in all places that we're checking the browser support of the `justify-content` property. 

Here is more info https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content

templates/footer_social_media.twig
GP-Dan-Tovbein: I've already updated the PR. Removed all related to BEM. Thanks guys!

assets/src/scss/base/_mixins.scss
Inwerpsel: We probably don't need to check supports anymore for this, the previous code was probably a leftover from when that used to be needed.

Inwerpsel: https://caniuse.com/?search=flex-start

themes/climate.json
Inwerpsel: Keep in mind we also need to update the other 2 themes. But probably it's going to be easier to do this last, in case we change names so you don't have to keep updating these 4 files.

assets/src/scss/layout/_footer.scss
```diff
@@ -30,137 +35,114 @@
...
+    @include large-and-up {
+      column-count: 2;
+    }
   }
 
-  @include medium-and-up {
-    font-size: $font-size-xxs;
-  }
+  .copyright {
+    --footer--copyright-- {
```
Inwerpsel, line 74: This is something we probably do want to change. These are some of the earlier variables I added, before there was the shorthand. If we use it here then the resulting variable name will be `--site-footer--copyright`, which is more specific because there may be many kinds of footers.

themes/climate.json
GP-Dan-Tovbein: Got it! I'll keep this conversation open, then.

assets/src/scss/base/_mixins.scss
GP-Dan-Tovbein: Removed the support check about the `justify-content` property

assets/src/scss/layout/_footer.scss
GP-Dan-Tovbein: Got it! It's already changed

GP-Dan-Tovbein:
@Inwerpsel it's updated the PR according to your review. Looking forward to your feedback 👍

GP-Dan-Tovbein:
@Inwerpsel FYI, I've pushed a new commit about all [requested changes](https://jira.greenpeace.org/browse/PLANET-6189?focusedCommentId=243327&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-243327)  by Houssam.

GP-Dan-Tovbein:
@Inwerpsel I've changed the label from **UAT Needed** to **UAT Passed**

assets/src/scss/layout/_footer.scss
```diff
@@ -1,25 +1,30 @@
...
 
-  @include medium-and-up {
-    padding: $n25 0 $n20;
-  }
+  a {
+    transition: color 100ms linear;
 
-  @include large-and-up {
-    padding: 55px 0 $n25;
+    --site-footer--link-- {
```
Inwerpsel, line 18: I think I earlier gave these a custom variable prefix because `a` is not very descriptive as name. On the other hand, this could also be solved by using custom labels in a UI that needs to have more readable names. The convenience of auto generating the name maybe outweighs the limited benefit of having `link` in the name, and if we use `a` consistently it can be easy to get used to.

@GP-Dan-Tovbein What do you think?

templates/footer.twig
```diff
@@ -13,9 +13,7 @@
 						</a>
 					</li>
 				{% endfor %}
-			</ul>
-
-			<ul class="list-unstyled footer-links-secondary">
+
```
Inwerpsel, line 16: It's a bit odd that we end up with 2 different menus that just concatenated at the front end. I suppose we don't want to keep that situation indefinitely and ideally want to have people merge these to a single menu?

GP-Dan-Tovbein: I agree with that. Also take a look [this](https://jira.greenpeace.org/browse/PLANET-6189?focusedCommentId=243206&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-243206) thread.

GP-Dan-Tovbein:
@Inwerpsel it's updated the PR with all requested changes!

PR #1471, opened 2021-09-29
 PLANET-6257: New country selector design implementation

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-6257

Implementation of the new Country Selector as described in the design system https://p4-designsystem.greenpeace.org/05f6e9516/v/0/p/16a899-footer

## Feature

__This is implemented behind a feature toggle__, so that every child-theme has time to adapt. 
In the admin panel, section _Planet 4 > Features_:
Screenshot from 2021-11-24 12-37-18

This will remove the country selector from the navigation bar and add it to the footer.
Screenshot from 2021-11-04 19-51-48
## Modifications

- `h3` tag was removed, the list structure and labels should be enough to express the meaning of elements
- Links have an `hreflang` attribute to indicate the main language of the URL
- JS script was dropped in favor of a backend generated template (cf `templates/footer_country_selector.twig`)

## For child themes extension

- A twig template [`templates/footer_country_selector.twig`](https://github.com/greenpeace/planet4-master-theme/pull/1471/files#diff-26a6ff6c33f468b55994daf5fd0006c20cef6073c1f06feec8200a30fff33c8a) is included in the `templates/footer.twig`.  
  Its content is generated from a backend variable [`countries_by_initials`](https://github.com/greenpeace/planet4-master-theme/pull/1471/files#diff-d5a71432d6e94abc1260a448df306d8dfc76d15fa4047646a4d133c38cc131d7R558-R561), coming from the json file `templates/countries.json`.
- The country selector colors should adapt to the footer colors by default.

Screenshot from 2021-11-17 20-32-52

## Test

- Activate the feature in Planet4 > Features "Enable  Country selector design"
- Check the Country selector in the footer of the frontend - it should also adapt to campaigns colors, and doesn't appear in minimal navigation mode


planet-4:

:watch: 2021.10.01 13:21:51

GP-Dan-Tovbein:
@lithrel I highly recommend to set the origin branch to [PLANET-6189_new_footer_design](https://github.com/greenpeace/planet4-master-theme/tree/PLANET-6189_new_footer_design), which is includes the latest approved changes.

Also, this is the "almost done" [PR](https://github.com/greenpeace/planet4-master-theme/pull/1466).

assets/src/scss/layout/_country-selector.scss
```diff
@@ -0,0 +1,162 @@
...
+    &::after {
+      --country-selector--chevron-- {
+        background-color: var(--site-footer--color, $white);
+      }
+      background-repeat: no-repeat;
+      content: "";
+      display: inline-block;
+      height: 8px;
+      margin-inline-start: 8px;
+      mask: url("../../images/chevron-down.svg") 0 0/8px 8px;
```
comzeradd, line 59: Is this different than the existing [chevron icon](https://github.com/greenpeace/planet4-master-theme/blob/master/assets/src/images/icons/chevron.svg)? Maybe we can use the same and just rotate it?

lithrel: Yep :+1: working on fixes right now, it's in it :)

templates/footer.twig
```diff
@@ -1,4 +1,9 @@
 <footer id="footer" class="site-footer site-footer--{{ nav_type }}">
+	{% if new_design_country_selector and nav_type != 'minimal' %}
+	<div id="country-selector" class="site-footer__country-selector">
```
GP-Dan-Tovbein, line 3: Could you change `site-footer__country-selector` to `footer-country-selector` or sth like that? since `site-footer__country-selector` is referring to a BEM naming fomat but we're not using that. I initially used BEM on the new Footer design but I had to changed it and respect the style guide of class names we have.

assets/src/scss/layout/_country-selector.scss
```diff
@@ -0,0 +1,200 @@
...
+    list-style: none;
+    margin: 0;
+    padding-inline-start: $space-lg;
+
+    @include medium-and-up {
+      column-count: 3;
+    }
+
+    & li --country-selector--item-- {
+      font-size: $font-size-sm;
```
GP-Dan-Tovbein, line 129: Do we want to allow editors to customize these two attributes?

templates/footer_country_selector.twig
```diff
@@ -0,0 +1,65 @@
...
+			aria-label="{{ data_nav_bar.country_dropdown_toggle }}"
+			data-ga-category="Menu Navigation"
+			data-ga-action="Open Country Selector"
+			data-ga-label="{{ page_category }}">
+				<span class="visually-hidden">{{ __( 'Selected', 'planet4-master-theme' ) }}:</span> {{ website_navbar_title }}
+				<span class="visually-hidden">{{ __( 'Change Country', 'planet4-master-theme' ) }}</span>
+		</button>
+	</div>
+</div>
+<div id="countries-list" class="countries-list">
```
GP-Dan-Tovbein, line 17: Not sure why do you need an `id` here. Correct me if I'm wrong, but I don't see any use with it.

templates/navigation-bar.twig
```diff
@@ -19,6 +19,7 @@
 			</span>
 		</button>
 		<ul id="navbar-dropdown" class="navbar-dropdown list-unstyled">
```
GP-Dan-Tovbein, line 21: Not sure why do you need an id here. Correct me if I'm wrong, but I don't see any use with it.

templates/base.twig
```diff
@@ -49,6 +49,8 @@
 
 		{{ function('wp_footer') }}
 	{% endblock %}
-	<script id="countries_script" type="application/json">{% include "countries.json" %}</script>
+	{% if not new_design_country_selector %}
+		<script id="countries_script" type="application/json">{% include "countries.json" %}</script>
```
GP-Dan-Tovbein, line 53: A good option is to set as `async` to this `script`. This allows to stop the HTML parsing only when the script is executed.

GP-Dan-Tovbein (commented):
I left some comments!

templates/navigation-bar.twig
lithrel: Not my code :grin:  but it's used by the current country selector script as a selector (although probably not needed even for that) - I will remove this in the new navbar design implementation

assets/src/scss/layout/_country-selector.scss
```diff
@@ -0,0 +1,200 @@
+
+#country-selector {
+  --country-selector-- {
+    color: var(--site-footer--color, $white);
```
Inwerpsel, line 4: What is the motivation for using the footer variables here?

Inwerpsel: As I understand, it could be to have the country selector color default to the general color of the site footer, but still allow changing it individually.

However it only has a small benefit to add `--site-footer--color` variable here, because you can also just put the same color into both names in the theme.

Perhaps we also don't want to have this color be separate for both. In that case we can just remove these rules and rely on CSS cascade of the `--site-footer` rules.

assets/src/scss/layout/_country-selector.scss
Inwerpsel: Semantic remark: You'll probably still see it referred to with "templates" in the code and UI, but that was inaccurate naming, as these are actually "themes".

Like Dan said, these might not be needed. Though it is possible that certain color combinations require a font change (mostly `font-weight`) to look nice.

assets/src/scss/layout/_country-selector.scss
lithrel: > As I understand, it could be to have the country selector color default to the general color of the site footer, but still allow changing it individually.

yup

> you can also just put the same color into both names in the theme.

Right, I was adapting my code to themes with the intent of not modifying any theme config, because that was the easiest thing I found giving me a hint on what properties we need to be easily modifiable. Checking other sites, [nordic themes](https://www.greenpeace.org/sweden/) have a different color for footer and country selector, so that usage exists (but of course I don't know what they will do when the country selector moves to the footer).
Sooo, I think it makes sense like this, but removing the default fallback and adding country-selector variables to campaign themes will also work for me so I can also do it if you prefer ? and keeping those variables separated seems useful

Inwerpsel: Yeah it makes sense as a temporary measure, in case there are some downstream themes that can't be updated.

For now [all themes we need to support](https://github.com/greenpeace/planet4-master-theme/tree/master/themes) are still inside the same repo, so we can adjust their config files and keep our code cleaner.

You can try using the experimental theme editor to generate the new config file. Should be straightforward if it uses colors that are already in the theme, you can select it from the suggestions. Unfortunately that only works for colors so with other things like font size it's a bit more manual. But might be easier to check some things than just using the JSON file.

I don't think any child theme depends on the CSS variables yet, they still use plain old CSS overrides. We still need to be sure we can offer them good usage steps so that we can handle new variables.

assets/src/scss/layout/_country-selector.scss
```diff
@@ -0,0 +1,195 @@
...
+#country-selector {
+  --country-selector-- {
+    color: $white;
+    font-size: $font-size-sm;
+  }
+  margin-bottom: 24px;
+  text-align: start;
+  width: 100%;
+
+  a --country-selector--a-- {
```
GP-Dan-Tovbein, line 11: You can also use 
```
a _-- {
    color: $white;
```

And will be parsed as `--country-selector--a--color`

assets/src/scss/layout/_country-selector.scss
```diff
@@ -0,0 +1,195 @@
...
+    max-height: 0;
+    overflow: hidden;
+    transition: max-height 0.5s cubic-bezier(0, 1.04, .38, .37) 0s;
+  }
+
+  .countries-list .container {
+    padding: 24px 24px $space-lg 16px;
+
+    @include medium-and-up {
+      padding: 24px 0 $space-lg 0;
```
GP-Dan-Tovbein, line 92: Use this shorthand `padding 24px 0 $space-lg`;

templates/footer_country_selector.twig
```diff
@@ -0,0 +1,65 @@
...
+{% if countries_by_initials is not empty %}
+	{% set international = countries_by_initials|first|first %}
+	<div class="container">
+	<button class="country-selector-toggle control-close"
+		data-bs-toggle="open"
+		data-bs-target="#country-selector"
+		aria-expanded="false"
+		aria-label="{{ __( 'Close country selector', 'planet4-master-theme' ) }}">
+	</button>
+	<ul class="countries" aria-label="{{ __('Worldwide site list', 'planet4-master-theme') }}" role="list">
```
GP-Dan-Tovbein, line 27: Don't need this `role="list"` because it's a list.

GP-Dan-Tovbein (changes requested):
I've added some comments. I see that you've defined potentially variables to themes but I don't see any theme with those vars. You should try to edit at least one theme and test, maybe oceans or plastics.

assets/src/scss/layout/_country-selector.scss
```diff
@@ -0,0 +1,195 @@
+
+#country-selector {
+  --country-selector-- {
```
Inwerpsel, line 3: Here you can use `_--` and it will have the same variable names.

assets/src/scss/layout/_country-selector.scss
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    .countries-list {
+      max-height: 5000px;
+      transition: max-height 1s linear 0s;
+    }
+  }
+
+  ul.countries {
+    column-count: 1;
```
Inwerpsel, line 113: `column-count` has a drawback we discovered on the footer, which probably is hard to get around. Not sure if it leads to other similar issues.

Check how the focus outline is split over 2 columns.

Screenshot from 2021-11-16 13-38-46

assets/src/scss/layout/_country-selector.scss
```diff
@@ -0,0 +1,195 @@
...
+    outline-offset: 0;
+  }
+
+  button:focus:not(:focus-visible) {
+    outline: 0;
+  }
+
+  .country-selector-toggle-container {
+    --country-selector--toggle-container-- {
+      border-bottom: 1px solid rgba(255, 255, 255, 0.5);
```
GP-Dan-Tovbein, line 32: I don't see this necessary to implement cause I'm not sure if the editors want to edit the border-bottom

assets/src/scss/layout/_country-selector.scss
```diff
@@ -0,0 +1,195 @@
...
+    align-self: flex-start;
+    background: transparent;
+    border: none;
+    color: var(--site-footer--color, $white);
+    cursor: pointer;
+    text-align: start;
+    white-space: nowrap;
+    width: 100%;
+
+    &::before {
```
GP-Dan-Tovbein, line 54: `&::before` and `&::after` are sharing some attributes that you can put them together like this

```
&::before, 
&::after {
background-color: var(--site-footer--color, $white);
content: "";
display: inline-block;
width: 16px;
}
```

assets/src/scss/layout/_country-selector.scss
```diff
@@ -0,0 +1,195 @@
...
+      transform: rotate(0.25turn);
+      transition: transform 300ms cubic-bezier(.86, 0, .07, 1);
+      width: 16px;
+    }
+  }
+
+  .countries-list {
+    --country-selector--list-- {
+      background-color: $active-blue;
+      font-weight: 500;
```
GP-Dan-Tovbein, line 81: I don't see this attribute to be necessary to add to the theme.

assets/src/scss/layout/_country-selector.scss
```diff
@@ -0,0 +1,195 @@
...
+  li.international {
+    margin: 0;
+    padding: 0 0 $space-lg 0;
+  }
+
+  li.country-group-letter {
+    .initial {
+      --country-selector--initial-- {
+        color: rgba(255, 255, 255, 0.5);
+        font-weight: 700;
```
GP-Dan-Tovbein, line 137: The same here. If it's really necessary for editor, then add it, if not remove it from this block

assets/src/scss/layout/_country-selector.scss
lithrel: Seems limited to Chrome (doesn't happen on Firefox or Safari for me) ? I checked in their bugs, on a semi-related bug they [mention](https://bugs.chromium.org/p/chromium/issues/detail?id=269061#c31) working on a new thing that would fix rendering issues with columns, but not sure it would include this one.

The country selector also has this crop from the top list element, but it doesn't appear on the bottom of the previous column.
Adding a sufficient `padding-top` to the `li` element would work, that probably also means changing other sizes. Considering the low impact I would prefer to open a bug ticket for this and look for a proper solution later.

assets/src/scss/layout/_country-selector.scss
```diff
@@ -28,10 +28,7 @@
   }
 
   .country-selector-toggle-container {
-    --country-selector--toggle-container-- {
-      border-bottom: 1px solid rgba(255, 255, 255, 0.5);
-    }
-
+    border-bottom: 1px solid rgba(255, 255, 255, 0.5);
```
GP-Dan-Tovbein, line 31: Not necessary to change, but you can also use `transparentize($white, .5);`

GP-Dan-Tovbein: The same with the other cases you use `rgba` colors

assets/src/scss/layout/_country-selector.scss
```diff
@@ -0,0 +1,192 @@
...
+    margin: 0;
+    padding: 0 0 $space-lg 0;
+  }
+
+  li.country-group-letter {
+    .initial {
+      --country-selector--initial-- {
+        color: rgba(255, 255, 255, 0.5);
+      }
+      font-size: 0.75rem;
```
GP-Dan-Tovbein, line 135: You can also use `$font-size-xxxs`

GP-Dan-Tovbein (changes requested):
Code looks good but I left some small requested changes

GP-Dan-Tovbein (approved):
Code looks good to me.

We should validate this implementation with campaign pages, if it's included

assets/src/scss/layout/_country-selector.scss
```diff
@@ -0,0 +1,192 @@
...
+      height: 12px;
+      margin-inline-start: 8px;
+      mask: url("../../images/chevron.svg") 0 0/16px 12px no-repeat;
+      transform: rotate(0.25turn);
+      transition: transform 300ms cubic-bezier(.86, 0, .07, 1);
+    }
+  }
+
+  .countries-list {
+    --country-selector--list-- {
```
Inwerpsel, line 76: You can also use the shorthand here.

assets/src/scss/layout/_country-selector.scss
```diff
@@ -0,0 +1,192 @@
...
+      line-height: 1;
+    }
+  }
+
+  li.international {
+    margin: 0;
+    padding: 0 0 $space-lg 0;
+  }
+
+  li.country-group-letter {
```
Inwerpsel, line 130: We don't need to add `li` here, and we can move this entire selector outside of `.country-selector` since it already has a nice long name. This will make the deeper parts much more readable.

assets/src/scss/layout/_country-selector.scss
```diff
@@ -0,0 +1,192 @@
...
+    border-radius: var(--link--focus--border-radius, 4px);
+    outline: var(--link--focus--border, 2px solid transparentize($white, 0.6));
+    outline-offset: 0;
+  }
+
+  button:focus:not(:focus-visible) {
+    outline: 0;
+  }
+
+  .country-selector-toggle-container {
```
Inwerpsel, line 30: This selector can be moved out of `#country-selector`.

assets/src/scss/style.scss
```diff
@@ -38,6 +38,7 @@ Text Domain: planet4-master-theme
 @import "layout/blocks";
 @import "layout/cookies";
 @import "layout/cookies-new";
+@import "layout/country-selector";
```
Inwerpsel, line 41: Maybe we should look for a way to split this conditional CSS from the main CSS? If you see above it we also have `cookies-new`, and we may want to add more things behind a feature toggle.

There's a relatively high performance penalty because this is all blocking CSS.

Could be nice to have a mechanism that avoids this problem, but might be complicated to implement, as each separate file needs to have a entrypoint to build it.

lithrel: I created a new style-new.scss file for the navigation bar implementation, because there's too many conflicts with the current CSS that are too hard to avoid, and it's a bigger file. I'll look for a more maintainable way to do that in the navigation bar PR

Inwerpsel: Nice :+1: Maybe we can set up code splitting similar to how our blocks work? We currently only do it for blocks, but there's a lot of areas where we can apply it. And it potentially allows us to avoid the majority of the current blocking CSS in most cases.

Inwerpsel: For example here it would allow us to avoid duplicating the main entrypoint. But also afterwards we could exclude all this CSS if the minimal navigation is used.

assets/src/scss/layout/_country-selector.scss
```diff
@@ -0,0 +1,192 @@
...
+  font-size: $font-size-sm;
+  margin-bottom: 24px;
+  text-align: start;
+  width: 100%;
+
+  a _-- {
+    color: inherit;
+  }
+
+  a:hover,
```
Inwerpsel, line 15: We can probably remove these selectors for hover and focus, because they set the same value as the generic ones, and use the same variable.

assets/src/scss/layout/_country-selector.scss
```diff
@@ -0,0 +1,192 @@
...
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  padding-bottom: 24px;
+  transition: border 0s linear 0.5s;
+  vertical-align: middle;
+  width: 100%;
+}
+
+.country-group-letter {
```
Inwerpsel, line 166: This name seems like it refers to the letter for the group, but it refers to the group itself (which is grouped by letter). It's probably less confusing to call it `country-group`. Then you can auto generate the name of the `initial` class below and it will actually make sense `--country-group--initial`.

assets/src/scss/layout/_country-selector.scss
Inwerpsel: Interesting, I think last time I checked Firefox for this and it also had the issue, but now indeed it doesn't. Perhaps they did a fix.

assets/src/scss/style.scss
comzeradd: Just a quick comment on the performance aspect. We don't plan to keep that under a feature flag for very long. Actually it would be activated by default in almost all NROs. It would only be a handful of NROs that would need to update their child theme markup before they toggle the switch.

assets/src/scss/layout/_country-selector.scss
lithrel: I removed the `a` one, but the generic styles we have for `.btn` are a lot more specific than just those border/outline fixes, that would lead to more specific declarations in the end

assets/src/scss/layout/_country-selector.scss
```diff
@@ -0,0 +1,189 @@
...
+      transform: rotate(-0.25turn);
+    }
+
+    .countries-list {
+      max-height: 5000px;
+      transition: max-height 1s linear 0s;
+    }
+  }
+
+  ul.countries {
```
Inwerpsel, line 97: Can you remove this redundant `ul`, and the other redundant elements where a class is also specified in this file?

The problem with high CSS specificity is that you'll have to at least match the specificity in any overrides you do afterwards (and there are a few in this file). It results in more complex code that is prone to specificity-related regressions.

It also means that any child theme that wants to override CSS needs to also include a selector that is specific enough, and so will have the same issues.

We create a class for this specific purpose anyway, we don't intend to use it elsewhere. If `countries` by itself is not semantic enough, we can rename it.

assets/src/scss/layout/_country-selector.scss
Inwerpsel: Definitely not a blocker for me, just highlighting it so we can check whether an alternative makes sense.

assets/src/scss/layout/_country-selector.scss
```diff
@@ -0,0 +1,189 @@
...
+    &::after {
+      height: 12px;
+      margin-inline-start: 8px;
+      mask: url("../../images/chevron.svg") 0 0/16px 12px no-repeat;
+      transform: rotate(0.25turn);
+      transition: transform 300ms cubic-bezier(.86, 0, .07, 1);
+    }
+  }
+
+  .countries-list {
```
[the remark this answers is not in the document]
lithrel: > (if there is any)

PTSD from trying to not conflict with existing rules when implementing this and the navbar :grin:  Having everything in one specific selector made it easy to not have side effects on other existing themes/components and make 2 codebases live together. But I see the other side of the coin, I'll take stuff out of the main selector.

src/Features.php
```diff
@@ -156,6 +161,13 @@ public static function is_active( string $name ): bool {
 	 * Add hooks related to Features activation
 	 */
 	public static function hooks() {
+		add_action(
+			'cmb2_options-page_process_fields_' . Settings::METABOX_ID,
+			__CLASS__ . '::on_pre_process',
```
Inwerpsel, line 166: You can also use the array notation to pass the callable. It has better IDE support.
```php
[ self::class, 'on_pre_process' ]
```
On PHPStorm for me the current notation using `__CLASS__` is not picked up as a usage, which causes unused elements warnings and doesn't allow for navigating to the declaration.

assets/src/scss/layout/_country-selector.scss
```diff
@@ -0,0 +1,189 @@
...
+    margin-inline-start: -#{$space-lg};
+    opacity: 0.5;
+    position: absolute;
+
+    html[dir="rtl"] & {
+      text-transform: lowercase;
+    }
+  }
+
+  & > ul {
```
Inwerpsel, line 161: I think you don't need to add the `&` here, it's only for joining a selector that needs to be added without a space.

assets/src/scss/layout/_country-selector.scss
```diff
@@ -0,0 +1,189 @@
...
+  font-size: $font-size-sm;
+  margin-bottom: 24px;
+  text-align: start;
+  width: 100%;
+
+  a _-- {
+    color: inherit;
+  }
+
+  button:hover {
```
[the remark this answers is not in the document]
Inwerpsel: Can also be done in a follow up.

Inwerpsel (approved):

assets/src/scss/layout/_country-selector.scss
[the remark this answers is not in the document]
Inwerpsel: Just to have some reference on the PR from things we discussed elsewhere:

Because the new and old CSS were conflicting, which forced us to complexify the new CSS so that both could be loaded together without affecting each other, I gave the suggestion I made earlier in this PR a shot. I [split off the country selector related code to 2 separate entrypoints](https://github.com/greenpeace/planet4-master-theme/pull/1521), and conditionally made it use either the old or the new version. That way it never has both versions loaded at the same time.

assets/src/scss/style.scss
Inwerpsel: @comzeradd I ended up doing [a PR to set this up](https://github.com/greenpeace/planet4-master-theme/pull/1521), after seeing that having both variants loaded at the same time caused specificity issues, which would force us to make the new CSS more complex.

Also it allowed us to merge this PR earlier without affecting production CSS payload size.

PR #1483, opened 2021-10-15
PLANET-6005 Sync svg icons with Design System

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-6005

---

Mostly replacements and some css changes to reflect the different icon set.

- Simplified icons css code that use `mask-image`, since all supported browsers do have support for that. Some still need the `-webkit` prefix, but babel is taking care of that.
- Moved all icons in the same place, for consistency and de-duplication.

### Testing

Some places where this new icons can be seen:
- On posts the comment icon right below the title.
- Search icon on navbar and on search page button.
- Clear/Delete icon on mobile navbar.


comzeradd:
Good catch! I copied the icons we use directly to the images folder, like they used to be.

PR #1499, opened 2021-10-26
PLANET-6257: Trigger Clouflare purge on feature toggle

lithrel:
Necessary for https://jira.greenpeace.org/browse/PLANET-6257

Since the new designs hidden behind toggle features change some markup of the frontend,
we have to purge the relevant Cloudflare cache when we switch those features.

## Changes 

- Toggle is now only for Country Selector, as it can be released by itself
- Extract Cloudflare purge execution from command
- Run purge on Country Selector feature toggle

## Test

Purge command should still work as expected
```console
> wp p4-cf-purge --urls='https://planet4.test/'
About to purge 1 urls.
Chunk 0: ok

> wp p4-cf-purge --all
About to purge 1402 urls.
Chunk 0: ok
...
Chunk 46: ok
```

### DIFF
--- a/src/Cloudflare.php
+++ b/src/Cloudflare.php
@@ -1,0 +1,100 @@
+<?php
+
+namespace P4\MasterTheme;
+
+use CF\Integration\DefaultConfig;
+use CF\Integration\DefaultIntegration;
+use CF\Integration\DefaultLogger;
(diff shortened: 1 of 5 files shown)


Inwerpsel (approved):
:beers:

PR #1500, opened 2021-10-26
PLANET-6008 Button and link focus outline

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6008

---

This PR also includes a fix for the regular link focus state, making it also use `outline` instead of `border` to avoid changing the layout dimensions.

The PR adds the same outline as links to buttons, and removes all the other `:focus` styles, which were always the same as the hover styles.

Still investigating:
- (Probably a separate ticket/PR) We don't have a ticket for adding the focus state around images that are a link, however it could be an improvement to include those. If you check an articles block and focus through its links, it's odd to have no indication of where your tab landed for the image only. Check video below.
- ~~The current navigation background makes the focus color almost invisible around the donate button. For the new design it's only an issue for the dark variant, and maybe it's less there.~~ I simplified the code to use the same focus color on all elements in the navigation bar.

I removed the previous videos from this PR description as they didn't have a offset, might add new ones.

### DIFF
--- a/assets/src/scss/base/_typography.scss
+++ b/assets/src/scss/base/_typography.scss
@@ -117,3 +117,7 @@ a --link-- {
     color: $link-color-visited;
   }
 }
+
+a:focus:not(:focus-visible) {
+  outline: 0;
+}
--- a/assets/src/scss/base/_typography.scss
+++ b/assets/src/scss/base/_typography.scss
@@ -102,7 +102,7 @@ a --link-- {
   text-decoration: none;
 
   &:focus {
-    border: 2px solid rgba(0, 106, 255, 0.4);
+    outline: 2px solid rgba(0, 106, 255, 0.4);
     border-radius: 4px;
   }
 
@@ -116,8 +116,4 @@ a --link-- {
   &:visited {
     color: $link-color-visited;
   }
-}
-
-a:focus-visible {
-  outline: 0;
 }
--- a/assets/src/scss/base/_typography.scss
+++ b/assets/src/scss/base/_typography.scss
@@ -110,7 +110,6 @@ a --link-- {
   &:active {
     color: $link-color;
     text-decoration: underline;
-    border: none;
   }
 
   &:visited {
--- a/assets/src/scss/base/_typography.scss
+++ b/assets/src/scss/base/_typography.scss
@@ -103,6 +103,7 @@ a --link-- {
 
   &:focus {
     outline: 2px solid rgba(0, 106, 255, 0.4);
+    outline-offset: 2px;
     border-radius: 4px;
   }
 
--- a/assets/src/scss/layout/_navbar.scss
+++ b/assets/src/scss/layout/_navbar.scss
@@ -78,10 +78,20 @@ $navbar-default-height: 60px;
   a, .btn, input {
     &:focus {
       --top-navigation--ui-controls-- {
-        outline: 2px solid rgba(255, 255, 255, 0.4);
+        outline-color: rgba(255, 255, 255, 0.4);
       }
-      // Need to undo the bootstrap outline-offset to get a consistent result.
-      outline-offset: 2px;
+    }
+  }
+
+  // Override bootstrap's default form styles because it's not accessible in the navbar.
+  // Instead use the same outline style as buttons, with offset.
+  #search_input:focus {
+    outline-style: solid;
+    outline-width: 2px;
+    outline-offset: 2px;
+
+    &:not(:focus-visible) {
+      outline: 0;
     }
   }
 }
--- a/assets/src/scss/layout/_navbar.scss
+++ b/assets/src/scss/layout/_navbar.scss
@@ -326,11 +326,6 @@ $navbar-default-height: 60px;
       @include medium-and-up {
         height: $menu-height-large;
       }
-
-      &:focus,
-      &:hover {
-        outline: 1px dotted transparentize($white, 0.5);
-      }
     }
 
     &::before {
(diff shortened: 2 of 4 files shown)


assets/src/scss/base/_typography.scss
```diff
@@ -102,7 +102,7 @@ a --link-- {
   text-decoration: none;
 
   &:focus {
-    border: 2px solid rgba(0, 106, 255, 0.4);
+    outline: 2px solid rgba(0, 106, 255, 0.4);
```
Inwerpsel, line 105: `:focus-visible` has somewhat nicer behavior in our case, because it avoids the hover style being added as you mouse-click or tap the button to navigate. In that case the outline flashes for a moment and looks weird.

However it doesn't have full browser support yet, specifically Safari. https://caniuse.com/css-focus-visible

So maybe something we can get back to later.

assets/src/scss/layout/_navbar.scss
```diff
@@ -75,6 +74,15 @@ $navbar-default-height: 60px;
       top: 32px;
     }
   }
+
+  a, .btn, input {
+    &:focus {
+      --top-navigation--ui-controls-- {
```
Inwerpsel, line 80: I made this into a single declaration, so that they can all be changed in one go if the top navigation color changes.

assets/src/scss/base/_typography.scss
Inwerpsel: Oh btw, while the normal links are not in scope of this ticket, I needed to fix this because I had to remove `a:focus-visible { outline: 0 }` below to get buttons working correctly.

assets/src/scss/layout/_navbar.scss
```diff
@@ -321,7 +329,7 @@ $navbar-default-height: 60px;
 
       &:focus,
       &:hover {
-        border: 1px dotted transparentize($white, 0.5);
+        outline: 1px dotted transparentize($white, 0.5);
```
mleray, line 332: Is this `outline` supposed to be applied on hover too? 🤔

Inwerpsel: Good point. I just replaced it with `outline`, but after checking this style does not result in any difference. Both with `outline` and with `border`. It's also not really needed because it's for mobile.

I'll remove this rule :broom:

assets/src/scss/layout/_navbar.scss
```diff
@@ -75,6 +74,16 @@ $navbar-default-height: 60px;
...
+
+  a, .btn, input {
+    &:focus {
+      --top-navigation--ui-controls-- {
+        outline: 2px solid rgba(255, 255, 255, 0.4);
+      }
+      // Need to undo the bootstrap outline-offset to get a consistent result.
+      outline-offset: 0;
+    }
+  }
```
Inwerpsel, line 86: I made this into a single declaration, so that they can all be changed in one go if the top navigation color changes.

mleray (changes requested):
I've tested this on local, and found a couple of places where the outline isn't properly applied on buttons, e.g. the "Load more" button from the Articles block (maybe all secondary buttons? 🤔) or the "Search" button on the search page... There might be more 🤷‍♀️

Inwerpsel:
> I've tested this on local, and found a couple of places where the outline isn't properly applied on buttons, e.g. the "Load more" button from the Articles block (maybe all secondary buttons? thinking) or the "Search" button on the search page... There might be more woman_shrugging

Can you define "isn't properly applied"?

mleray:
> > I've tested this on local, and found a couple of places where the outline isn't properly applied on buttons, e.g. the "Load more" button from the Articles block (maybe all secondary buttons? thinking) or the "Search" button on the search page... There might be more woman_shrugging
> 
> Can you define "isn't properly applied"?

"doesn't appear at all" 😬

Inwerpsel:
> > > I've tested this on local, and found a couple of places where the outline isn't properly applied on buttons, e.g. the "Load more" button from the Articles block (maybe all secondary buttons? thinking) or the "Search" button on the search page... There might be more woman_shrugging
> > 
> > 
> > Can you define "isn't properly applied"?
> 
> "doesn't appear at all" grimacing

Interestingly, they are in fact more properly applied :smile: They don't show up when you use chrome dev tools to set the focus state, because Bootstrap adds a rule to remove the outline if the element is not also in the `:focus-visible` state.

The result of that Bootstrap rule is the focus ring doesn't show when clicking the element. The user already found the element, since he clicked it, so the focus doesn't help UX and looks off.

It's only on `button` elements, hence the examples you give. I'll check if I can get the same behavior on `a` buttons too.

assets/src/scss/base/_typography.scss
comzeradd: Does it make sense to use a `@supports` clause so at least we have the nice behavior on all other browsers?

Inwerpsel: I found another fix in the meanwhile that plays well with non-supporting browsers, using `a:focus:not(:focus-visible)` to remove the outline again.

If the browser doesn't support `:focus-visible`, unlike what you might expect the `:not()` selector to do, the selector won't match. So those would still keep the focus styles.

Still looking for a confirmation of this browser behavior in the specs but it seems to work.

Inwerpsel: It's similar to what's suggested in the example here: https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#selectively_showing_the_focus_indicator

mleray (approved):
Looks good to me 👍

PR #1504, opened 2021-10-28
PLANET-5799 Align search results title color with design system

mleray:
### Description

See https://jira.greenpeace.org/browse/PLANET-5799

When working on this ticket, it seems that some occurrences of the search results component were not properly updated, for instance the press releases or stories pages.

### Testing

Make sure the title links behave as expected in the fixed version (in the broken one, they become blue on hover):
- [Broken](https://www-dev.greenpeace.org/test-uranus/story/)
- [Fixed](https://www-dev.greenpeace.org/test-jupiter/story/)

### DIFF
--- a/assets/src/scss/pages/search/_search-results.scss
+++ b/assets/src/scss/pages/search/_search-results.scss
@@ -181,7 +181,9 @@
     font-size: 1.625rem;
   }
 
-  &:visited {
+  &:visited,
+  &:hover,
+  &:active {
     color: $grey-80;
   }
 }


assets/src/scss/pages/search/_search-results.scss
```diff
@@ -181,7 +181,9 @@
     font-size: 1.625rem;
   }
 
-  &:visited {
+  &:visited,
```
Inwerpsel, line 184: Out of all places where we would remove the visited state, the search results seem to be the one where such a state has the most value. For example Google search results. I don't see why we would keep an visited color in regular page links but remove it here.

mleray: That's a conversation to have with the design team 🤷‍♀️ But I think this is for consistency, since other links that look this way on the site (such as Columns block headers and Articles block titles) have this behaviour.

comzeradd (approved):
I missed that PR 🤦🏻‍♂️ Looks good 👍🏻

assets/src/scss/pages/search/_search-results.scss
comzeradd: That's my assumption too. But @Inwerpsel feel free to leave a comment to the ticket for the Design team to see it. We can always re-visit later or revert that.

Inwerpsel: Yeah, didn't specify but it was indeed not a remark for these changes, just something that struck me.

Inwerpsel (approved):

PR #1509, opened 2021-11-09
PLANET-6256: New navigation bar design

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-6256

New navbar design, as specified in the [design system](https://p4-designsystem.greenpeace.org/05f6e9516/v/0/p/106cdb-navigation-bar).

## Feature
The new design is activated by a feature toggle.

_Planet 4 > Features_
Screenshot from 2021-11-09 17-32-32

## Styles

It has 2 styles integrated, Dark and Light, selectable in the admin panel.

_Planet 4 > Navigation_
Screenshot from 2021-11-29 10-14-23

Screenshot from 2021-11-26 17-20-12

Screenshot from 2021-11-26 17-20-58


planet-4:

:watch: 2021.11.26 09:05:40

templates/navigation-bar.twig
```diff
@@ -1,4 +1,137 @@
 {% block navigation_bar %}
+{% if new_design_navigation_bar %}
```
Inwerpsel, line 2: Would be a bit nicer if we had the old and new version in a separate file.

That way, once we remove the old design, we don't need to remove the if statement from around all markup of the new version, which would change indention of all lines.

assets/src/scss/base/_mixins.scss
```diff
@@ -40,6 +40,27 @@
   }
 }
 
+// Mobile and small width (less than 768px)
+@mixin small-and-less {
```
Inwerpsel, line 44: Since we use a mobile first approach, our code is structured by first including the styles for the smallest screens, then use `min-width` queries. 

Is there any code here that really needs a `max-width` query and can not be rewritten to also use a mobile first approach? Maybe the CSS targeting the mobile touch menu?

lithrel: I ended up using this because the min-width approach forces to overwrite some properties that are actually useless in other widths, especially for something like the menu which has 2 very different styles, and 2 sub-styles for each. I'm gonna give it a shot at mobile first, see if I can remove those mixins.

lithrel: I removed small-and-less and large-and-less :+1: . For medium-and-less, this allows to have no side effect when you stretch your window from a medium width with menu open to larger width - the alternatives seems a lot more complex to me.

webpack.config.js
```diff
@@ -25,6 +25,9 @@ module.exports = {
     bootstrap: './assets/src/scss/bootstrap-build.scss',
     "country-selector-old": './assets/src/scss/partials/country-selector-old.scss',
     "country-selector-new": './assets/src/scss/partials/country-selector-new.scss',
+    "navigation-bar-old": './assets/src/scss/layout/_navbar-old.scss',
+    "navigation-bar-dark": './assets/src/scss/partials/navigation-bar-dark.scss',
+    "navigation-bar-light": './assets/src/scss/partials/navigation-bar-light.scss',
```
Inwerpsel, line 30: As an alternative approach, [using inline styles](https://github.com/greenpeace/planet4-master-theme/blob/40aac328a327e3565c38232c0a3604192af081b0/templates/css-variables.twig#L1-L7) to set different root properties has a few advantages.

- No need to build an additional CSS file for each style.
- It's generally recommended to load critical CSS inline.
- Easier to make compatible with themes on campaign pages, which already work that way.

But we would need to use a JSON object instead, or a PHP array, for the key value pairs. It's fine for now to go with the separate CSS files :+1:

Inwerpsel (approved):
Most changes look good to me for using in prod. Left a few suggestions.

We should do a follow up focusing just on some CSS tweaks (e.g. breaking up the large `_navbar.scss` file, maybe reducing specificity in some places) before we ask NRO devs to adapt their child themes, so that we don't have to ask them to adapt the code multiple times.

webpack.config.js
lithrel: :ok_hand:  I'll be fine with anything that put those rules in explicitely named files, as long as there's a comment somewhere pointing at where they come from; finding the origin of inline styles can be really painful, especially if the property names are generated programmatically

PR #1511, opened 2021-11-11
Use 5.8.1 for PHP tests

Inwerpsel:
* We should make this more robust. The challenge is the WP version now
needs to be the same in multiple places in multiple repos.


comzeradd:
Closing this in favor of #1512

PR #1517, opened 2021-11-16
Fix space between breadcrumb tags

comzeradd:
This was removed in a previous commit causing tags to be too close to each other (eg. in the Articles block).

According to the [Design System](https://p4-designsystem.greenpeace.org/05f6e9516/v/15009/p/835f95-post-page/i/40590918), these still need some space in between.

### DIFF
--- a/assets/src/scss/layout/_breadcrumbs.scss
+++ b/assets/src/scss/layout/_breadcrumbs.scss
@@ -25,6 +25,10 @@
       }
     }
     display: inline-block;
+
+    &:not(:only-child):not(:last-child):not(.page-type) {
+      margin-inline-end: 8px;
+    }
   }
 
   .tag-item--main {


planet-4:

:watch: 2021.11.16 08:14:27

assets/src/scss/layout/_breadcrumbs.scss
```diff
@@ -25,6 +25,10 @@
       }
     }
     display: inline-block;
+
+    &:not(:only-child):not(:last-child) {
+      margin-inline-end: 7px;
```
mleray, line 30: Shouldn't it be 8px instead of 7px? 🤔

comzeradd: Ah yes, that would make more sense. I used the old value, but better use multiples of 8. Commit amended.

mleray (changes requested):
The separator bullets also have margins, which now makes it too much space for the main tag  🤔 

Screenshot 2021-11-16 at 09 32 32

comzeradd:
Yes, because they share the same class. We used to have an exception for that, so I added this one too, to remove the margin in those cases.

mleray (approved):

PR #1518, opened 2021-11-17
POC Matching custom country selector background using theme config file

Inwerpsel:
Use the configuration file to set another color on the country list depending on the color chosen for the navigation.

I only set this up on the `climate` theme.

Example page: https://www-dev.greenpeace.org/test-sinope/campaign/climate-emergency/

This is mostly a POC to verify the config file can be used in this way.

In fact I found out that the current campaign page themes don't need a variable, because they only change the colors if the minimal navigation is selected. In that case the country selector is not displayed.

We still use a separate meta field for every CSS variable, which is annoying because they all need to be registered for the rest API. If we would address this, then we can add similar dependent fields by only adding them to the config file.

### DIFF
--- a/src/PostCampaign.php
+++ b/src/PostCampaign.php
@@ -29,6 +29,7 @@ class PostCampaign {
 		'campaign_body_font',
 		'campaign_footer_theme',
 		'footer_links_color',
+		'country-selector--countries-list--background',
 	];
 
 	public const LEGACY_THEMES = [
--- a/templates/footer.twig
+++ b/templates/footer.twig
@@ -1,5 +1,5 @@
 <footer id="footer" class="site-footer site-footer--{{ nav_type }}">
-	{% if new_design_country_selector and nav_type != 'minimal' %}
+	{% if new_design_country_selector %}
 	<div id="country-selector" class="footer-country-selector">
 		{% include 'footer_country_selector.twig' %}
 	</div>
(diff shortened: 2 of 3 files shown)


templates/footer.twig
```diff
@@ -1,5 +1,5 @@
 <footer id="footer" class="site-footer site-footer--{{ nav_type }}">
-	{% if new_design_country_selector and nav_type != 'minimal' %}
+	{% if new_design_country_selector %}
```
Inwerpsel, line 2: When I was almost done I actually realized that we don't allow to choose any custom colors unless the navigation type is minimal. In which case the country selector is not displayed.

So for campaign themes we don't need to add this variable yet. But it's working already anyway so I just removed this check for now to demonstrate.

src/PostCampaign.php
```diff
@@ -29,6 +29,7 @@ class PostCampaign {
 		'campaign_body_font',
 		'campaign_footer_theme',
 		'footer_links_color',
+		'country-selector--countries-list--background',
```
Inwerpsel, line 32: This is a consequence of the current way it stores the CSS variables set through the sidebar: each one gets an individual meta field.

We need to add this here because all items of this `const` are registered for the rest API, otherwise they won't be stored, and silently ignored.

We should rewrite this to use a single field that holds a JSON object with all variables, so that we can easily add more.

planet-4:

:watch: 2021.11.17 11:03:48

PR #1521, opened 2021-11-23
Move front end assets logic out of MasterSite, create separate partial CSS files depending on country selector feature

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6256

This branch combines a few changes:
* Add an extra option so we can turn off the auto purge on test instances. It likely causes us to hit a purge rate limit which also can affect production's ability to purge.
* Move the logic relating to loading front end assets to a separate function (in a class because autoloading). That way the code is more manageable than having everything inside of `MasterSite`.
* Create a separate entry point for both variants of the country selector, and load only the relevant one.

### Split CSS
If the option is turned off, the front end should look exactly the same as before, only difference is the country selector related CSS is in a separate file.
If the option is turned on, because the required other changes are not there, it will still output the old version and lack styles. This is not a problem because on production no site will turn this on before we have finished the new version.
### Toggle to control purging behavior
You should see a fast response time of saving the features page if this is turned off, and slow if on.


planet-4:

:watch: 2021.11.23 10:16:14

src/PublicAssets.php
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Wrapper class for the enqueue function because we can't autoload functions.
+ */
+final class PublicAssets {
+	/**
+	 * Load styling and behaviour on website pages.
+	 */
+	public static function enqueue(): void {
+		$theme_dir = get_template_directory_uri();
```
Inwerpsel, line 13: For now I preserved this to not change the logic too much while moving the code, but it can probably also use `Loader:enqueue_versioned_style`, which gets this internally.

assets/src/scss/partials/country-selector-new.scss
```diff
@@ -0,0 +1,6 @@
+@import "../base/mixins";
+@import "../base/variables";
+@import "../base/colors";
+@import "../base/fonts";
+
+//@import "../layout/country-selector";
```
Inwerpsel, line 6: I for now commented this import, you can choose whether you use it, or move the new code directly in this file.

assets/src/scss/partials/country-selector-old.scss
```diff
@@ -0,0 +1,175 @@
+@import "../base/mixins";
+@import "../base/variables";
+@import "../base/colors";
+@import "../base/fonts";
+
+$menu-height-large: 3.75rem;
```
Inwerpsel, line 6: It's not ideal to copy these here from the navbar, but it's the old design so we don't intend to change it and will remove soon.

lithrel (approved):
:+1:

PR #1522, opened 2021-11-23
Move palette classes to separate file

Inwerpsel:
The classes for the palette were being duplicated by every entry point that included it, because it looked as though the file only contained SASS variable definitions.

I moved those to a separate file which is imported only in the front end and editor style entry points.

This needs to be bumped in the plugin as submodule.


lithrel (approved):
:+1:

PR #1524, opened 2021-11-23
Add container class to page-template elements

mleray:
### Description

We have some content width/padding issues in pages on mobile, which are due to our blocks having the `container` class but not the "non-blocks" items (paragraphs for example). Also because of this we had to manually add the page width to the `page-template` element, which doesn't make sense since we use the same ones as Bootstrap anyway so we might as well use that directly 🙂 So a solution would be to remove the `container` class from the blocks (except full-width ones) (done in [this PR](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/737)) and add it to the `page-template` div instead.

### Testing

You can see the new layout in this [test instance](https://www-dev.greenpeace.org/test-uranus), everything should look pretty much the same except with a bit less padding than before on mobile. In all screen sizes, all elements (blocks and non-blocks) of the page content should be aligned.

### DIFF
--- a/assets/src/scss/pages/_page.scss
+++ b/assets/src/scss/pages/_page.scss
@@ -1,28 +1,4 @@
 div.page-template {
-  width: 540px;
-  max-width: 100%;
-  margin-left: auto;
-  margin-right: auto;
-  padding: 0 16px;
-
-  @include medium-and-up {
-    width: 720px;
-    padding: 0;
-  }
-
-  @include large-and-up {
-    width: 960px;
-  }
-
-  @include x-large-and-up {
-    font-size: 1.125rem;
-    width: 1140px;
-  }
-
-  @include xx-large-and-up {
-    width: 1320px;
-  }
-
   > * {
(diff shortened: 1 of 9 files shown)


planet-4:

:new_moon: [uranus](https://www-dev.greenpeace.org/test-uranus) | [admin](https://www-dev.greenpeace.org/test-uranus/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-uranus/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-uranus) | [composer-local.json](https://github.com/greenpeace/planet4-test-uranus/blob/main/composer-local.json)

:watch: 2021.11.24 14:31:33

Inwerpsel (approved):
Code looks good and didn't spot any issues on the test instance.

This might require some small adjustments in child themes. Perhaps makes sense to release to all dev sites first.

PR #1527, opened 2021-11-26
Faster update command for data sync to test instances

lithrel:
The [command used to fix post authors](https://github.com/greenpeace/planet4-builder/blob/main/src/bin/sync_to_test_site.sh#L95-L97) during data sync takes a ridiculous amount of time. (approx 1.7sec per post, more than 20 min for 800 posts)
It makes some data sync fail: https://app.circleci.com/pipelines/github/greenpeace/planet4-test-pandora/209/workflows/6d9e5226-aada-47f3-aa55-45b96820a1cb/jobs/1108
Some others take 40 minutes: https://app.circleci.com/pipelines/github/greenpeace/planet4-test-pandora/152/workflows/4096c607-bc6c-48b7-bda8-aa84c5ee3ed9/jobs/773

Using a simple batch request and bypassing other update ripples should only take a few seconds in total.

### DIFF
--- a/src/Commands.php
+++ b/src/Commands.php
@@ -8,6 +8,7 @@ namespace P4\MasterTheme;
 use P4\MasterTheme\Commands\CloudflarePurge;
 use P4\MasterTheme\Commands\RunActivator;
 use P4\MasterTheme\Commands\SaveCloudflareKey;
+use P4\MasterTheme\Commands\FixOrphans;
 
 /**
  * Class with a static function just because PHP can't autoload functions.
@@ -23,5 +24,6 @@ class Commands {
 		RunActivator::register();
 		SaveCloudflareKey::register();
 		CloudflarePurge::register();
+		FixOrphans::register();
 	}
 }
--- a/src/Commands/FixOrphans.php
+++ b/src/Commands/FixOrphans.php
@@ -1,0 +1,58 @@
+<?php
+
+namespace P4\MasterTheme\Commands;
+
+use WP_CLI;
+
+/**
+ * Class PostUpdate
+ *
+ * @example wp p4-post fix-orphans
+ */
+class FixOrphans extends Command {
+	/**
+	 * The name to access the command.
+	 *
+	 * @return string The command name.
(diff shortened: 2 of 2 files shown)


src/Commands/PostUpdate.php
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * @example wp p4-post update 1 2 3 --post_author=1
+ */
+class PostUpdate extends Command {
+	/**
+	 * Fields allowed for update
+	 *
+	 * @var string[]
+	 */
+	private static $allowed_fields = [ 'post_author' ];
```
Inwerpsel, line 18: You can use a `const` here since it's not intended to change dynamically.

Inwerpsel (commented):
Looks promising!

Can't we take this even a bit further and avoid [these lines](https://github.com/greenpeace/planet4-builder/blob/main/src/bin/sync_to_test_site.sh#L92-L94) as well? It seems we can make the command without ID arguments, so that it internally gets the right IDs. Or maybe even a single update query using `where post_author = %post_author%`? :thinking:

planet-4:

:watch: 2021.11.29 10:07:34

lithrel:
> Can't we take this even a bit further and avoid [these lines](https://github.com/greenpeace/planet4-builder/blob/main/src/bin/sync_to_test_site.sh#L92-L94) as well? It seems we can make the command without ID arguments, so that it internally gets the right IDs. Or maybe even a single update query using `where post_author = %post_author%`? thinking

Yep, I tried to make a generic update post command, but I think a simpler command would be to just fix orphan posts, posts without an author or with an author not present in the list of authors in the instance. This command probably wouldn't need any arguments and do the job of those 6 lines.

lithrel:
@Inwerpsel  New proposition :grin:  `wp p4-post fix-orphans` finds the first admin user, and assigns to this user the posts with a 0 author or an author not present in the users table.

Inwerpsel (approved):
:beers:

PR #1531, opened 2021-11-30
Simplify navbar SCSS

Inwerpsel:
* Give navigation bar donate button a separate variable, since it now
has a different style than donate buttons added using a class.
* Remove heights as variables for the donate button, since we need to
adapt other CSS before the height can be reliably changed.
* Remove redundant rules, which are possibly a result of the code
changing from max-width to min-width queries.
* Remove rules from generic `.top-navigation button` selector. These
were being overridden by most buttons, and caused specificity issues,
forcing overrides to use a more specific selector where it's really not
needed.
* As a result, all element selectors that were only added to match the
specificity could be removed.
* Use transparent to get a transparent background, instead of addding
the CSS variable of the parent background color.


planet-4:

:watch: 2021.11.30 13:14:39

assets/src/scss/layout/_navbar.scss
```diff
@@ -129,11 +129,6 @@ $navbar-default-height: 60px;
         display: flex;
       }
 
-      .nav-search-toggle-container {
-        cursor: unset;
-        pointer-events: none;
```
lithrel, line 141: On medium width, menu opened, clicking on the search toggle doesn't close the menu/remove the overlay if you remove the pointer-events rule. (In a previous version it was also opening the search bar while keeping the overlay)

Inwerpsel: Right, with the current structure that's indeed needed. I guess ideally we would add the overlay over everything except the opened menu, but I'll change it back for now.

lithrel: The overlay is over everything outside of nav-main, might be easier to have a different search toggle outside of the nav-main for medium and smaller widths

Inwerpsel:
I'll close this as the changes were included in https://github.com/greenpeace/planet4-master-theme/pull/1509

PR #1532, opened 2021-11-30
Preserve nav file

Inwerpsel:


templates/html-header.twig
```diff
@@ -18,6 +18,7 @@
 	<meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no">
 	<meta name="robots" content="max-snippet:-1, max-image-preview:large, max-video-preview:-1">
 	<link rel="shortcut icon" type="image/ico" href="{{ theme.uri }}/favicon.ico"/>
+	<script src="https://unpkg.com/source-map@0.7.3/dist/source-map.js"></script>
```
Inwerpsel, line 21: Ah woops, committed this by accident from something I was testing locally.

Inwerpsel:
I'll close this again since the issue of child themes overriding templates by name prevent us from using the old template name, which would take some effort to get around, probably not worth it.

PR #1536, opened 2021-12-03
[PoC] PLANET-5934 Define 8px spacers

comzeradd:
This is just a Proof of Concept for [PLANET-5934](https://jira.greenpeace.org/browse/PLANET-5934).

This approach removes all `$nxx` units (which are based on 5px intervals anyway) and replaces `$space-breakpoint` variables with a new naming scheme (check `_variables` file in this diff).

The new naming scheme uses a default `8px` spacer as the base unit and then we create variations of that. 

To make the variables easier to parse when reading the code, I used this pattern:
- `$sp-n`: means `n * 8px`
- `$sp-nx`: means `n.5 * 8px`

So for instance:
- `$sp-4` is `32px`
- `$sp-5x` is `44px`


assets/src/scss/base/_variables.scss
```diff
@@ -10,28 +10,21 @@ $large-width:             992px;
...
-
 // Spacing
-$space-xxs: 4px;
-$space-xs:  8px;
-$space-sm:  12px;
-$space-md:  16px;
-$space-lg:  32px;
-$space-xl:  48px;
-$space-xxl: 64px;
+$spacer: 8px;
```
Inwerpsel, line 14: We can also use numbers and write a CI job that validates them to be multiples of 8.

planet-4:

:watch: 2021.12.10 14:52:57

assets/src/scss/base/_variables.scss
comzeradd: Yeah we could some time on setting up a CI validation, although this feels a bit late in the design process.

assets/src/scss/base/_mixins.scss
```diff
@@ -55,10 +55,10 @@
 }
 
 // Margins and padding. Use to consistently apply gutters
-// use like this: @include padding($left: $n15, $right: $n15); wich will be compiled as padding-left: 15px; padding-right: 15px;
+// use like this: @include padding($left: $sp-2, $right: $sp-2); wich will be compiled as padding-left: 15px; padding-right: 15px;
 
 // To use as negative numbers interpolate these as follows...
-// @include margin($top: -#{$n10}, $right: $n20, $bottom: $n10, $left: 0); which will be compiled as margin: -10, 20px 10px 0px;
+// @include margin($top: -#{$sp-1}, $right: $sp-2x, $bottom: $sp-1, $left: 0); which will be compiled as margin: -10, 20px 10px 0px;
```
mleray, line 61: The comment here needs to be updated since it still uses multiples of 10 🙃

mleray: (same on the line above btw, which uses multiples of 5)

comzeradd: Good catch :grin: Comment updated.

assets/src/scss/base/_variables.scss
```diff
@@ -10,28 +10,21 @@ $large-width:             992px;
...
+$sp-x: calc(0.5 * #{$spacer});
```
GP-Dan-Tovbein, line 15: Why don't we move these calculation to a function?
```
@function pow($base) {
  @return calc($base) * #{$spacer});
}
```

Does it make sense?

GP-Dan-Tovbein (commented):
Code looks good, but just a small suggestion.

assets/src/scss/base/_variables.scss
comzeradd: That's a great suggestion. Makes the code significantly more readable. I adjusted the code. I put the function in `_functions` just for consistency.

GP-Dan-Tovbein (approved):
Code looks good to me!

Inwerpsel:
> Those few that are not already based on 8px were by most 2px off from their nearest variable, which seems acceptable.

I don't know how this seems acceptable to do in a bulk fix in so many places at the same time. We have a setup where child themes are overriding our CSS. You can definitely have significant regressions even with a 2px (or 1px) change, and it's very hard to predict where that will happen.

We should at least thorougly check this on all NRO dev sites before deploying.

PR #1537, opened 2021-12-06
Ensure a separate size is always created

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6372

* ~~It seems there's a bug in case you upload an image that is exactly
2560px wide. In that case for WPML sites the Carousel Header block gets
a non CDN URL, i.e. the PHP server's wp-uploads folder. This folder is
not persistent.~~

### Possible related issue on WP stateless plugin: https://github.com/udx/wp-stateless/issues/596
This probably explains why in the media detail page for the affected images you get a 404 link to the original. While this is not a big problem by itself (unless we would need the original to regenerate sizes in the future), ~~it's probably related to the issue of this size using a non CDN URL~~ (turns out this is not the case).

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -183,3 +183,5 @@ add_filter(
 	10,
 	2
 );
+
+simple_value_filter( 'wp_image_resize_identical_dimensions', true );


functions.php
```diff
@@ -183,3 +183,5 @@ function ( bool $allow_css, $css_test_string ) {
 	10,
 	2
 );
+
+simple_value_filter( 'wp_image_resize_identical_dimensions', true );
```
Inwerpsel, line 187: Relevant core code
https://github.com/WordPress/WordPress/blob/master/wp-includes/media.php#L635-L653

planet-4:

:watch: 2021.12.07 12:53:56

lithrel (approved):
:+1:

lithrel (approved):
:+1:

Inwerpsel:
We found that the issue of disappearing images is unrelated to this, and was instead caused by a WPML plugin that was recently added.

We can probably close this PR, likely it won't solve the issue of wp-stateless not preserving the original.

PR #1541, opened 2021-12-10
Support Analytics values using Google Sheet

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6451
* Add Google Docs API client lib.
* Add option to choose between Smartsheet and Google for the
integration.
* Add settings field for the local Google Sheet.
* hardcode the global doc ID for now. Maybe later env var but it's on the Jira ticket anyway so it can be public.
* I'll set up the authentication in CircleCI context.

Before we initiate the release we need to verify the key is functional. I had some issues while testing because it requires a specific format of base64.

### DIFF
--- a/src/AnalyticsValues.php
+++ b/src/AnalyticsValues.php
@@ -34,25 +34,28 @@ final class AnalyticsValues {
 	 *
 	 * @param Smartsheet      $global_smartsheet Smartsheet containing global projects data.
 	 * @param Smartsheet|null $local_smartsheet Smartsheet containing local projects data.
+	 * @param bool            $uses_google Google sheets have different boolean columns.
 	 *
 	 * @return static
 	 */
 	public static function from_smartsheets(
 		Smartsheet $global_smartsheet,
-		Smartsheet $local_smartsheet = null
+		Smartsheet $local_smartsheet = null,
+		bool $uses_google = false
 	): self {
 		$project_name_column = $global_smartsheet->get_column_index( 'Global Project Standard' );
 		$approved_column     = $global_smartsheet->get_column_index( 'Standard Approved' );
 		$tracking_id_column  = $global_smartsheet->get_column_index( 'Tracking ID' );
 
-		$global_projects = $global_smartsheet->filter_by_column( $approved_column, true )->sort_on_column(
-			$project_name_column
-		)->export_columns(
-			[
-				$project_name_column => 'global_project_name',
-				$tracking_id_column  => 'tracking_id',
-			]
-		);
+		$column_names = [
+			$project_name_column => 'global_project_name',
+			$tracking_id_column  => 'tracking_id',
+		];
+
+		$global_projects = $global_smartsheet
+			->filter_by_column( $approved_column, $uses_google ? 'yes' : true )
+			->sort_on_column( $project_name_column )
+			->export_columns( $column_names );
 
 		$local_projects = null;
 		if ( null !== $local_smartsheet ) {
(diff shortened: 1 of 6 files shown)


src/AnalyticsValues.php
```diff
@@ -45,13 +45,10 @@ public static function from_smartsheets(
...
 
-		$global_projects = $global_smartsheet->filter_by_column( $approved_column, true )->sort_on_column(
-			$project_name_column
-		)->export_columns(
-			[
-				$project_name_column => 'global_project_name',
-				$tracking_id_column  => 'tracking_id',
-			]
+		$global_projects = $global_smartsheet
+			->filter_by_column( $approved_column, true, true )
```
Inwerpsel, line 49: The API response of Smartsheet contained a boolean, while for Docs it's just the cell content.

For now the implementation requires to use a value that PHP type casts to either true or false.

I might change this to do a different filter depending on the source sheet.

planet-4:

:watch: 2021.12.10 15:48:58

src/Smartsheet.php
```diff
@@ -37,27 +37,54 @@ private function __construct( array $columns, array $rows ) {
...
+	public function filter_by_column( int $column_index, $expected, bool $convert_boolean = false ): self {
 		$rows = array_filter(
 			$this->rows,
-			function ( $row ) use ( $column_index, $column_value ) {
-				return $row['cells'][ $column_index ]['value'] === $column_value;
+			function ( $row ) use ( $convert_boolean, $column_index, $expected ) {
+				$value = $row['cells'][ $column_index ]['value'];
+
+				if ( $convert_boolean ) {
+					return ! $expected === ! $value;
```
Inwerpsel, line 84: Apparently this [trips out the Yoda conditions check](https://app.circleci.com/pipelines/github/greenpeace/planet4-master-theme/4384/workflows/47ff3711-ce46-4539-9d17-477e08fdad39/jobs/25602), which fails no matter which one is first.

src/Features.php
```diff
@@ -32,6 +32,7 @@ class Features {
 
 	public const PURGE_ON_FEATURE_CHANGES = 'purge_on_feature_changes';
 
+	public const GOOGLE_SHEET_REPLACES_SMARTSHEET = 'google_sheet_replaces_smartsheet';
```
comzeradd, line 35: Since this has to be released early next week, why add a feature flag that we would have to manually switch for all the websites?

Inwerpsel: How else would we switch and still provide continuity? :thinking: 

We need to release the settings field before people can enter their local sheets Google URL. This feature toggle allows to do that in a single release.

Also when we release it on Monday we still have a few days overlap where Smartsheet still works. So if an NRO would find a problem with their local sheet, they can switch it back to Smartsheet, fix it, and re-enable.

Inwerpsel: Keep in mind that we also need to enter the local sheets IDs in each site anyway.

comzeradd: Ah right, I was thinking mostly about the Global Sheet. But for local projects, NROs should gradually update and switch :+1:

composer.json
```diff
@@ -36,7 +36,8 @@
     }
   },
   "require": {
-    "alek13/slack": "^2.0"
+    "alek13/slack": "^2.0",
+    "google/apiclient": "^2.0"
```
Inwerpsel, line 40: Still sorting an issue with auto loading. That breaks the test instance.

Inwerpsel: This turned out an issue with test instance deployments. For now we worked around it by merging the package first.

src/Smartsheet.php
```diff
@@ -138,10 +161,12 @@ function ( $carry, $cell ) use ( $columns, &$cell_index ) {
 	 * @return int|null The id of the column, or null if no column was found.
 	 */
 	public function get_column_index( string $column_title ): ?int {
+		$i = 0;
```
lithrel, line 164: I think `return array_search( $column_title, array_column( $this->columns, 'title' ) ) ?? null;` also does the job

lithrel (approved):
:+1:

src/Smartsheet.php
Inwerpsel: Changed it, but `??` only works for null, so I had to use `?:` and allow its syntax here.

lithrel: Ah my bad :facepalm:  but then `a ?: b` returns b if a is 0, which can clearly happen with array indexes
Probably better in 2 steps 
```php
$index = array_search( $column_title, array_column( $this->columns, 'title' ) );
return $index !== false ? $index : null;
```

PR #1525, opened 2021-11-24
Load master-theme partial styles before any child-theme

lithrel:
Prioritize master-theme styles enqueue, so that the cascading order is stable.

Action `wp_enqueue_scripts` is used without any priority, and because the child-theme `functions.php` file is loaded before the master-theme one, its styles are enqueued before any style it does not declare as a dependency.
Adding a priority 0 to master-theme enqueue fixes the issue.

_Before fix, the child theme css loads before the country-selector one, which can lead to overwriting child theme rules_
Screenshot from 2021-11-24 08-38-58

_After fix, the country-selector style loads with master-theme main style file_
Screenshot from 2021-11-24 08-38-10

edit: Adding a priority < 10 makes some unit tests fail (`Trying to get property 'queue' of non-object` on global `$wp_styles` in function `wp_maybe_inline_styles()`), so we keep the default priority for now and will ask child themes to use a higher one.

PR #1543, opened 2021-12-13
Fix main & partial styles enqueue order

lithrel:
`master-theme` styles are loaded after `child-theme` styles in some situations.
This leads to style overrides in an unexpected order.

Follows: https://github.com/greenpeace/planet4-master-theme/pull/1525

## Fix
Mostly changing a default priority, and moving some enqueue code to skip it if the right file is not there.

- Ensure master-theme styles are enqueued before any child-theme style, by adding priority 0.
  Most child-themes are enqueuing their styles with priority 99, but some have no priority or priority 10, which interferes with our default priority.
- Fix unit tests by dropping style enqueue if the main style is not built.
  Unit tests are currently testing some frontend features and trigger styles inclusion, but they run without assets built. This way of testing triggers errors down the Timber chain that make those tests fail.

## Test
_From #1525_

_Before fix, the child theme css loads before the country-selector one, which can lead to overwriting child theme rules_
Screenshot from 2021-11-24 08-38-58

_After fix, the country-selector style loads with master-theme main style file_
Screenshot from 2021-11-24 08-38-10

### DIFF
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -122,7 +122,7 @@ class MasterSite extends TimberSite {
 		// Load the editor scripts only enqueuing editor scripts while in context of the editor.
 		add_action( 'enqueue_block_editor_assets', [ $this, 'enqueue_editor_assets' ] );
 		// Load main theme assets before any child theme.
-		add_action( 'wp_enqueue_scripts', [ PublicAssets::class, 'enqueue' ] );
+		add_action( 'wp_enqueue_scripts', [ PublicAssets::class, 'enqueue' ], 0 );
 		add_filter( 'safe_style_css', [ $this, 'set_custom_allowed_css_properties' ] );
 		add_filter( 'wp_kses_allowed_html', [ $this, 'set_custom_allowed_attributes_filter' ], 10, 2 );
 		add_action( 'add_meta_boxes', [ $this, 'add_meta_box_search' ] );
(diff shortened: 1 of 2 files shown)


src/MasterSite.php
```diff
@@ -122,7 +122,7 @@ protected function hooks() {
 		// Load the editor scripts only enqueuing editor scripts while in context of the editor.
 		add_action( 'enqueue_block_editor_assets', [ $this, 'enqueue_editor_assets' ] );
 		// Load main theme assets before any child theme.
-		add_action( 'wp_enqueue_scripts', [ PublicAssets::class, 'enqueue' ] );
+		add_action( 'wp_enqueue_scripts', [ PublicAssets::class, 'enqueue' ], 0 );
```
Inwerpsel, line 125: I suppose this means the child themes that use priority 10 or less should be checked as their loading order will change?

planet-4:

:watch: 2021.12.13 13:40:46

Inwerpsel (approved):
We could use the assets from the `create-zip` step, by putting them in `/tmp/workspace`. I think we can even attach the workspace folder directly to `assets/build`. 

Checking the file exists makes sense anyway :+1:

src/MasterSite.php
lithrel: Yep, but we already have problems with those (cf [dev japan](https://www-dev.greenpeace.org/japan/)), that's why I'm working on this

PR #1551, opened 2021-12-16
Remove enhanced donate button on new navbar, fix search toggle

lithrel:
In Planet 4 > Donate button we can set a fixed position for the donate button
This situation doesn't exist in the new navbar, and if active the button appears at the bottom of the page. We're hiding this button on the new style, and will remove the option from the scripts and backend after the transition.

Search field is not hidden when a user clicks anywhere else on the page after opening it.
This comes from a double click generated by the `closeInactiveMenus` function, that runs on the 2 toggles.

### DIFF
--- a/assets/src/js/header.js
+++ b/assets/src/js/header.js
@@ -32,13 +32,9 @@ export const setupHeader = function($) {
       return attr === 'false' ? 'true' : 'false';
     });
 
-    // Set same attributes for all search toggles
+    // Propagate attributes to all search toggles
     if (evt.currentTarget.classList.contains('nav-search-toggle')) {
-      document.querySelectorAll('.nav-search-toggle').forEach(e => {
-        e.setAttribute('aria-expanded', $button.attr('aria-expanded'));
-        e.setAttribute('data-ga-action', $button.attr('aria-expanded') === 'false' ? 'Open search' : 'Close search');
-        e.classList.toggle('open', $button.hasClass('open'));
-      });
+      setSearchToggles($button.attr('aria-expanded') === 'true');
     }
 
     // Lock scroll when navigation menu is open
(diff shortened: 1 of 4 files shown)


planet-4:

:watch: 2021.12.16 11:15:35

assets/src/js/header.js
```diff
@@ -93,6 +98,20 @@ export const setupHeader = function($) {
     $('.nav-menu-toggle').trigger('click');
   });
 
+  /**
+   * Propagate attributes to all search toggles
+   *
+   * @param {bool} expanded Toggle is expanded
+   */
+  function setSearchToggles(expanded) {
```
Inwerpsel, line 106: You can declare this function outside of `setupHeader`. Probably also a few other functions in this file.

assets/src/scss/layout/_navbar.scss
```diff
@@ -258,6 +258,10 @@ a.nav-link --nav-link-- {
   }
 }
 
+.btn-donate-top {
+  display: none;
```
Inwerpsel, line 262: Why do we have to hide it? Can't we instead adjust the code to not add the class in case the feature is turned on? https://github.com/greenpeace/planet4-master-theme/blob/b61e9e0e039a77ad6db77596c2ca6f2a6194bbea/src/MasterSite.php#L553-L557

assets/src/js/header.js
```diff
@@ -93,6 +98,20 @@ export const setupHeader = function($) {
     $('.nav-menu-toggle').trigger('click');
```
Inwerpsel, line 98: Is this maybe related to the double click you mention in the PR description? I think it's best to avoid triggering the click event just because the handler has the logic we want to execute. We can instead put that logic in a function and call it from multiple places.

assets/src/scss/layout/_navbar.scss
lithrel: Sure that works too :+1:

assets/src/js/header.js
lithrel: No the double click comes from the closeInactiveMenus() function clicking on everything and the fact that there's 2 different search toggles controlling the target, so a click on each toggle but then a revert in the effect on the target

assets/src/js/header.js
```diff
@@ -54,10 +50,19 @@ export const setupHeader = function($) {
 
   // Close all menus when clicking somewhere else
   $(document).on('click', function closeInactiveMenus(evt) {
+    let searchToggled = false;
     const clickedElement = evt.target;
     $('button[aria-expanded="true"]').each((i, button) => {
       const $button = $(button);
       const buttonTarget = $($button.data('bs-target')).get( 0 );
+
+      if ($button.get(0).classList.contains('nav-search-toggle')) {
```
Inwerpsel, line 59: Shouldn't we instead tackle the root cause, which is the usage of `.trigger('click')`?

lithrel: My view was more to do that when we rewrite this script without jquery

Inwerpsel: As discussed on Slack, the script needs a significant rewrite, so we can go with this solution for now.

Another way would be to only select visible buttons.
```js
$('button[aria-expanded="true"]:visible')

Inwerpsel (approved):

PR #1558, opened 2021-12-23
PLANET-5968 Ensure translated media gets Stateless meta

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-5968

* WPML creates copies to use in other languages which use the same file.
However they don't copy all meta. As a result the Stateless plugin
doesn't detect it should put the CDN URL.
* Remove action that was only needed to debug this issue.

Tested on the BE site and confirmed that it does create the stateless meta in all cases. I used [this image](https://www-dev.greenpeace.org/belgium/wp-admin/upload.php?item=25838) uploaded to a French post.

Interestingly the bug of the missing media only occurred when you upload an image to a post. Uploading from the media page using "add new" already seemed to create the right WP stateless meta data. Probably the previous filter (which is removed in this PR) only ran in that case.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -184,19 +184,11 @@ add_filter(
 	2
 );
 add_action(
-	'post_updated',
-	function ( $post_id, $post_after, $post_before ) {
-		if (
-			is_plugin_active( 'wp-stateless/wp-stateless-media.php' )
-			&& get_option( 'sm_custom_domain' )
-			&& 'page' === $post_after->post_type
-			&& $post_after->post_content
-			&& strpos( $post_after->post_content, 'wp-content/uploads' )
-			&& ! strpos( $post_before->post_content, 'wp-content/uploads' )
-		) {
-			trigger_error( "Post $post_id updated with a local URL instead of CDN.", E_USER_WARNING ); //phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_trigger_error,WordPress.Security.EscapeOutput.OutputNotEscaped
-		}
+	'wpml_after_update_attachment_texts',
+	function ( $original_attachment_id, $translation ) {
+		$original_sm_cloud = get_post_meta( $original_attachment_id, 'sm_cloud', true );
+		update_post_meta( $translation->element_id, 'sm_cloud', $original_sm_cloud );
 	},
 	1,
-	3
+	2
 );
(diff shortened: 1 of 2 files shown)


functions.php
```diff
@@ -184,18 +184,10 @@ function ( bool $allow_css, $css_test_string ) {
...
-			is_plugin_active( 'wp-stateless/wp-stateless-media.php' )
-			&& get_option( 'sm_custom_domain' )
-			&& 'page' === $post_after->post_type
-			&& $post_after->post_content
-			&& strpos( $post_after->post_content, 'wp-content/uploads' )
-			&& ! strpos( $post_before->post_content, 'wp-content/uploads' )
-		) {
-			trigger_error( "Post $post_id updated with a local URL instead of CDN.", E_USER_WARNING ); //phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_trigger_error,WordPress.Security.EscapeOutput.OutputNotEscaped
-		}
+	'wpml_after_update_attachment_texts',
```
Inwerpsel, line 187: I still need to verify this is the filter we need.

planet-4:

:watch: 2021.12.23 13:44:50

functions.php
Inwerpsel: Looks like it is in fact this filter, it produces the right result and I didn't find any more appropriate alternative.

functions.php
```diff
@@ -184,18 +184,10 @@ function ( bool $allow_css, $css_test_string ) {
...
+	function ( $original_attachment_id, $translation ) {
+		$original_sm_cloud = get_post_meta( $original_attachment_id, 'sm_cloud', true );
+		update_post_meta( $translation->element_id, 'sm_cloud', $original_sm_cloud );
 	},
 	1,
 	3
```
lithrel, line 193: accepted_args should be a 2 now I think

lithrel (approved):
:+1: 
I can see this hook in the code but not in [the doc](https://wpml.org/documentation/support/wpml-coding-api/wpml-hooks-reference/) :shrug:

Inwerpsel:
> I can see this hook in the code but not in [the doc](https://wpml.org/documentation/support/wpml-coding-api/wpml-hooks-reference/)

It's probably a recent addition, and overall WPML's documentation is a bit lacking/lagging. For this filter I even only find [a single Google result](https://www.google.com/search?q=%22wpml_after_update_attachment_texts%22).

PR #1559, opened 2021-12-23
PLANET-5758 Remove jQuery from header.js

mleray:
### Description

See [PLANET-5758](https://jira.greenpeace.org/browse/PLANET-5758)

I've removed quite a big chunk of the file, these parts did not seem to be used anymore as I couldn't find any use of the classes in our code.

### Testing

You can test everything either on local or on the [test instance](https://www-dev.greenpeace.org/test-deimos), making sure that everything behaves as expected. You should check both the old and new navigation bar and menu, since both should still be working!

### DIFF
--- a/assets/src/js/header.js
+++ b/assets/src/js/header.js
@@ -1,165 +1,5 @@
-export const setupHeader = function($) {
-  'use strict';
-
-  $(document).on('click', [
-    '.navbar-dropdown-toggle',
-    '.nav-menu-toggle',
-    '.country-dropdown-toggle',
-    '.country-selector-toggle',
-    '.navbar-search-toggle',
-    '.nav-search-toggle',
-    '.nav-languages-toggle',
-  ].join(), function toggleNavDropdown(evt) {
-    evt.preventDefault();
-    evt.stopPropagation();
-
-    const $button = $(this);
-    const target = $button.data('bs-target');
-    if (!target) {
-      throw new Error('Missing `data-bs-target` attribute: specify the container to be toggled');
-    }
-    const toggleClass = $button.data('bs-toggle');
-    if (!toggleClass) {
-      throw new Error('Missing `data-bs-toggle` attribute: specify the class to toggle');
-    }
-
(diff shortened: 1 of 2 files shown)


assets/src/js/header.js
```diff
@@ -1,177 +1,146 @@
-export const setupHeader = function($) {
-  'use strict';
+const updateGaAction = (element, elementName) => {
+  element.dataset.gaAction = `${element.ariaExpanded === 'false' ? 'Open' : 'Close'} ${elementName}`;
```
lithrel, line 2: I got confused while testing, but this access to aria attributes is [not supported on firefox](https://caniuse.com/?search=ariaExpanded) at the moment :grimacing:  (seems to be [Aria 1.2 recommendations](https://www.w3.org/TR/wai-aria-1.2/#idl-reflection-attribute-values), still recent)
Without it I don't see other ways than the [get/setAttribute](https://w3c.github.io/aria/#idl_example_usage) function that its supposed to replace

lithrel (changes requested):

assets/src/js/header.js
```diff
@@ -1,177 +1,146 @@
...
+  if (element.classList.contains('nav-search-toggle')) {
+    setSearchToggles(element.ariaExpanded === 'true');
+  }
 
-    // Lock scroll when navigation menu is open
-    if (evt.currentTarget.classList.contains('nav-menu-toggle')) {
-      document.body.classList.toggle('no-scroll-nav-open', evt.currentTarget.getAttribute('aria-expanded') === 'true');
+  // We need to focus the search input when showing it
+  const searchInput = document.querySelector('#search_input');
+  if (element.classList.contains('nav-search-toggle') || element.classList.contains('.navbar-search-toggle')) {
```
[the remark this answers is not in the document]
mleray: Good catch, fixed!

assets/src/js/header.js
mleray: Oopsie 😬 Updated!

lithrel (approved):
:+1:

PR #1561, opened 2021-12-24
PLANET-5761 Remove jQuery from search

mleray:
### Description

See [PLANET-5761](https://jira.greenpeace.org/browse/PLANET-5761)

### DIFF
--- a/templates/tease-search.twig
+++ b/templates/tease-search.twig
@@ -1,10 +1,12 @@
 {% set ga_page_type = post.post_type|capitalize %}
+{% set post_link = post.link %}
 {% if ( 'action' == post.content_type ) %}
 	{% set is_action = true %}
 	{% set ga_page_type = 'Take Action' %}
 {% elseif ( 'document' == post.content_type ) %}
 	{% set is_document = true %}
 	{% set ga_page_type = 'Document' %}
+	{% set post_link = fn('wp_get_attachment_url', post.id) %}
 {% elseif ( 'campaign' == post.content_type ) %}
 	{% set is_campaign = true %}
 {% elseif ( 'archive' == post.post_type ) %}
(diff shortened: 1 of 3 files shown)


planet-4:

:watch: 2021.12.24 13:18:03

templates/tease-search.twig
```diff
@@ -25,7 +27,9 @@
 			alt="{{ fn('esc_attr', post.thumbnail_alt|default( post.post_title ) )|raw }}"
 			data-ga-category="Search Results"
 			data-ga-action="Image"
-			data-ga-label="{{ ga_page_type }}" />
+			data-ga-label="{{ ga_page_type }}"
+			onclick="window.location.href='{{ post_link }}'"
```
Inwerpsel, line 31: If we want the image to be a clickable link, we should use an `a` tag around it.

Implementing our own link with JS misses a lot of things about the UX of links:
- focus states are different
- not accessible
- open in a new tab won't work
- probably more

Inwerpsel:
@mleray we currently have a ticket to solve the issue of duplicate results @lithrel is working on. That will likely involve making search use the Ajax calls exclusively, so also for the first page of results.

I suppose we should do both at the same time to avoid converting code that will be removed/significantly altered.

lithrel:
> That will likely involve making search use the Ajax calls exclusively

@Inwerpsel  Still working on this topic and not sure I will do that, I still think it's better to understand the difference between the 2 search requests and properly unify the query behind them; adding an ajax call is just adding a request for the client, I don't really see a benefit to this at the moment

Inwerpsel:
> > That will likely involve making search use the Ajax calls exclusively
> 
> @Inwerpsel Still working on this topic and not sure I will do that, I still think it's better to understand the difference between the 2 search requests and properly unify the query behind them; adding an ajax call is just adding a request for the client, I don't really see a benefit to this at the moment

Good point, we should definitely sort out what the query is supposed to do. Using the Ajax calls seemed like the quickest way to get there and we would be able to already drop much of the complexity of search.

It has a few more benefits. We could likely cache the initial (empty) search HTML in Cloudflare. So the total time would not be significantly more, and perhaps even less because the initial call can be made more efficient than the current request that creates the first page HTML.

That being said, it would also need to have the sidebar filters, which may turn out more complicated (I guess it would need to be added to the first page data). So we can indeed still consider fixing the separate query.

PR #1562, opened 2022-01-03
Clean up (some) search dead code

Inwerpsel:
I removed everything that could easily be determined is not used. We should be able to merge this without affecting any live code.

Easier to verify [with whitespace ignored](https://github.com/greenpeace/planet4-master-theme/pull/1562/files?diff=split&w=1).

There surely is more dead code so expect a part II :grin:

### DIFF
--- a/templates/search.twig
+++ b/templates/search.twig
@@ -353,20 +353,15 @@
 							</div>
 						</div>
 						<div class="multiple-search-result">
-							{% if ( load_more.async ) %}
-								{% set displayed_posts = paged_posts %}
-							{% else %}
-								{% set displayed_posts = posts %}
-							{% endif %}
 							<ul class="list-unstyled">
-								{% for post in displayed_posts %}
+								{% for post in paged_posts %}
 									{% include ['tease-search.twig', 'tease-'~post.post_type~'.twig', 'tease.twig'] %}
 								{% endfor %}
 							</ul>
 							{% if ( load_more and found_posts > paged_posts|length ) %}
 								<div class="col-lg-12 mb-5 load-more-button-div">
 									<button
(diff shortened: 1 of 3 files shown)


mleray:
You'll need to rebase the PR in order to fix the broken acceptance test 🙃

Inwerpsel:
> You'll need to rebase the PR in order to fix the broken acceptance test upside_down_face

Thanks! :raised_hands: Also had to [fix the post deploy script](https://github.com/greenpeace/planet4-base/commit/ddab47027e56f945714859c468d350a21adfa8a5) in base.

templates/search.twig
```diff
@@ -377,24 +372,7 @@
 										</button>
 								</div>
 							{% endif %}
-							{% if ( pagination ) %}
```
mleray, line 380: Since this is no longer used, we should probably update the few comments in `Search.php` that still mention it (such as [this one](https://github.com/greenpeace/planet4-master-theme/blob/master/src/Search.php#L354))?

Inwerpsel: The comment is referring to the pagination we do use (which is what JS sends when you click the button based on what is shown).

I checked our code but nothing would something named `pagination` to the Twig context, so it's never called.

I doubt that it could even work, WP's `the_posts_pagination` function likely doesn't give the right output for the ElasticSearch query.

Inwerpsel: Checking locally if I remove the condition, the output does have use the right total amount of pages.

But the URL doesn't work for us because our current search implementation only uses the button for pagination.

Inwerpsel: Updated the comments to only use the general term :+1:

planet-4:

:watch: 2022.01.03 17:23:05

mleray (approved):
Checked on the test instance, all seems to be working as expected 👍 Just needs a small fix for the PHP tests!

PR #1563, opened 2022-01-03
Fix country selector acceptance test

lithrel:
Add a try/catch to test the new country selector, if the old one is not found.


mleray (approved):
Looks good to me 👍

PR #1570, opened 2022-01-10
PLANET-6263 Implement country selector banner A/B test

mleray:
### Description

See [PLANET-6263](https://jira.greenpeace.org/browse/PLANET-6263).

The banner template, JS and CSS code are now added to the GPI child theme, see https://github.com/greenpeace/planet4-child-theme-international/pull/1. This PR only takes care of including the template to the navbar.


planet-4:

:watch: 2022.01.10 11:08:42

assets/src/scss/layout/_navbar.scss
```diff
@@ -272,3 +272,4 @@ a.nav-link {
 @import "navbar/site-logo";
 @import "navbar/languages";
 @import "navbar/search";
+@import "navbar/country-selector-banner";
```
Inwerpsel, line 275: Since we only would actually use this on a single site, we should probably avoid loading the unused CSS on all the other sites.

templates/country_selector_banner.twig
```diff
@@ -0,0 +1,13 @@
+<div id="country-selector-banner" class="collapse">
```
Inwerpsel, line 1: Similar here. We would be adding this markup to every page of every site, and it will only be used on a single page of a single site.

assets/src/scss/layout/_navbar.scss
comzeradd: Would it make things easier if we move this code [here](https://github.com/greenpeace/planet4-child-theme-international)?

We can only keep the `{% include 'country_selector_banner.twig' %}` here. It should failed silently if the template is missing.

mleray: I guess we could do that yes, if we're sure it won't cause any crash! @Inwerpsel what do you think?

Inwerpsel: Sounds good :+1: We could also add the template to the child theme if we include an empty one in master theme and override it with a non empty one in the international child theme.

assets/src/js/header.js
```diff
@@ -108,6 +108,29 @@ const closeElement = (event, buttonClass) => {
   closeButton.click();
 };
 
+const setupCountrySelectorBanner = () => {
```
Inwerpsel, line 111: Can we also add this JS to the child theme?

mleray: Done!

templates/navigation-bar-new.twig
```diff
@@ -147,4 +147,5 @@
 		</button>
 	</div>
 	<div class="top-navigation-overlay"></div>
+	{% include 'country_selector_banner.twig' ignore missing %}
```
comzeradd, line 150: You are using underscores here for the template name, but dashes on the child theme 😁

mleray: Very good point haha, I was confused as to why it didn't show while testing 🤦‍♀️

mleray: I've updated the name in the child-theme, that made me realize that we also need to add there the svg for the close icon

comzeradd (approved):

PR #1573, opened 2022-01-11
Adjust CSS Customizer typography for readability

comzeradd:
Customizer font is really big since it's inherited from our theme.
Since it's on a sidebar a decreased size makes it a bit easier to read.

Changes:
- Decreases font-size
- Decreases line-height

### DIFF
--- a/assets/src/scss/editorOverrides.scss
+++ b/assets/src/scss/editorOverrides.scss
@@ -245,3 +245,9 @@ input.describe[type=text][data-setting=caption] {
 .btn {
   user-select: auto;
 }
+
+// Adjust CSS Customizer for readability
+.CodeMirror-code {
+  font-size: $font-size-xs;
+  line-height: 1rem;
+}


GP-Dan-Tovbein (approved):
Code looks good to me!

However, I wonder what is the motivation of using `rem`s and `px`s units both together. I see it in a lot of places. If we use `rem` we should also be sure to set `font-size` to `100%` into the `html`, to be more consistent. This is something that we should follow up in the future.

comzeradd:
> I wonder what is the motivation of using `rem`s and `px`s units both together

Yea, I guess we need to clean up some old code. Ideally we should always use `rem` for font sizes.

planet-4:

:watch: 2022.01.11 13:08:43

GP-Dan-Tovbein:
> > I wonder what is the motivation of using `rem`s and `px`s units both together
> 
> Yea, I guess we need to clean up some old code. Ideally we should always use `rem` for font sizes.

Completely agree with that 👍

PR #1577, opened 2022-01-13
PLANET-6508: Limit menu items count and title length in nav menu

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-6508

**WIP**

Error messages on navbar menu edition.


assets/src/js/menu_editor.js
```diff
@@ -0,0 +1,349 @@
...
+    return (conf.depthConf && conf.depthConf[depth]) ? conf.depthConf[depth] : conf;
+  };
+
+  /**
+   * Change item depth, using nav-menu.js API
+   *
+   * @param NodeElement item The menu item
+   */
+  const normalizeDepth = (item, maxDepth) => {
+    let obj = jQuery(`#${item.id}`);
```
Inwerpsel, line 235: `document.getElementById` is actually a simpler syntax :grin:

assets/src/js/menu_editor.js
```diff
@@ -0,0 +1,349 @@
...
+  if (!p4_menu_config) {
+    return;
+  }
+
+  // Get divs to check and write in
+  const menuContent = document.getElementById('menu-to-edit');
+  const menuEditorFooter = document.getElementById('nav-menu-footer');
+
+  // Add instruction for submenus
+  let instruction = document.createElement('p');
```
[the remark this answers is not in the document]
Inwerpsel: It's the same for every other usage of `let` in the PR. You only need `let` once you re-assign the variable, not when you change properties of an object or elements of an array.

Inwerpsel (commented):
Some small quick wins suggested here, but otherwise code looks pretty good. Still going through the rest.

assets/src/js/menu_editor.js
```diff
@@ -0,0 +1,349 @@
...
+  const getItemChildren = (item) => {
+    const depth = getItemDepth(item);
+    const itemsAfter = menuContent.querySelectorAll(
+      `#${item.id} ~ li.menu-item:not(.sortable-placeholder)`
+    );
+    if (itemsAfter.length <= 0 ) {
+      return null;
+    }
+
+    let children = [];
```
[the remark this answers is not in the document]
lithrel: I think I can't break out of a `.filter()`, I use a `for` loop here because I want to stop when I reach a different level - so trying to get the sub items of an item only on the specific sublevel
```
- Item A
- Item B
  - Item B-1
  - Item B-2
    - Item B-2-1
  - Item B-3
- Item C
```
So here for `getItemChildren(Item B)`, only the B-\* items, not A/C or B-\*-\*. And now that I write this I realize the issue in my code that will stop too early in some cases (it's not visible because of the depth limit currently used) :grimacing: And also missing a reset of the globalMaxDepth, fixing that :ok_hand:

assets/src/js/menu_editor.js
lithrel: The [nav-menu API](https://github.com/WordPress/WordPress/blob/master/wp-admin/js/nav-menu.js#L78) is not available if I only use getElementById, I don't know if there's another way to access it ?

assets/src/js/menu_editor.js
```diff
@@ -188,19 +189,20 @@ const menuEditorRestrictions = () => {
     const itemsAfter = menuContent.querySelectorAll(
       `#${item.id} ~ li.menu-item:not(.sortable-placeholder)`
     );
-    if (itemsAfter.length <= 0 ) {
-      return null;
-    }
 
-    let children = [];
+    const children = [];
     for (let child of [...itemsAfter]) {
```
Inwerpsel, line 194: Also this one can be a `const`

Inwerpsel (approved):

assets/src/js/menu_editor.js
Inwerpsel: Ah didn't realize it needs jQuery, that's fine then :+1:

PR #1579, opened 2022-01-17
POC Use Query Loop instead of Articles on posts

Inwerpsel:
* Just to test how easily we can start using this block.


Inwerpsel:
Closing this as a similar approach is being used in this PR for archive pages in https://github.com/greenpeace/planet4-master-theme/pull/1631. The difference with this draft it that here we set the filters dynamically to the same ones on the Articles block, instead of using `{"inherit": true}`.

PR #1581, opened 2022-01-19
Update pseudo selectors for buttons

GP-Dan-Tovbein:
Removed because of a11y

### DIFF
--- a/assets/src/scss/components/_share-buttons.scss
+++ b/assets/src/scss/components/_share-buttons.scss
@@ -26,7 +26,6 @@
     background: $twitter;
 
     &:hover,
-    &:focus,
     &:active {
       background: darken($twitter, 20%);
     }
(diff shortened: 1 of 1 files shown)


comzeradd:
Wouldn't this remove focus and active states completely from buttons?

Inwerpsel:
> Wouldn't this remove focus and active states completely from buttons?

This is only on share buttons. Those already get the [general focus outline we use on all buttons](https://github.com/greenpeace/planet4-master-theme/blob/e88a064a5bd81a146b53b99096eb9eb009d11120/assets/src/scss/components/_buttons.scss#L14-L17), which needs to be distinct from the hover state for a11y.

As for the active state we may want to find an alternative that is distinct from the other states. We could also only remove the different focus style for now and preserve the active style.

planet-4:

:watch: 2022.01.20 19:27:40

Inwerpsel (approved):

PR #1584, opened 2022-01-20
Re-add core block patterns

Inwerpsel:
* Just for easily creating a page on a test instance to evaluate which
core block patterns (if any) we would use.

Once I enabled the core patterns, I noticed on the test instance only 2 of them are available, because most of them use core blocks which we currently don't allow.

I then added a setting to this PR with all core blocks we currently don't enable, just to easily be able to see all the core blocks on the test instance.

But then I noticed it still didn't have all core blocks yet which I saw locally. That's because I was using the latest WP 5.9 RC.

Instead of setting the RC up on this test instance, I changed strategy and [created a feature toggle](https://github.com/greenpeace/planet4-master-theme/pull/1585) which we can enable on the [`gutenberg` instance](https://www-dev.greenpeace.org/gutenberg/). That one already has all blocks + 5.9 RC.

I'll keep this open for now as it might be useful when we evaluate core blocks.

**Observation**: There are a lot of core blocks we currently don't allow. This is because a lot of the recent complex Gutenberg blocks consist of many small component blocks which can be re-arranged.

Given this, maybe an exclusion list would work better than the [inclusion list](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/38812dde95747278dbc45316dc16b65a7cb3be27/planet4-gutenberg-blocks.php#L185-L227) we currently have?
Screenshot from 2022-01-21 11-39-33

To get to the list I just extracted all block types from the Gutenberg repo and removed the ones which we already allow. So there might be a few that are not interesting or older deprecated blocks that were still in the repo.

### DIFF
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -109,8 +109,6 @@ class MasterSite extends TimberSite {
 		if ( planet4_get_option( FEATURES::WP_TEMPLATE_EDITOR ) ) {
 			add_theme_support( 'block-templates' );
 		}
-		// Disable wp5.5 Block Patterns.
-		remove_theme_support( 'core-block-patterns' );
 
 		add_post_type_support( 'page', 'excerpt' );  // Added excerpt option to pages.
 
--- a/src/Settings.php
+++ b/src/Settings.php
@@ -23,6 +23,83 @@ class Settings {
 
 	public const SLACK_WEBHOOK = 'slack_webhook';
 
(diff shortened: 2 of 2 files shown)


planet-4:

:watch: 2022.01.20 11:42:44

Inwerpsel:
I moved this to https://github.com/greenpeace/planet4-master-theme/pull/1584 with only the core blocks settings change. Will move and update descriptions later.

PR #1591, opened 2022-01-24
PLANET-6510: Navigation bar - Tabs menu on mobile

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-6510

> Navigation Bar: implement a tabs menu on mobile
> The tabs should lead to each parent/category page of the site.
> They should have a selected state on the frontend regardless the level of hierarchy (example: Parent page > sub page).
> To view all items users would be able to swipe horizontally.
> The tabs menu should be sticky at the top and disappear when users start scrolling down the page.

### New settings page 
for all Information Architecture options, in _Planet 4 > Information architecture_

Activate this navigation menu:
Screenshot from 2022-02-01 16-31-29

### New menu on mobile

Menu is visible on `width < 992px`

Screenshot from 2022-01-24 16-05-55

### Additional changes

- Menu height is now 60px on width < 992px, (+50px if tabs menu is enabled)
- Menu links alignement with the Design iteration 2 (underline color, offset and animation, switch from padding to margin for spacing horizontal items)
- Added some CSS variables to more easily tweak theme and mobile menu

## Test
Test on [tavros](https://www-dev.greenpeace.org/test-tavros/) or locally
- Activate menu, add more menu items if necessary
- Check that current navigation still works

### DIFF
--- a/assets/src/js/header.js
+++ b/assets/src/js/header.js
@@ -120,11 +120,11 @@ const setMobileTabsMenuScroll = () => {
     return;
   }
 
-  const distToClose = 50
-    , distToOpen = 50;
-  let lastScrollDir = null
-    , lastScrollTop = window.pageYOffset || document.documentElement.scrollTop
-    , lastScrollRef = lastScrollTop;
+  const distToClose = 50;
+  const distToOpen = 50;
+  let lastScrollDir = null;
+  let lastScrollTop = window.pageYOffset || document.documentElement.scrollTop;
+  let lastScrollRef = lastScrollTop;
 
   /**
    * Get scroll direction, distance from the lastScrollTop, and distance from a reference point
@@ -132,16 +132,16 @@ const setMobileTabsMenuScroll = () => {
    * @return {Object} {scroll direction, top position, distance scrolled, distance scrolled from ref point}
    */
   const scrollData = () => {
-    let scrollTop = window.pageYOffset || document.documentElement.scrollTop;
+    const scrollTop = window.pageYOffset || document.documentElement.scrollTop;
     if (scrollTop === lastScrollTop) {
       return {};
     }
 
-    let dir = scrollTop > lastScrollTop ? 'down' : 'up';
-    let dist = Math.abs(scrollTop - lastScrollTop);
-    let ref = Math.abs(scrollTop - lastScrollRef);
+    const dir = scrollTop > lastScrollTop ? 'down' : 'up';
+    const dist = Math.abs(scrollTop - lastScrollTop);
+    const ref = Math.abs(scrollTop - lastScrollRef);
     lastScrollTop = Math.max(0, scrollTop);
-    return {dir: dir, pos: scrollTop, dist: dist, ref: ref};
+    return { dir, pos: scrollTop, dist, ref };
   };
 
   /**
@@ -172,9 +172,7 @@ const setMobileTabsMenuScroll = () => {
   };
 
   ['touchmove', 'scroll'].forEach((eventName) => {
-    document.addEventListener(eventName, () => {
-      toggleMobileTabsMenu();
-    });
+    document.addEventListener(eventName, toggleMobileTabsMenu);
   });
 };
 
--- a/assets/src/js/header.js
+++ b/assets/src/js/header.js
@@ -110,6 +110,74 @@ const closeElement = (event, buttonClass) => {
   closeButton.click();
 };
 
+/**
+ * Set the mobile tabs menu behavior on scroll
+ * Mobile tabs menu should hide when user scrolls down, and reappear when they scroll up
+ */
+const setMobileTabsMenuScroll = () => {
+  const menu = document.getElementById('nav-mobile');
+  if (!menu) {
+    return;
+  }
+
+  const distToClose = 50
+    , distToOpen = 50;
+  let lastScrollDir = null
+    , lastScrollTop = window.pageYOffset || document.documentElement.scrollTop
+    , lastScrollRef = lastScrollTop;
+
+  /**
+   * Get scroll direction, distance from the lastScrollTop, and distance from a reference point
+   *
+   * @return {Object} {scroll direction, top position, distance scrolled, distance scrolled from ref point}
+   */
+  const scrollData = () => {
+    let scrollTop = window.pageYOffset || document.documentElement.scrollTop;
+    if (scrollTop === lastScrollTop) {
+      return {};
+    }
+
+    let dir = scrollTop > lastScrollTop ? 'down' : 'up';
+    let dist = Math.abs(scrollTop - lastScrollTop);
+    let ref = Math.abs(scrollTop - lastScrollRef);
+    lastScrollTop = Math.max(0, scrollTop);
+    return {dir: dir, pos: scrollTop, dist: dist, ref: ref};
+  };
+
+  /**
+   * Show/hide tabs menu depending on scroll direction and distance scrolled from ref point
+   * Update reference point and reference direction only on state change
+   */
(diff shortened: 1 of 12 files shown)


assets/src/js/header.js
```diff
@@ -110,6 +110,74 @@ const closeElement = (event, buttonClass) => {
...
+ * Mobile tabs menu should hide when user scrolls down, and reappear when they scroll up
+ */
+const setMobileTabsMenuScroll = () => {
+  const menu = document.getElementById('nav-mobile');
+  if (!menu) {
+    return;
+  }
+
+  const distToClose = 50
+    , distToOpen = 50;
```
mleray, line 124: It's probably just me, but I find it weird to have the commas at the beginning of a line like this 😅 Is that a rule we have or something?

assets/src/js/header.js
```diff
@@ -110,6 +110,74 @@ const closeElement = (event, buttonClass) => {
...
+    let scrollTop = window.pageYOffset || document.documentElement.scrollTop;
+    if (scrollTop === lastScrollTop) {
+      return {};
+    }
+
+    let dir = scrollTop > lastScrollTop ? 'down' : 'up';
+    let dist = Math.abs(scrollTop - lastScrollTop);
+    let ref = Math.abs(scrollTop - lastScrollRef);
+    lastScrollTop = Math.max(0, scrollTop);
+    return {dir: dir, pos: scrollTop, dist: dist, ref: ref};
```
mleray, line 144: You can shorten this here to `{ dir, pos: scrollTop, dist, ref }`

assets/src/js/header.js
```diff
@@ -110,6 +110,74 @@ const closeElement = (event, buttonClass) => {
...
+    , lastScrollTop = window.pageYOffset || document.documentElement.scrollTop
+    , lastScrollRef = lastScrollTop;
+
+  /**
+   * Get scroll direction, distance from the lastScrollTop, and distance from a reference point
+   *
+   * @return {Object} {scroll direction, top position, distance scrolled, distance scrolled from ref point}
+   */
+  const scrollData = () => {
+    let scrollTop = window.pageYOffset || document.documentElement.scrollTop;
```
mleray, line 135: I see there are several places where you used `let` where you could have used `const` because the value is never changed, I think that would be better for readability

assets/src/js/header.js
```diff
@@ -110,6 +110,74 @@ const closeElement = (event, buttonClass) => {
...
+    if (dir === 'up' && ref >= distToOpen) {
+      lastScrollDir = dir;
+      lastScrollRef = lastScrollTop;
+      menu.classList.remove('mobile-menu-hidden');
+      return;
+    }
+  };
+
+  ['touchmove', 'scroll'].forEach((eventName) => {
+    document.addEventListener(eventName, () => {
```
mleray, line 175: I think this can be shortened to `document.addEventListener(eventName, toggleMobileTabsMenu)`

src/Settings.php
```diff
@@ -3,6 +3,7 @@
 namespace P4\MasterTheme;
 
 use CMB2_Field;
+use P4\MasterTheme\Settings\InformationArchitecture;
```
mleray, line 6: In the [MasterSite](https://github.com/greenpeace/planet4-master-theme/pull/1591/files#diff-d5a71432d6e94abc1260a448df306d8dfc76d15fa4047646a4d133c38cc131d7R14) file you renamed this as `IA`, maybe it would be good to do that here too for consistency?

src/Settings/InformationArchitecture.php
```diff
@@ -0,0 +1,83 @@
...
+	 *
+	 * @return array  The fields.
+	 */
+	public static function get_fields(): array {
+		$fields = [
+			[
+				'id'   => self::MOBILE_TABS_MENU,
+				'name' => __( 'Enable mobile tabs menu', 'planet4-master-theme-backend' ),
+				'desc' => __(
+					'Display a sticky tabs menu visible only on mobile size.',
```
mleray, line 49: It's also on tablets right? Maybe better to just specify the screen size?

templates/navigation-bar-new.twig
```diff
@@ -146,6 +150,27 @@
 			</span>
 		</button>
 	</div>
+	{% if mobile_tabs_menu %}
+	<div id="nav-mobile">
+		<nav id="nav-mobile-menu">
+			<ul>
+				{% for key,item in navbar_menu_items %}
+				<li class="nav-item {{ item.class }} {{ item == item.current ? 'active' : '' }}">
+					{% set link_ga_action = item.title %}
```
mleray, line 159: I see that [in the actual navbar](https://github.com/greenpeace/planet4-master-theme/blob/829716999de29a001b93909880dd26e76f8b46f2/templates/navigation-bar-new.twig#L28-L31) we manually set "Act" and "Explore" for the `link_ga_action`, I'm guessing because of potential translations? Maybe we should do the same here? 🤔 Otherwise this variable is pretty useless here and I would just remove it.

comzeradd: When we drop Act/Explore customization we will rely on item title to set the `link_ga_action` even on the main navbar. It's part of [PLANET-6497[(https://jira.greenpeace.org/browse/PLANET-6497). Since this is hidden behind a feature flag, I think it's fine to have this here already.

assets/src/scss/layout/_navbar.scss
```diff
@@ -1,6 +1,6 @@
 $menu-height: 68px;
```
mleray, line 1: Why do we have both `$menu-height` and `$menu-height-large`, if they're the same? 🤔

assets/src/scss/layout/_navbar.scss
```diff
@@ -209,33 +210,97 @@ $navbar-default-height: 60px;
...
   @include large-and-up {
     line-height: $menu-height;
+    margin: 0 20px;
   }
 }
 
+:root {
+  --nav-link--color: $white;
+  --nav-link--opacity: 1;
+  --nav-link--padding: 16px 0;
```
mleray, line 220: Maybe nice here to use our new spacing variables (i.e. `$sp-2`)

assets/src/scss/layout/navbar/_search.scss
```diff
@@ -175,8 +175,10 @@
   }
 
   @include large-and-up {
-    display: inline-block;
     border-inline-start: var(--top-navigation--separation, 1px solid transparentize($white, 0.5));
+    display: inline-block;
+    height: $menu-height-large;
+    padding: 12px;
```
mleray, line 181: Same 😬

assets/src/scss/layout/_navbar.scss
lithrel: I'd say legacy code :grimacing: We clearly need to normalize those variables, I'd be inclined to do it in a specific PR as I have no idea what the extent of these changes would be

mleray: Fair enough!

assets/src/scss/layout/navbar/_search.scss
```diff
@@ -165,7 +165,7 @@
 .nav-search-toggle-container {
   display: none;
   height: $menu-height-small;
-  padding: 8px 12px;
+  padding: #{$sp-1} #{$sp-1x};
```
mleray, line 168: You should be able to use directly `$sp-1` here 🤔

mleray (approved):
Looks good to me 👍

assets/src/scss/layout/navbar/_search.scss
Inwerpsel: There is also a nice syntax for this called numbers :wink:

PR #1593, opened 2022-01-26
Planet 6606 fix navbar issue

GP-Dan-Tovbein:
Ticket: https://jira.greenpeace.org/browse/PLANET-6606

The issue basically related of `Safari` by versions `< 13`, and initially reported by some iOS devices that have installed the OS version about < 12. _It's good to mention that the version of Safari, is the same as the OS_. It's been reported in iPhone and iPad.

**What is fixed?**

# Issue 1: Incompatible properties

### Description
There are some properties that work only on versions higher than `14`. 
Se an example about `inset-inline-start`, which is available only on versions higher than 14.

### Fix
Checked if the `property` is supported through this code
```
@supports not (property: value) {}
```

# Issue 2: Navigation's width

### Description
The `X` is hidden. The problem was about that the main navigation has been set to 400px of width, but smallest screens are about 320px. 

### Fix
For smallest screens, the width has been set to `100vw`

All these "workarounds" were been applied also to `ltr` and `rtl` html directions.

Demo page: https://www-dev.greenpeace.org/test-rhea

Some screenshots

**iPhone 5s (OS 12)**
Screenshot 2022-01-27 at 08 53 55
Screenshot 2022-01-27 at 08 54 01
Screenshot 2022-01-27 at 08 54 08
Screenshot 2022-01-27 at 08 54 30
Screenshot 2022-01-27 at 08 54 39
Screenshot 2022-01-27 at 09 40 08
Screenshot 2022-01-27 at 09 40 12
Screenshot 2022-01-27 at 09 40 18
Screenshot 2022-01-27 at 09 43 46

**iPad 7 gen (OS 13)**
Screenshot 2022-01-27 at 13 14 25

### DIFF
--- a/assets/src/scss/base/_mixins.scss
+++ b/assets/src/scss/base/_mixins.scss
@@ -12,6 +12,13 @@
   }
 }
 
+// Mobile to Large (From 576px to 992px)
+@mixin small-to-large {
+  @media (min-width: #{$small-width}) and (max-width: #{$large-width}) {
+    @content;
+  }
+}
+
 // Tablet & Up (greater than 768px)
 @mixin medium-and-up {
   @media (min-width: #{$medium-width}) {
--- a/assets/src/scss/base/_mixins.scss
+++ b/assets/src/scss/base/_mixins.scss
@@ -14,7 +14,7 @@
 
 // Mobile to Large (From 576px to 992px)
 @mixin small-to-large {
-  @media (min-width: #{$small-width}) and (max-width: #{$large-width}) {
+  @media (min-width: #{$small-width}) and (max-width: #{$large-width - 1}) {
     @content;
   }
 }
--- a/assets/src/scss/layout/_navbar.scss
+++ b/assets/src/scss/layout/_navbar.scss
@@ -96,7 +96,7 @@ $navbar-default-height: 60px;
 
     @supports not (inset-inline-start: -400px) {
       transform: translateX(-400px);
-      transition: transform 0.2s ease-out;
+      transition: transform 0.2s ease-in;
 
       html[dir="rtl"] & {
         transform: translateX(400px);
--- a/assets/src/scss/layout/_navbar.scss
+++ b/assets/src/scss/layout/_navbar.scss
@@ -88,17 +88,15 @@ $navbar-default-height: 60px;
     flex-direction: column-reverse;
     height: 100%;
     inset-inline-start: -400px;
-    overflow-y: auto;
-    overflow-x: hidden;
     position: fixed;
     transition: inset-inline-start 0.2s ease-in;
     top: 0;
     width: 400px;
+    z-index: 99;
 
     &.open {
       inset-inline-start: 0;
       transition: inset-inline-start 0.2s ease-out;
-      z-index: 99;
 
       ~ .top-navigation-overlay {
         display: block;
@@ -119,6 +117,26 @@ $navbar-default-height: 60px;
   @include small-and-less {
     width: 100vw;
     inset-inline-start: -100vw;
+
+    @supports not (inset-inline-start: -100vw) {
+      transform: translateX(-100vw);
+      transition: transform 0.2s ease-out;
+
+      html[dir="rtl"] & {
+        transform: translateX(100vw);
+      }
+    }
+
+    &.open {
+      @supports not (inset-inline-start: 0) {
+        transform: translateX(0);
+        transition: transform 0.2s ease-out;
+
+        html[dir="rtl"] & {
+          transform: translateX(0);
+        }
+      }
+    }
   }
 
   @include large-and-up {
@@ -134,15 +152,6 @@ $navbar-default-height: 60px;
   top: 0;
   width: auto;
   z-index: 4;
-
-  inset-block-start: 0;
-  @supports not (inset-block-start: 0) {
-    left: 0;
-
-    html[dir="rtl"] & {
-      left: 100vw;
-    }
-  }
 
   @include medium-and-less {
     display: block;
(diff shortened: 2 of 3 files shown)


planet-4:

:watch: 2022.01.26 12:49:34

assets/src/scss/layout/_navbar.scss
```diff
@@ -227,6 +226,7 @@ a.nav-link {
 
   @include medium-and-less {
     display: block;
+    right: 0;
```
lithrel, line 229: If you switch to `dir="rtl"` the close button is stuck on the wrong side, I think `insert-inline-end` would work better ?

assets/src/scss/layout/_navbar.scss
```diff
@@ -117,9 +117,8 @@ $navbar-default-height: 60px;
   }
 
   @include small-and-less {
-    &.open {
-      width: 100%;
-    }
+    width: 100vw;
```
lithrel, line 120: I have a small glitch if you take `width` out of the `.open` selector, if you resize your window to a smaller width the left menu appears and disappears for no reason
Peek 2022-01-26 14-18

assets/src/scss/layout/_navbar.scss
GP-Dan-Tovbein: I figured out some minutes ago that the issue is bacause`insert-inline-end` is not supported to that iOS used version (12) (https://developer.mozilla.org/en-US/docs/Web/CSS/inset-inline-start). Only it has a full supported from version `14.5`

I'm just working on that doing the same as you suggested !

assets/src/scss/layout/_navbar.scss
```diff
@@ -117,8 +115,27 @@ $navbar-default-height: 60px;
   }
 
   @include small-and-less {
+    width: 100vw;
+    inset-inline-start: -100vw;
+
+    @supports not (inset-inline-start: -100vw) {
```
Inwerpsel, line 121: You can probably use a single `@supports` rule testing for `insert-inline-start: -100vw`. I don't think there's cases where a browser would support only 1 of the 2.

comzeradd: Just make sure that `@supports` is supported by the browser in question :grin:

Otherwise the code should be reverse to avoid using `not`.

GP-Dan-Tovbein: I've added `not` since `insert-inline-start` is available in most of the browsers but there some few that aren't. Doing this we'll avoid to add extra code in most cases.

Does it make sense?

GP-Dan-Tovbein:
@comzeradd @Inwerpsel FYI, Stefan has reported the same issue by using iPad 7 gen (OS 13). I fixed that and validated by him through this [thread](https://greenpeace.slack.com/archives/C014UMRC4AJ/p1643290339082000).

assets/src/scss/layout/_navbar.scss
comzeradd: Yes. But I meant that if that browser doesn’t supports `@supports` either then this won’t work.

I see it’s [supported](https://caniuse.com/css-featurequeries) since Safari 9, so we are good 👍🏻

assets/src/scss/layout/_navbar.scss
comzeradd: I can reproduce this too. It happens at exactly `767px` width.

GP-Dan-Tovbein: Oh sorry! I completely missed this conversation. Working on this

GP-Dan-Tovbein:
Also approved by Andre (the reporter) https://greenpeace.slack.com/archives/C014UMRC4AJ/p1643314372094600?thread_ts=1643290339.082000&cid=C014UMRC4AJ

assets/src/scss/base/_mixins.scss
```diff
@@ -12,6 +12,13 @@
   }
 }
 
+// Mobile to Large (From 576px to 992px)
+@mixin small-to-large {
+  @media (min-width: #{$small-width}) and (max-width: #{$large-width - 1}) {
+    @content;
+  }
+}
+
```
GP-Dan-Tovbein, line 21: The production released behavior is not taking account the behavior for smaller screens because it has set `400px` of width, and if any user navigate with an `iPhone 5`, which is the initially reported device, some parts of the navbar will be hidden (included the X button), since the width of the navbar (`400px`) exceeds the width of the viewport (`320px`).

But *Why I've added this new @media?* 
We need to specify behaviors to this kind of implementations. _It's a very particular issue_.
For one hand, there is a rule `@media medium-and-less` that set the `navbar` to `400px`, on the other hand there is another rule `@media small-and-less`, which it was included on this PR, that set to `100vw` the `navbar` width for smaller screens. In consequence, the `width` attribute is set in two different `@media queries` through the same viewport width and causes this [unexpected behavior](https://github.com/greenpeace/planet4-master-theme/pull/1593/files#r792628708) reported by Flo.

Finally, I assume that when Safari will have support to all these unsupported features in `< OS 13`, we could remove all these unloved [workarounds](https://github.com/greenpeace/planet4-master-theme/pull/1593/files#diff-91009267fa7864d1b148d657a615793b252f3e148c4001b639eab5a4c51324e4R128-R159).

lithrel (approved):
:+1: 
Tested on Chromium 97.0 & Firefox 96.0

assets/src/scss/layout/_navbar.scss
```diff
@@ -143,6 +179,7 @@ $navbar-default-height: 60px;
   }
 
   @include large-and-up {
+    display: none;
```
lithrel, line 182: As the frontend test notices https://app.circleci.com/pipelines/github/greenpeace/planet4-master-theme/4756/workflows/4ba298db-80d2-40d7-88b5-4c74300a797a/jobs/28292 there are 2 display declarations here

GP-Dan-Tovbein: Yes! I figured out today about this. I just updated the PR.

PR #1594, opened 2022-01-26
POC Use WordPress's new template mechanism

Inwerpsel:
This adds an Author page template using a new template mechanism in WordPress. I hope one day each of these mechanisms will get a clear own name, but for now we need to explain in more detail to avoid confusion.

This mechanism works by checking the `block-templates` directory for a file with a certain name. For example `author.html` for the Author page.

It's checked before any of the other mechanisms, so I didn't have to remove `author.php` which it would use if we didn't provide this new file.

There's just 1 problem, but a big one. We have a lot of elements of our site in Twig templates, in fact most of them. Ideally all of these should be converted to blocks so that we can just add those blocks to the new WP template files.

But that's a lot of work before we can start using this. To get around the limitation, I created a new ["Framework" block](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/768) which holds the entire Planet4 UI, except the content part. The content is created by rendering the block's inner blocks to a HTML string and output it raw in Twig.

This was surprisingly easy and could directly render the `base.twig` template with almost no modification.

### Observations
* To fully create the author page currently it misses a block to output author name as a title. Should be relatively easy to create this block ourselves as we only need to define a back end render function for this use case.
* An author bio block exists but it's more created as a secondary element on a post page. Maybe only needs styling changes.
* "Inherit from template" Query Loop block seems to work for the author page. Still need to confirm other page types.
* Pagination works and uses the items per page defined in WP's reading settings. On the test instance I changed this to 12 to better line out.
* By default it uses the path and not a query string for the page numbers. Which is nice because it makes caching easier. We can just invalidate all pages from cache every time any list entry is added, changed, or removed. Could be the CF plugin already does this to some extent.

### [Test instance demo](https://www-dev.greenpeace.org/test-venus/author/dtovbein/)
I imported data from Denmark. Keep in mind that the author IDs on posts are not updated, so the author will change to whatever user has that ID on the test instance.

### WordPress docs
* https://developer.wordpress.org/block-editor/how-to-guides/themes/block-theme-overview/

### Gutenberg repo related issues
* https://github.com/WordPress/gutenberg/issues/37407

### Other WP themes that use this feature
* https://github.com/WordPress/theme-experiments/tree/master/tt1-blocks

### DIFF
--- a/assets/src/scss/editorOverrides.scss
+++ b/assets/src/scss/editorOverrides.scss
@@ -19,12 +19,13 @@
   }
 }
 
+.wp-block {
+  max-width: 90%;
+  margin-left: auto;
+  margin-right: auto;
+}
+
 .edit-post-visual-editor {
-  .wp-block {
-    max-width: 90%;
-    margin-left: auto;
-    margin-right: auto;
-  }
 
   li {
     font-size: unset;
--- a/assets/src/scss/editorStyle.scss
+++ b/assets/src/scss/editorStyle.scss
@@ -21,3 +21,4 @@
 @import "layout/forms";
 
 @import "editorOverrides";
+@import "layout/block-classes";
--- a/assets/src/scss/layout/_block-classes.scss
+++ b/assets/src/scss/layout/_block-classes.scss
@@ -1,5 +1,52 @@
 
 .inline-container > * {
   display: inline;
-  margin-inline-end: 4px;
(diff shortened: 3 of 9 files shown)


planet-4:

:watch: 2022.01.27 09:38:01

PR #1597, opened 2022-01-31
Fix tags spacing in breadcrumbs

mleray:
### Description

The space between tags in the breadscrumbs had been removed by mistake when adding the new separators between categories in post navigation links. 

### Testing

You can find these tags in several places, the easiest one to check is probably in an Articles block. Make sure the new post navigation links also still look as expected!

### DIFF
--- a/assets/src/scss/layout/_breadcrumbs.scss
+++ b/assets/src/scss/layout/_breadcrumbs.scss
@@ -12,8 +12,12 @@
   .tag-wrap-bullet {
     pointer-events: none;
     font-weight: 400;
-    margin: 0 $sp-1;
+    margin-inline-end: $sp-1;
     color: $link-color;
+  }
+
+  .tag-wrap-bullet {
+    margin-inline-start: $sp-1;
   }
 
   .tag-item {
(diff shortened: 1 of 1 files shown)


planet-4:

:watch: 2022.01.31 13:07:48

assets/src/scss/layout/_breadcrumbs.scss
```diff
@@ -26,6 +30,10 @@
       }
     }
     display: inline-block;
+
+    &:not(:only-child):not(:last-child):not(.page-type) {
+      margin-inline-end: $sp-1;
+    }
```
comzeradd, line 36: Do you think this can be reversed? Otherwise it may be difficult for child theme to override `.tag-item` spacing. 

So define a default `margin-inline-end` and then override it for:
````
&:only-child,
&:last-child,
&:page-type
```

mleray: This is actually just adding [the lines that were there before](https://github.com/greenpeace/planet4-master-theme/commit/18776d064d5c704ad0ed0caa770c967bdc5a7301#diff-edf8cadc9b9e8d401ce446b027742bd8a026dc2b6b42064f3d596935e541c875L29-L31), so I think it was like this for quite a while 🤔

comzeradd: Ah ok then, so we know that child theme overrides won't break :+1:

comzeradd (approved):

PR #1598, opened 2022-02-02
POC Add Listing Page post type

Inwerpsel:
* Use it in tag redirect page dropdown.
* Check if it's published before using. This would need to be shown in
the UI.
* Use campaigns caps for now to avoid needing database updates.
* Lacks a migration script for existing content, which we would need to
release together to ensure the existing pages are shown as a value in
the dropdown.

### DIFF
--- a/src/Campaigns.php
+++ b/src/Campaigns.php
@@ -88,9 +88,10 @@ class Campaigns {
 			$dropdown_args = [
 				'show_option_none' => __( 'Select Page', 'planet4-master-theme-backend' ),
 				'hide_empty'       => 0,
-				'hierarchical'     => true,
-				'selected'         => $redirect_page,
+				'selected'         => esc_attr( $redirect_page ),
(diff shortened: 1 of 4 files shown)


src/Campaigns.php
```diff
@@ -88,9 +88,10 @@ public function add_taxonomy_form_fields( $wp_tag ) {
 			$dropdown_args = [
 				'show_option_none' => __( 'Select Page', 'planet4-master-theme-backend' ),
 				'hide_empty'       => 0,
-				'hierarchical'     => true,
```
Inwerpsel, line 91: I removed this because `wp_dropdown_pages` only works on hierarchical post types.

For listing pages, hierarchy doesn't make sense, the page is associated with a post, not a term.

As a workaround I made it hierarchical anyway, but we could use an alternative to `wp_dropdown_pages` that doesn't have this tight coupling.

PR #1600, opened 2022-02-03
PLANET-6611 Hide post type in admin listings

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6611

Make `post_type` a hidden column by default. The Post Type Switcher plugin doesn't do this, even though it doesn't depend on the column being visible to work.

Showing the column is a bit pointless because all listings only ever have a single post type.

The only advantage of having it a visible column is it allows you to see the different type after switching. But only when using "Quick edit" in the listing view. It's because quick edit doesn't refresh the page but updates in place.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -192,3 +192,21 @@ add_action(
 	1,
 	2
 );
+
+/**
+ * This is not a column in WordPress by default, but is added by the Post Type Switcher plugin.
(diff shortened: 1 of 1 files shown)


planet-4:

:watch: 2022.02.03 13:01:46

PR #1601, opened 2022-02-03
PLANET-6614: Save post tags as ordered

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-6614

> it is impossible to set the main tag (and the other page tags) in a custom order, because when there is more than 1 tag upon page save the tags get automatically rearranged in alphabetical order. 

Post tags are not ordered by default, so order is lost on save, and never used on fetch.

## Fix

Using [`register_taxonomy_args`](https://developer.wordpress.org/reference/functions/register_taxonomy/) hook to
- set `sort` parameter for keeping order when saving post tags
- set `args` parameter for using order when fetching post tags

This configuration forces the use of the DB field `wp_term_relationships.term_order`.

## Test

- Edit a new page
- Add tags in a different order than alphabetical (ie `Ocean, Coal, Forest`)
- Clicking on the `Save draft` button shouldn't reorder the tags
- Tags should be displayed on the same order on the frontend

### DIFF
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -264,6 +264,9 @@ class MasterSite extends TimberSite {
 			}
 		);
 
+		// Make post tags ordered.
+		add_filter( 'register_taxonomy_args', [ $this, 'set_post_tags_as_ordered' ], 10, 2 );
+
 		$this->register_meta_fields();
 	}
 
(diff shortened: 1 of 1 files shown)


planet-4:

:watch: 2022.02.03 13:14:16

PR #1604, opened 2022-02-04
PLANET-6509: Navigation Bar styling drop-down menus

GP-Dan-Tovbein:
Ref: https://jira.greenpeace.org/browse/PLANET-6509

👉 [Parent PR](https://github.com/greenpeace/planet4-master-theme/pull/1628)

**How to test**
Navigate to the [instance](https://www-dev.greenpeace.org/test-nix/) and play around with the menu.

Also go to:
`Planet4 > Navigation` and change the **Website Navigation Style** to `dark|light`

**What to test**
- Validate the Design iteration2_IA & nav MVP from [here](https://www.figma.com/file/Gn8wyqgto7608rlESePXZE/Foundations-%26-Components?node-id=0%3A103).
- Analytics vars set to the sub-menu items.

**New CSS variables**
```
--submenu-nav-link--color
--submenu-nav-link--hover--color
--submenu-nav-link--font-size
--submenu-nav-link--font-weight
```

### DIFF
--- a/assets/src/js/header.js
+++ b/assets/src/js/header.js
@@ -229,3 +229,4 @@ export const setupHeader = () => {
 
   setMobileTabsMenuScroll();
 };
+
--- a/assets/src/scss/layout/_navbar.scss
+++ b/assets/src/scss/layout/_navbar.scss
@@ -204,6 +204,91 @@ $navbar-default-height: 60px;
   }
 }
 
+.nav-submenu {
+  position: fixed;
+  display: none;
+  min-width: 193px;
+  padding-top: $sp-1;
+
+  .nav-submenu-wrapper {
+    border-radius: $sp-1;
+    width: 100%;
+    box-shadow: 0 1px $sp-x rgba(0, 0, 0, 0.12);
+  }
+
+  ul {
+    width: 100%;
+    // The `height` value represents the value of 6 items's height
+    // and this selectors's padding
+    height: auto;
+    max-height: 320px;
+    display: block;
+    flex-direction: column;
+    padding: $sp-2 0;
+    background: $white;
+    border-radius: $sp-1;
+    overflow-y: auto;
+
+    // Only applied to Firefox
+    scrollbar-width: thin;
+    scrollbar-color: transparent;
+
+    &::-webkit-scrollbar {
+      width: $sp-1;
+    }
+
+    &::-webkit-scrollbar-track {
+      border-radius: $sp-x;
+    }
+
+    &::-webkit-scrollbar-thumb {
+      border-radius: $sp-x;
+      background: rgba(0, 0, 0, 0.15);
+    }
+
+    &::-webkit-scrollbar-thumb:hover {
+      background: rgba(0, 0, 0, 0.2);
+    }
+  }
+
+  li {
+    width: 100%;
+    text-align: left;
+    white-space: nowrap;
+    color: $grey-80;
+    height: 48px;
+
+    a.nav-link {
+      --submenu-nav-link-- {
+        color: $grey-80;
+        font-size: $sp-2;
+        font-weight: 500;
+
+        &:hover {
+          _-- {
+            color: $grey-80;
+          }
+        }
+      }
+      text-decoration: none;
+      height: 100%;
+      display: flex;
+      align-items: center;
+      padding-right: $sp-3;
+      padding-left: $sp-3;
+
+      &:hover {
+        background: $grey-10;
+      }
+
+      &:before {
+        // Disable nav-link animation on hover
+        display: none;
+      }
+    }
+  }
+}
+
 .nav-item {
   color: inherit;
   font-size: $font-base-size;
@@ -212,6 +297,13 @@ $navbar-default-height: 60px;
   @include large-and-up {
     line-height: $menu-height;
     margin: 0 20px;
+
+    &:hover {
+      .nav-submenu {
+        display: flex;
+        z-index: 999;
+      }
+    }
   }
 }
 
@@ -221,7 +313,6 @@ a.nav-link {
 
     &:hover {
       color: $white;
-      text-decoration-color: $gp-green;
     }
   }
 
@@ -230,21 +321,10 @@ a.nav-link {
   padding: $sp-2 0;
   position: relative;
   width: 100%;
+  text-decoration: none;
 
   @include large-and-up {
     padding: 0;
-  }
-
-  &:hover {
-    opacity: 1;
-    text-decoration-line: underline;
-    text-decoration-thickness: 3px;
-    text-underline-offset: auto;
-    text-underline-position: under;
-
-    @include large-and-up {
-      text-decoration-line: none;
-    }
   }
 
   // Border-bottom for underline animation
@@ -261,34 +341,33 @@ a.nav-link {
     transform: scaleX(0);
     transition: transform 0.25s;
     z-index: -1;
+
+    @include large-and-up {
+      border-bottom-width: 4px;
+    }
   }
 }
 
 .active a.nav-link {
   --nav-link-active-- {
-    color: var(--nav-link--color, $white);
-    text-decoration-color: $gp-green;
+    color: $white;
+
+    &:hover {
+      color: $white;
+    }
   }
 
-  opacity: 1;
-  text-decoration-line: underline;
-  text-decoration-thickness: 3px;
-  text-underline-offset: auto;
-  text-underline-position: under;
-
-  @include large-and-up {
-    text-decoration-line: none;
-    text-decoration-thickness: 4px;
+  &:before {
+    transform: scaleX(1);
+    border-bottom-color: var(--nav-link--text-decoration-color, $gp-green);
   }
 }
 
 // Underline animation with border-bottom
-a.nav-link:hover:before,
-.active a.nav-link:before {
-  @include large-and-up {
-    transform: scaleX(1);
-    transition: transform 0.25s;
-  }
+.nav-item:hover a.nav-link:before {
+  transform: scaleX(1);
+  transition: transform 0.25s;
+  border-bottom-color: var(--nav-link--text-decoration-color, $gp-green);
 }
 
 .nav-menu-toggle {
--- a/src/AdminAssets.php
+++ b/src/AdminAssets.php
@@ -39,7 +39,7 @@ class AdminAssets {
 						'maxChars' => 18,
 					],
 					1 => [
-						'maxItems' => 5,
+						'maxItems' => 10,
 						'maxChars' => 18,
 					],
 				],
--- a/assets/src/scss/partials/navigation-bar-light.scss
+++ b/assets/src/scss/partials/navigation-bar-light.scss
@@ -14,6 +14,8 @@
   --top-navigation--input--placeholder--color: #{$grey-40};
   --nav-link--color: #{$black};
   --nav-link--hover--color: #{$black};
+  --nav-link-active--color: #{$black};
+  --nav-link-active--hover--color: #{$black};
 }
 
 #nav-mobile-menu {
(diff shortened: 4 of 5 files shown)


planet-4:

:watch: 2022.02.04 20:09:00

assets/src/js/burger_menu.js
```diff
@@ -0,0 +1,29 @@
+export const setupBurgerMenu = () => {
+
+  /**
+   * This event is fired once the user clicks on the chevron icon
+   * To collapse or expand the sub list through each item from
+   * The burger menu
+   * @param {*} evt
+   */
+  const onClickItem = evt => {
```
mleray, line 9: Can't we use bootstrap in order to do this? 🤔

assets/src/js/header.js
```diff
@@ -227,5 +227,19 @@ export const setupHeader = () => {
     }
   });
 
+  /**
+   * Setup max height to submenues.
+   * The limits should be 6 for scrolleable and 10 for maximum into the list
```
lithrel, line 232: We settled for soft limitations in [PLANET-6508](https://github.com/greenpeace/planet4-master-theme/pull/1577), you can implement those in the menu editor like this (it includes a fix for a bug in the current code :grimacing: )
```diff
diff --git a/assets/src/js/menu_editor.js b/assets/src/js/menu_editor.js
index a6e98cef..4dd39abb 100644
--- a/assets/src/js/menu_editor.js
+++ b/assets/src/js/menu_editor.js
@@ -160,7 +160,7 @@ const menuEditorRestrictions = () => {
     return locations.map(location => {
       return document.querySelector(`input[name="menu-locations[${location}]"]`).checked
         ? location : null;
-    })[0] || null;
+    }).filter(l => l)[0] || null;
   };
 
   /**
diff --git a/src/AdminAssets.php b/src/AdminAssets.php
index f4a6a094..7fb7dcd8 100644
--- a/src/AdminAssets.php
+++ b/src/AdminAssets.php
@@ -24,8 +24,9 @@ class AdminAssets {
 		$theme_dir       = get_template_directory_uri();
 		$menus           = get_nav_menu_locations();
 		$navbar_location = 'navigation-bar-menu';
+		$donate_location = 'donate-menu';
 
-		if ( ! isset( $menus[ $navbar_location ] ) ) {
+		if ( ! isset( $menus[ $navbar_location ] ) && ! isset( $menus[ $donate_location ] ) ) {
 			return;
 		}
 
@@ -44,6 +45,11 @@ class AdminAssets {
 					],
 				],
 			],
+			$donate_location => [
+				'maxDepth' => 0,
+				'maxItems' => 5,
+				'maxChars' => 18,
+			],
 		];
 
 		$conf = apply_filters( 'planet4_menu_config', $default_conf );
```

GP-Dan-Tovbein: Thanks for point me this! 
I've added the sub menu setup just for setting the `scrolleable` class in cases where the list would have more than 6 items. See what the ticket mentions:
> A drop down menu should show 6 visible items and have a scroll bar when there are more than 6 items on desktop screens. 

Let me know if I'm in the correct way!

lithrel: Sounds good :+1: and in my diff `maxItems` should be `6` then I guess

GP-Dan-Tovbein: Yep! I've already changed that

src/AdminAssets.php
```diff
@@ -39,7 +39,7 @@ private static function enqueue_menu_editor_script(): void {
 						'maxChars' => 18,
 					],
 					1 => [
-						'maxItems' => 5,
+						'maxItems' => 10,
```
GP-Dan-Tovbein, line 42: As it's been requested by the ticket, the maximum items should be set to `10` but scrollable greater than `6`.

assets/src/scss/layout/_navbar.scss
```diff
@@ -261,24 +324,24 @@ a.nav-link {
...
     z-index: -1;
+
+    @include large-and-up {
+      border-bottom-width: 4px;
+    }
   }
 }
 
 .active a.nav-link {
   --nav-link-active-- {
```
GP-Dan-Tovbein, line 335: Removed all this code since the border bottom is assigned to the `:after` pseudo element.

assets/src/scss/layout/_navbar.scss
```diff
@@ -230,23 +304,12 @@ a.nav-link {
   padding: $sp-2 0;
   position: relative;
   width: 100%;
+  text-decoration: none;
 
   @include large-and-up {
     padding: 0;
   }
```
GP-Dan-Tovbein, line 312: Removed all this code since the border bottom is assigned to the `:after` pseudo element.

assets/src/js/header.js
```diff
@@ -227,5 +227,44 @@ export const setupHeader = () => {
...
+      // Find its parent menu item
+      const menuItem = document.querySelector(`*[data-menu="${submenu.dataset.parentMenu}"]`);
+      const list = submenu.querySelector('ul');
+
+      // The limits should be 6 for a scrollable list
+      if(list.children.length >= 6) {
+        list.classList.add('scrollable');
+      }
+
+      submenu.addEventListener('mouseover', evt => {
```
mleray, line 246: I think this could be achieved in CSS directly with the `:hover` pseudo-element on the `li` elements, by changing the `display` value of their corresponding dropdown menus

assets/src/js/header.js
```diff
@@ -227,5 +227,44 @@ export const setupHeader = () => {
...
+        evt.preventDefault();
+
+        if(menuItem) {
```
mleray, line 249: I think here you should check if the menu item doesn't already have the `active` class before adding it

assets/src/js/header.js
```diff
@@ -227,5 +227,44 @@ export const setupHeader = () => {
...
+   * Mobile should render an accordion list
+   */
+  const setupSubmenu = () => {
+    document.querySelectorAll('.nav-submenu').forEach(submenu => {
+      // Find its parent menu item
+      const menuItem = document.querySelector(`*[data-menu="${submenu.dataset.parentMenu}"]`);
+      const list = submenu.querySelector('ul');
+
+      // The limits should be 6 for a scrollable list
+      if(list.children.length >= 6) {
```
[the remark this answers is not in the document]
mleray: I see that you actually already know the `max-height` it could be, I think if you just set [this value](https://github.com/greenpeace/planet4-master-theme/pull/1604/files#diff-91009267fa7864d1b148d657a615793b252f3e148c4001b639eab5a4c51324e4R228) as max-height to all of them with the auto overflow-y then it should do the same behaviour without the need for this extra JS and class 🤔

assets/src/scss/layout/_navbar.scss
```diff
@@ -204,6 +204,93 @@ $navbar-default-height: 60px;
...
+
+    &.scrollable {
+      height: 320px;
+      overflow-y: auto;
+      display: block;
+
+      // Only applied to Firefox
+      scrollbar-width: thin;
+      scrollbar-color: transparent;
+      scrollbar-gutter: stable both-edges;
```
mleray, line 235: This one I actually see applied in Chrome as well, and it looks a bit strange I think:
Screenshot 2022-02-24 at 09 44 37

assets/src/scss/layout/_navbar.scss
mleray: Could this be the reason why the underline is now gone in the mobile menu? 🤔

mleray (changes requested):
I've left a few suggestions for improvements, let me know if anything is not clear 🙂 

I've also noticed on the test instance that if you open the search bar in screens between 992px & 1200px with that many menu items, since the navbar spreads on 2 lines it looks pretty bad:
Screenshot 2022-02-24 at 10 05 05

assets/src/js/header.js
GP-Dan-Tovbein: Good catch!

assets/src/scss/layout/_navbar.scss
GP-Dan-Tovbein: The `text-decoration` feature is not used anymore. The underline effect is now applied to the `a`'s `:after` pseudoselector. See [here](https://github.com/greenpeace/planet4-master-theme/blob/fdc54acd1d45f16e90f3f3f7238b5da358787fb2/assets/src/scss/layout/_navbar.scss#L372-L373)

mleray: Yes I saw that, but there used to be an underline effect in the mobile menu too and that is gone now with this PR, so I was wondering if it could be due to this particular removal

GP-Dan-Tovbein:
> I've left a few suggestions for improvements, let me know if anything is not clear 🙂
> 
> I've also noticed on the test instance that if you open the search bar in screens between 992px & 1200px with that many menu items, since the navbar spreads on 2 lines it looks pretty bad: Screenshot 2022-02-24 at 10 05 05

**This is something that I was discussing with Houssam some days ago!**
Here is the chat:

> Dan: I think is too long the max length of characters set to each item. What do you think?
> Houssam: well, depending on each NRO. These NROs have the longest characters.
> Houssam: We can keep this option for NROs who will definitely use short words. What do you think?
> Dan: Yep! But how do you avoid for NROs who use large words?
> Dan: We can keep this but assume the human error when they’re editing the menus
> Houssam: yes! we have this concern :grimacing: some NROs won't follow the best practices for the nav bar items

assets/src/scss/layout/_navbar.scss
GP-Dan-Tovbein: Check it again! I rolled back this effect and included to mobile. [Here is the code](https://github.com/greenpeace/planet4-master-theme/pull/1604/files#diff-91009267fa7864d1b148d657a615793b252f3e148c4001b639eab5a4c51324e4R365-R369)

assets/src/js/header.js
```diff
@@ -227,5 +227,38 @@ export const setupHeader = () => {
...
+   */
+  const setupSubmenu = () => {
+    document.querySelectorAll('.nav-submenu').forEach(submenu => {
+      // Find its parent menu item
+      const menuItem = document.querySelector(`*[data-menu="${submenu.dataset.parentMenu}"]`);
+
+      submenu.addEventListener('mouseover', evt => {
+        evt.preventDefault();
+
+        if(menuItem && !menuItem.classList.contains('active')) {
```
mleray, line 243: I think it would be best to avoid adding the `active` class, since it should be reserved for the page the user is actually currently on... I think it's doable in CSS using something like:

```
.nav-item:hover a.nav-link:before {
    transform: scaleX(1);
    transition: transform 0.25s;
    border-bottom-color: var(--nav-link--text-decoration-color, $gp-green);
}
```

GP-Dan-Tovbein: If we add your suggested css it would be only applied when the nav item is hovered. Let me know if I'm correct

mleray: The `nav-item` element is the "root" list item in the navbar, so the submenu is technically part of it, therefore hovering on the submenu is also hovering on the `nav-item`... Does that make sense? 😅

assets/src/scss/layout/_navbar.scss
```diff
@@ -261,34 +341,32 @@ a.nav-link {
...
-  text-underline-position: under;
+    &:hover {
+      color: $white;
+    }
+  }
 
-  @include large-and-up {
-    text-decoration-line: none;
-    text-decoration-thickness: 4px;
+  &:before {
```
mleray, line 360: I think you need to add `transform: scaleX(1);` here too otherwise the border doesn't show for the active item

mleray (approved):
Looks good to me :+1: Maybe you can just squash some of the commits before merging!

PR #1608, opened 2022-02-08
PLANET-6611 Add config for whether type can be switched

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6611
* This needs to use args, the plugin only exposes the args and doesn't
directly accept a post type slug.

I'm still checking the sidebar, it doesn't seem to take this into account for whether the widget is displayed. Though the choice list is in fact properly filtered. So it's only a problem on the post page. But could be annoying because you can't switch back afterwards :grimacing: 

I then wanted to hide it on post editor pages with CSS, but the markup has nothing we can use to reliably target it. It only re-uses classes also present in other sidebar components.

Probably the best solution is to dequeue the plugin's editor script and provide our own component. This way it's not needed to fork the whole plugin just for this component. Should be manageable as it's a [simple standalone component](https://github.com/JJJ/post-type-switcher/blob/master/src/components/post-type-switcher.js) and we only need minor modifications to it.

[Example post page](https://www-dev.greenpeace.org/test-pluto/wp-admin/post.php?post=47198&action=edit
) (same on any post)

The filter does work as intended for bulk and quick edit.

I also noticed the plugin does an [unsafe GET when switching the post type](https://github.com/JJJ/post-type-switcher/blob/master/src/components/post-type-switcher.js#L39). I think this endpoint is only used for this sidebar. We should probably make this use POST.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -224,4 +224,6 @@ add_filter(
 	4
 );
 
+add_action( 'wp_print_scripts', fn() => wp_dequeue_script( 'pts_blockeditor' ), 100 );
+
 simple_value_filter( 'pts_post_type_filter', [ 'allow_switch_type' => true ], 1 );
--- a/functions.php
+++ b/functions.php
@@ -210,3 +210,18 @@ add_filter(
 	10,
 	1
 );
+
+add_filter(
+	'register_post_type_args',
+	function ( $args, $name ) {
+		if ( 'page' === $name ) {
+			$args['allow_switch_type'] = true;
+		}
+
+		return $args;
+	},
+	10,
+	4
+);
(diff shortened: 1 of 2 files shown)


planet-4:

:watch: 2022.02.08 16:01:44

comzeradd:
Hm.. yeah maybe we can leave this for now. It would be nice if the plugin filter could allow us to do that easily, but doesn't worth adding more complexity to cover an edge case.

Inwerpsel:
Makes sense :+1: Also the plugin does have a browser prompt which makes it pretty hard to switch by accident.

PR #1625, opened 2022-02-16
Fix navbar search position in large screens

mleray:
### Description

It should adapt to the bigger height of the navbar for screen widths between 992px and 1200px.

### Testing

You can see the fixed version on the [sinope test instance](https://www-dev.greenpeace.org/test-sinope/) or on local. You can see the broken version on the [GPI site](https://www.greenpeace.org/international/) for example!


lithrel (approved):
:+1:

PR #1626, opened 2022-02-16
PLANET-6579: Allow a drop down menu for the Donate button through the Navigation Bar

GP-Dan-Tovbein:
Ref: https://jira.greenpeace.org/browse/PLANET-6579

👉 [Parent PR](https://github.com/greenpeace/planet4-master-theme/pull/1628)

**How to test**
- First, check if the Donate menu has items already added going to `Appearence > Menus`
- Navigate to the [instance](https://www-dev.greenpeace.org/test-rhea/) and play around with the Donate menu.

Also go to:
`Planet4 > Navigation` and change the **Website Navigation Style** to `dark|light`

**What to test**
- Validate the Design iteration2_IA & nav MVP from [here](https://www.figma.com/file/Gn8wyqgto7608rlESePXZE/Foundations-%26-Components?node-id=0%3A103).
- Analytics vars set to the sub-menu items.

### DIFF
--- a/assets/src/js/menu_editor.js
+++ b/assets/src/js/menu_editor.js
@@ -157,7 +157,7 @@ const menuEditorRestrictions = () => {
    * @return string The current location slug
    */
   const getCurrentLocation = () => {
-    return locations.map(location => {
+    return locations.filter(location => {
       return document.querySelector(`input[name="menu-locations[${location}]"]`).checked
         ? location : null;
     })[0] || null;
--- a/assets/src/scss/layout/_navbar.scss
+++ b/assets/src/scss/layout/_navbar.scss
@@ -71,10 +71,27 @@ $navbar-default-height: 60px;
 
     height: 44px;
     line-height: 40px;
+
+    @include large-and-up {
+      &.dropdown {
+        &::after {
+          content: "";
+          display: inline-block;
+          width: 10px;
+          height: 10px;
+          margin-inline-start: $sp-x;
+          mask: url("../../images/chevron.svg") 0 0/10px 10px;
+          transform: rotate(90deg);
+          transition: transform 150ms linear;
+          background-color: $white;
+          background-repeat: no-repeat;
+        }
+      }
+    }
   }
 
   .nav-donate {
-    padding: 16px 0;
+    padding: $sp-2 0;
 
     @include large-and-up {
       padding: 0;
@@ -196,7 +213,8 @@ $navbar-default-height: 60px;
     @include large-and-up {
       display: flex;
       flex-grow: 1;
-      justify-content: end;
+      align-items: center;
+      justify-content: flex-end;
       padding-inline-end: 0;
       padding-top: 0;
       text-align: end;
@@ -205,9 +223,11 @@ $navbar-default-height: 60px;
 }
 
 .nav-submenu {
+  _-- {
+    width: 193px;
+  }
   position: fixed;
   display: none;
-  min-width: 193px;
   padding-top: $sp-1;
 
   .nav-submenu-wrapper {
@@ -254,9 +274,7 @@ $navbar-default-height: 60px;
   li {
     width: 100%;
     text-align: left;
-    white-space: nowrap;
     color: $grey-80;
-    height: 48px;
 
     a.nav-link {
       --submenu-nav-link-- {
@@ -274,8 +292,9 @@ $navbar-default-height: 60px;
       height: 100%;
       display: flex;
       align-items: center;
-      padding-right: $sp-3;
-      padding-left: $sp-3;
+      padding: 14px $sp-3;
+      font-size: 16px;
+      line-height: 19px;
 
       &:hover {
         background: $grey-10;
@@ -290,13 +309,24 @@ $navbar-default-height: 60px;
 }
 
 .nav-item {
+  @include large-and-up {
+    margin-inline-start: $sp-2x;
+  }
+}
+
+.nav-donate:hover .dropdown::after {
+  transform: rotate(-90deg);
+}
+
+.nav-item,
+.nav-donate {
   color: inherit;
   font-size: $font-base-size;
   font-weight: 700;
 
   @include large-and-up {
     line-height: $menu-height;
-    margin: 0 20px;
+    margin-inline-end: $sp-2x;
 
     &:hover {
       .nav-submenu {
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -170,6 +170,7 @@ class MasterSite extends TimberSite {
 		register_nav_menus(
 			[
 				'navigation-bar-menu' => __( 'Navigation Bar Menu', 'planet4-master-theme-backend' ),
+				'donate-menu'         => __( 'Donate Button', 'planet4-master-theme-backend' ),
 			]
 		);
 
@@ -540,6 +541,23 @@ class MasterSite extends TimberSite {
 				return ! in_array( 'wpml-ls-item', $item->classes ?? [], true );
 			}
 		);
+
+		// Donate button menu dropdown.
+		// If no Donate menu is defined, we use the old settings from Planet 4 > Donate.
+		$donate_menu_items[] = [
+			'link'  => planet4_get_option( 'donate_button', '#' ),
+			'title' => planet4_get_option( 'donate_text', __( 'Donate', 'planet4-master-theme' ) ),
+		];
+
+		if ( has_nav_menu( 'donate-menu' ) ) {
+			$donate_menu = new TimberMenu( 'donate-menu' );
+
+			if ( ! empty( $donate_menu->get_items() ) ) {
+				$donate_menu_items = $donate_menu->get_items();
+			}
+		}
+
+		$context['donate_menu_items'] = $donate_menu_items;
 
 		$languages                 = function_exists( 'icl_get_languages' ) ? icl_get_languages() : [];
 		$context['site_languages'] = $languages;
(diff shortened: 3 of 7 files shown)


planet-4:

:watch: 2022.02.16 13:08:40

src/MasterSite.php
```diff
@@ -530,6 +531,23 @@ function ( $item ) {
 			}
 		);
 
+		// Donate button menu dropdown.
+		// If no Donate menu is defined, we use the old settings from Planet 4 > Donate.
+		$donate_menu_items[] = [
+			'link'  => planet4_get_option( 'donate_button' ) ?? '#',
```
mleray, line 537: I since found out that this function has a [second param for default value](https://github.com/greenpeace/planet4-master-theme/blob/master/functions.php#L73), so here instead of using `??` you could write `planet4_get_option( 'donate_button', '#' )`... Same on the following line 🙂

src/Settings.php
```diff
@@ -647,6 +651,16 @@ public function admin_page_display( string $plugin_page ): void {
 		);
 	}
 
+	/**
+	 * Show notice to inform users that the Donate button settings are now in Appearance > Menus.
+	 */
+	public function show_donate_menu_notice(): void {
```
mleray, line 657: I came up with the copy for this on my own when working on the ticket, but probably someone (maybe @comzeradd?) should review it 😬

mleray: Maybe we also need documentation for this 🤔 It is a bit complicated since you need to create a new menu etc

mleray: Actually just saw that there is [documentation](https://planet4.greenpeace.org/manage/admin-setup/#donate-button) for the Donate button already, that will probably need to be updated?

assets/src/scss/layout/_navbar.scss
```diff
@@ -69,12 +69,27 @@ $navbar-default-height: 60px;
       }
     }
 
-    height: 44px;
-    line-height: 40px;
+    height: $sp-5x;
```
mleray, line 72: I don't think we should use these variables for height or line-height, they are meant to be used for spacing values (so `margin`, `padding`, etc)

assets/src/scss/layout/_navbar.scss
```diff
@@ -290,13 +306,24 @@ $navbar-default-height: 60px;
...
+.nav-item,
+.nav-donate {
   color: inherit;
   font-size: $font-base-size;
   font-weight: 700;
 
   @include large-and-up {
     line-height: $menu-height;
-    margin: 0 20px;
+    margin-right: $sp-2x;
```
mleray, line 326: You probably want to use `margin-inline-end` here instead, so that it works for RTL sites as well

assets/src/scss/layout/_navbar.scss
GP-Dan-Tovbein: Good point!

src/Settings.php
```diff
@@ -654,6 +666,18 @@ public function admin_page_display( string $plugin_page ): void {
 		);
 	}
 
+	/**
+	 * Show notice to inform users that the Donate button settings are now in Appearance > Menus.
+	 *
+	 * @return string
+	 */
+	private function show_donate_menu_notice(): string {
+		return '<p>These settings have been moved to <a href="/wp-admin/nav-menus.php?action=locations">a new menu location</a>
```
mleray, line 675: I mentioned it in the previous PR, so I'll say it here again 😁 I came up with this copy when I first worked on this, it should probably be reviewed by @comzeradd or Suzi!

[Our documentation](https://planet4.greenpeace.org/manage/admin-setup/#donate-button) for this section will also likely need an update, and we should add somewhere some directions for how to add a Donate button dropdown menu since it's a bit complicated.

assets/src/scss/layout/_navbar.scss
```diff
@@ -69,12 +69,27 @@ $navbar-default-height: 60px;
       }
     }
 
-    height: 44px;
-    line-height: 40px;
+    height: $sp-5x;
+    line-height: $sp-5;
+
+    &.dropdown {
```
mleray, line 75: This chevron is currently also showing in the mobile side menu, I don't think it should?

Screenshot 2022-02-25 at 15 01 37

src/Settings.php
GP-Dan-Tovbein: I agree! I left this suggestion into the description of the [base PR](https://github.com/greenpeace/planet4-master-theme/pull/1628) just to follow up before merge into master.

templates/navigation-submenu.twig
```diff
@@ -0,0 +1,25 @@
+<nav class='nav-submenu'>
+	<div class='nav-submenu-wrapper'>
+		<ul>
+		{% for key,item in item.children %}
```
mleray, line 4: I think it would make more sense to call this parameter `menu` rather than `item` (so it would become `for key,item in menu.children`), the current notation is a bit confusing imo

GP-Dan-Tovbein: Completely agree!

templates/navigation-submenu.twig
```diff
@@ -0,0 +1,25 @@
...
+			<li>
+				{% if fn('get_post_meta', item.ID, '_menu_item_object_id', true ) == act_page_id %}
```
mleray, line 6: I think we can remove these special cases already since we're dropping the Act/Explore navigation, and just use `item.title` everywhere

templates/navigation-submenu.twig
GP-Dan-Tovbein: @mleray I was working on this suggestion (initially agreed by me) but if we change the name to `menu` instead `item` each link inside menu will take `menu.*` values but it seems to be a bit inconsistent that, since it's not a menu, it's an link (or item). What do you think?

templates/navigation-submenu.twig
GP-Dan-Tovbein: Shouldn't they behave as a link? And what about Donate links?

mleray: They will still behave as links 🤔 This is just for the Google Analytics action parameter, and if I understood correctly with the new IA we no longer treat Act and Explore as special pages so here we should use `item.title` everywhere... Anyway, I see that we still use this in the navigation bar template so I guess we can keep this change for later in the process

GP-Dan-Tovbein: Sorry I miss explained 😶 for links I referred to `data-*` values. I asked about these values a week ago to Julia, so, is this [answer](https://jira.greenpeace.org/browse/PLANET-6509?focusedCommentId=245076&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-245076) solve your doubt?

assets/src/scss/layout/_navbar.scss
```diff
@@ -196,6 +213,7 @@ $navbar-default-height: 60px;
     @include large-and-up {
       display: flex;
       flex-grow: 1;
+      align-items: center;
       justify-content: end;
```
mleray, line 217: Not really related to the Donate button menu but can be quickly solved, here it should be `flex-end` instead of `end` (which doesn't work in Safari)

mleray: Actually it looks like `end` is [not even a valid value](https://css-tricks.com/almanac/properties/j/justify-content/) for this property, it's lucky that it works in some browsers 😄

templates/navigation-submenu.twig
mleray: Well not really, it looks like she agrees with me? Julia says to use `item.title`, she doesn't mention exceptions for the Act and Explore pages... But this is not related to the Donate button so it can be solved in another PR 🙃

templates/navigation-bar-new.twig
```diff
@@ -42,46 +42,27 @@
...
+						href="{{ donate_menu.link }}"
 						data-ga-category="Menu Navigation"
 						data-ga-action="Donate"
 						data-ga-label="{{ page_category }}">
-							{{ donatetext }}
+							{{ donate_menu.title }}
 					</a>
-				</li>
+					{% if donate_menu.children is not empty %}
+						{% include 'navigation-submenu.twig' with { menu: donate_menu } only %}
```
mleray, line 61: Why do you use `only` here? 🤔 We don't use it anywhere else, including in the other [inclusion of this same template](https://github.com/greenpeace/planet4-master-theme/pull/1626/files#diff-3ad24a02d9e63bde53c8195a030484bfb200dac551677c03584386ba3394c92aR45)

assets/src/js/menu_editor.js
```diff
@@ -157,7 +157,7 @@ const menuEditorRestrictions = () => {
    * @return string The current location slug
    */
   const getCurrentLocation = () => {
-    return locations.map(location => {
+    return locations.filter(location => {
```
GP-Dan-Tovbein, line 160: @lithrel I had to change the `map` to `filter`. Bacause `map` will always returns a value and in consequence if you try to get the zero index of the returned array the value will be always `null` in cases where you have more than 1 item into the list.
i.e. Having this array: `['navigation-bar-menu', 'donate-menu']`, then:

If you want to check `navigation-bar-menu` the result will be:
```
['navigation-bar-menu', null] // Zero index = 'navigation-bar-menu'
```
If you want to check `donate-menu` the result will be:
```
[null, 'donate-menu'] // Zero index = null
```

So, it will never get the config of the `donate-menu`

lithrel: Yes I saw that a bit late, the fix was also part of the code I had left there https://github.com/greenpeace/planet4-master-theme/pull/1604#discussion_r810962793, works for me :+1:

mleray (approved):
Looks good to me 👍

PR #1630, opened 2022-02-21
PLANET-6513 Implement the new post tags

mleray:
### Description

See [PLANET-6513](https://jira.greenpeace.org/browse/PLANET-6513)
These new tags look more like buttons and are positioned at the end of the post content, before the author block

### Testing

You can check the new post tags on local in any post, or on the test instance also on any post, for example [this one](https://www-dev.greenpeace.org/test-phobos/story/51526/animations-we-loved-this-year-7-short-films-that-inspire-action/).

### DIFF
--- a/assets/src/scss/components/_buttons.scss
+++ b/assets/src/scss/components/_buttons.scss
@@ -261,3 +261,31 @@ button.load-more-mt {
     transition: opacity 400ms;
   }
 }
+
+.post-tag-button {
+  _-- {
+    border-color: $link-color;
+    color: $link-color;
+    background-color: $white;
+    font-weight: 500;
+
+    &:visited {
+      color: $link-color;
+    }
+
+    &:hover {
+      background-color: $link-color;
+      color: $white;
+    }
+
+    &:active {
+      background-color: $link-color;
+      color: $white;
+    }
+  }
+  margin-inline-end: $sp-2;
+  margin-bottom: $sp-2;
+  line-height: 2.3;
+  padding: 0 $sp-2;
+  transition-duration: 100ms;
+}
--- a/assets/src/scss/pages/post/_post.scss
+++ b/assets/src/scss/pages/post/_post.scss
@@ -2,19 +2,16 @@
 .single-attachment,
 .single-idea {
   @include large-and-up {
-    .post-content,
-    .author-block,
-    .page-header {
+    .narrow-container {
       margin-left: auto;
       margin-right: auto;
       max-width: 776px;
+      padding: 0;
     }
   }
 
   @include x-large-and-up {
-    .post-content,
-    .author-block,
-    .page-header {
+    .narrow-container {
       max-width: 736px;
     }
   }
@@ -91,5 +88,9 @@
     @include medium-and-up {
       padding-top: 24px;
     }
+  }
+
+  .post-tags {
+    margin-bottom: $sp-2;
   }
 }
--- a/templates/single-password.twig
+++ b/templates/single-password.twig
@@ -8,7 +8,7 @@
 			<div class="page-header-background"></div>
 		</header>
 
-		<div class="container">
+		<div class="container narrow-container">
 			<div class="post-content">
 				<form id="password-form" class="password-form" action="{{login_url}}?action=postpass&_wp_http_referer={{ post.link|url_encode }}" method="post">
 					<div class="mb-3">
--- a/templates/single.twig
+++ b/templates/single.twig
@@ -6,7 +6,7 @@
 
 	<article id="post-{{ post.ID }}">
 
-		<div class="container">
+		<div class="container narrow-container">
 			<header class="page-header">
 				<div class="top-page-tags">
 					{% if ( page_type ) %}
@@ -20,7 +20,7 @@
 						</a>
 					{% endif %}
 
-					{% if ( page_type and (post.issues_nav_data or post.tags) ) %}
+					{% if ( page_type and post.issues_nav_data ) %}
 						<span class="main-tag-chevron" aria-hidden="true"></span>
 					{% endif %}
 
@@ -83,7 +83,7 @@
 		</div>
 
 		<!-- Post Block Start -->
-		<div class="container">
+		<div class="container narrow-container">
 			<div class="post-content">
 				<article class="post-details clearfix">
 					{{ post.content|e('wp_kses_post')|raw }}
@@ -91,6 +91,20 @@
 						{{ fn('do_blocks', post.take_action_boxout )|raw }}
 					{% endif %}
 				</article>
+			</div>
+			<div class="post-tags">
+				{% if ( post.tags ) %}
+					{% for tag in post.tags %}
+						<a
+							class="btn post-tag-button"
+							href="{{ tag.link }}"
+							data-ga-category="Post"
+							data-ga-action="Navigation Tag"
+						>
+							<span aria-label="hashtag">#</span><span>{{ tag.name|e('wp_kses_post')|raw }}</span>
+						</a>
+					{% endfor %}
+				{% endif %}
 			</div>
 		</div>
 		<!-- Post Page Block End -->
--- a/templates/blocks/author_profile.twig
+++ b/templates/blocks/author_profile.twig
@@ -1,5 +1,5 @@
 {% if ( post.author.name and post.author.description ) %}
-	<div class="container clearfix">
+	<div class="container narrow-container clearfix">
 		<div class="author-block" itemscope itemtype="http://schema.org/Person">
 			{% if ( post.author.avatar ) %}
 				<figure class="author-block-image d-none d-sm-none d-md-block">
--- a/templates/single-attachment.twig
+++ b/templates/single-attachment.twig
@@ -6,7 +6,7 @@
 
 	<article id="post-{{ post.ID }}">
 
-		<div class="container">
+		<div class="container narrow-container">
 			<header class="page-header">
 				<h1 class="page-header-title">{{ post.title|raw }}</h1>
 				<div class="single-post-meta">
@@ -23,7 +23,7 @@
 		</div>
 
 		<!-- Attachment Post Block Start -->
-		<div class="container">
+		<div class="container narrow-container">
 			<div class="post-content">
 				<article class="post-details clearfix">
 					{{ post.content|e('wp_kses_post')|raw }}


planet-4:

:watch: 2022.02.21 15:59:00

assets/src/scss/components/_buttons.scss
```diff
@@ -261,3 +261,27 @@ button.load-more-mt {
...
 }
+
+.btn-post-tag {
+  --button-post-tag-- {
+    border-color: $link-color;
+    color: $link-color;
+    background-color: $white;
+    font-weight: 500;
+
+    &:visited {
```
Inwerpsel, line 272: I think we have other places where a visited state could apply not intentionally, since [the rule](https://github.com/greenpeace/planet4-master-theme/blob/af9998a39132464a872959ffa0dbf58dcdc16906/assets/src/scss/base/_typography.scss#L116-L118) applies to every `a` element on the page.

Perhaps we can make the "main" visited selector more specific. I.e. only apply it to links in the post content?
```css
.post-content a:visited {
    color: $link-color-visited;
}

File: assets/src/scss/components/_buttons.scss [hunk 2]
```diff
@@ -261,3 +261,27 @@ button.load-more-mt {
     transition: opacity 400ms;
   }
 }
+
+.btn-post-tag {
+  --button-post-tag-- {
```
Inwerpsel, line 266: Since this is a new class we're introducing, it could be nicer to just choose a class name that plays well as a variable prefix. To reduce the amount of variations of the name we need to deal with.

templates/single.twig
```diff
@@ -91,6 +91,19 @@
...
 					{% endif %}
 				</article>
+				{% if ( post.tags ) %}
+					{% for tag in post.tags %}
+						<a
+							class="btn btn-post-tag"
+							href="{{ tag.link }}"
+							data-ga-category="Post"
+							data-ga-action="Navigation Tag"
+							data-ga-label="n/a"
```
Inwerpsel, line 101: `data-ga-label` is [not required](https://www.hotjar.com/google-analytics/glossary/events/). We can probably exclude the data attribute instead of putting `n/a` everywhere.

But I didn't check if it's really not required by our GA setup, perhaps it somehow crashes while parsing the data attributes if the value is not there.

If that's the case, we should update the analytics container code to not require it. Or maybe make it add `n/a` by default in case the attribute is missing.

assets/src/scss/components/_buttons.scss
Inwerpsel: It seems overall safer to do, since most elements outside the post content have a different background anyway, and the main link visited color is really made for against a white background.

templates/single.twig
mleray: This should probably be done in a separate ticket, since we use this in a lot of other places, and we'd need to double-check with Julia and/or the new data analyst before making the changes 🤔

assets/src/scss/components/_buttons.scss
mleray: I'm not sure what you mean here by "plays well as a variable"?

templates/single.twig
Inwerpsel: You mean the part of updating the code in the GA container, right? Sure, that's not something I expect we can quickly change.

But we could already just confirm what happens if the `data-ga-label` is not present. I don't expect this to be a problem, in which case we can already exclude the useless `n/a` value.

assets/src/scss/components/_buttons.scss
mleray: In this particular case it would not solve the issue, since the post tags are in the `post-content` div... But yeah maybe it would be good for other buttons! Might be better in a separate ticket since it will need some testing?

assets/src/scss/components/_buttons.scss
Inwerpsel: I mean you could use `_--` so that it uses the class name which is already on the previous line. I assume you didn't use it because `btn-post-tag` doesn't look nice as a variable name?

mleray: Oh, no, I think it's fine but other button classes used a separate variable prefix so I did the same 😅 But we could use `_--`

assets/src/scss/components/_buttons.scss
Inwerpsel: Do the post tags need to be inside the `post-content` div? `post_content` is a specific field on a post that contains what you see in the editor.

mleray: If we move them outside of the `post-content` div then they are no longer correctly aligned :/ Of course that could be solved by adding some duplicated CSS to them but that's not ideal

templates/single.twig
```diff
@@ -91,6 +91,19 @@
 						{{ fn('do_blocks', post.take_action_boxout )|raw }}
 					{% endif %}
 				</article>
+				{% if ( post.tags ) %}
+					{% for tag in post.tags %}
```
Inwerpsel, line 95: These tags do not have a wrapper element, which probably makes alignment harder.
Screenshot from 2022-02-28 12-52-32

assets/src/scss/components/_buttons.scss
Inwerpsel: IMHO that makes sense to do anyway. I don't think we would need to duplicate a lot (if any) CSS to do that.

assets/src/scss/components/_buttons.scss
mleray: [Updated](https://github.com/greenpeace/planet4-master-theme/pull/1630/commits/f35df4cbed49cacfe5f7f962aaf55b65823a03ab)!

assets/src/scss/components/_buttons.scss
mleray: [Done](https://github.com/greenpeace/planet4-master-theme/pull/1630/commits/3e9d157f4bde38957469f9337230cd088392a678)

templates/single.twig
mleray: Well I meant more the part about updating all the places where we use this `n/a` value, I don't think we should do this as part of this PR

Inwerpsel: No indeed, I just meant to exclude it from then new code if we find it doesn't need to be there.

assets/src/scss/components/_buttons.scss
mleray: However I'm not sure about using
```
.post-content a:visited {
    color: $link-color-visited;
}
```
since links could also be in blocks for example and there they would need the visited state too 🤔

templates/single.twig
mleray: [Removed](https://github.com/greenpeace/planet4-master-theme/pull/1630/commits/ddf32c299e0ef4104034f993e22140546c09fb0c)

assets/src/scss/components/_buttons.scss
Inwerpsel: Pages also use this class, though interestingly Campaign Pages don't. `post-content` here is not referring to the "post" post type. It's called like that because the content field is called `post_content` on posts of all types.

Blocks are by definition inside of the post content, so that shouldn't be a problem.

Inwerpsel: Right, I was put on the wrong track as I assumed this class referred to the `post_content` field which is present on posts of all types. But I see for pages it uses `page-content`. So we currently don't have a consistent class around the editor content across post types.

We could of course add one, or just target both in the CSS selector.

As for blocks, those should by definition be inside of the post content.

Inwerpsel: Anyway, it was just a suggestion to see if it could make our live easier, seems a little more complex than I initially hoped. I'm ok with the current code and getting back to this later.

Houssam-saleh:
I agree with you @mleray & @Inwerpsel to use `0.1s` transition for the tags buttons, since we will have multiple tags on the same row and it would be great to hover on them quickly.

Inwerpsel (commented):
Just noticed I still had this comment. I'll go through the rest and probably approve :wink:

assets/src/scss/components/_buttons.scss
```diff
@@ -261,3 +261,30 @@ button.load-more-mt {
...
+      background-color: $link-color;
+      color: $white;
+    }
+
+    &:active {
+      background-color: $link-color;
+      color: $white;
+    }
+  }
+  margin-inline-end: $sp-2;
```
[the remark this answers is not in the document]
mleray: I don't think it's necessary since buttons have `display: inline-block;` 🤔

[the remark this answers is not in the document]
Inwerpsel: No not needed. Just seems more accurate to express this as a `gap`, rather than a `margin-inline-end`. Just a suggestion btw.

templates/single.twig
```diff
@@ -84,14 +84,28 @@
 
 		<!-- Post Block Start -->
 		<div class="container">
-			<div class="post-content">
+			<div class="post-content narrow-container">
```
Inwerpsel, line 87: Perhaps it makes more sense to replace the outer `container` `div` with this class?

mleray: That would not work on mobile

Inwerpsel: I see. Would using both classes work?
```
<div class="container narrow-container">

Comment (mleray) [reply templates/single.twig:87]: That results in a small padding difference, which messes up the alignment 😕

Comment (Inwerpsel) [reply templates/single.twig:87]: Can we fix that with further changes? Or would it be too many changes?

Comment (mleray) [reply templates/single.twig:87]: I would rather address this later on, it's already a few unrelated changes with the addition of the `narrow-container` class and they impact quite a lot of pages

Comment (Inwerpsel) [reply templates/single.twig:87]: I don't see how these changes are "unrelated"... We're trying to build something that makes sense, and putting an extra container class, and then nesting it inside of the existing one doesn't really make sense to me.

Comment (mleray) [reply templates/single.twig:87]: "Unrelated" in the sense that it wasn't necessary for the ticket itself and it could have been done separately... The changes we made so far already simplify the code, but as I said they will impact a lot of pages so I'm a bit hesitant to keep changing things right now

Comment (mleray) [reply templates/single.twig]: [Updated](https://github.com/greenpeace/planet4-master-theme/pull/1630/commits/228ee5e710b7a4d3b6aaad7003b54d838b7d98d4)!

File: templates/single-password.twig [hunk 7]
```diff
@@ -4,11 +4,13 @@
 
 	<article id="post-{{ post.ID }}">
 
-		<header class="page-header">
-			<div class="page-header-background"></div>
-		</header>
+		<div class="container narrow-container">
+			<header class="page-header">
+				<div class="page-header-background"></div>
```
Inwerpsel, line 9: Do we need to put this empty markup here? :thinking:

mleray: Removed the container div [there](https://github.com/greenpeace/planet4-master-theme/pull/1630/commits/476299d44b3ff60cc62036694ad564ee93bc286b)

assets/src/scss/pages/post/_post.scss
```diff
@@ -2,19 +2,16 @@
 .single-attachment,
 .single-idea {
   @include large-and-up {
-    .post-content,
-    .author-block,
-    .page-header {
+    .narrow-container {
```
Inwerpsel, line 5: Now that the class is only present where we include it, we can move it outside of the parent selector.

Inwerpsel (approved):
Looks good, just a suggestion for `.narrow-container` but we can do it afterwards too.

PR #1631, opened 2022-02-21
PLANET-6645 Add paginated list on tag pages

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6645

As a first step of using paginated lists, the new list is only added to tag pages.

Define query loop block in HTML and use it to output almost the same markup as our Articles block.

We can gradually rewrite the CSS to make more sense with the core block, but using Article's existing CSS classes makes ensuring the same layout much easier.

I used a few **feature toggles** to aid development (just a few small if-statements really). We can probably remove them before merging.
* [Toggle old server generated pages / paginated list](https://github.com/greenpeace/planet4-master-theme/blob/e1cbd068bb0e5d31cef28f531d1d71885961a7c0/tag.php#L70-L70) (could be useful for releasing)
* [Toggle Grid / list](https://github.com/greenpeace/planet4-master-theme/blob/e1cbd068bb0e5d31cef28f531d1d71885961a7c0/tag.php#L71-L71)
* [Turn off all "redirect pages"](https://github.com/greenpeace/planet4-master-theme/blob/244993165c943fa60042897d7dd0c5410701839a/tag.php#L27-L27) to avoid needing to do it manually on each tag (maybe also useful for pilot NROs to test this functionality, they have a redirect page on every tag)

### Approach
It uses the new templates on the **server side only**.

I found that **we don't need to add block HTML comments** for additional markup, as long as it's only server side. This makes the markup a lot more manageable. Only the [outer Query block](https://github.com/greenpeace/planet4-master-theme/blob/be2366211ce6d58c85a69289a70f96289b189877/parts/query-list.html#L1-L2) has both attributes and wrapper markup.

To be a completely "correct" block template we'd need to do this for all HTML elements you see in the file. Each class of these elements needs to occur in both the attributes and the markup. And besides `div` all tag names need to be passed in the Group block's `tagName` attribute. Not very maintenance friendly...

But we only need that once we want to load these templates in the block editor. At that point we'll be able to edit the templates using the block editor and save the result to a file, making the extra code complexity a non issue. So we can add them at the point we start using the template editor.

Server side only blocks are very easy to add and can easily reuse existing server side logic. They only need an array of config and a function. It's like React components on the server side, but they render just render once.

### Examples (Mars env, data from GPI, tag redirect pages disabled)
https://www-dev.greenpeace.org/test-mars/tag/climate/
https://www-dev.greenpeace.org/test-mars/tag/consumption/
https://www-dev.greenpeace.org/test-mars/tag/oceans/

### TODO

- [x] Responsive image:
  - [x] [Create custom block to output same markup as Articles image](https://github.com/greenpeace/planet4-master-theme/pull/1631/commits/be2366211ce6d58c85a69289a70f96289b189877)
  - [x] Generalize block to support different sizes for grid view (follow up)
- [ ] Cloudflare cache purges: Ensure it purges all pages at the right time. Investigate amount of purges.
  - [x] Go through existing CF plugin purge logic
  - [x] Check if purges happen reliably (I've seen it not happen a few times, but then it does again)
  - [ ] Ensure purges for all page numbers
- [x] Other archive pages (follow up for other types, but already tested current approach there)
- [ ] Which feature toggles?
- [ ] Which options pages?
- [ ] Release steps

### Small fixes
- [x] Excerpt style: Core block has inner paragraph preventing Articles style. Either adapt CSS or create our own block with same markup as Articles

### Images
There is a core ticket to add block attributes and maybe additional context to `wp_calculate_image_sizes`. Unfortunately there's no recent development. But potentially we can use a filter to achieve this behavior already.

### DIFF
--- a/assets/src/scss/layout/_block-classes.scss
+++ b/assets/src/scss/layout/_block-classes.scss
@@ -1,0 +1,50 @@
+.wp-block-query-pagination {
+  a {
+    color: black;
+    font-weight: bold;
+  }
+
+  font-family: $roboto;
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+  width: 100%;
+  margin-bottom: 16px;
+}
+
+.page-numbers {
+  &:not(:last-child) {
+    margin-inline-end: 8px;
+  }
+
+  &.current {
+    border: 1px solid black;
+    border-radius: 4px;
+    padding: 8px 12px;
+  }
+}
+
+.wp-block-post-terms {
+  font-weight: 500;
+}
+
+.taxonomy-post_tag a {
+  display: inline-block;
+
+  &:before {
+    content: "#";
+    speak: none;
+  }
+}
+
+.wp-block-post {
+  margin-bottom: 32px;
+
+  figure {
+    margin-bottom: 0;
+  }
+}
+
+.wp-block-post-excerpt__excerpt {
+  font-size: 1rem;
+}
--- a/functions.php
+++ b/functions.php
@@ -79,6 +79,7 @@ function planet4_get_option( $key = '', $default = null ) {
 use P4\MasterTheme\ImageArchive\Rest;
 use P4\MasterTheme\Loader;
 use P4\MasterTheme\Notifications\Slack;
+use P4\MasterTheme\Post;
 use Timber\Timber;
 
 if ( ! class_exists( 'Timber' ) ) {
@@ -249,3 +250,62 @@ add_filter(
 	10,
 	2
 );
+
+/**
+ * I'll move this somewhere else in master theme.
+ *
+ * @return void
+ */
+function register_more_blocks() {
+	register_block_type(
+		'p4/reading-time',
+		[
+			'render_callback' => [ Post::class, 'reading_time_block' ],
+			'uses_context'    => [ 'postId' ],
+		]
+	);
+	register_block_type(
+		'p4/post-author-name',
+		[
+			'render_callback' => function ( array $attributes, $content, $block ) {
+				$author_override = get_post_meta( $block->context['postId'], 'p4_author_override', true );
+				$post_author_id  = get_post_field( 'post_author', $block->context['postId'] );
+
+				$is_override = ! empty( $author_override );
+
+				$name = $is_override ? $author_override : get_the_author_meta( 'display_name', $post_author_id );
+				$link = $is_override ? '#' : get_author_posts_url( $post_author_id );
+
+				$block_content = $author_override ? $name : "<a href='$link'>$name</a>";
+
+				return "<span class='article-list-item-author'>$block_content</span>";
+			},
+			'uses_context'    => [ 'postId' ],
+		]
+	);
+	// Like the core block but with an appropriate sizes attribute.
+	register_block_type(
+		'p4/post-featured-image',
+		[
+			'render_callback' => function ( array $attributes, $content, $block ) {
+				$post_id        = $block->context['postId'];
+				$post_link      = get_permalink( $post_id );
+				$featured_image = get_the_post_thumbnail(
+					$post_id,
+					null,
+					// For now hard coded sizes to the ones from Articles, as it's the single usage.
+					// This can be made a block attribute, or even construct a sizes attr with math based on context.
+					// For example, it could already access displayLayout from Query block to know how many columns are
+					// being rendered. If it then also knows the flex gap and container width, it should have all needed
+					// info to support a large amount of cases.
+					[ 'sizes' => '(min-width: 1600px) 389px, (min-width: 1200px) 335px, (min-width: 1000px) 281px, (min-width: 780px) 209px, (min-width: 580px) 516px, calc(100vw - 24px)' ]
+				);
+
+				return "<a href='$post_link'>$featured_image</a>";
+			},
+			'uses_context'    => [ 'postId' ],
+		]
+	);
+}
+
+add_action( 'init', 'register_more_blocks' );
--- a/functions.php
+++ b/functions.php
@@ -76,6 +76,7 @@ function planet4_get_option( $key = '', $default = null ) {
 	return $options[ $key ] ?? $default;
 }
 
+use P4\MasterTheme\Features\Dev\ListingPagePagination;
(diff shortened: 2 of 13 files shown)


planet-4:

:watch: 2022.02.21 12:02:06

__block-templates/query-grid.html
```diff
@@ -0,0 +1,35 @@
+<!-- wp:query { "query":{ "inherit": true }, "displayLayout": { "type": "flex", "columns":4 } } -->
```
Inwerpsel, line 1: I preserved this grid template as it was already in the POC. For now the code will use only the list view.

src/Post.php
```diff
@@ -498,6 +499,21 @@ public function reading_time(): ?int {
...
+	/**
+	 * Server side render for the reading time block.
+	 *
+	 * @param $attributes array Block attributes, unused.
+	 * @param $content string Content which apparently no core block uses.
+	 * @param $block WP_Block With all block properties.
+	 *
+	 * @return string Formatted reading time.
+	 */
+	public static function reading_time_block( array $attributes, $content, $block ) {
```
Inwerpsel, line 511: I don't trust WordPress enough to put `WP_Block` as a type here :grimacing:

functions.php
```diff
@@ -210,3 +211,23 @@ function ( $hidden ) {
 	10,
 	1
 );
+
+// Might move later.
+function register_more_blocks() {
+	register_block_type( 'p4/reading-time', [
```
Inwerpsel, line 217: Since it's easy to add many back end only blocks, it's a good time to reconsider our block namespace. `planet4-blocks` is very long and the suffix `-blocks` is redundant. Using `p4/` instead saves us a lot of typing and long lines of code, and it's already a commonly known abbreviation in the project.

parts/query-grid.html
```diff
@@ -0,0 +1,34 @@
+<!-- wp:query { "query":{ "inherit": true }, "displayLayout": { "type": "flex", "columns":4 } } -->
```
Inwerpsel, line 1: I preserved this grid template as it was already in the POC. For now the code will use only the list view.

Inwerpsel: I now included a feature toggle to load this template instead.

tag.php
```diff
@@ -34,6 +34,12 @@
 		include 'page.php';
 
 	} else {
+		$view = \P4\MasterTheme\Features::is_active(\P4\MasterTheme\Features::LISTING_PAGE_GRID_VIEW) ? 'grid' : 'list';
+		$query_template = file_get_contents( get_template_directory() . "/parts/query-$view.html" );
+
+		$content = do_blocks( $query_template );
+
+		$context['query_loop'] = $content;
```
Inwerpsel, line 43: I assume the previous content of the page is no longer relevant, I'll remove it later to simplify the changes in the PR for now.

parts/query-grid.html
Inwerpsel: For now I'm focusing on the list view template, so this one is a bit behind.

assets/src/scss/layout/_block-classes.scss
```diff
@@ -0,0 +1,45 @@
+.wp-block-query-pagination {
```
Inwerpsel, line 1: For now it's the only style, so I added it directly. In the future we could turn this into a block style, once we use the block in the editor. Or when we add another back end generated pagination style.

tag.php
```diff
@@ -64,8 +66,24 @@
...
+			$query_template = file_get_contents( get_template_directory() . "/parts/query-$view.html" );
+
+			$content = do_blocks( $query_template );
+
+			$context['query_loop'] = $content;
+		}
+
 		$campaign = new TaxonomyCampaign( $templates, $context );
 
+		if ( ListingPagePagination::is_active() ) {
```
Inwerpsel, line 81: For now I just made it render with the mystic `TaxonomyCampaign` class. We can remove this class once we're sure we can remove the old path.

tag.php
Inwerpsel: I for now added a feature toggle instead of removing the code.

I did this because I realized that the previous auto generated page has different post types, and the paginated list for now only contains posts.

Another reason was to unblock this PR. The CF changes are still in progress but don't have to block merging everything else because of the feature toggle.

assets/src/scss/layout/_block-classes.scss
```diff
@@ -0,0 +1,45 @@
...
+    border-radius: 4px;
+    padding: 8px 12px;
+  }
+}
+
+.wp-block-post-terms {
+  font-weight: 500;
+}
+
+.taxonomy-post_tag a:before {
```
mleray, line 30: This doesn't seem to work as expected on RTL sites:
Screenshot 2022-04-12 at 16 29 31

Inwerpsel: Nice catch, I fixed it with `display: inline-block`, but still don't see why it would do something differently for 1 of the tags :exploding_head: afaik they have the same CSS.

functions.php
```diff
@@ -249,3 +250,127 @@ function ( array $editor_settings, WP_Block_Editor_Context $block_editor_context
...
+			'render_callback' => function ( array $attributes, $content, $block ) {
+				$author_override = get_post_meta( $block->context['postId'], 'p4_author_override', true );
+				$post_author_id  = get_post_field( 'post_author', $block->context['postId'] );
+
+				$is_override = ! empty( $author_override );
+
+				$name = $is_override ? $author_override : get_the_author_meta( 'display_name', $post_author_id );
+				$link = $is_override ? '#' : get_author_posts_url( $post_author_id );
+
+				$tag_content = $author_override ? $name : "<a href='$link'>$name</a>";
```
mleray, line 279: Shouldn't this variable be called `$author_name` or something similar, rather than `$tag_content`? 🤔

Inwerpsel: No, it's intentional. I mean "tag content" in the sense that the HTML tag's content will be either a link, or just text.

Inwerpsel: Changed it to `$block_content`

mleray (approved):
Tested on the instance, looks good to me 👍

PR #1632, opened 2022-02-21
PLANET-6651 Increase padding top in posts when mobile tabs are active

mleray:
### Description

See [PLANET-6651](https://jira.greenpeace.org/browse/PLANET-6651)
This fix had been implemented for pages already, but not for posts


planet-4:

:watch: 2022.02.21 15:13:30

lithrel (approved):
:+1: 
I think a selector without attribute like `.single-post .with-mobile-tabs ~ .page-header` would also work, but I'm actually not sure it's specific to posts only

mleray:
> 👍 I think a selector without attribute like `.single-post .with-mobile-tabs ~ .page-header` would also work, but I'm actually not sure it's specific to posts only

I think then it wouldn't work for attachment and idea pages that use respectively `single-attachment` and `single-idea` classes... I could however add it to [this file](https://github.com/greenpeace/planet4-master-theme/blob/master/assets/src/scss/pages/post/_post.scss) I guess, do you think that would be a better approach? Maybe it would be more clear 🤔

lithrel:
I honestly have no idea what would be a best approach :grimacing:  I know I added this one too for page content https://github.com/greenpeace/planet4-master-theme/blob/master/assets/src/scss/pages/_page.scss#L53 , so that's multiple places for a header/content top padding, but the intent and scope of each file is not always clear to me

assets/src/scss/pages/_404.scss
```diff
@@ -1,12 +1,6 @@
 .page-404-page {
   .page-header {
-    color: $grey-80;
```
mleray, line 3: This was inherited from `_page-header.scss`

assets/src/scss/pages/_404.scss
```diff
@@ -1,12 +1,6 @@
...
-    min-height: 1200px;
-    padding-top: 80px;
-
-    @include medium-and-up {
-      padding-top: 110px;
-    }
+    min-height: 100vh;
```
mleray, line 3: This looks better on mobile than before, where `1200px` was really long (also seems to be a random value)

lithrel (approved):
:+1:

PR #1636, opened 2022-03-01
PLANET-6656 Hide Follow us link when no social media links

mleray:
### Description

See [PLANET-6656](https://jira.greenpeace.org/browse/PLANET-6656)
In the backend it is possible to remove all social media links, in this case we should completely hide the social media part of the footer including the Follow us text.

### Testing

I've removed all social media links from the [deimos test instance](https://www-dev.greenpeace.org/test-deimos/), with the fix you can see that in the footer the `Follow us` link no longer shows! Before the fix it shows like this:
Screenshot 2022-03-01 at 14 33 24

### DIFF
--- a/assets/src/scss/layout/_footer-social-media.scss
+++ b/assets/src/scss/layout/_footer-social-media.scss
@@ -13,6 +13,7 @@
       margin-bottom: 24px;
       display: block;
       line-height: 1;
+      white-space: nowrap;
     }
   }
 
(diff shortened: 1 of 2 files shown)


assets/src/scss/layout/_footer-social-media.scss
```diff
@@ -13,6 +13,7 @@
       margin-bottom: 24px;
       display: block;
       line-height: 1;
+      white-space: nowrap;
```
mleray, line 16: I added this line because otherwise when there are less than 3 social media icons the text `Follow us` splits into 2 lines which looks ugly 😬

comzeradd (approved):

PR #1637, opened 2022-03-02
PLANET-6647: Youtube Embed - block not showing in the editor

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-6647

> Videos added using the Youtube embed block disappear in the editor.
> The block is still there, but has a height of 0 in Gutenberg as a preview. Videos do show up in the frontend, so it's only the preview and editing that doesn't work.

Gutenberg editor is forcing an iframe for any embed, which doesn't allow for lite-youtube styles to be applied.
Cf. Gutenberg [embed preview component](https://github.com/WordPress/gutenberg/blob/aa05bd47df3247730065b197a170ab84e4c6761f/packages/block-library/src/embed/embed-preview.js#L96) and [sandbox component](https://github.com/WordPress/gutenberg/blob/d5915916abc45e6682f4bdb70888aa41e98aa395/packages/components/src/sandbox/index.js#L94) 

Removing lite-youtube integration in the editor fixes the issue, while keeping its integration on the frontend.

## Test
:warning: Pay attention to cache, **use a different youtube URLs for each test**. :warning: In details, there are 2 caches you can hit:
- embed cache in wp_postmeta table, it should not contain any reference to lite-youtube, but you can also empty it with `docker-compose exec php-fpm wp embed cache clear {postID}`
- rest api cache (a call to `https://www.planet4.test/wp-json/oembed/1.0/proxy?url={youtubeurl}` used by the editor to generate the preview), might return the same response to the same url call even after code change  and mutliple refresh

To test:
- Add a youtube embed in a post
- The video should show in the editor
  - Control with your browser inspector that the iframe contains the youtube iframe and not a lite-youtube tag
- The video should show on the frontend (preview or published)
  - Control with your browser inspector that a `<lite-youtube>` tag is used

### DIFF
--- a/assets/src/scss/editorStyle.scss
+++ b/assets/src/scss/editorStyle.scss
@@ -3,7 +3,6 @@
 @import "~bootstrap/scss/carousel";
 @import "~bootstrap/scss/tooltip";
 @import "~bootstrap/scss/buttons";
-@import "~lite-youtube-embed";
 
 @import "base/variables";
 @import "base/colors";
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -812,13 +812,6 @@ class MasterSite extends TimberSite {
 	 */
 	public function enqueue_editor_assets(): void {
 		Loader::enqueue_versioned_style( 'assets/build/editorStyle.min.css' );
-		wp_enqueue_script(
-			'youtube',
-			get_template_directory_uri() . '/assets/build/lite-yt-embed.js',
-			[],
-			1,
-			true
-		);
 	}
 
 	/**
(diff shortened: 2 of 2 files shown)


planet-4:

:watch: 2022.03.02 15:54:46

mleray (approved):

lithrel:
> However, will that solve the issue for already existing blocks? The complaint in the ticket was the impossibility to edit them confused

If I understand the code correctly (which is a big if), the resulting embed is [kept in cache for a day](https://github.com/WordPress/wordpress-develop/blob/b28dcfd83243459e9cc08f46f70418ebfde9dde8/src/wp-includes/class-wp-oembed-controller.php#L239-L241) - and I believe we flush the cache on every deployment, so if I check on Atlas test instance ([see youtube blocks list](https://www-dev.greenpeace.org/test-atlas/wp-admin/admin.php?page=plugin_blocks_report_beta&s=youtube)) youtube blocks seems to be working in the editor (apart from urls linking to playlists, considered invalid urls on the backend, not sure why :thinking: ).
If this doesn't work as expected I can write a patch to clear those caches, but I think it should be fine

PR #1640, opened 2022-03-08
PLANET-6619 Ensure "redirect page" has no duplicate URL by redirecting to the tag URL it's overriding

mleray:
https://jira.greenpeace.org/browse/PLANET-6619

### DIFF
--- a/page.php
+++ b/page.php
@@ -28,6 +28,7 @@
  */
 
 use P4\MasterTheme\Context;
+use P4\MasterTheme\Features\RedirectRedirectPages;
 use P4\MasterTheme\Post;
 use Timber\Timber;
 
@@ -35,6 +36,30 @@ $context        = Timber::get_context();
 $post           = new Post(); // phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
 $page_meta_data = get_post_meta( $post->ID );
 $page_meta_data = array_map( 'reset', $page_meta_data );
+
+// Ensure redirect is only performed if we're not already on a tag URL. Because tag.php includes this file.
+if ( ! is_tag() && RedirectRedirectPages::is_active() ) {
+	$args  = [
+		// Ensure the term is returned even if no posts are tagged with it.
+		'hide_empty' => false,
+		'meta_query' => [
+			[
+				'key'     => 'redirect_page',
+				'value'   => $post->ID,
+				'compare' => '=',
+			],
+		],
+		'fields'     => 'ids',
+	];
+	$terms = get_terms( $args );
+
+	if ( ! empty( $terms ) ) {
+		$term_link = get_term_link( $terms[0] );
+
+		wp_safe_redirect( $term_link, 301 );
+		exit();
+	}
+}
 
 // Set Navigation Issues links.
 $post->set_issues_links();
--- a/src/Features/RedirectRedirectPages.php
+++ b/src/Features/RedirectRedirectPages.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace P4\MasterTheme\Features;
+
+use P4\MasterTheme\Feature;
+
+/**
+ * @see description().
+ */
+class RedirectRedirectPages extends Feature {
+
+	/**
+	 * @inheritDoc
+	 */
+	public static function id(): string {
+		return 'redirect_redirect_pages';
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	protected static function name(): string {
+		return __( 'Redirect Tags redirect pages', 'planet4-master-theme-backend' );
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	protected static function description(): string {
+		return __(
+			'Do a HTTP Redirect from Tags redirect pages to the tag URL, to avoid duplicate content.',
+			'planet4-master-theme-backend'
+		);
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public static function show_toggle_production(): bool {
+		return true;
+	}
+}
--- a/src/Settings/Features.php
+++ b/src/Settings/Features.php
@@ -22,6 +22,7 @@ use P4\MasterTheme\Features\NewDesignCountrySelector;
 use P4\MasterTheme\Features\NewDesignNavigationBar;
 use P4\MasterTheme\Features\PurgeOnFeatureChanges;
 use P4\MasterTheme\Features\PostPageCategoryLinks;
+use P4\MasterTheme\Features\RedirectRedirectPages;
 use P4\MasterTheme\Loader;
 use P4\MasterTheme\Settings;
 
@@ -114,6 +115,7 @@ class Features {
 			NewDesignCountrySelector::class,
 			NewDesignNavigationBar::class,
 			GoogleSheetReplacesSmartsheet::class,
+			RedirectRedirectPages::class,
 
 			// Dev only.
 			DisableDataSync::class,


src/Campaigns.php
```diff
@@ -271,10 +271,18 @@ public function save_taxonomy_meta( $term_id ) {
 		}
 
 		$redirect_page = filter_input( INPUT_POST, 'redirect_page', FILTER_VALIDATE_INT ) ?? 0;
+
 		if ( $redirect_page ) {
 			update_term_meta( $term_id, 'redirect_page', $redirect_page );
+			update_post_meta( $redirect_page, 'redirect_tag_page', $term_id );
```
Inwerpsel, line 277: We don't need to add the same information to the post's meta. Instead we should query the tag meta for a tag that has this post set.

page-templates/evergreen.php
```diff
@@ -26,6 +26,11 @@
 $page_meta_data = get_post_meta( $post->ID );
 $page_meta_data = array_map( 'reset', $page_meta_data );
 
+$redirect_tag_id = get_post_meta( $post->ID, 'redirect_tag_page', true );
```
Inwerpsel, line 29: I'm still checking if there's a function for this. In the end we want it to execute the following query:
```sql
select * from wp_termmeta where meta_key='redirect_page' and meta_value={{post_id}}

File: page.php [hunk 3]
```diff
@@ -36,6 +36,19 @@
 $page_meta_data = get_post_meta( $post->ID );
 $page_meta_data = array_map( 'reset', $page_meta_data );
 
+$redirect_tag_id = get_post_meta( $post->ID, 'redirect_tag_page', true );
+
+if ( $redirect_tag_id ) {
+	$redirect_tag_page = get_term( $redirect_tag_id );
+	$redirect_tag_url  = home_url( '/tag/' . $redirect_tag_page->slug );
```
Inwerpsel, line 43: This needs to use `get_tag_link` which you can just pass the tag ID.

page.php
```diff
@@ -36,6 +36,22 @@
 $page_meta_data = get_post_meta( $post->ID );
 $page_meta_data = array_map( 'reset', $page_meta_data );
 
+if ( ! is_tag() ) {
```
Inwerpsel, line 39: This works because `is_tag` still returns true.

The fact it does is not as straightforward as it seems. In `tag.php` we [re-assign the "queried object"](https://github.com/greenpeace/planet4-master-theme/blob/f79fda3337458efdc60a52c05bb8b927c22d93c0/tag.php#L26-L29).

This results in WordPress considering the page as being the object that is rendered (e.g. the admin UI will point to this page instead of the tag). But doing so doesn't re-assign `$wp-query->is_tag`, so our check still works.

planet-4:

:watch: 2022.03.08 14:20:15

src/Campaigns.php
mleray: Would that be really be better? It feels more efficient to set it here rather than query all tags in all pages 🤔

Inwerpsel: Can you elaborate what you mean with "more efficient"?

mleray: It feels like it would be better performance-wise to just check if that meta attribute is set, rather than loop through all available tags

Inwerpsel: In the end both approaches lead to a SQL query on a table.

Using post meta:
```SQL
select meta_value from wp_postmeta where meta_key='redirect_tag_page' and post_id={{post_id}}
```

Using tag meta:
```SQL
select term_id from wp_termmeta where meta_key='redirect_page' and meta_value={{post_id}}
```

Inwerpsel: It's a query with similar complexity. Even if there's a performance difference, it should be small, as neither query will ever take long.

Keep in mind that our pages are also cached in Cloudflare, and this includes redirects. So even if the difference was significant, it would only affect a tiny portion of page visits.

mleray: Alright, thanks for clarifying! 😄

Inwerpsel: There's probably a way after all to use a function instead of raw SQL.
https://wordpress.stackexchange.com/questions/231212/get-term-by-custom-term-meta-and-taxonomy

I checked the code and WordPress should fetch the result from cache before using SQL.

https://github.com/WordPress/wordpress-develop/blob/trunk/src%2Fwp-includes%2Fclass-wp-term-query.php#L745-L750

Inwerpsel:
I added the DO NOT MERGE label because we need to validate this redirect is ok to do on all sites.

If it's not ok, we may need to feature toggle it.

Inwerpsel:
> I added the DO NOT MERGE label because we need to validate this redirect is ok to do on all sites.
> 
> If it's not ok, we may need to feature toggle it.

I ended up adding a feature toggle because it's easy to add in this case, and we found some NROs are using a similar fix but then redirect to the page URL instead of the tag URL. With our fix it would result in a redirect loop. The feature toggle allows these sites to implement a fix / decide whether to switch how they redirect.

comzeradd:
> I ended up adding a feature toggle because it's easy to add in this case, and we found some NROs are using a similar fix

Since it's just two NROs, let's have that feature flag as opt-out (aka enabled by default), and we can send them a heads up or disable it during release.

Inwerpsel:
> > I ended up adding a feature toggle because it's easy to add in this case, and we found some NROs are using a similar fix
> 
> Since it's just two NROs, let's have that feature flag as opt-out (aka enabled by default), and we can send them a heads up or disable it during release.

Can we do a follow up to switch all instances? That way the risk can be spread.

mleray:
Maybe it would be better to define this new feature (`Redirect redirect pages`) so that it's not in the middle of dev-only features?

Screenshot 2022-04-15 at 15 02 16

mleray:
I'm not sure whether or not that's an issue, but I think this fix will not work if trying to redirect from a tag page to another tag page? Since we check `is_tag()` 🤔

src/Features/RedirectRedirectPages.php
```diff
@@ -0,0 +1,42 @@
...
+	 */
+	public static function id(): string {
+		return 'redirect_redirect_pages';
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	protected static function name(): string {
+		return __( 'Redirect redirect pages', 'planet4-master-theme-backend' );
```
comzeradd, line 23: Just to make it more explicit: _"Redirect unused Tag pages"_.

(and by "explicit" I mean that the title should be about "Tag pages")

src/Features/RedirectRedirectPages.php
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	protected static function description(): string {
+		return __(
+			'Not a typo: if turned on this will do a HTTP redirect from a tag redirect page URL to the tag URL, to avoid duplicate content.',
```
comzeradd, line 31: A suggestion to make this sorter: _"Redirect from the Tags redirect pages to the tag URL, to avoid duplicate content."_

Inwerpsel: I preserved "HTTP" as it makes it more obvious there's 2 different meanings.

comzeradd:
This one looks like it's good to go. It's behind a feature flag, so we can gradually enable it on websites that don't have their own custom redirect override in place.

We do use `is_tag()`, which will not capture cases where a tag page is used as a redirect page on another tag. But I think that's really an edge case that is unlikely to be happening in any website.

lithrel (approved):
:+1:

PR #1646, opened 2022-03-10
Unwrap cookies CSS

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6664

Was needlessly wrapping specific selectors inside the element they happened to be in.

Also duplicate media queries instead of selectors. That way it's always close to the thing it's overriding.

[Hide whitespace changes](https://github.com/greenpeace/planet4-master-theme/pull/1646/files?diff=split&w=1)

It's easier if we merge this simplification first, as it should not change any behavior. After that I'll open a PR to address the button ellipsis issue.

### DIFF
--- a/assets/src/scss/components/_buttons.scss
+++ b/assets/src/scss/components/_buttons.scss
@@ -31,7 +31,6 @@
   position: relative;
   font-size: $font-size-sm;
   text-decoration: none;
-  color: $white;
   border: 1px solid transparent;
   cursor: pointer;
   line-height: 3;
@@ -40,7 +39,6 @@
   &:hover,
   &:active,
   &:visited {
-    color: $white;
     text-decoration: none;
   }
 
--- a/templates/cookies.twig
+++ b/templates/cookies.twig
@@ -28,7 +28,7 @@
 			</div>
 			{% if show_settings and cookies.enable_reject_all_cookies %}
 				<button
-					class="btn reject-all-cookies"
+					class="btn btn-discrete reject-all-cookies"
 					data-ga-category="Cookies box"
 					data-ga-action="Reject all cookies"
 				>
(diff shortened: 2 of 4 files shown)


planet-4:

:watch: 2022.03.10 11:03:04

mleray:
I think you still need some level of nesting for the `reject-all-cookies` class, because it should have different styles based on whether it's the one on the first tab or on the settings tab:
Screenshot 2022-03-10 at 13 06 14

Or maybe we could add a new class to one of them to differentiate them without need for nesting.

mleray:
Also, these buttons should not be this tall:
- In this PR:
Screenshot 2022-03-10 at 13 10 05

- Expected:
Screenshot 2022-03-10 at 13 11 05

Inwerpsel:
Yeah the second commit is still WIP, I made the PR draft again.

templates/cookies.twig
```diff
@@ -28,7 +28,7 @@
 			</div>
 			{% if show_settings and cookies.enable_reject_all_cookies %}
 				<button
-					class="btn reject-all-cookies"
+					class="btn btn-discrete reject-all-cookies"
```
Inwerpsel, line 31: I instead turned this into a new class, to mirror the usage of `btn-secondary` on the other button.

Inwerpsel: See https://github.com/greenpeace/planet4-master-theme/pull/1646#issuecomment-1063986146

assets/src/scss/components/_buttons.scss
```diff
@@ -40,7 +39,6 @@
   &:hover,
   &:active,
   &:visited {
-    color: $white;
```
Inwerpsel, line 43: I removed this because otherwise it applied to the new button style class, where I just want the same color on hover. It should not be a problem as all buttons use a style.

mleray (approved):

PR #1638, opened 2022-03-04
PLANET-6555 Style Gravity Forms

mleray:
### Description

See [PLANET-6555](https://jira.greenpeace.org/browse/PLANET-6555)
This includes the following form fields: textarea, input, checkbox, radio, select.

### Testing
You can test a form example on [this page](https://www-dev.greenpeace.org/test-nix/gravity-forms-tests/), you can also play around in the backend with editing the existing GF form or creating a new one! Please make sure to also test the validation messages when submitting an invalid form.

PR #1647, opened 2022-03-10
Revert "PLANET-6555 Style Gravity Forms"

Inwerpsel:
Reverts greenpeace/planet4-master-theme#1638

I noticed locally on the cookies settings that suddenly the amount of CSS selectors is massive. I don't even have the GF plugin installed.

This is just one example I stumbled on. `@extend` is likely not the right approach here. And we should ensure that if we need to serve a significant amount of additional CSS, it should only be served if there's a GF form on the page that needs it.

Currently production:
Screenshot from 2022-03-10 12-53-02

Latest master:
Screenshot from 2022-03-10 12-52-46


Inwerpsel:
I checked the resulting file sizes and they are more than double because of this change
```bash
$ ls -lh assets/build/style.min.css assets/build/editorStyle.min.css
-rw-rw-r-- 1 pieter pieter 244K Mar 10 13:00 assets/build/editorStyle.min.css
-rw-rw-r-- 1 pieter pieter 201K Mar 10 13:00 assets/build/style.min.css
$ go revert-1638-style-gravity-forms-6555 
Switched to branch 'revert-1638-style-gravity-forms-6555'
Your branch is up to date with 'origin/revert-1638-style-gravity-forms-6555'.
$ ls -lh assets/build/style.min.css assets/build/editorStyle.min.css
-rw-rw-r-- 1 pieter pieter 155K Mar 10 13:00 assets/build/editorStyle.min.css
-rw-rw-r-- 1 pieter pieter  97K Mar 10 13:00 assets/build/style.min.css

mleray (approved):
Right now, I don't see how we can avoid adding a significant amount of CSS, because [according to the GF docs](https://docs.gravityforms.com/category/user-guides/design-and-layout/css-selectors/) we either need to use these super-long selectors in order to apply our changes, or we need to add `!important` everywhere 😕 But maybe I'm wrong and there's a better way!
If not, indeed we should think of a way to load these styles only when there is a GF form on the page before releasing.

Inwerpsel:
> Right now, I don't see how we can avoid adding a significant amount of CSS, because [according to the GF docs](https://docs.gravityforms.com/category/user-guides/design-and-layout/css-selectors/) we either need to use these super-long selectors in order to apply our changes, or we need to add `!important` everywhere confused But maybe I'm wrong and there's a better way! If not, indeed we should think of a way to load these styles only when there is a GF form on the page before releasing.

I'm not saying we can't add CSS. But if you see that what you added results in more than double the size of our main CSS file, which was already large at 97K, while the PR only adds 174 lines of SCSS. Something is clearly off here.

planet-4:

:watch: 2022.03.10 12:42:50

PR #1651, opened 2022-03-17
PLANET-6668 Implement Highlighted CTA pattern

mleray:
### Description

See [PLANET-6668](https://jira.greenpeace.org/browse/PLANET-6668), the [example page](https://www-dev.greenpeace.org/gutenberg/new-p4-blocks/) and the [designs](https://www.figma.com/file/leueo1LMrlPabJkZYs2NgW/IA%26nav_Final-mockups?node-id=1%3A1650)
This PR includes adding the "Dark Green" color to our palette, and creating the transparent button styles.

**Related PR:** https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/814

### Testing

You can test the new pattern on [this page](https://www-dev.greenpeace.org/test-oberon/51007-2/) for example, or on your local.

### DIFF
--- a/assets/src/scss/base/_palette.scss
+++ b/assets/src/scss/base/_palette.scss
@@ -11,7 +11,8 @@ $palette: (
   "dark-blue": $dark-blue,
   "blue": $blue,
   "orange-hover": $orange-hover,
-  "yellow": $yellow
+  "yellow": $yellow,
+  "dark-green": $dark-green,
 );
 
 @each $name, $color in $palette {
@@ -21,5 +22,8 @@ $palette: (
 
   .has-#{$name}-background-color {
     background-color: $color;
+
+    --transparent-button--hover--color: #{$color};
+    --transparent-button--active--color: #{$color};
   }
 }
--- a/assets/src/scss/editorOverrides.scss
+++ b/assets/src/scss/editorOverrides.scss
@@ -126,6 +126,14 @@
   }
 }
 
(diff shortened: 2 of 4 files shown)


assets/src/scss/base/_palette.scss
```diff
@@ -21,5 +22,8 @@ $palette: (
 
   .has-#{$name}-background-color {
     background-color: $color;
+
+    --transparent-button--hover--color: #{$color};
```
mleray, line 26: This is the only way I could think about in order to have the correct color on hover in those new buttons, however this will only work if the selected background color is from our own palette... It won't work if editors choose instead a color from the default WP palette, in that case for now I'm using a [dark grey as hover color](https://github.com/greenpeace/planet4-master-theme/pull/1651/files#diff-2bc1c6e089a7a01997bb250bca7b661343c35a1512e49d22a11b1f00ef03b783R300) to make sure that it's easy to read with the white background and still sort of fits in with our colors.

assets/src/scss/editorOverrides.scss
```diff
@@ -126,6 +126,14 @@
   }
 }
 
+.wp-block-button.transparent-button .wp-block-button__link[role="textbox"] {
```
mleray, line 129: I haven't added the `:hover` styles here because it looked weird when editing the button

sagarsdeshmukh (approved):

PR #1660, opened 2022-04-04
PLANET-6730 Split post css

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6730

Easier [with whitespace hidden](https://github.com/greenpeace/planet4-master-theme/pull/1660/files?diff=split&w=1).

Put all CSS that is only needed on posts in a separate entry point. This is then included in the same cases by checking post type before enqueuing (post, attachment, and idea).

TODO:

- [x]  Check child themes on dev sites.
- [x]   * Check whether decreased specificity interferes with overrides.
- [x]   * Check if the separate file is properly enqueued in the right order.
- [x] * Further reduce now obsolete CSS specificity. Perhaps better to do this in a follow up to reduce the chance of child theme regressions. (follow up)

### DIFF
--- a/assets/src/scss/pages/post/_article-content.scss
+++ b/assets/src/scss/pages/post/_article-content.scss
@@ -58,11 +58,6 @@
     }
   }
 
-  blockquote,
-  blockquote > p {
-    @include blockquote;
-  }
-
   p {
     margin: 0 0 18px 0;
     overflow-wrap: break-word;
@@ -218,20 +213,6 @@
 
     br + br {
       display: none;
-    }
-  }
-
-  .more-link {
-    @extend .btn;
-    @extend .btn-secondary;
-    @extend .btn-large;
-  }
-
-  .page-links {
-    a {
-      @extend .btn;
-      @extend .btn-secondary;
-      @extend .btn-small;
     }
   }
 
--- a/assets/src/scss/pages/post/_post.scss
+++ b/assets/src/scss/pages/post/_post.scss
@@ -1,96 +1,90 @@
-.single-post,
-.single-attachment,
-.single-idea {
-  @include large-and-up {
-    .narrow-container {
-      margin-left: auto;
-      margin-right: auto;
-      max-width: 776px;
-      padding: 0;
-    }
+@include large-and-up {
+  .narrow-container {
+    margin-left: auto;
+    margin-right: auto;
+    max-width: 776px;
+    padding: 0;
+  }
+}
+
+@include x-large-and-up {
+  .narrow-container {
+    max-width: 736px;
+  }
+}
+
+.skewed-overlay {
+  mask-image: none;
+}
+
+.page-header-title {
+  color: $grey-80;
+  hyphens: none;
+  max-width: none;
+  margin-bottom: 8px;
+  line-height: 1.2;
+}
+
+.single-post-meta {
+  color: $grey-40;
+  font-size: 14px;
+  font-family: $roboto;
+  border-bottom: 1px solid transparentize($grey-20, 0.5);
+  padding-bottom: 16px;
+  line-height: 24px;
+  margin-bottom: 16px;
+  display: flex;
+  gap: 8px;
+  align-items: end;
+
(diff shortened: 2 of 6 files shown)


webpack.config.js
```diff
@@ -90,29 +91,6 @@ module.exports = {
       ignoreOrder: false, // Enable to remove warnings about conflicting order
       filename: './[name].min.css'
     }),
-    new RemovePlugin({
```
Inwerpsel, line 93: There's no point removing these, they don't affect performance as they're not used, and probably could be used to improve our CSS efficiency.

planet-4:

:watch: 2022.04.04 10:45:14

assets/src/scss/pages/post/_article-content.scss
```diff
@@ -58,11 +58,6 @@
     }
   }
 
-  blockquote,
```
Inwerpsel, line 61: This is already here
https://github.com/greenpeace/planet4-master-theme/blob/5099dcc5464132874f346ae3e501446c897138f2/assets/src/scss/components/_blockquote.scss#L19-L22
which is loaded on all pages
https://github.com/greenpeace/planet4-master-theme/blob/5099dcc5464132874f346ae3e501446c897138f2/assets/src/scss/style.scss#L30-L30

assets/src/scss/pages/post/_article-content.scss
```diff
@@ -221,20 +216,6 @@
     }
   }
 
-  .more-link {
```
Inwerpsel, line 224: The only place in our Github repos where this is still added to the template is on the Lux's site's contact form, where custom CSS overrides all styles anyway.
https://github.com/search?q=org%3Agreenpeace+more-link&type=code

Inwerpsel: The only difference is the large style. This could be solved by giving the [single usage](https://github.com/greenpeace/planet4-child-theme-luxembourg/blob/caf1f950debf95cb8b3a1d201140a3ebc3ccbca0/templates/contact.twig#L144-L146) the `.btn-large` class.

assets/src/scss/pages/post/_article-content.scss
```diff
@@ -221,20 +216,6 @@
...
-    @extend .btn;
-    @extend .btn-secondary;
-    @extend .btn-large;
-  }
-
-  .page-links {
```
Inwerpsel, line 230: This one is not used at all anymore: https://github.com/search?q=org%3Agreenpeace+page-links&type=code

webpack.config.js
Inwerpsel: I added it back again to reduce the scope of the PR.

mleray (approved):
Checked on the test instance, looks good to me 👍

assets/src/scss/pages/post/_post.scss
```diff
@@ -1,96 +1,90 @@
...
-          top: -2px;
-        }
-      }
-    }
+  svg {
+    top: -2px;
   }
+}
 
-  .post-articles-block .articles-block {
```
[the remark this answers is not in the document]
Inwerpsel: Can we check in a follow up? The `single.twig` template is [overridden by a few child themes](https://github.com/search?q=org%3Agreenpeace+post-articles-block&type=code
). 
Also the specificity change maybe affects[ CSS that targets `.articles` block only](https://github.com/search?q=org%3Agreenpeace+articles-block&type=code).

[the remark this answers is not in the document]
mleray: Yes, I'm fine with removing this in a follow-up! However the specificity changes should not affect existing Articles blocks right? Since the Articles block is currently not allowed in posts, and this CSS is only for posts 🤔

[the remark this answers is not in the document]
Inwerpsel: It should be fine in our base CSS, and probably also in the child themes, but it's harder to tell without looking at each in detail.

PR #1628, opened 2022-02-17
PLANET-6673: Navigation bar drop down menus

GP-Dan-Tovbein:
Ref: [Ticket](https://jira.greenpeace.org/browse/PLANET-6673)

🎨 [Figma Design](https://www.figma.com/file/Gn8wyqgto7608rlESePXZE/Foundations-%26-Components?node-id=1126%3A1834) 

Fully implementation of drop down menus (already approved). It will work as a final review before merging into the `master` branch.

Go to the [Demo page](https://www-dev.greenpeace.org/test-nix) and play around the main navigation.

**Please be sure that the feature is enabled**, if not then go to Planet 4 > Information architecture > Enable dropdown menu

**Tickets**
- ~Desktop version~ ✅ : [PLANET-6509](https://jira.greenpeace.org/browse/PLANET-6509)  
- ~Mobile version~ ✅ [PLANET-6511](https://jira.greenpeace.org/browse/PLANET-6511)
- ~Donate Button~ ✅: [PLANET-6579](https://jira.greenpeace.org/browse/PLANET-6579)

**Related PRs status**
- Desktop version : https://github.com/greenpeace/planet4-master-theme/pull/1604
   Current status: ~In Development~ | ~UAT Passed~ | ~In Review~ | ~Review Passed~
- Mobile version : https://github.com/greenpeace/planet4-master-theme/pull/1634
   Current status: ~In Development~ | ~UAT Passed~ | ~In Review~ | ~Review Passed~
- Donate Button : https://github.com/greenpeace/planet4-master-theme/pull/1626
   Current status: ~In Development~ | ~UAT Passed~ | ~In Review~ | ~Review Passed~

**Documentation**
[P4 settings](https://planet4.greenpeace.org/manage/admin-setup/)
[P4 Key content](https://planet4.greenpeace.org/manage/administer/content-setup/)

**Important**
[Dropdown feature on Dev instances analysis file before merge](https://docs.google.com/spreadsheets/d/1uwWTto_tHptQHqWe4XyaTz7XcjtZ9WhcOajNCGBxkSg/edit#gid=0).

**To follow up before merge**
- ✅ ~Update the [Main Navigation Documentation](https://planet4.greenpeace.org/manage/admin-setup/#main-p4-navigation--title--act--explore-)~ [Luca's Slack message](https://greenpeace-gpi.slack.com/archives/G015K63081W/p1647610662009389)
[...]

PR #1672, opened 2022-04-13
Clean up burger menu code

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6673 

Cleanup PR after we merged https://github.com/greenpeace/planet4-master-theme/pull/1628

We glossed over some parts of the code to unblock merging the branch. There's a few things that would be nice to address before we need to worry about child theme overrides. Mainly the CSS specificity as overrides would likely copy it.

I'll start with addressing some basic things, like using tabs inside Twig files and indentation. After that we can see if there's more room for improvement, but maybe it's better to merge these already to provide a cleaner basis to work on.

Maybe easier to see the CSS changes [with whitespace hidden](https://github.com/greenpeace/planet4-master-theme/pull/1672/files?diff=split&w=1).

I'm pretty sure that after we make the CSS readable, we'll find we can remove a lot of unsemantic wrapper elements in favor of a cleaner CSS solution. I expect there are also many redundant rules that can be removed / replaced with a better alternative.

### Testing
This should still work the exact same way. On top of reducing specificity, I [only changed 1 class name](https://github.com/greenpeace/planet4-master-theme/pull/1672/files#r849341105).

### DIFF
--- a/assets/src/scss/layout/navbar/_burger-menu.scss
+++ b/assets/src/scss/layout/navbar/_burger-menu.scss
@@ -49,170 +49,6 @@
     width: auto;
   }
 
-  .burger-menu-header {
-    display: flex;
-    flex-direction: column;
-    flex-shrink: 0;
-    border-bottom: 1px solid rgba(0, 0, 0, 0.2);
-    padding-block-end: $sp-2;
-
-    .site-logo {
-      display: block;
-      width: fit-content;
-      margin-inline-start: 0;
-      margin-bottom: $sp-1x;
-      line-height: 1;
-
-      img {
-        height: 23px;
-      }
-    }
-
-    .nav-languages {
-      width: 100%;
-      margin-top: $sp-1x;
-      margin-bottom: $sp-1x;
-
-      ul {
-        margin-inline: 0;
-        padding: 0;
-        margin-bottom: 0;
-      }
-    }
-  }
-
-  .nav-wrapper {
-    overflow-y: scroll;
-
-    .nav-items {
-      width: 100%;
-      padding: $sp-1x 0 0;
-    }
-
-    @include scrollbar-layout;
-  }
-
-  .burger-menu-footer {
-    display: flex;
-    justify-self: flex-end;
-    flex-shrink: 0;
-    margin-top: auto;
-    width: 100%;
-    padding: $sp-3 0 0;
-    border-top: 1px solid rgba(0, 0, 0, 0.2);
-
-    .btn-donate {
-      width: 100%;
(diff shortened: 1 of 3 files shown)


planet-4:

:watch: 2022.04.13 10:11:20

templates/burger-menu-items.twig
```diff
@@ -8,7 +8,7 @@
 		{% set link_ga_action = item.title %}
 	{% endif %}
 	<li class="burger-menu-item nav-item accordion-item {{ item.class }} {{ item == item.current ? 'active' : '' }}">
-		<div class="item-wrapper accordion-header" id="burger-menu-accordion-{{ targetId }}">
+		<div class="burger-menu-item-wrapper accordion-header" id="burger-menu-accordion-{{ targetId }}">
```
Inwerpsel, line 11: This is the only markup change. The `item-wrapper` class is too broad to make any sense.

I'll probably do another PR after this one to address the many unsemantic wrapper elements, so might change it again there.

GP-Dan-Tovbein (approved):
Code looks good to me!

assets/src/scss/layout/navbar/_burger-menu.scss
```diff
@@ -49,220 +49,219 @@
...
-    transition: inset-inline-start 0.2s ease-in;
-
-    @supports not (inset-inline-start: -400px) {
-      transform: translateX(-400px);
-      transition: transform 0.2s ease-in;
-
-      html[dir="rtl"] & {
-        transform: translateX(400px);
-      }
+.nav-menu-close {
```
GP-Dan-Tovbein, line 234: If you move outside the `.burger-menu` class, you're converting this as a general class and I think it would be nice (only in this case) to set the specificity. What do you think?

GP-Dan-Tovbein (commented):
I noticed something after my approval.
What do you thing about that?

assets/src/scss/layout/navbar/_burger-menu.scss
Inwerpsel: We only use this class [in 1 place in 1 Twig template](https://github.com/search?q=org%3Agreenpeace+%22nav-menu-close%22&type=code).

And if you think about it, why would these styles only make sense if the close button is inside a burger menu? What if we put the same close button on something that is not a burger menu item.

If somebody would decide to in our code start using the class name in another place, they'll immediately see that style applied. If it doesn't make sense for that item, you can just use another class. If the styles already work on the new element, you don't have to adjust any CSS. Lower specificity = CSS that you can actually reuse.

PR #1673, opened 2022-04-13
PLANET-6673 Burger menu fixes

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6673

Some further low hanging fruit improvements to the code, mostly making it do the same with less lines of CSS.

I also used `d-lg-none` in favor of CSS rules to hide other elements.

~~Each commit is a separate fix, please refer to the commit messages for more detailed info.~~ I'll probably rebase it as now the latest fix supersedes some of the previous commits.

I ended up using transforms instead for the open and close animation. If you do `translateX(-100%)`, it always uses 100% of the width of that element. Whereas `inset-inline-start` needed CSS per screen size, as % there uses the parent container as a reference.

### DIFF
--- a/assets/src/scss/base/_mixins.scss
+++ b/assets/src/scss/base/_mixins.scss
@@ -12,13 +12,6 @@
   }
 }
 
-// Mobile to Large (From 576px to 992px)
-@mixin small-to-large {
-  @media (min-width: #{$small-width}) and (max-width: #{$large-width - 1}) {
-    @content;
-  }
-}
-
 // Tablet & Up (greater than 768px)
 @mixin medium-and-up {
   @media (min-width: #{$medium-width}) {
--- a/assets/src/scss/layout/navbar/_burger-menu.scss
+++ b/assets/src/scss/layout/navbar/_burger-menu.scss
@@ -1,3 +1,20 @@
+@mixin slide-from-start() {
+  &:not(.open) {
+    transition: transform 0.2s ease-in, visibility;
+    transition-delay: 0s, 0.2s;
+    transform: translateX(-100%);
+    visibility: hidden;
+
+    html[dir="rtl"] & {
+      transform: translateX(100%);
+    }
+  }
+
+  &.open {
+    transition: transform 0.2s ease-out;
+  }
+}
+
 .burger-menu {
   _-- {
     background: $white;
@@ -11,28 +28,15 @@
   justify-content: flex-start;
   top: 0;
   left: 0;
-  inset-inline-start: -100%;
   padding: 18px $sp-3;
   z-index: 9999999999;
   overflow: hidden;
   height: 100%;
 
-  &.open {
-    inset-inline-start: 0;
-    transition: inset-inline-start 0.2s ease-out;
+  @include slide-from-start;
 
-    @supports not (inset-inline-start: 0) {
-      transform: translateX(0);
-      transition: transform 0.2s ease-out;
-
-      html[dir="rtl"] & {
-        transform: translateX(0);
-      }
-    }
-
-    ~ .burger-menu-overlay {
-      display: block;
-    }
+  &.open ~ .burger-menu-overlay {
+    display: block;
   }
 
   ul {
@@ -59,31 +63,10 @@
 
   @include small-and-less {
     width: 100vw;
-    inset-inline-start: -100vw;
-
-    @supports not (inset-inline-start: -100vw) {
-      transform: translateX(-100vw);
-      transition: transform 0.2s ease-in;
-
-      html[dir="rtl"] & {
-        transform: translateX(100vw);
-      }
-    }
   }
 
-  @include small-to-large {
+  @include small-and-up {
     width: 400px;
-    inset-inline-start: -400px;
-    transition: inset-inline-start 0.2s ease-in;
-
-    @supports not (inset-inline-start: -400px) {
-      transform: translateX(-400px);
-      transition: transform 0.2s ease-in;
-
-      html[dir="rtl"] & {
-        transform: translateX(400px);
-      }
-    }
   }
 }
 
--- a/assets/src/scss/layout/navbar/_burger-menu.scss
+++ b/assets/src/scss/layout/navbar/_burger-menu.scss
@@ -32,10 +32,6 @@
 
     ~ .burger-menu-overlay {
       display: block;
-
-      @include large-and-up {
-        display: none;
-      }
     }
   }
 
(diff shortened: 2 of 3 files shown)


assets/src/scss/layout/navbar/_burger-menu.scss
```diff
@@ -12,14 +12,14 @@
   top: 0;
   left: 0;
   inset-inline-start: -100%;
+  transition: inset-inline-start 0.2s ease-out;
```
GP-Dan-Tovbein, line 15: If you add this attribute, you'll see a weird burger menu behavior whenever you resize the browser. That's why wasn't added by its default state.

GP-Dan-Tovbein: Uploading Screen Recording 2022-04-13 at 12.49.49.mov…

GP-Dan-Tovbein: https://user-images.githubusercontent.com/77975803/163220335-976cedbf-432f-4f39-90a3-6d226905ad10.mov

Inwerpsel: I agree that doesn't look great, but that's only a tiny fraction of cases, in case someone switches from portrait to landscape. The switching animation itself probably makes this unnoticeable.

As for resizing the window, that's also a very rare event. So I'm not sure if it's worth not having a nice animation in all other cases.

Inwerpsel: But let me check if I can fix it :eyes:

GP-Dan-Tovbein: But if you see this is not a big problem, then we could add it.

Inwerpsel: This seems to work, didn't find any drawbacks yet :crossed_fingers:  https://github.com/greenpeace/planet4-master-theme/pull/1673/commits/bebd164f118c2d640a2e2e05e9f1b623cff937de

Inwerpsel: I changed this to use `:not(.open)`, it needs less rules. https://github.com/greenpeace/planet4-master-theme/blob/377053af1ba308a97dd4bcb009e4be7a86ccb8c6/assets/src/scss/layout/navbar/_burger-menu.scss#L20-L22

assets/src/scss/layout/navbar/_burger-menu.scss
```diff
@@ -11,16 +11,28 @@
...
   z-index: 9999999999;
   overflow: hidden;
   height: 100%;
 
-  &.open {
-    inset-inline-start: 0;
-    transition: inset-inline-start 0.2s ease-out;
+  &:not(.open) {
+    visibility: hidden;
+    transition-delay: 0s, 0.2s;
```
GP-Dan-Tovbein, line 22: The transition to close through medium screens is so fast

assets/src/scss/layout/navbar/_burger-menu.scss
```diff
@@ -11,16 +11,28 @@
   justify-content: flex-start;
   top: 0;
   left: 0;
-  inset-inline-start: -100%;
+  transition: inset-inline-start 0.2s ease-out, visibility;
```
GP-Dan-Tovbein, line 14: The transition to close through medium screens is so fast

Inwerpsel: Yes still looking into it, I'll push a fix later.

GP-Dan-Tovbein (commented):
I've tested this new changes on a tablet and it looks different than before

assets/src/scss/layout/navbar/_burger-menu.scss
```diff
@@ -63,29 +71,20 @@
...
-      transform: translateX(-100vw);
-      transition: transform 0.2s ease-in;
-
-      html[dir="rtl"] & {
-        transform: translateX(100vw);
-      }
-    }
   }
 
-  @include small-to-large {
```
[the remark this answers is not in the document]
Inwerpsel: Yes that was because I removed both the `100vw` and `400px` variant of the code, whereas only the `100vw` had the same effect as `100%`. I re-added it in the latest push, can you confirm it's ok now?

assets/src/scss/base/_mixins.scss
```diff
@@ -256,3 +249,20 @@ $nested-comment-left-margin: 50px;
     background: rgba(0, 0, 0, 0.2);
   }
 }
+
+@mixin slide-from-start() {
```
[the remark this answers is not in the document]
Inwerpsel: It seems to make sense as a mixin, you could put it on any other element that uses the `.open` class to make it slide in.

A secondary reason is that it helps keep the other file readable.

But maybe I'll move the mixin into that file while it's only used there.

templates/burger-menu.twig
```diff
@@ -81,4 +81,4 @@
 		</div>
 	{% endif %}
 </div>
-<div class="burger-menu-overlay"></div>
+<div class="burger-menu-overlay d-lg-none"></div>
```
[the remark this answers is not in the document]
GP-Dan-Tovbein: From my point of view those are different cases, the burger menu starts displayed to `flex`, but the overlay [is started to none](https://github.com/greenpeace/planet4-master-theme/pull/1673/files#diff-30638aecdc1bb63b61eab64177e65ff6123f16d8f032495ff5ef8a97b66272b8R237)).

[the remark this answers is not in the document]
GP-Dan-Tovbein: BTW, the overlay is not being shown

Screenshot 2022-04-14 at 11 46 44

assets/src/scss/layout/navbar/_burger-menu.scss
```diff
@@ -1,3 +1,20 @@
+@mixin slide-from-start() {
```
GP-Dan-Tovbein, line 1: Going back to the previous question, what's the motivation to add this `mixin`? Is only used here and only one time

Inwerpsel: https://github.com/greenpeace/planet4-master-theme/pull/1673#discussion_r850512997

GP-Dan-Tovbein: TBH, if it's moved to this file it doesn't make sense to use a `mixin`

Inwerpsel: In this case it's used more than once, just mean that it's common to put a mixin in a file if that makes it more readable.

Inwerpsel: We can move this later once we have a second use case.

GP-Dan-Tovbein (approved):

PR #1674, opened 2022-04-14
PLANET-6705 Treat gp_nro cookie as 1st party cookie

mleray:
### Description

See [PLANET-6705](https://jira.greenpeace.org/browse/PLANET-6705)
After the recent Cookies changes we did, the cookie we set for the last visited NRO is considered as 1st party, so it should be fired right away without waiting for user consent. It should also be updated when switching sites.

### Testing

On local, you can switch to this branch and open http://www.planet4.test/ in an incognito window, then in the development tools you can check the cookies and you should see the `gp_nro` cookie set right away, along with the `greenpeace` one. On `master` branch you should notice only the `greenpeace` one is set on the first visit.

### DIFF
--- a/assets/src/js/cookies.js
+++ b/assets/src/js/cookies.js
@@ -61,7 +61,8 @@ export const setupCookies = () => {
 
   const cookie = readCookie('active_consent_choice');
   const cookiesBox = document.querySelector('#set-cookie');
-  const nro = document.body.dataset.nro;
+  const currentNRO = document.body.dataset.nro;
+  const previousNRO = readCookie('gp_nro');
   const greenpeace = readCookie('greenpeace');
   const noTrack = readCookie('no_track');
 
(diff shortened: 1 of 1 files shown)


planet-4:

:watch: 2022.04.14 07:30:27

assets/src/js/cookies.js
```diff
@@ -97,26 +97,22 @@ export const setupCookies = () => {
     }
   };
 
-  // Make the necessary cookies checked by default on user's first visit.
-  // Here if the No cookies set(absence of 'greenpeace' & 'no_track' cookies) consider as first visit of user.
+  // Make the necessary cookies checked by default on user's first visit, also create cookie to store last visited nro.
+  // Here if the No cookies set (absence of 'greenpeace' & 'no_track' cookies) consider as first visit of user.
   if (greenpeace === null && noTrack === null) {
     createCookie('greenpeace', ONLY_NECESSARY, 365);
+    createCookie('gp_nro', nro, 30);
```
lithrel, line 104: I'm realizing with this that our cookies are domain and not path specific, so if you visited `/luxembourg` first those cookies are set, and if you then go to `/mexico` `gp_nro` won't change.
I think we need to put this line out of any condition so it stays accurate, keeping its current behavior of changing its content when you visit a different nro.

mleray: Good point 🤔 Maybe we could check the value of `gp_nro` and if we're on a different NRO site then we update it?

lithrel: Yep should work too :+1:

lithrel (approved):
:+1:

PR #1680, opened 2022-04-15
Use opacity instead of visibility for better performance

Inwerpsel:
Follow up on https://github.com/greenpeace/planet4-master-theme/pull/1673 and https://jira.greenpeace.org/browse/PLANET-6673

* This way the off screen element is still not visibile while the screen
size is changing. Opacity has better render performance.

I also added a transition of this opacity for better performance. It should be safe, as it will be below each element unless it has a lower z index.

You can't animate opacity if the element is `display: none`, so I had to move the overlay under all elements with `z-index`. So far this seemed reliable.

### Before

https://user-images.githubusercontent.com/7604138/164056181-1f350e07-0425-4596-bc17-6470cde32ddc.mp4

### After

https://user-images.githubusercontent.com/7604138/164056216-f224128b-14e7-4df3-a0f1-f6f93608144d.mp4

### DIFF
--- a/assets/src/scss/layout/navbar/_burger-menu.scss
+++ b/assets/src/scss/layout/navbar/_burger-menu.scss
@@ -1,9 +1,11 @@
 @mixin slide-from-start() {
+  $duration: .2s;
+
   &:not(.open) {
-    transition: transform 0.2s ease-in, visibility;
-    transition-delay: 0s, 0.2s;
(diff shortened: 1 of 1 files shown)


planet-4:

:watch: 2022.04.15 13:19:28

GP-Dan-Tovbein (approved):
Code looks good to me!

(but still disagree to use a mixin here LOL)

PR #1684, opened 2022-04-18
Change <section> to <header> from Header

GP-Dan-Tovbein:
This is a semantic (probably a nice to have) change and opened for suggestions.
Is it not better to use a `<header>` instead of a `<section>` tag? basically because of SEO.

Maybe you implemented it previously for a particular reason.


planet-4:

:watch: 2022.04.18 14:44:07

comzeradd (approved):
Definitely better semantically :+1:

I don't think we rely on `section` for the header css, so this change seems transparent

PR #1687, opened 2022-04-22
For now only style this on the tag pages

Inwerpsel:
* CSS was written assuming it's the only usage of the block, but now
people are starting to allow it in the editor. Ensure the CSS doesn't
conflict anymore.
* I didn't change the pagination style, because that one definitely
should not be used in regular pages.

It's just a quick fix, we can later properly register this as block styles.

Example tag page on the instance https://www-dev.greenpeace.org/test-oberon/tag/50-years/ Styles should still be applied there.

### DIFF
--- a/assets/src/scss/layout/_block-classes.scss
+++ b/assets/src/scss/layout/_block-classes.scss
@@ -24,27 +24,32 @@
   }
 }
 
-.wp-block-post-terms {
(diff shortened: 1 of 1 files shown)


GP-Dan-Tovbein (approved):
Code looks good to me!

PR #1671, opened 2022-04-13
PLANET-6552: Rewrite Search filters to follow native taxonomies

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-6652

> - We could re-purpose the filters to follow native taxonomies. So in the end we can have these filters:
>    "Categories" (actual categories)
>    "Tags" (actual tags) leave out for now
>    "Post Types" (eg. Stories, Publications, etc)
>    "Action Types" (eg. Petition, Event)
>    "Content Types" (WP post types. eg. pages, posts, actions, archive, document, etc)
> - Hide taxonomies and post types if there is no content

## Modifications

- Extracted filters to their own classes
- Renamed some filters to follow code taxonomy
- Implemented p4_action post type (only appears if the post type is active)

<table>
<tr>
<td><em>Master branch</em></td>
<td><em>This branch</em></td>
</tr>
<td>

</td>
<td>

</td>
</tr>
</table>

## Test

Test on [Phobos](https://www-dev.greenpeace.org/test-phobos)

Locally, run the elastic container and the indexation
```shell
> docker-compose -f docker-compose.full.yml up -d elasticsearch
> docker-compose exec php-fpm wp option update ep_host $(ELASTICSEARCH_HOST)
> docker-compose exec php-fpm wp elasticpress index --setup --quiet 
```

## Notes

Those mixed mentions are also noted in
- page.php l.25
- tag.php l.5

PR #1688, opened 2022-04-26
Add setting to link to category directly

Inwerpsel:
### Ref: [PLANET-6747](https://jira.greenpeace.org/browse/PLANET-6747)

Investigate whether it's easy to replace the weird logic of which links are shown after the tags, with a simple link to the category page itself.

Note that our category page atm still needs to get pagination and a new listing style. We should also first check the content of these pages, to be sure our new alternative is sufficient.

Since the toggle is only a few lines of code, it's easy to provide already on dev envs.

There's some things an NRO should watch out for when enabling:
* "Uncategorized" is a category, and will be shown as such. Also you need to manually unassign it after assigning an actual category.
* Some categories which didn't have a page will start getting shown. They should maybe be removed if they're not intended to show up. If they can't be removed we could maybe check a visibility setting (or create one) per category.

The old mechanism does effectively the same as tag "redirect pages", with small differences:
* If no page exists, the category is not linked to by itself.
* It doesn't replace the content of the category page, but the links to it.

### DIFF
--- a/src/Post.php
+++ b/src/Post.php
@@ -2,6 +2,7 @@
 
 namespace P4\MasterTheme;
 
+use P4\MasterTheme\Features\PostPageCategoryLinks;
 use Timber\Post as TimberPost;
 use Timber\Term as TimberTerm;
 use WP_Block;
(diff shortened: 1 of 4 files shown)


planet-4:

:watch: 2022.04.26 11:51:26

src/Post.php
```diff
@@ -178,6 +179,17 @@ public function set_issues_links() {
 		$explore_page_id = $options['explore_page'] ?? '';
 		$categories      = get_the_category( $this->ID );
 
+		if ( PostPageCategoryLinks::is_active() ) {
+			$this->issues_nav_data = array_map(
```
lithrel, line 183: we could filter out `uncategorised` there (I just did that on [search filters](https://github.com/greenpeace/planet4-master-theme/pull/1671/files#diff-59c92d1bd26aebd2cddd3879fe484d83f59324074b2d50bad013fca4f6a1d60eR26-R28))

lithrel (approved):
:+1:

PR #1691, opened 2022-04-27
PLANET-6652: Extract search filters

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-6652
Ref: https://jira.greenpeace.org/browse/PLANET-6797
Ref/ https://jira.greenpeace.org/browse/PLANET-6578

Clarify search code by matching native taxonomy with class structure 
- Extract search filters to their own classes
- Rename page_types to post_types
- Hide tags search
- Add p4_action type to search

### DIFF
--- a/src/Search.php
+++ b/src/Search.php
@@ -625,22 +625,14 @@ abstract class Search {
 							break;
 						case 'ctype':
 							switch ( $filter['id'] ) {
-								case 0:
-									$args['post_type']   = 'page';
-									$args['post_status'] = 'publish';
-									$act_page            = planet4_get_option( 'act_page', 0 );
-									$args['post_parent'] = esc_sql( $act_page );
-									break;
 								case 1:
 									$args['post_type']      = 'attachment';
 									$args['post_status']    = 'inherit';
 									$args['post_mime_type'] = self::DOCUMENT_TYPES;
 									break;
 								case 2:
-									$args['post_type']             = 'page';
-									$args['post_status']           = 'publish';
-									$act_page                      = planet4_get_option( 'act_page', 0 );
-									$args['post_parent__not_in'][] = esc_sql( $act_page );
+									$args['post_type']   = 'page';
+									$args['post_status'] = 'publish';
 									break;
 								case 3:
 									$args['post_type']   = 'post';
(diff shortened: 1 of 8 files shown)


src/Search.php
```diff
@@ -727,85 +727,21 @@ protected function set_general_context( &$context ) {
...
 
-		// Tag <-> Campaign.
-		$tags = get_terms(
-			[
-				'taxonomy'   => 'post_tag',
-				'hide_empty' => false,
-			]
+		// Content Types.
+		$context['content_types'] = Search\Filters\ContentTypes::get_filters(
+			self::should_include_archive()
```
Inwerpsel, line 739: We could include the check whether it should include the archive inside of `get_filters`. I assume that if it would be called from another place, it would need the same value anywhere?

src/Search/Filters/Tags.php
```diff
@@ -0,0 +1,52 @@
...
+	public static function get_all(): array {
+		$tags = get_terms(
+			[
+				'taxonomy'   => 'post_tag',
+				'hide_empty' => false,
+			]
+		);
+
+		if ( is_wp_error( $tags )
+			|| ! is_array( $tags )
```
Inwerpsel, line 25: If the args are defined here statically in code, it's probably not needed to do the check for whether it's an array. WP should already reliably return an array. It's only a string if you pass in an argument to count instead of fetch.

Because of our args we can probably also assume it's always going to be a `WP_Term` if `is_wp_error()` returns false.

In case it returns a non empty array that has other stuff than `WP_Term`, I guess we would rather like to throw an error, instead of continuing with an empty array?

src/Search.php
lithrel: I don't know, that's why I made a parameter of it :grimacing:  if I move it to the filter, should I reverse the usage of this value in the Search class (use the filter method to define the archive usage), or keep a duplicate method in Search and Filter ?

comzeradd:
@lithrel is Actions showing up for you locally? I tried various things but couldn't make it work. 

I replicated my steps on the test instance.
1. Searched for "climate" and found one page that was showing up.
2. Converted it to an Action with the Post Switcher plugin.
3. Re-sync ES
4. Searched again for "climate".
Result: the page no longer shows up in results and no "Actions" filter is visible on the sidebar

lithrel:
@comzeradd right, :+1: the post type was missing for the general search args, fixed now :wrench:  the "Action" content type [appears](https://www-dev.greenpeace.org/test-phobos/?s=&orderby=_score) for your page
You will see the "Action types" section if you create and assign an [Action type](https://www-dev.greenpeace.org/test-phobos/wp-admin/edit-tags.php?taxonomy=action-type&post_type=p4_action)

comzeradd (approved):
Code looks good and tested locally again :+1:

PR #1694, opened 2022-04-28
PLANET-6645 Add pagination on other listing pages

Inwerpsel:
[PLANET-6645](https://jira.greenpeace.org/browse/PLANET-6645)

---

* All uses the same template for now.
* Create a feature per type, so that people can adjust their content if
needed.
* ~~For the post type, the pagination is not working. Probably due to our
permalink structure.~~ Looks like it was only a local issue, maybe cache related.

[Author page](https://www-dev.greenpeace.org/test-mars/author/amelekou/)
[Tag page](https://www-dev.greenpeace.org/test-mars/tag/consumption/)
[Category page](https://www-dev.greenpeace.org/test-mars/category/issues/) (doesn't make much sense with our old content structure)
[Post type page](https://www-dev.greenpeace.org/test-mars/press-release/)

The test instance makes use of a few dev only feature toggles to make testing easier:
* **"Post page category links"**, else there's no link to category pages
* **"Disable Tag Redirect Pages"**, to avoid doing it manually on many tags

### Todo
Move feature toggles. Before doing that I'll check to decouple the admin page from the option name, so that we can change the page independently.


planet-4:

:watch: 2022.04.28 12:41:09

src/Settings/Features.php
```diff
@@ -96,7 +99,10 @@ public static function all_features(): array {
 			NewDesignCountrySelector::class,
 			NewDesignNavigationBar::class,
 			GoogleSheetReplacesSmartsheet::class,
-			ListingPagePagination::class,
+			NewListingPagesTags::class,
```
Inwerpsel, line 102: I added a toggle per type, because we need to adjust the caching to work for all the types.

src/Settings/Features.php
```diff
@@ -113,7 +116,10 @@ public static function all_features(): array {
 			NewDesignCountrySelector::class,
 			NewDesignNavigationBar::class,
 			GoogleSheetReplacesSmartsheet::class,
-			ListingPagePagination::class,
+			TagPagePagination::class,
```
mleray, line 119: These new features show as `Dev only`, is that intended? Because from the code it seems they shouldn't 🤔 

Screenshot 2022-05-03 at 09 34 04

Inwerpsel: You mean the file location, not being in `/Dev/` namespace? If that, it's intended, creating that namespace was a mistake as we would need to rename stuff too much, it's nicer to have it grouped logically.

mleray: No, I meant that it's not located under the `// Dev only` comment a few lines below

Inwerpsel: It's a similar issue, it's not nice to maintain when a feature goes from dev to prod. Like the PR description mentions I'm still checking the improve the exact location and for these features.

mleray (approved):
Checked on the test instance, looks good to me 👍

sagarsdeshmukh:
The PR is rebased and tested on local and [test instance](https://www-dev.greenpeace.org/test-mars/author/amelekou/page/4/), the listing page pagination functionality is working as intended.
Just to double-check on the feature toggle options, 
_Should we remove all feature flags (tag, author, category, post type) and add a single checkbox (enable Listing page pagination) or we don’t need this feature toggle anymore?_

comzeradd:
> _Should we remove all feature flags (tag, author, category, post type) and add a single checkbox (enable Listing page pagination) or we don’t need this feature toggle anymore?_

Let's have on feature flag for all listing pages. We have a follow up ticket to to UI adjustments, so that will allow us to hold the release a bit till everything is in place. But we can move that feature flag to normal production set, in case some NROs are eager to try it out.

mleray (approved):
Checked on the test instance, works as expected and the code looks good to me 👍 
Only small thing is maybe we should put this `👷 Listing page grid view` setting closer to the `👷 Listing page pagination` one, since they are related? Right now I see it at the bottom of the list. Maybe the description could be improved too, not sure if the current one is super clear 🤔
Not a blocker though, just a suggestion!

Screenshot 2022-07-05 at 13 53 14

src/Settings/Features.php
```diff
@@ -116,6 +115,7 @@ public static function all_features(): array {
 			NewDesignNavigationBar::class,
 			GoogleSheetReplacesSmartsheet::class,
 			RedirectRedirectPages::class,
+			DisableTagRedirectPages::class,
```
comzeradd, line 118: I think that one is not needed at all

PR #1695, opened 2022-04-28
PLANET-6805: Change donate button text through the burger menu

GP-Dan-Tovbein:
Ref: https://jira.greenpeace.org/browse/PLANET-6805

Use the default provided from Planet4 > Donate button instead the "Make a donation" text. It avoids to use a duplicated donate button into the burger menu.

**New Logic**

- If the Donate button has not been created, then use the default one through the `navbar` and burger menu.
- If the donate button has been created, then show on the `navbar`. And if it has at least one item, then show it also in the burger menu.
- The "Make a donation" button is now replaced by the default settings.

**How to test**
Play around within the donate dropdown menu (Appearance > Menu) and the default donate button (Planet4 > Donate Button).

### DIFF
--- a/templates/burger-menu.twig
+++ b/templates/burger-menu.twig
@@ -47,11 +47,11 @@
 			<ul id="accordion-list" class="accordion">
 				<li class="burger-menu-item nav-item">
 					<a
-							class="nav-link"
-							href="{{ site.url }}"
-							data-ga-category="Menu Navigation"
-							data-ga-action="Home"
-							data-ga-label="{{ page_category }}">
+						class="nav-link"
+						href="{{ site.url }}"
+						data-ga-category="Menu Navigation"
+						data-ga-action="Home"
+						data-ga-label="{{ page_category }}">
 						{{ __( 'Home', 'planet4-master-theme' ) }}
 					</a>
 				</li>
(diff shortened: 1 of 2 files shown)


planet-4:

:watch: 2022.04.29 11:40:09

templates/burger-menu.twig
```diff
@@ -68,15 +68,14 @@
 	</div>
 
 	{% if donate_menu_items|first is not null %}
-		{% set donate_menu = donate_menu_items|first %}
```
comzeradd, line 71: Don't we still need that to get `href` value below?

templates/burger-menu.twig
```diff
@@ -60,23 +60,22 @@
...
+				{% if donate_menu_items is not empty and donate_menu_items[0].children|length >= 1 and dropdown_menu %}
 					{% include "burger-menu-items.twig" with { menu: donate_menu_items } %}
 				{% endif %}
 			</ul>
 		</nav>
 	</div>
 
-	{% if donate_menu_items|first is not null %}
-		{% set donate_menu = donate_menu_items|first %}
+	{% if donate_menu_default|first is not null %}
```
mleray, line 70: Because of the default values, `donate_menu_default` should never be empty right? Maybe we can remove this check 🤔

templates/burger-menu.twig
```diff
@@ -60,23 +60,22 @@
...
 		<div class="burger-menu-footer">
 			<a
-					class="btn btn-donate"
-					href="{{ donate_menu.link }}"
-					data-ga-category="Menu Navigation"
-					data-ga-action="Donate"
-					data-ga-label="{{ page_category }}">
-				{{ __( 'Make a donation', 'planet4-master-theme' ) }}
+				class="btn btn-donate"
```
mleray, line 73: The button "Donate" style will be removed, and these now use our primary style... We can keep `btn-donate`, especially for child themes, but maybe we should use `btn-primary` too? 🤔  Right now [this is duplicating some primary button styles](https://github.com/greenpeace/planet4-master-theme/blob/6923bd518aa3597b1686803e9719d87958a59f22/assets/src/scss/layout/_navbar.scss#L57-L69)
Could be done in a separate PR though 🙂

mleray (approved):
Tested on local, works as expected 👍 
I've left a couple of comments, but nothing blocking!

templates/burger-menu.twig
GP-Dan-Tovbein: If we change to `btn-primary`, we will have also to change something else from [here](https://github.com/greenpeace/planet4-master-theme/blob/278c0f242bb25bb646e78b9225d2533b0bb7f86b/assets/src/scss/components/_buttons.scss#L174-L176) and that styles are also being with other components. So, I agree to do it in a separate PR whenever we'll work on this removal.

GP-Dan-Tovbein:
There is a [ticket](https://jira.greenpeace.org/browse/PLANET-6698) already created about removing the donate button styles.

PR #1705, opened 2022-05-06
Ensure unique HTTPS URLs get purged

Inwerpsel:
[PLANET-6770](https://jira.greenpeace.org/browse/PLANET-6770)

* We hit the rate limit and are still sending batches with http URLs.
It's unclear if it would even still send the HTTPS URL.

Tested locally by commenting out [this code](https://github.com/greenpeace/planet4-master-theme/blob/0259049bab34a55957c023b59c0c547e6d0921ff/src/CloudflarePurger.php#L32-L45), so that it also works without having the plugin setup.


lithrel (approved):
:+1:

PR #1714, opened 2022-05-12
PLANET-6712 Add transparent style to core Button block

mleray:
### Description

See [PLANET-6712](https://jira.greenpeace.org/browse/PLANET-6712)
This is needed for some block patterns, for now only the Highlighted CTA

**Related PR**: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/849

### Testing

You can test both the Highlighted CTA pattern and regular buttons on this [page](https://www-dev.greenpeace.org/test-titan/transparent-buttons/) I made for UAT, with several background colors from our palette.

### DIFF
--- a/assets/src/scss/components/_buttons.scss
+++ b/assets/src/scss/components/_buttons.scss
@@ -296,7 +296,8 @@ button.load-more-mt {
   transition-duration: 100ms;
 }
 
-.wp-block-button.transparent-button a {
+.wp-block-button.transparent-button a,
+.wp-block-button.is-style-transparent a {
   --transparent-button-- {
     border-color: $white;
     color: $white;
--- a/assets/src/scss/editorOverrides.scss
+++ b/assets/src/scss/editorOverrides.scss
@@ -133,7 +133,8 @@
   }
 }
 
-.wp-block-button.transparent-button .wp-block-button__link[role="textbox"] {
+.wp-block-button.transparent-button .wp-block-button__link[role="textbox"],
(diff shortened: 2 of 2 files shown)


assets/src/scss/components/_buttons.scss
```diff
@@ -296,7 +296,8 @@ button.load-more-mt {
   transition-duration: 100ms;
 }
 
-.wp-block-button.transparent-button a {
+.wp-block-button.transparent-button a,
```
mleray, line 299: For now I've kept the old class to make sure existing Highlighted CTA patterns still look fine, but I'm not sure what would be best? This pattern probably hasn't been used a lot yet, so we should be able to manually update them, but I don't know if there's an easy way to find them since patterns don't appear in the blocks report... Maybe we need to run a migration script? 🤔

planet-4:

:watch: 2022.05.13 08:30:52

assets/src/scss/components/_buttons.scss
lithrel: I can think of ways to detect patterns, more or less reliable :grimacing: , but it would be a lot easier if we were adding a unique class to the container of each one. 
Sounds like a good idea to manually update while we can (or ping people on slack, they probably know better where their content is), by creating a ticket to do that and remove this rule after ?

lithrel (approved):
:+1:

assets/src/scss/components/_buttons.scss
mleray: Created a [ticket](https://jira.greenpeace.org/browse/PLANET-6780)

PR #1716, opened 2022-05-17
PLANET-6767: Remove control panel overlay blocking clicks

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-6767

> Icons on dashboard (flush part) are so big, I can't flush the cache anymore.

WP 5.9 background image is displayed over our links and prevent clicks. It is because we use the `.welcome-panel` class, that WP uses to display its "welcome new 5.9 version" background.

<table>
<tr><td>Before fix</td><td>After fix</td></tr>
<tr>
<td></td>
<td></td>
</tr></table>

- Removing this class fixes the issue
- Changing markup a bit to fit other panel boxes 

## Test

On the admin dashboard, check the Planet4 Control Panel.
- Links should be visible and clickable

### DIFF
--- a/admin/css/dashboard.css
+++ b/admin/css/dashboard.css
@@ -1,5 +1,14 @@
+.cp-item {
+  font-size:  14px;
+  padding:  5px 0;
+}
+
 .cp-item div {
   background-color: rgb(250, 250, 250);
+}
+
+#dashboard-widgets .cp-item h3 {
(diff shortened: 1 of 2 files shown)


mleray (approved):
Tested on local, looks good to me 👍 Just a small typo to fix in the commit title 😉

PR #1733, opened 2022-06-10
PLANET-6887 Fix placeholder color for Button block

mleray:
### Description

See [PLANET-6687](https://jira.greenpeace.org/browse/PLANET-6687)
Without this fix, the placeholder is white regardless of the actual text color.

### Testing

On local, you can test the fix with any empty Button block (without text) by changing the style (esp. `secondary` since it doesn't have white text color) or by picking a new text color from the palette. On this branch you should see the text color always being applied to the placeholder, whereas on `master` branch it would remain white.


GP-Dan-Tovbein (approved):
Code looks good to me!

PR #1734, opened 2022-06-13
Fix php notice when IA option unavailable

lithrel:
A PHP Notice appears when an IA option is not yet set

https://output.circle-artifacts.com/output/job/7c8a73d8-065a-4632-aa3b-b8f29ac061a3/artifacts/0/artifacts/codeception/report.html
Screenshot from 2022-06-10 11-18-20


comzeradd (approved):
Tested locally where indeed I was getting the warning before this fix :+1:

lithrel:
Ah thanks, I couldn't remember how to replicate it :man_facepalming:

PR #1737, opened 2022-06-13
Fix clear text button from search input

GP-Dan-Tovbein:
Ref: https://jira.greenpeace.org/browse/PLANET-6819

The issue is reproduced on the international live site but affected to all NROs
Screenshot 2022-06-13 at 12 21 53

Video:
https://user-images.githubusercontent.com/77975803/173387606-b9cffa8d-3681-4cfc-ab13-0440ee14eada.mov

### DIFF
--- a/assets/src/scss/layout/navbar/_search.scss
+++ b/assets/src/scss/layout/navbar/_search.scss
@@ -63,17 +63,13 @@
     border-inline-start: var(--top-navigation--separation, 1px solid transparentize($white, 0.5));
     border-top: none;
     box-shadow: none;
-    display: inline-block;
+    display: inline-flex;
     height: 100%;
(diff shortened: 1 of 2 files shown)


assets/src/scss/layout/navbar/_search.scss
```diff
@@ -63,17 +63,13 @@
...
+    display: inline-flex;
     height: 100%;
     inset-inline-end: auto;
     min-width: 25%;
     padding: 0 16px;
     position: unset;
     width: auto;
-
-    &.open {
-      display: inline-block;
```
GP-Dan-Tovbein, line 75: I don't think we need this rule applied to XL screens

lithrel (approved):
:+1:

PR #1741, opened 2022-06-15
Allow articles block in posts

GP-Dan-Tovbein:
Ticket: https://jira.greenpeace.org/browse/PLANET-6723

According to this Slack thread, we need to apply new styles only when the Article is included into a post.

This PR is blocked this other [PR](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/870). More info there!

IMPORTANT: We don't need to use the Ieda's instance for a demo page.

### DIFF
--- a/assets/src/scss/pages/post/_article-content.scss
+++ b/assets/src/scss/pages/post/_article-content.scss
@@ -219,6 +219,58 @@
   q {
     font-style: italic;
   }
+
+  .article-list-item {
+    .top-page-tags {
+      margin-bottom: $sp-1;
+    }
+
+    .article-list-item-headline {
+      padding-top: 0;
+      padding-bottom: 0;
+      -webkit-line-clamp: 2;
+    }
+
+    .article-list-item-content {
+      width: 100%;
+      -webkit-line-clamp: 3;
+
+      @include medium-and-up {
+        -webkit-line-clamp: 2;
+      }
+
+      @include large-and-up {
+        -webkit-line-clamp: 3;
+      }
+
+      @include x-large-and-up {
+        -webkit-line-clamp: 3;
+      }
+    }
+
+    .article-list-item-headline,
+    .article-list-item-content {
+      margin-bottom: $sp-1;
+      display: -webkit-box;
+      -webkit-box-orient: vertical;
+      overflow: hidden;
+    }
+
+    .article-list-item-image img {
+      @include medium-and-up {
+        height: 180px;
+      }
+
+      @include large-and-up {
+        height: 210px;
+      }
+    }
+
+    .article-list-item-meta {
+      margin-bottom: 0;
+      line-height: 24px;
+    }
+  }
 }
 
 .post-details {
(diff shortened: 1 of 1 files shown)


mleray (commented):
Looking at the designs, it seems the comment form should be full width right? Currently we still have [some `max-width` values](https://github.com/greenpeace/planet4-master-theme/blob/master/assets/src/scss/pages/post/_comments-block.scss#L66-L76) set for bigger screen sizes, which don't always align with the post content width 🤔 They can probably be removed!

templates/single.twig
```diff
@@ -106,18 +106,12 @@
...
+
+			{% if ( show_comments ) %}
+				{% include "comments_section.twig" with {comments:post.get_comments()} %}
+			{% endif %}
 		</div>
 		<!-- Post Page Block End -->
-
-		{% include "blocks/author_profile.twig" with {post:post} %}
-
-		{% if ( post.articles ) %}
```
lithrel, line 114: We need to keep this functionality "Include articles in post" (option at the bottom of the post editor), so you should either keep everything here and put it in a specific div to give proper width, or move that conditional code with the rest

templates/comments_section.twig
```diff
@@ -1,4 +1,4 @@
-<div class="container" id="comments">
```
GP-Dan-Tovbein, line 1: I removed the `.container` since I prefer to its width be controlled by the [parent](https://github.com/greenpeace/planet4-master-theme/blob/9da984962ad54767037327c55f5a50909f7bcf37/templates/single.twig#L86).
Also I've checked and only the post template is the only template that uses tho comment_section.

assets/src/scss/pages/post/_article-content.scss
```diff
@@ -284,3 +284,23 @@
...
+  .top-page-tags {
+    margin-bottom: $sp-1;
+  }
+
+  .article-list-item-headline {
+    padding-bottom: 0;
+    margin-bottom: $sp-1;
+  }
+
+  .article-list-item-content {
```
GP-Dan-Tovbein, line 298: There is a similar behavior [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/PLANET-6723_allow_articles_block_in_posts/assets/src/styles/blocks/Articles.scss#L125-L134) but only applied to `mobile`. In our case we need to apply new styles only when the Article is included into a post.

lithrel (approved):
:+1:

comzeradd (changes requested):
On the [demo page](https://www-dev.greenpeace.org/test-oberon/story/51096/post-title-goes-here-it-can-be-this-long/) I see that the whole section outside the post (Author, Comments, Related Articles) is altered to be using a narrow container. That's outside the scope of this ticket. See [discussion](https://greenpeace-gpi.slack.com/archives/G015K63081W/p1655815987151759?thread_ts=1655810778.399749&cid=G015K63081W).

comzeradd (approved):

assets/src/scss/pages/post/_article-content.scss
```diff
@@ -219,6 +219,57 @@
...
+      -webkit-box-orient: vertical;
+      overflow: hidden;
+    }
+
+    .article-list-item-image img {
+      @include medium-and-up {
+        height: 180px;
+      }
+
+      @include large-and-up {
```
GP-Dan-Tovbein, line 264: Unfortunately, we'll need to duplicate this rule from [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/styles/blocks/Articles.scss#L44-L46), since the image's height is overridden. The other approach is to create a new `@media` only to M screens with something like this:

```
@mixin medium-only {
  @media (min-width: #{$medium-width}) and (max-width: #{$large-width - 1}) {
    @content;
  }
}
```
I'd suggest to implement it like this, and whenever we modify the `Articles block` (I hope soon) we would revert and standardize the code, by using only the [Articles styles](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/styles/blocks/Articles.scss) and remove all these overrides.

PR #1753, opened 2022-07-07
PLANET-6845: Fix menu config loading

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-6845

> Houssam and I noticed that WPML sites, such as Belgium or MENA, has the Menu restrictions not working as expected.

Because of a mistake in a condition, if the Donate menu location doesn't exist on the instance, the menu configuration limitations are not applied to the editor.
Fixing the condition to work if at least one of the menus location exists.

### DIFF
--- a/src/AdminAssets.php
+++ b/src/AdminAssets.php
@@ -26,7 +26,9 @@ class AdminAssets {
 		$navbar_location = 'navigation-bar-menu';
 		$donate_location = 'donate-menu';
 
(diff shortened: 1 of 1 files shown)


GP-Dan-Tovbein (approved):
Code looks good to me! Nice fix :) 

Also validated against the [MENA's dev instance](https://github.com/greenpeace/planet4-mena/blob/main/development.json#L3) , which has been leading with this issue, and it worked as expected.
Screenshot 2022-07-11 at 07 14 09

PR #1762, opened 2022-07-29
PLANET-6666: Load GravityForms styles only with block on page

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-6666

> GravityForms (and the Addons we use) may enqueue some styles and scripts by default even if there is no form block in the page.
> Our own GF styling from [PLANET-6555](https://jira.greenpeace.org/browse/PLANET-6555) also needs to be enqueued only when a form is present on a page.

Gravity Forms already loads its scripts and styles only when a form is called the current page.  
This PR extracts our own styles to load them only when a GF block is there.


planet-4:

:watch: 2022.08.12 09:03:26

assets/src/scss/base/_typography.scss
```diff
@@ -1,3 +1,5 @@
+@import "fonts";
+
```
comzeradd, line 2: Couldn't we just import `_fonts` too in GF styles below? (like we do in the [main style file](https://github.com/greenpeace/planet4-master-theme/blob/master/assets/src/scss/style.scss#L20-L21))

lithrel: Yes good point

comzeradd (approved):

PR #1776, opened 2022-08-24
PLANET-6853 Add parallax to PageHeader images

mleray:
### Description

See [PLANET-6853](https://jira.greenpeace.org/browse/PLANET-6853)
This is done via JS. There is also [another PR](https://github.com/greenpeace/planet4-master-theme/pull/1783) to investigate using a library instead.

**Related PR:**
- https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/931

### Testing

You can test the new parallax effect either on your local with any Page Header pattern, or on [this page](https://www-dev.greenpeace.org/test-leda/multiple-page-headers/) that I made for UAT!


planet-4:

:watch: 2022.08.24 11:42:01

assets/src/js/page_header_parallax.js
```diff
@@ -0,0 +1,8 @@
+export const setupPageHeaderParallax = () => {
+  const pageHeaders = [...document.querySelectorAll('.is-pattern-p4-page-header > .wp-block-media-text__media > img')];
+
+  window.addEventListener('scroll', () => {
+    const scrollPosition = window.scrollY;
+    pageHeaders.forEach(pageHeader => pageHeader.style.transform = 'translateY(' + scrollPosition * 0.6 + 'px)');
```
GP-Dan-Tovbein, line 6: What do you think if you change to this ?   
```
pageHeaders.forEach(pageHeader => pageHeader.style.transform =`translateY(${scrollPosition * 0.6}px)`
```
Do you see it more readable?

assets/src/js/page_header_parallax.js
```diff
@@ -0,0 +1,8 @@
+export const setupPageHeaderParallax = () => {
+  const pageHeaders = [...document.querySelectorAll('.is-pattern-p4-page-header > .wp-block-media-text__media > img')];
```
GP-Dan-Tovbein, line 2: I don't see necessary a spread syntax (...) here.
`const pageHeaders = document.querySelectorAll('.is-pattern-p4-page-header > .wp-block-media-text__media > img');` it's fine

mleray: It works fine indeed, but it's not an array, it's a `NodeList` which is a bit different. It's ok for now, but it might cause issues in case we want to use some other array-specific functions 🙂 I can change it if you prefer 👍

mleray:
One limitation of this implementation is that it will only work if the Page Header pattern is placed first in the page (also without a page title) but I guess we should take into account that it could be anywhere on the page? I don't think we enforce its positioning at the moment 🤔

assets/src/js/page_header_parallax.js
GP-Dan-Tovbein: > One limitation of this implementation is that it will only work if the Page Header pattern is placed first in the page (also without a page title) but I guess we should take into account that it could be anywhere on the page? I don't think we enforce its positioning at the moment 🤔

Gotcha! Did you try if setting `background-attachment` as `fixed` to the image behaves similar?

mleray: It's not gonna work here, because the Media & Text block doesn't use `background` CSS properties but an actual image tag... I think it should be doable however to find the position of the Page Header, and determine the scrolling effect based on that!

mleray:
@GP-Dan-Tovbein I [updated](https://github.com/greenpeace/planet4-master-theme/pull/1776/commits/6d96010f72e8a9db53524d2a4eaccd01e1ee9db4) the code to take the pattern's position into account!
I've also added a [commit](https://github.com/greenpeace/planet4-master-theme/pull/1776/commits/42da2076c2b70336ae72a0502f534d2bcdb7c7ba) to make sure that this parallax is only applied to large screens and up, since we don't want that behaviour in mobile/tablet devices.

assets/src/js/page_header_parallax.js
```diff
@@ -0,0 +1,38 @@
+const isSmallWindow = () => window.innerWidth < 992;
```
GP-Dan-Tovbein, line 1: What do you think to nest these local functions into the `setupPageHeaderParallax` scope?

mleray: Then they would get defined every single time we call the function, whereas here we only define them once 🤔

GP-Dan-Tovbein (approved):
Code looks good to me! Good job :)

assets/src/js/page_header_parallax.js
GP-Dan-Tovbein: Just in case, I meant this https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Functions#nested_functions_and_closures. We could keep it for now!

mleray:
Closing this PR in favor of using the Rellax library (https://github.com/greenpeace/planet4-master-theme/pull/1783)

PR #1795, opened 2022-09-19
PLANET-6907 Update Gravity Forms settings initial values

mleray:
### Description

See [PLANET-6907](https://jira.greenpeace.org/browse/PLANET-6907)
By default we want the sub-label and description to be above inputs instead of below

### Testing

I've created a new form on the test instance for UAT, in its [settings](https://www-dev.greenpeace.org/test-phoebe/wp-admin/admin.php?page=gf_edit_forms&view=settings&subview=settings&id=1) you can see that now by default both fields are set to "Above inputs". On local or on the instance, you can also add a new form yourself and make sure that the initial values are properly set.

### DIFF
--- a/src/GravityFormsExtensions.php
+++ b/src/GravityFormsExtensions.php
@@ -2,6 +2,8 @@
 
 namespace P4\MasterTheme;
 
+use GFAPI;
+
 /**
  * Class P4\MasterTheme\GravityFormsExtensions
  * The Gravity form plugin extension, use to add custom functionality like Planet4 form type.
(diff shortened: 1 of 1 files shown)


sagarsdeshmukh (dismissed):
LGTM 👍

sagarsdeshmukh:
Hi @mleray , I tested the changes on local, looks good to me 👍 , In my opinion, if we keep all Gravity form related changes/customisation in one place like a [GravityFormsExtensions.php](https://github.com/greenpeace/planet4-master-theme/blob/master/src/GravityFormsExtensions.php) file will be better to maintain in future. 🙂

sagarsdeshmukh (approved):
👍

PR #1800, opened 2022-09-22
Check if share platforms are set for GF confirmation message

mleray:
### Description

Without this check, if editors didn't manually save the confirmation message then the buttons don't show in the frontend (and you see some warnings on local). If not defined, we set them to true since we want to show them by default. 

### Testing

You can test this behaviour on local by creating a new form and not editing/saving the confirmation message. On this branch the share buttons should then show in the confirmation message, whereas on `master` they won't. Another option to fix this issue would maybe be to set the share buttons values for new forms in [this after save function](https://github.com/greenpeace/planet4-master-theme/blob/master/src/GravityFormsExtensions.php#L167), but that wouldn't solve the problem for existing forms which is the reason why I've gone for this fix instead!

### DIFF
--- a/src/GravityFormsExtensions.php
+++ b/src/GravityFormsExtensions.php
@@ -278,11 +278,11 @@ class GravityFormsExtensions {
 		$confirmation_fields = [
 			'confirmation'    => $default_confirmation['message'],
 			'share_platforms' => [
-				'facebook' => '1' === $default_confirmation['facebook'] ?? false,
-				'twitter'  => '1' === $default_confirmation['twitter'] ?? false,
-				'whatsapp' => '1' === $default_confirmation['whatsapp'] ?? false,
-				'native'   => '1' === $default_confirmation['native'] ?? false,
-				'email'    => '1' === $default_confirmation['email'] ?? false,
+				'facebook' => $default_confirmation['facebook'] ?? true,
+				'twitter'  => $default_confirmation['twitter'] ?? true,
+				'whatsapp' => $default_confirmation['whatsapp'] ?? true,
+				'native'   => $default_confirmation['native'] ?? true,
+				'email'    => $default_confirmation['email'] ?? true,
 			],
 			'share_text'      => $default_confirmation['p4_gf_share_text_override'] ?? '',
 			'share_url'       => $default_confirmation['p4_gf_share_url_override'] ?? '',


planet-4:

:watch: 2022.09.22 13:05:11

src/GravityFormsExtensions.php
```diff
@@ -278,11 +278,11 @@ function ( $conf ) use ( $confirmation ) {
...
-				'facebook' => '1' === $default_confirmation['facebook'] ?? false,
-				'twitter'  => '1' === $default_confirmation['twitter'] ?? false,
-				'whatsapp' => '1' === $default_confirmation['whatsapp'] ?? false,
-				'native'   => '1' === $default_confirmation['native'] ?? false,
-				'email'    => '1' === $default_confirmation['email'] ?? false,
+				'facebook' => '1' === isset( $default_confirmation['facebook'] ) ? $default_confirmation['facebook'] : true,
+				'twitter'  => '1' === isset( $default_confirmation['twitter'] ) ? $default_confirmation['twitter'] : true,
+				'whatsapp' => '1' === isset( $default_confirmation['whatsapp'] ) ? $default_confirmation['whatsapp'] : true,
+				'native'   => '1' === isset( $default_confirmation['native'] ) ? $default_confirmation['native'] : true,
+				'email'    => '1' === isset( $default_confirmation['email'] ) ? $default_confirmation['email'] : true,
```
sagarsdeshmukh, line 285: Damm!!! 😬 I under look this condition because of  Null Coalescing Operator(??) operator usage, but because of 2 expression next to each other and as per Operator Precedence the first expression evaluated first (`'1' === $default_confirmation['email']`) which of-course produces undefined variable warning.

Here we can also simplifies it as -
`'email'    => $default_confirmation['email'] ?? true`

sagarsdeshmukh (approved):
Looks good, just a small suggestion :)

src/GravityFormsExtensions.php
```diff
@@ -278,11 +278,11 @@ function ( $conf ) use ( $confirmation ) {
 		$confirmation_fields = [
 			'confirmation'    => $default_confirmation['message'],
 			'share_platforms' => [
-				'facebook' => '1' === $default_confirmation['facebook'] ?? false,
-				'twitter'  => '1' === $default_confirmation['twitter'] ?? false,
-				'whatsapp' => '1' === $default_confirmation['whatsapp'] ?? false,
-				'native'   => '1' === $default_confirmation['native'] ?? false,
-				'email'    => '1' === $default_confirmation['email'] ?? false,
+				'facebook' => '1' === isset( $default_confirmation['facebook'] ) ? $default_confirmation['facebook'] : true,
```
GP-Dan-Tovbein, line 281: Correct if I'm wrong, but this condition will always return `true` or am I missing something?

mleray: Yeah I made a mistake in the first implementation, the PR has been updated!

src/GravityFormsExtensions.php
```diff
@@ -278,11 +278,11 @@ function ( $conf ) use ( $confirmation ) {
 		$confirmation_fields = [
 			'confirmation'    => $default_confirmation['message'],
 			'share_platforms' => [
-				'facebook' => '1' === $default_confirmation['facebook'] ?? false,
-				'twitter'  => '1' === $default_confirmation['twitter'] ?? false,
-				'whatsapp' => '1' === $default_confirmation['whatsapp'] ?? false,
-				'native'   => '1' === $default_confirmation['native'] ?? false,
-				'email'    => '1' === $default_confirmation['email'] ?? false,
+				'facebook' => isset( $default_confirmation['facebook'] ) ? '1' === $default_confirmation['facebook'] : true,
```
GP-Dan-Tovbein, line 281: I have one comments and one comment about this,

1. What is supposed to returns each `$default_confirmation[...]`? `0` or `1`?
2. Because all this social media have applied the same logic, you can move it to a function, like:
```
function fnName (value): boolean {
 return isset(value) ? '1' === value : true
}
```
And then 
```
'facebook' => fnName( $default_confirmation['facebook'] ),
'twitter' => fnName( $default_confirmation['twitter'] ),
...
```
What do you think?

mleray: Added a [commit](https://github.com/greenpeace/planet4-master-theme/pull/1800/commits/eea944b0ee45bb8a0aab31586f296ae545b16a30) with the suggested refactoring!

mleray: I actually went with [Sagar's suggested refactoring](https://github.com/greenpeace/planet4-master-theme/pull/1800#discussion_r977767326) since it simplifies the code quite a bit and seems to work just as well 🙂

sagarsdeshmukh (approved):
👍

PR #1825, opened 2022-10-20
PLANET-6678 Alphabetically order Planet 4 settings menu

Osong-Michael:
**Description**

See [PLANET-6678](https://jira.greenpeace.org/browse/PLANET-6678)

This PR adds a filter to alphabetically sort the Planet 4 settings menu.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -462,3 +462,12 @@ add_filter(
 		return $vars;
 	}
 );
+
+// Action to filter P4 settings menu.
+add_action(
+	'admin_head',
+	function() {
+		global $submenu;
+		uasort( $submenu['planet4_settings_navigation'], fn ( $a, $b ) => $a[0] <=> $b[0] );
+	}
+);


planet-4:

:watch: 2022.10.20 12:43:42

lithrel (commented):
@Osong-Michael  A few comments, let me know if you need more details :wave: 
WordPress has 2 different hooks, [actions and filters](https://developer.wordpress.org/plugins/hooks/#actions-vs-filters); filters are usually used to modify a value before it is used (that value is modified and returned by the function declared for the hook). For example the [`custom_menu_order`](https://developer.wordpress.org/reference/hooks/custom_menu_order/) hook is useful to declare if you allow or not to modify the top level menu order, and it has to return a `bool` to say that. In PHP if a function returns nothing it will [always return `null`](https://www.php.net/manual/en/functions.returning-values.php), and `null` is a loose equivalent of `false`.  
Knowing all of that, it means that your function will set that value to `false` [there](https://github.com/WordPress/wordpress-develop/blob/419e8248d61137d42b537bf2e6880f93a1ea361a/src/wp-admin/includes/menu.php#L273) which is not what you intended :) (and could cause unexpected behavior if an NRO child theme has previously declared this value to `true`).
You are modifying data in this function, but it's not data coming from the filter itself (and I couldn't find one that passes submenus :panda_face:), so I would use an action hook for that, the best I could see is `admin_head`, so `add_action('admin_head', your function);`.
For the sorting part, it does work because of the [default way PHP will sort arrays](https://www.php.net/manual/en/language.operators.comparison.php#language.operators.comparison.types), but
- because it's not explicit you could add the sorting method in comments (Sorting by title),
- and because we might have issues if all the entries have a different count of elements you could write an explicit callback with [`uasort()`](https://www.php.net/manual/en/function.uasort.php) like `fn ( $a, $b ) => $a[0] <=> $b[0]` knowing that the Title is the first element of each entry.

lithrel (approved):
:+1:

PR #1842, opened 2022-11-14
PLANET-6914 Prepare to upgrade to Wordpress 6.1.1

mleray:
### Description

See [PLANET-6914](https://jira.greenpeace.org/browse/PLANET-6914)
Some small fixes were needed before the upgrade:
- remove the new `text-transform` added to labels in the editor
- force `0 1rem` padding in the editor block area
- move the `text-decoration` changes for links to `theme.json`

I've created 2 follow-up tickets that would require more work and/or some investigation:
- [PLANET-6958](https://jira.greenpeace.org/browse/PLANET-6958) to look into our RichText components that currently use the deprecated `multiline` prop, since it's probably gonna require quite a lot of changes and it's less urgent because we have until WP 6.3 before that prop is completely removed
- [PLANET-6965](https://jira.greenpeace.org/browse/PLANET-6965) to adapt the page creation modal to all post types

[The PR for the actual upgrade](https://github.com/greenpeace/planet4-base/pull/226) is in the base repo.

### Testing

You can test the changes either on local or on the [janus test instance](https://www-dev.greenpeace.org/test-janus). To see how it is before these fixes, you can compare it with the [gutenberg instance](https://www-dev.greenpeace.org/gutenberg/) which is already running WP 6.1!

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -1,6 +1,25 @@
 {
   "$schema": "https://schemas.wp.org/trunk/theme.json",
   "version": 2,
+  "styles": {
+		"elements": {
+			"link": {
+				"typography": {
+					"textDecoration": "none"
+				},
+				":hover": {
+					"typography": {
+						"textDecoration": "underline"
+					}
+				},
+				":active": {
+					"typography": {
+						"textDecoration": "underline"
+					}
+				}
+			}
+		}
+	},
   "settings": {
     "color": {
       "custom": false,
--- a/assets/src/scss/base/_typography.scss
+++ b/assets/src/scss/base/_typography.scss
@@ -106,7 +106,6 @@ h6 {
 
 a --link-- {
   color: $link-color;
-  text-decoration: none;
 
   &:focus {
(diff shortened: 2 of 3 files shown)


planet-4:

:watch: 2022.11.15 11:07:11

theme.json
```diff
@@ -1,6 +1,25 @@
 {
   "$schema": "https://schemas.wp.org/trunk/theme.json",
   "version": 2,
+  "styles": {
```
mleray, line 4: Maybe we could move the link color changes there as well, but since these are not broken by the upgrade I figured it might be better to do this in a separate PR!
Also, I'm not sure that we would want this since it might make it more difficult for NROs to customise? I don't know if they can use `theme.json` for that 🤔

comzeradd: Can you create a follow up ticket with what we can move over to `theme.json`? It seems that's technically possible to ovveride things through its own `theme.json` file ([example](https://wpastra.com/docs/override-astras-theme-json/)). We could document that process.

comzeradd (approved):

PR #1843, opened 2022-11-16
PLANET-6961: Action Type listing page empty when no pagination

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-6961

Queries ending up in the `taxonomy.php` template are the [Custom taxonomy](https://developer.wordpress.org/themes/template-files-section/taxonomy-templates/#custom-taxonomy) ones.
We have 2 different custom taxonomies,
- `p4-page-type` (cf. `/edit-tags.php?taxonomy=p4-page-type`) 
- `action-type` (cf. `/edit-tags.php?taxonomy=action-type`).

## Fix

- Change query args to fetch any valid post type, not only `post`
- Rebase query on taxonomy query instead of static  `p4-page-type` taxonomy
- Force `tax-p4-page-type` class [used for font-family](https://github.com/greenpeace/planet4-master-theme/blob/1dc9c0714f32112fc5a05ee433541cdc5c9e29ac/assets/src/scss/pages/_page_type.scss#L1)

## Test

- Activate option `Action post type` in _Planet 4 > Information architecture_
- Create a new Action type in _Actions > Action type_
- Create a new Action and associate it to your new Action type
- Check the listing page, with option `Listing page pagination` in _Planet 4 > Information architecture_ enabled and disabled
  - with listing disabled, you should still see some content

### DIFF
--- a/taxonomy.php
+++ b/taxonomy.php
@@ -13,6 +13,8 @@
 use P4\MasterTheme\Features\Dev\ListingPageGridView;
 use P4\MasterTheme\Features\ListingPagePagination;
 use P4\MasterTheme\Post;
+use P4\MasterTheme\ActionPage;
+use P4\MasterTheme\Features\ActionPostType;
 use Timber\Timber;
 
 $templates = [ 'taxonomy.twig', 'index.twig' ];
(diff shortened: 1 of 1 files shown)


comzeradd (approved):
Looks good, and tested locally.

There are some other issues I noticed with Action Type listing pages, that have different UI based on whether pagination is enabled or not. But I'll open a follow up ticket for that.

PR #1851, opened 2022-11-24
PLANET-6981 Remove jQuery from menu_editor.js

mleray:
### Description

The functionality looks a bit different now, but it's more consistent with the other possible errors. We could also look into adding `pointer-events: none;` to the save button when there are errors, to make sure that editors cannot submit the new menu.

### Testing

In the backend in the Menus section, when editing the main navigation you should now see an error message show if you try to add menu items at a depth of more than 1.

PR #1852, opened 2022-11-29
PLANET-6936 Remove jQuery from frontend

mleray:
### Description

See [PLANET-6936](https://jira.greenpeace.org/browse/PLANET-6936)
This should not cause any issues.
I've opened some follow-up PRs to stop using jQuery in a few other places in the admin area (so not totally related to this ticket):
- [dashboard.js](https://github.com/greenpeace/planet4-master-theme/pull/1853)
- [addButtonLinkPasteWarning.js](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/977)
- [menu_editor.js](https://github.com/greenpeace/planet4-master-theme/pull/1851)

### Testing

All should still work as expected in the site (blocks, admin code, plugins, etc)

### DIFF
--- a/src/PublicAssets.php
+++ b/src/PublicAssets.php
@@ -12,20 +12,6 @@ final class PublicAssets {
 	public static function enqueue_js() {
 		$js_creation = filectime( get_template_directory() . '/assets/build/index.js' );
 
-		$jquery_should_wait = is_plugin_active( 'planet4-plugin-gutenberg-blocks/planet4-gutenberg-blocks.php' ) && ! is_user_logged_in();
-
-		$jquery_deps = $jquery_should_wait ? [ 'planet4-blocks-script' ] : [];
-
-		// JS files.
-		wp_deregister_script( 'jquery' );
-		wp_register_script(
-			'jquery',
-			'https://cdnjs.cloudflare.com/ajax/libs/jquery/3.3.1/jquery.min.js',
-			$jquery_deps,
-			'3.3.1',
-			true
-		);
-
 		$theme_dir = get_template_directory_uri();
 		// Variables reflected from PHP to the JS side.
 		$localized_variables = [
--- a/package-lock.json
+++ b/package-lock.json
@@ -12525,11 +12525,6 @@
         }
       }
     },
-    "jquery": {
-      "version": "3.5.1",
-      "resolved": "https://registry.npmjs.org/jquery/-/jquery-3.5.1.tgz",
-      "integrity": "sha512-XwIBPqcMn57FxfT+Go5pzySnm4KWkT1Tv7gjrpT1srtf8Weynl6R273VJ5GjkRb51IzMp5nbaPjJXMWeju2MKg=="
-    },
     "js-base64": {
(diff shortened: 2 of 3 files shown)


planet-4:

:watch: 2022.11.29 13:21:41

comzeradd:
We should probably also update `package-lock.json` in this PR. I think that if your run `make install-deps` locally it should re-create it.

src/PublicAssets.php
```diff
@@ -12,20 +12,6 @@ final class PublicAssets {
 	public static function enqueue_js() {
 		$js_creation = filectime( get_template_directory() . '/assets/build/index.js' );
 
-		$jquery_should_wait = is_plugin_active( 'planet4-plugin-gutenberg-blocks/planet4-gutenberg-blocks.php' ) && ! is_user_logged_in();
-
-		$jquery_deps = $jquery_should_wait ? [ 'planet4-blocks-script' ] : [];
-
-		// JS files.
-		wp_deregister_script( 'jquery' );
```
lithrel, line 20: The version in WP will still be included if you don't remove the dependency on l.41/27 when registering index.js

src/PublicAssets.php
```diff
@@ -38,7 +24,7 @@ public static function enqueue_js() {
 		wp_register_script(
 			'main',
 			$theme_dir . '/assets/build/index.js',
-			[ 'jquery' ],
+			[],
```
comzeradd, line 27: Hey @lithrel @mleray if we leave this intact for a while would that include WP jquery version? Just thinking about letting this for now to give a bit more time on child themes for migration.

lithrel: right, it does force the inclusion of jQuery 3.6.1 coming from `wp-includes/js`.

mleray: Even when we remove this it's still included from `wp-includes/js` 🤔

mleray: I don't know how/if we can solve that

mleray: Looks like maybe we could use WordPress' [wp_deregister_script](https://developer.wordpress.org/reference/functions/wp_deregister_script/) function

lithrel: > Even when we remove this it's still included from `wp-includes/js` thinking

It's not a cache issue ? works on local for me, might be another block/plugin loading it on it's side depending on your page ?

mleray: It works as expected for me on local now that I've added [this commit](https://github.com/greenpeace/planet4-master-theme/pull/1852/commits/42bddbaf25f4460f9670a0d0905395a45ed71843)

mleray: Of course if we want to keep it for a while for NROs as suggested by @comzeradd, we would need to remove that

comzeradd: Yes, let's revert this last commit and re-add the change above in the diff to make sure jquery is still enqueued in the frontend. We can then merge the PR and re-test this when we want to completely remove the library.

PR #1860, opened 2022-12-08
Add command for WPML media path migration

comzeradd:
Adding this command to be able to re-run the media path fix migration after WPML activation on site the migration has already run.

### Testing

I couldn't reproduce Aotearoa's issue locally even with their latest production db. So the command does run properly locally but didn't see any visible changes.

I'll deploy this branch in Aotearoa's dev site, so we can test it before merge.


lithrel (approved):
:+1:

comzeradd:
Tested it on Aotearoa's dev site and media libraries older images seem fixed for Maori language.

PR #1870, opened 2022-12-22
Remove duotone feature the proper way

comzeradd:
This is a follow up from 24cc5ec49fe0cec09289deb9680e186274423104

It was breaking theme.json scheme validation. This change is how duotone feature is supposed to be disabled.

This also includes a fist commit to re-indent `theme.json` for removing tabs.

### Testing

Image block should not have the [duotone button](https://wordpress.com/go/tutorials/how-to-use-the-duotone-image-filter-with-wordpress-com/) in its toolbar

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -50,11 +50,12 @@
   "settings": {
     "color": {
       "custom": false,
+      "duotone": [],
       "customDuotone": false,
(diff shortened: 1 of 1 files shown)


mleray (approved):

PR #1871, opened 2022-12-23
Add border radius feature to core/image

GP-Dan-Tovbein:
By Doing this we'll be able to potentially remove some custom classes for borders from [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/main/assets/src/styles/blocks/core-overrides/Image.scss#L1-L12).

```sass
@mixin rounded-image-size($size) {
  figure,
  img {
    border-radius: 50%
    height: $size
    width: $size
  }

  img {
    object-fit: cover
  }
}
```
and
```sass
&.is-style-rounded-180 {
    @include rounded-image-size(180px)
  }

  &.is-style-rounded-90 {
    @include rounded-image-size(90px)
  }
```
## What to do?
1) Merge this feature
2) Open a new PR removing all the appearances of those classes. For example from the [Quick Links block pattern](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/main/assets/src/block-templates/quick-links/template.js#L8).

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -169,7 +169,12 @@
             }
           ]
         }
-      }
+      },
+      "core/image": {
+        "border": {
+          "radius": true
+        }
+      },
     },
     "layout": {
       "contentSize": "736px"


planet-4:

:watch: 2022.12.26 17:38:40

mleray (commented):
This change will solve the border-radius issue, but the classes were also added to force the size of the images, is there a solution for that now too? 🤔 Otherwise we won't be able to completely remove the classes, we'll just be able to simplify their CSS code a bit

GP-Dan-Tovbein:
> This change will solve the border-radius issue, but the classes were also added to force the size of the images, I guess we would need to add a class for each size like the `w-40` one [you added for the Issues template](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/988/files#diff-8342cb8eaccd7517ed01f546e094c49fe382ffafe11e8261d1b97f63c09323f5R37)?

Correct me if I see correctly your point! We wouldn't need to add custom class for `border-radius` since it's currently a default feature. I think the editor could now control that doesn't matter which image size he/she prefers. Am I right?

PR #1879, opened 2023-01-18
PLANET-6982 Add Gravity forms language translation files

sagarsdeshmukh:
Copy GF language files from https://www.gravityforms.com/my-account/downloads/translations/


comzeradd (approved):

comzeradd:
Can you also adjust the [zip size check](https://github.com/greenpeace/planet4-master-theme/blob/main/.circleci/config.yml#L382)? It seems that the zip file now is bigger than 5M. But since these zip files are used for deployment I'm wondering whether we could actually exclude these files 🤔

sagarsdeshmukh:
> Can you also adjust the [zip size check](https://github.com/greenpeace/planet4-master-theme/blob/main/.circleci/config.yml#L382)? It seems that the zip file now is bigger than 5M. But since these zip files are used for deployment I'm wondering whether we could actually exclude these files 🤔 Which we probably need there to rsync them to the right place.

Indeed the file size increased by 3.3MB because of plugin translation files, I updated the zip size check with 9MB, and its working fine now 👍

planet-4:

:watch: 2023.01.26 14:56:40

PR #1914, opened 2023-02-08
PLANET-7020 Add new site identity feature toggle

mleray:
### Description

See [PLANET-7020](https://jira.greenpeace.org/browse/PLANET-7020) & [PLANET-6984](https://jira.greenpeace.org/browse/PLANET-6984)

### Testing

When checking the new site identity toggle in `Appearance > Customize > Site identity`, you should see in the frontend and in the editor that the new value of `--body--color` is applied (#1c1c1c instead of #020202). You'll need to also use [this branch](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/1014) from the blocks plugin to see the change applied properly.

### DIFF
--- a/assets/src/scss/editorStyle.scss
+++ b/assets/src/scss/editorStyle.scss
@@ -22,3 +22,6 @@
 @import "layout/gravity-forms";
 
 @import "editorOverrides";
+
+// CSS variables
+@import "base/css-variables";
--- a/assets/src/scss/style.scss
+++ b/assets/src/scss/style.scss
@@ -73,3 +73,6 @@ Text Domain: planet4-master-theme
 @import "vendors/usabilla";
 @import "vendors/gtm";
 @import "~lite-youtube-embed";
+
+// CSS variables
+@import "base/css-variables";
--- a/assets/src/scss/base/_css-variables.scss
+++ b/assets/src/scss/base/_css-variables.scss
@@ -1,0 +1,12 @@
+:root {
+  --body--font-family: #{$lora};
+  --headings--font-family: #{$roboto};
+  --headings--font-weight: bold;
+  --body--background-color: #{$white};
+  --body--color: #{$grey-80};
+  --link--color: #{$link-color};
+  --link--hover--color: #{$link-color};
+  --link--text-decoration: none;
+  --link--hover--text-decoration: underline;
+  --link--visited--color: #{$link-color-visited};
+}
--- a/assets/src/scss/base/_variables.scss
+++ b/assets/src/scss/base/_variables.scss
@@ -30,17 +30,3 @@ $sp-6x: space(6.5);
 $sp-7: space(7);
 $sp-7x: space(7.5);
 $sp-8: space(8);
-
-// Variables
-:root {
-  --body--font-family: #{$lora};
-  --headings--font-family: #{$roboto};
-  --headings--font-weight: bold;
-  --body--background-color: #{$white};
-  --body--color: #{$grey-80};
-  --link--color: #{$link-color};
-  --link--hover--color: #{$link-color};
-  --link--text-decoration: none;
-  --link--hover--text-decoration: underline;
-  --link--visited--color: #{$link-color-visited};
-}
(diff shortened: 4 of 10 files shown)


planet-4:

:watch: 2023.02.08 10:01:02

planet-4:

:watch: 2023.02.15 12:28:54

planet-4:

:watch: 2023.02.16 09:59:16

comzeradd:
Looks good. Can you adjust the label to be on the same level as the checkbox? It's probably its `display` value that pushes it down.

This whole section seems to have a oversized font compared to the rest of the admin panel, but I couldn't find out where are we overwritting this. It doesn't seem to be the default WP font size for the Customizer.

comzeradd:
> This whole section seems to have a oversized font compared to the rest of the admin panel, but I couldn't find out where are we overwritting this. It doesn't seem to be the default WP font size for the Customizer.

On a quick inspector test, something like this seems to adjust typography to meet the rest of the admin panel look. Maybe there is a better selector for this, 

```css
.customize-pane-child p,
.customize-pane-child li {
  font-size: 0.7rem;
  line-height: inherit;
}
```

mleray:
> > This whole section seems to have a oversized font compared to the rest of the admin panel, but I couldn't find out where are we overwritting this. It doesn't seem to be the default WP font size for the Customizer.
> 
> On a quick inspector test, something like this seems to adjust typography to meet the rest of the admin panel look. Maybe there is a better selector for this,
> 
> ```css
> .customize-pane-child p,
> .customize-pane-child li {
>   font-size: 0.7rem;
>   line-height: inherit;
> }
> ```

This is a bit too small in my opinion 🤔 I think we should ask the design team for some recommendations, and maybe it should be a separate ticket?

comzeradd:
> This is a bit too small in my opinion  I think we should ask the design team for some recommendations, and maybe it should be a separate ticket?

Yes, that's indeed out of scope for this ticket.

assets/src/scss/style.scss
```diff
@@ -73,3 +73,17 @@ Text Domain: planet4-master-theme
...
+  --headings--font-family: #{$roboto};
+  --headings--font-weight: bold;
+  --body--background-color: #{$white};
+  --body--color: #{$grey-80};
+  --link--color: #{$link-color};
+  --link--hover--color: #{$link-color};
+  --link--text-decoration: none;
+  --link--hover--text-decoration: underline;
+  --link--visited--color: #{$link-color-visited};
+}
```
comzeradd, line 89: Not sure I get this change 🤔

mleray: This is to make sure that we only initialize them once, when loading this file. The way they were defined before in `_variables.scss` resulted in this being declared multiple times, since that file is imported in several blocks... Which means that they would override the new `style.scss` loaded for the new identity styles! Does it make sense? 😅

comzeradd: Yes! Thanks :) And it wouldn't affect the plugin in cases where we use one of these variables? I think we only include `_variables.scss` from the submosule.

mleray: Hmmm you're right, the color change is currently not applied in the editor 😞 I'll see how to fix it

mleray: I've refactored this a bit so that now the CSS variables are defined in a new file `_css-variables.scss`, and then we include this file in both the `style.scss` and `editorStyle.scss`. It seems to work as expected, and the new identity styles are also applied to the editor

assets/src/scss/editorStyle.scss
```diff
@@ -22,3 +22,6 @@
 @import "layout/gravity-forms";
 
 @import "editorOverrides";
+
+// CSS variables
+@import "base/css-variables";
```
comzeradd, line 27: And we don't need to import that to [blocks.scss](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/main/assets/src/styles/blocks.scss) too?

comzeradd: Hm.. testing locally it seems that it's not needed.

mleray: I think that the editor always loads `editorStyle.scss`, so that's enough to initialise the CSS variables that are used in the blocks

comzeradd (approved):
Great work 🍵

PR #1919, opened 2023-02-13
Fix some links hover/visited/active color

mleray:
### Description

These should always be the same color regardless of their state (I double-checked with the design team)

### Testing

You can see the broken versions on the GPI site:
- [Sitemap](https://www.greenpeace.org/international/sitemap/)
- [Post](https://www.greenpeace.org/international/press-release/58205/after-13-days-and-nearly-4000km-greenpeace-occupation-of-shell-oil-platform-ends-with-no-arrests/)

On local you can test the fixed versions in the Sitemap page and in any post.


lithrel (approved):
:+1:

PR #1921, opened 2023-02-15
PLANET-7077: Featured img not picked up/ missing when sharing on socials

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-7077

> On some posts the set featured image is not picked up and missing when sharing on socials

The loop going through post images can return an empty result in some cases.

A request in DB `select * from wp_postmeta where meta_key = 'p4_og_image_id'` will return some rows with empty `meta_value`; in the current loop, `isset` will return true and `get_og_image()` will return nothing, even if a featured image (thumbnail_id) is set.


comzeradd (approved):
:+1:

PR #1925, opened 2023-02-16
PLANET-7064: Playwright CI job

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-7064


planet-4:

:watch: 2023.02.16 16:28:11

tests/e2e/tools/lib/login.js
```diff
@@ -0,0 +1,22 @@
+const { test, expect } = require('@playwright/test');
```
Osong-Michael, line 1: You are importing this but not using it.

tests/e2e/tools/lib/rest.js
```diff
@@ -0,0 +1,40 @@
+import { nonce } from './nonce';
+
+async function rest(page, context, options) {
```
Osong-Michael, line 3: Looks like we are passing the page param to this function but not using it, is there a reason?

.circleci/config.yml
```diff
@@ -326,6 +339,74 @@ jobs:
             - booking.json
             - pr
 
+  request-test-user:
+    docker:
+      - image: cimg/base:2023.02
```
comzeradd, line 344: Why not use our image? `greenpeaceinternational/circleci-base:latest`

.circleci/config.yml
```diff
@@ -326,6 +339,74 @@ jobs:
...
+    steps:
+      - attach_workspace:
+          at: /tmp/workspace
+      - run:
+          name: Create test user
+          command: |
+            instance_repo="planet4-test-$(cat /tmp/workspace/test-instance)"
+            echo $instance_repo
+            echo $CIRCLE_WORKFLOW_ID
+            curl -vvv --request POST \
```
comzeradd, line 356: (Just thinking out loud) whether it would be "cleaner" to add a python script in our `circleci` image that handles these two API requests, using `pycircleci` (it's already included in that image). As an example see the [release pipeline script](https://github.com/greenpeace/planet4-release/blob/main/src/main.py#L25-L28).

lithrel: ah yes, makes sense :+1:

lithrel: @comzeradd  I added a script [in our circleci repo](https://github.com/greenpeace/planet4-circleci/pull/68), and used it here for user creation, [it works properly](https://app.circleci.com/pipelines/github/greenpeace/planet4-master-theme/7488/workflows/ef3a7808-644b-4bfe-b760-4cc507ec09b4/jobs/50364) :ok_hand: 
I tried a bit using it for deletion too, but with the fact that we're using a playwright specific image for e2e test, and that the deletion job should run always even if e2e fails, I'm not sure how to do this cleanly for the moment so I left the curl call.

comzeradd:
Looks great 👍🏻 

 Maybe @Osong-Michael @dantovbein want to have a final look before we approve/merge.

Osong-Michael (commented):
All good on my end too 👍

comzeradd (approved):

PR #1940, opened 2023-03-01
PLANET-7052: Integrate Design Tokens into the codebase

GP-Dan-Tovbein:
Ref: https://jira.greenpeace.org/browse/PLANET-7052

## Description
As an initial step, we are going to update only the updated colors.
It's also fixed white and grey colors with transparencies.

## Testing
Navigate into the test instance.
Nothing has been dramatically changed. The page layout should looks like the same.

### DIFF
--- a/assets/src/scss/base/_colors.scss
+++ b/assets/src/scss/base/_colors.scss
@@ -1,19 +1,19 @@
-$white: #fff;
+$white: var(--white);
 $black: #000;
 
 // Grey
 $grey-80: #020202;
 $grey: #23292d;
-$grey-60: #45494c;
-$grey-40: #6f7376;
-$grey-20: #a3a5a7;
-$grey-10: #ececec;
-$grey-05: #f5f7f8;
+$grey-60: var(--grey-800);
+$grey-40: var(--grey-600);
+$grey-20: var(--grey-500);
+$grey-10: var(--grey-200);
+$grey-05: var(--grey-100);
 
 // Green
-$gp-green: #66cc00;
+$gp-green: var(--gp-green-500);
 $gp-green-80: #8bcc49;
-$gp-green-20: #e3f2d3;
+$gp-green-20: var(--gp-green-100);
 
 // Blue
 $blue: #2077bf;
@@ -58,9 +58,9 @@ $table-font-color: #020202;
 
 // Background colors
 // Grey variant (default)
-$table-header-grey: #45494c;
-$table-odd-row-grey: #ececec;
-$table-even-row-grey: #f5f7f8;
+$table-header-grey: var(--grey-800);
+$table-odd-row-grey: var(--grey-200);
+$table-even-row-grey: var(--grey-100);
 $table-footer-grey: #e0e4e7;
 
 // Green variant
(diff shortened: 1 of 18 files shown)


assets/src/scss/base/_colors.scss
```diff
@@ -1,19 +1,19 @@
...
 
 // Grey
 $grey-80: #020202;
 $grey: #23292d;
-$grey-60: #45494c;
-$grey-40: #6f7376;
-$grey-20: #a3a5a7;
-$grey-10: #ececec;
-$grey-05: #f5f7f8;
+$grey-60: var(--grey-800);
```
GP-Dan-Tovbein, line 7: For now we can keep this variables, but I think we would need to remove all `$grey-*` afterwards and use directly the token var `--grey-*` but it's a `FLAGGED` implementation. 

The same with the rest of variables.

planet-4:

:watch: 2023.03.01 12:26:03

assets/src/scss/base/_css-variables.scss
```diff
@@ -2,8 +2,8 @@
   --body--font-family: #{$lora};
   --headings--font-family: #{$roboto};
   --headings--font-weight: bold;
-  --body--background-color: #{$white};
-  --body--color: #{$grey-80};
+  --body--background-color: var(--white);
+  --body--color: var(--color-text-body);
```
mleray, line 6: This is the new identity body text color, I don't think that we want to set it here since it should only be applied when the setting is toggled... You could use it [here](https://github.com/greenpeace/planet4-master-theme/blob/main/assets/src/scss/new-identity/style.scss#L2) though 🙂

assets/src/scss/pages/post/_post.scss
```diff
@@ -28,7 +28,7 @@
   color: $grey-40;
   font-size: 14px;
   font-family: var(--headings--font-family);
-  border-bottom: 1px solid transparentize($grey-20, 0.5);
+  border-bottom: 1px solid rgba(var(--grey-50), 0.5);
```
mleray, line 31: It seems that `--grey-50` is not defined, this doesn't show any border

mleray: Also, even when replacing `--grey-50` by one that is defined (such as `--grey-500`) the border actually still doesn't show, so maybe these functions (`rgba`, `transparentize`) don't work with CSS variables?

mleray: Actually why did you switch from `transparentize` to `rgba`? 🤔

assets/src/scss/layout/_country-selector.scss
```diff
@@ -28,7 +28,7 @@
 
   &.open {
     .country-selector-toggle-container {
-      border-bottom: 1px solid transparentize($white, 1);
+      border-bottom: 1px solid rgba(var(--white), 1);
```
mleray, line 31: Not sure why we use `transparentize` here (or `rgba`) since opacity is 1, seems we could just use the color directly 🤔

comzeradd:
Closing this as we already use the tokens as part of the new identity css changes. We can re-visit if any more change is needed when we roll out the new visual identity and remote the feature flag.

PR #1944, opened 2023-03-09
Fix tag creation when certain fields are missing

mleray:
### Description

This was crashing, my guess is that it is due to a [recent refactor of the validate function](https://github.com/greenpeace/planet4-master-theme/commit/d8303237ece741dc19c773c4f2c2ec25e3f6c73f#diff-09f0c1569665b6a290f7d056be1cabce24a3834d7940a7c16da9817bbd0de723R353). Maybe this is a partial fix for [PLANET-7099](https://jira.greenpeace.org/browse/PLANET-7099).

### Testing

If you try creating a tag on the `main` branch without selecting a redirect page or an image, it will crash (even though on reload the tag will be there). On this branch it should work as expected.


lithrel (approved):
:+1:

PR #1945, opened 2023-03-10
PLANET-5727: Limit of post revisions

GP-Dan-Tovbein:
Ref: [PLANET-5727](https://jira.greenpeace.org/browse/PLANET-5727)

## Description
We should define a limit of post `revisions` and avoid recording unnecessary records into the database. Currently, we don't have a limit for that which means that every time the editor save the post it's store a new revision.

I've reproduced the same as the ticket description, using the international site database and of course with different dates, and got worst results.

#### Using the planet4_international database as a part of this experiment I get (March 9, 2023)
Total size: **359,3 MB**
wp_posts table (with revisions): **272,8 MB**
wp_posts table (without revisions): **89,3 MB**
New size: ~**175,8 MB** which represents ~**51%** of size removed.

#### Example of query to get all revisions grouped by post
```sql
SELECT count(ID) AS total_revisions, ID, post_parent FROM wp_posts WHERE post_type = "revision" GROUP BY post_parent ORDER BY post_date ASC;
+-----------------+-------+-------------+
| total_revisions | ID    | post_parent |
+-----------------+-------+-------------+
|               1 | 61952 |       11442 |
|               1 | 60932 |       11452 |
|               1 | 61951 |        1320 |
|               1 | 61961 |       18532 |
|               1 | 61958 |       54800 |
|               1 | 61956 |       49031 |
|               1 | 61954 |       21155 |
|               1 | 61955 |       25526 |
|               1 | 61962 |       58548 |
|               1 | 60936 |       60796 |
|               1 | 61960 |       61511 |
|               3 | 45241 |       45227 |
|               3 | 45260 |       45226 |
|               3 | 45271 |       45269 |
|               2 | 45280 |       45278 |
|               3 | 61965 |       20416 |
|               3 | 61984 |       56556 |
|               3 | 61989 |       57850 |
|               3 | 61993 |       61276 |
|               3 | 62006 |       58980 |
|               3 | 62050 |        9129 |
|               3 | 62058 |       54446 |
|               3 | 62068 |       45080 |
|               3 | 62071 |       54964 |
|               1 | 61950 |       61949 |
|               3 | 62110 |       45224 |
+-----------------+-------+-------------+
```

#### Remove all `revisions`
```sql
mysql> select count(ID) AS total_revisions from wp_posts where post_type = 'revision';
+----------+
| total_revisions |
+----------+
|    19812 |
+----------+
```

```sql
mysql> delete from wp_posts where post_type = 'revision';
Query OK, 19812 rows affected (7.11 sec)
```

```sql
mysql> select count(ID) as total_revisions from wp_posts where post_type = 'revision';
+----------+
| total_revisions |
+----------+
|        0 |
+----------+
```
**Other queries.**
Total of revisions since the last week
```sql
mysql> SELECT COUNT(ID) as total_revisions FROM wp_posts AS posts WHERE posts.post_type = "revision" AND posts.post_modified BETWEEN date_sub(now(), INTERVAL 1 WEEK) AND now();
+----------+
| total_revisions |
+----------+
|        0 |
+----------+
```

Total of revisions since the last 2 weeks
```sql
mysql> SELECT COUNT(ID) AS total_revisions  FROM wp_posts AS posts WHERE posts.post_type = "revision" AND posts.post_modified BETWEEN date_sub(now(), INTERVAL 2 WEEK) AND now();
+----------+
| COUNT(*) |
+----------+
|       23 |
+----------+
```

Total of revisions since the last 3 weeks
```sql
mysql> SELECT COUNT(ID) AS total_revisions FROM wp_posts AS posts WHERE posts.post_type = "revision" AND posts.post_modified BETWEEN date_sub(now(), INTERVAL 3 WEEK) AND now();
+----------+
| total_revisions |
+----------+
|       74 |
+----------+
```

Total of revisions since the last 4 weeks
```sql
mysql> SELECT COUNT(*) FROM wp_posts AS posts WHERE posts.post_type = "revision" AND posts.post_modified BETWEEN date_sub(now(), INTERVAL 4 WEEK) AND now();
+----------+
| COUNT(*) |
+----------+
|      143 |
+----------+
```

I've also added an option, under `Settings > Writing`, which allows to editors to change the desired revisions in case they prefer more or less to keep.

Screenshot 2023-03-10 at 10 30 43

Then, just create a new page, type whatever you want and `save` the page more than 20 times. It will keep as maximum 20 revisions.

You can also test the migration by running `wp-env run cli wp p4-run-activator`.

## Testing (locally)
1. Make sure the field is already added into the `Settings > Writing` page.
2. Create a new page, add a dummy content and save it more than 20 times.
  a. If the `revisions_to_keep` is set you won't be able to save it more than 20 times.
3. Go back  `Settings > Writing` page and change the `revisions_to_keep` field to a lower value, then go back to the previous page.
4. Run the migration file by ` wp-env run cli wp p4-run-activator `.
5. If everything works as expected, you will see the value of `Revisions` of that page the same as the `revisions_to_keep` before updated.

Demo video

https://github.com/greenpeace/planet4-master-theme/assets/77975803/5b1609e2-a16a-4fa2-8d94-7af37b4bb6ac

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -617,4 +617,3 @@ add_filter(
         return $theme_json->update_with($new_color_palette);
     }
 );
-
--- a/functions.php
+++ b/functions.php
@@ -617,39 +617,52 @@ add_filter(
         return $theme_json->update_with($new_color_palette);
     }
 );
-// This filter overrides the default revisions value which by default it is set to 1.
-// That means that all revisions are stored and the db storage it's growing unnecessarily.
-add_filter(
-    'wp_revisions_to_keep',
-    function (): void {
-        register_setting(
-            'writing',
-            'revisions_to_keep',
-        );
 
-        add_settings_section(
-            'post-revisions-section',
-            '',
-            function (): void {
-                echo '';
-            },
-            'writing'
-        );
+// Registering into the Settings > Writing setting page
+add_action(
+    'admin_menu',
+    function (): void {
+        register_setting('writing', 'revisions_to_keep');
 
         add_settings_field(
-            'my-settings-field-id',
-            'Post Revisions',
-            function (): void {
-                echo '
-                    <label for="limit-post-revisions">
-                        <input id="limit-post-revisions" type="number" name="revisions_to_keep" value=' . get_option('revisions_to_keep') . '>  ' // phpcs:ignore Generic.Files.LineLength.MaxExceeded
-                        . __('Maximum number of revisions to store for each post.', 'planet4-master-theme-backend') .
-                    '</label>';
+            'post-revisions-field',
+            __('Post revisions', 'planet4-master-theme-backend'),
+            function ($val): void {
+                $id = $val['id'];
+                $option_name = $val['option_name'];
+                ?>
+                    <input
+                        type="number"
+                        name="<?php echo esc_attr($option_name) ?>"
+                        id="<?php echo esc_attr($id) ?>"
+                        value="<?php echo esc_attr(get_option($option_name)) ?>"
+                    />
+                    <span>
+                        <?php echo __(
+                            'Maximum number of revisions to store for each post.',
+                            'planet4-master-theme-backend'
+                        )?>
+                    </span>
+                <?php
             },
             'writing',
-            'post-revisions-section'
+            'default',
+            array(
+                'id' => 'post-revisions-field',
+                'option_name' => 'revisions_to_keep'
+            )
         );
     },
     10,
-    1
+    2,
+);
+
+// Maximum revisions to keep whenever the editor save a post
+add_filter(
+    'wp_revisions_to_keep',
+    function () {
+        return get_option('revisions_to_keep');
+    },
+    10,
+    2
 );
--- a/functions.php
+++ b/functions.php
@@ -618,13 +618,3 @@ add_filter(
     }
 );
 
-// Maximum revisions to keep whenever the editor save a post
-add_filter(
-    'wp_revisions_to_keep',
-    function ($revisions) {
-        $revisions_to_keep = get_option('revisions_to_keep');
-        return $revisions_to_keep ?: $revisions;
-    },
-    10,
-    2
-);
--- a/functions.php
+++ b/functions.php
@@ -622,9 +622,35 @@ add_filter(
 add_filter(
     'wp_revisions_to_keep',
     function () {
-        $revisions_to_keep = planet4_get_option('revisions_to_keep', 5);
-        return $revisions_to_keep;
-    },
+        register_setting(
+            'writing',
+            'revisions_to_keep',
+        );
+
+        add_settings_section(
+            'post-revisions-section',
+            '',
+            function () {
+                echo '';
+            },
+            'writing'
+        );
+
+        add_settings_field(
+            'my-settings-field-id',
+            'Post Revisions',
+            function (){
+                echo '
+                    <label for="limit-post-revisions">
+                        <input id="limit-post-revisions" type="number" name="revisions_to_keep" value=' . get_option('revisions_to_keep') . '>  '
+                        . __('Maximum number of revisions to store for each post.', 'planet4-master-theme-backend') .
+                    '</label>';
+            },
+            'writing',
+            'post-revisions-section'
+        );
+    }
+    ,
     10,
-    2
+    1
 );
--- a/functions.php
+++ b/functions.php
@@ -617,3 +617,14 @@ add_filter(
         return $theme_json->update_with($new_color_palette);
     }
 );
+// This filter overrides the default revisions value which by default it is set to 1.
+// That means that all revisions are stored and the db storage it's growing unnecessarily.
+add_filter(
+    'wp_revisions_to_keep',
+    function () {
+        $revisions_to_keep = planet4_get_option('revisions_to_keep', 5);
+        return $revisions_to_keep;
+    },
+    10,
+    2
+);
--- a/functions.php
+++ b/functions.php
@@ -618,50 +618,12 @@ add_filter(
     }
 );
 
-// Registering into the Settings > Writing setting page
-add_action(
-    'admin_menu',
-    function (): void {
-        register_setting('writing', 'revisions_to_keep');
-
-        add_settings_field(
-            'post-revisions-field',
-            __('Post revisions', 'planet4-master-theme-backend'),
-            function ($val): void {
-                $id = $val['id'];
-                $option_name = $val['option_name'];
-                ?>
-                    <input
-                        type="number"
-                        name="<?php echo esc_attr($option_name) ?>"
-                        id="<?php echo esc_attr($id) ?>"
-                        value="<?php echo esc_attr(get_option($option_name)) ?>"
-                    />
-                    <span>
-                        <?php echo __(
-                            'Maximum number of revisions to store for each post.',
-                            'planet4-master-theme-backend'
-                        )?>
-                    </span>
-                <?php
-            },
-            'writing',
-            'default',
-            array(
-                'id' => 'post-revisions-field',
-                'option_name' => 'revisions_to_keep'
-            )
-        );
-    },
-    10,
-    2,
-);
-
 // Maximum revisions to keep whenever the editor save a post
 add_filter(
     'wp_revisions_to_keep',
-    function () {
-        return get_option('revisions_to_keep');
+    function ($revisions) {
+        $revisions_to_keep = get_option('revisions_to_keep');
+        return $revisions_to_keep ?: $revisions;
     },
     10,
     2
--- a/functions.php
+++ b/functions.php
@@ -621,7 +621,7 @@ add_filter(
 // That means that all revisions are stored and the db storage it's growing unnecessarily.
 add_filter(
     'wp_revisions_to_keep',
-    function () {
+    function (): void {
         register_setting(
             'writing',
             'revisions_to_keep',
@@ -630,7 +630,7 @@ add_filter(
         add_settings_section(
             'post-revisions-section',
             '',
-            function () {
+            function (): void {
                 echo '';
             },
             'writing'
@@ -639,18 +639,17 @@ add_filter(
         add_settings_field(
             'my-settings-field-id',
             'Post Revisions',
-            function (){
+            function (): void {
                 echo '
                     <label for="limit-post-revisions">
-                        <input id="limit-post-revisions" type="number" name="revisions_to_keep" value=' . get_option('revisions_to_keep') . '>  '
+                        <input id="limit-post-revisions" type="number" name="revisions_to_keep" value=' . get_option('revisions_to_keep') . '>  ' // phpcs:ignore Generic.Files.LineLength.MaxExceeded
                         . __('Maximum number of revisions to store for each post.', 'planet4-master-theme-backend') .
                     '</label>';
             },
             'writing',
             'post-revisions-section'
         );
-    }
-    ,
+    },
     10,
     1
 );
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -122,6 +122,7 @@ class MasterSite extends TimberSite
         add_filter('get_twig', [$this, 'add_to_twig']);
         add_action('init', [$this, 'register_taxonomies'], 2);
         add_action('init', [$this, 'register_oembed_provider']);
+        add_action('admin_menu', [$this, 'add_post_revisions_setting']);
         add_action('admin_enqueue_scripts', [$this, 'enqueue_admin_assets']);
         // Load the editor scripts only enqueuing editor scripts while in context of the editor.
         add_action('enqueue_block_editor_assets', [$this, 'enqueue_editor_assets']);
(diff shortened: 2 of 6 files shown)


planet-4:

:watch: 2023.03.10 12:40:13

src/Settings.php
```diff
@@ -164,6 +164,20 @@ public function __construct()
...
+                    [
+                        'name' => __('Post Revisions', 'planet4-master-theme-backend'),
+                        'id' => 'revisions_to_keep',
+                        'type' => 'text',
+                        'attributes' => [
+                            'type' => 'number',
+                        ],
+                        'default' => 5,
+                        'desc' => __(
+                            'This is the maximum of revisions to store into the database.',
```
comzeradd, line 176: Minort re-phrasing suggestion: `'Maximum number of revisions to store for each post.'`

functions.php
```diff
@@ -501,3 +501,15 @@ function ($canonical_url, $post) {
     10,
     2
 );
+
+// This filter overrides the default revisions value which by default it is set to 1.
+// That means that all revisions are stored and the db storage it's growing unnecessarily.
+add_filter(
+    'wp_revisions_to_keep',
+    function () {
+        $revisions_to_keep = planet4_get_option('revisions_to_keep', 5);
```
comzeradd, line 510: We didn't define specific requirements in the ticket but maybe `5` is too low. Maybe `20` would be a more safe default options for editors?

comzeradd (commented):
Can we move the option setting in _Settings > Writing_? In the _Writing Settings_ section.

functions.php
```diff
@@ -501,3 +501,39 @@ function ($canonical_url, $post) {
...
+            'writing'
+        );
+
+        add_settings_field(
+            'my-settings-field-id',
+            'Post Revisions',
+            function (){
+                echo '
+                    <label for="limit-post-revisions">
+                        <input id="limit-post-revisions" type="number" name="revisions_to_keep" value=' . get_option('revisions_to_keep') . '>  '
```
sagarsdeshmukh, line 528: By default `revisionss_to_keep` value is empty at start, so you may pass a default value as a second arg in `get_option()` function.
(https://developer.wordpress.org/reference/functions/get_option/)

src/Migrations/M018UpdatePostRevisions.php
```diff
@@ -0,0 +1,51 @@
...
+            WHERE post_type = "revision"
+            GROUP BY post_parent
+            ORDER BY post_date ASC';
+
+        // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+        $prepared_sql = $wpdb->prepare($sql);
+        $results = $wpdb->get_results($prepared_sql); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+
+        foreach ((array) $results as $post) {
+            if ((int) $post->total_revisions <= $MAX_REVISIONS) {
```
sagarsdeshmukh, line 38: As a safe side could you add condition, `$MAX_REVISIONS` should not empty

src/Migrations/M018UpdatePostRevisions.php
```diff
@@ -0,0 +1,51 @@
...
+                continue;
+            }
+
+            echo "\nDelete " . $MAX_REVISIONS . " of " . $post->total_revisions . " revisions FROM " . $post->ID;
+            $sql = '
+                DELETE FROM wp_posts
+                WHERE post_type = "revision"
+                AND post_parent = ' . $post->post_parent . '
+                ORDER BY post_date DESC limit ' . $MAX_REVISIONS;
```
sagarsdeshmukh, line 47: The DELETE ORDER BY LIMIT query will delete a `$MAX_REVISIONS` only from filter condition of Delete query, i.e if we have 100 post revisions then as per our $MAX_REVISIONS = 20 , it will delete only 20 revisions.

src/Migrations/M022UpdatePostRevisions.php
```diff
@@ -0,0 +1,64 @@
...
+        // Check if exist
+        $MAX_REVISIONS = get_option('revisions_to_keep');
+        if (false === get_option('revisions_to_keep')) {
+            // If not, set 20 by default
+            $MAX_REVISIONS = 20;
+            update_option('revisions_to_keep', $MAX_REVISIONS);
+        }
+
+        $sql = '
+            DELETE FROM wp_posts
```
GP-Dan-Tovbein, line 32: This query removes all the revisions that exceeds the maximum for each post.

For example,
if a post has 22 revisions and another one has 54 revisions, this query will remove 36 unnecessary revisions. 2 for the first post, and 34 for the other post.

src/Migrations/M022UpdatePostRevisions.php
```diff
@@ -0,0 +1,64 @@
...
+            WHERE ID IN (
+                SELECT ID
+                FROM (
+                    SELECT
+                        post_parent,
+                        ID,
+                        post_date,
+                        row_number() OVER (PARTITION BY post_parent ORDER BY post_date desc) AS rn
+                    FROM (
+                        SELECT
```
GP-Dan-Tovbein, line 42: Select all posts which more than `$MAX_REVISIONS`

functions.php
```diff
@@ -617,3 +617,52 @@ function ($theme_json) {
         return $theme_json->update_with($new_color_palette);
     }
 );
+
+// Registering into the Settings > Writing setting page
```
sagarsdeshmukh, line 621: My preference will be, we should move this piece of code to some other place/file like `planet4-master-theme/src/MasterSite.php`, it will be better for future maintenance.

functions.php
```diff
@@ -617,3 +617,52 @@ function ($theme_json) {
...
+                'option_name' => 'revisions_to_keep'
+            )
+        );
+    },
+    10,
+    2,
+);
+
+// Maximum revisions to keep whenever the editor save a post
+add_filter(
```
sagarsdeshmukh, line 661: We can make this function a bit more safe, -(In case of `revisions_to_keep` found empty...)
```
add_filter(
    'wp_revisions_to_keep',
    function ($num) {
        $revisions_to_keep = get_option('revisions_to_keep');
        return $revisions_to_keep ?: $num;
    },
    10,
    1
);
```

src/Migrations/M022UpdatePostRevisions.php
```diff
@@ -0,0 +1,64 @@
...
+     * @param MigrationRecord $record Information on the execution, can be used to add logs.
+     * phpcs:disable SlevomatCodingStandard.Functions.UnusedParameter -- interface implementation
+     */
+    protected static function execute(MigrationRecord $record): void
+    {
+        global $wpdb;
+
+        // Check if exist
+        $MAX_REVISIONS = get_option('revisions_to_keep');
+        if (false === get_option('revisions_to_keep')) {
```
sagarsdeshmukh, line 25: The `not`(!) operator will be good fit for this `if` condition.(It covers both cases, the empty or false value)
```
if (! $MAX_REVISIONS) { 
...
}
```

src/Migrations/M022UpdatePostRevisions.php
```diff
@@ -0,0 +1,68 @@
...
+            // If not, set 20 revisions by default
+            $MAX_REVISIONS = 20;
+            update_option('revisions_to_keep', $MAX_REVISIONS);
+        } else {
+            $MAX_REVISIONS = get_option('revisions_to_keep');
+        }
+
+        $wpdb->query('SET @rownumber:=0');
+        $wpdb->query('SET @postparent:=NULL');
+        $sql = '
```
GP-Dan-Tovbein, line 33: In case we want upgrade to `MySQL 8.0`, it's better if we implement the `ROW_NUMBER` function.
```sql
DELETE FROM wp_posts
            WHERE ID IN (
                SELECT ID
                FROM (
                    SELECT
                        post_parent,
                        ID,
                        post_date,
                        row_number() OVER (PARTITION BY post_parent ORDER BY post_date desc) AS rn
                    FROM (
                        SELECT
                            ID,
                            post_parent,
                            post_date
                        FROM wp_posts
                        WHERE post_parent IN (
                            SELECT post_parent
                            FROM wp_posts
                            WHERE post_type = "%2$s"
                            GROUP BY post_parent
                            HAVING count(ID) > %1$s
                        ) AND post_type = "%2$s"
                        ORDER BY post_date ASC
                    ) AS r
                ) AS d WHERE d.rn > %1$s
```

src/Migrations/M022UpdatePostRevisions.php
```diff
@@ -0,0 +1,68 @@
...
+
+        if (!get_option('revisions_to_keep')) {
+            // If not, set 20 revisions by default
+            $MAX_REVISIONS = 20;
+            update_option('revisions_to_keep', $MAX_REVISIONS);
+        } else {
+            $MAX_REVISIONS = get_option('revisions_to_keep');
+        }
+
+        $wpdb->query('SET @rownumber:=0');
```
GP-Dan-Tovbein, line 31: The next query is compatible with `MySQL 5.7`, that's why it's required to query a variable before.

functions.php
```diff
@@ -617,3 +617,14 @@ function ($theme_json) {
...
+    'wp_revisions_to_keep',
+    function ($revisions) {
+        $revisions_to_keep = get_option('revisions_to_keep');
+        return $revisions_to_keep ?: $revisions;
+    },
+    10,
+    2
```
sagarsdeshmukh, line 629: As we are accepting 1 argument in function, we need to pass 1 instead of 2, 
Ref. https://developer.wordpress.org/reference/functions/add_filter/

functions.php
sagarsdeshmukh: We can move this add_filter hook to  `src/MasterSite.php` file as well.
Note: The reason we move it out of `functions.php` is, this file will grow very fast if we add filters and hooks in it.

src/Migrations/M022UpdatePostRevisions.php
```diff
@@ -0,0 +1,68 @@
...
+     * Perform the actual migration.
+     *
+     * @param MigrationRecord $record Information on the execution, can be used to add logs.
+     * phpcs:disable SlevomatCodingStandard.Functions.UnusedParameter -- interface implementation
+     */
+    protected static function execute(MigrationRecord $record): void
+    {
+        global $wpdb;
+
+        if (!get_option('revisions_to_keep')) {
```
sagarsdeshmukh, line 23: We can make it little less code, if we make it -
```
        $MAX_REVISIONS = get_option('revisions_to_keep');
        if (!$MAX_REVISIONS) {
            // If not, set 20 revisions by default
            $MAX_REVISIONS = 20;
            update_option('revisions_to_keep', $MAX_REVISIONS);
        }
```
This way we will call `get_option()` function once.

GP-Dan-Tovbein: You are correct!

src/MasterSite.php
```diff
@@ -304,6 +304,56 @@ function ($tests) {
...
+                return $revisions_to_keep ?: $revisions;
+            },
+            10,
+            2
+        );
+
+        // Registering into the Settings > Writing setting page
+        add_action(
+            'admin_menu',
+            function (): void {
```
sagarsdeshmukh, line 321: We can move this function code into a new function , as [like](https://github.com/greenpeace/planet4-master-theme/blob/f3d2d05e187ef887d14614b149b6e6a8817cf254/src/MasterSite.php#L125) -
`add_action('admin_enqueue_scripts', [$this, 'enqueue_admin_assets']);`

src/MasterSite.php
```diff
@@ -304,6 +304,56 @@ function ($tests) {
...
+                    'writing',
+                    'default',
+                    array(
+                        'id' => 'post-revisions-field',
+                        'option_name' => 'revisions_to_keep'
+                    )
+                );
+            },
+            10,
+            2,
```
sagarsdeshmukh: You can skip the last 2 arguments in `add_action()` function as the priority 10 is a default value and we dont have function accepted arguments.

src/MasterSite.php
```diff
@@ -304,6 +304,56 @@ function ($tests) {
...
 
+        // Maximum revisions to keep whenever the editor save a post
+        add_filter(
+            'wp_revisions_to_keep',
+            function ($revisions) {
+                $revisions_to_keep = get_option('revisions_to_keep');
+                return $revisions_to_keep ?: $revisions;
+            },
+            10,
+            2
```
sagarsdeshmukh: It should be 1 as we have single argument(i.e. $revisions) pass to the function.

src/MasterSite.php
GP-Dan-Tovbein: That sounds logic!

sagarsdeshmukh: I mean to say -
we can make it
`add_action(‘admin_menu’, [$this, ‘add_post_revisions_setting’]); `
and we will add a new function with -
```
public function add_post_revisions_setting() {
...
}
```
Sorry, if I confuse you earlier

sagarsdeshmukh (approved):
Tested on local, the PR looks good to me 👍

PR #1955, opened 2023-03-21
PLANET-6990 Apply the new style to body copy links

Osong-Michael:
### Description: [PLANET-6990](https://jira.greenpeace.org/browse/PLANET-6990)

**Testing**

Check this [post](https://www-dev.greenpeace.org/test-tavros/story/29458/peak-oil-decline-coronavirus-economy/) on the tavros instance. Because the _New Site Identity_ has been enabled, it reflects the new changes for body copy links.

### DIFF
--- a/assets/src/scss/new-identity/style.scss
+++ b/assets/src/scss/new-identity/style.scss
@@ -121,15 +121,24 @@ table.spreadsheet-table.is-color-gp-green {
   }
 }
 
-a:where(.standalone-link) {
+.standalone-link {
   width: fit-content;
-  border-bottom: 2px solid var(--gp-green-500);
   font-weight: 700 !important;
-  transition: 0.3s cubic-bezier(0.5, 0, 0, 1);
+}
 
-  &:hover {
-    box-shadow: inset 0 calc(-22px) var(--gp-green-100);
-    text-decoration: none !important;
+// Body Copy links selectors
+.post-content,
+.page-content {
+  .standalone-link,
+  p > a,
+  li > a {
+    transition: 0.2s cubic-bezier(0.5, 0, 0, 1);
+    border-bottom: 2px solid var(--gp-green-500);
+
+    &:hover {
+      box-shadow: inset 0 calc(-22px) var(--gp-green-100);
+      text-decoration: none !important;
+    }
   }
 }
 
--- a/tests/e2e/404.spec.js
+++ b/tests/e2e/404.spec.js
@@ -1,6 +1,7 @@
-const { test, expect } = require('@playwright/test');
+const {test, expect} = require('@playwright/test');
 
-test('check the 404 page', async ({ page }) => {
+test('check the 404 page', async ({page}) => {
+  test.setTimeout(240 * 1000);
   const response = await page.goto('./thispagereallywillnotexist');
 
   // Check the page status.
@@ -9,8 +10,9 @@ test('check the 404 page', async ({ page }) => {
   // Check the page text.
   const settingsText = await page.evaluate('window.p4bk_vars.page_text_404');
   await expect(settingsText).toBeDefined();
+  const settingsTextUpdated = settingsText.replace(/\s+/g, ' ').replaceAll('&', '&amp;');
   const pageContent = await page.locator('.speech-bubble').innerHTML();
-  await expect(pageContent.replace(/\s+/g, ' ')).toContain(settingsText.replace(/\s+/g, ' '));
+  await expect(pageContent.replace(/\s+/g, ' ')).toContain(settingsTextUpdated.trim());
 
   // Check the page image background.
   const settingsImage = await page.evaluate('window.p4bk_vars.page_bg_image_404');


planet-4:

:watch: 2023.03.22 11:31:23

assets/src/scss/new-identity/style.scss
```diff
@@ -10,3 +10,14 @@
   --block-columns--step-number--color: var(--grey-900);
   --block-columns--card-header--background: var(--p4-dark-green-800);
 }
+
+a.external-link, a.pdf-link {
+  color: var(--grey-900) !important;
+  text-decoration: none !important;
```
GP-Dan-Tovbein, line 16: You can override the current CSS variable like [here](https://github.com/greenpeace/planet4-master-theme/pull/1959/files#diff-b418b2241107b99a37ed0291b0385d4ad8fe3762dd4d859be83bf83c070428d7R20). But not necessary if we will merge the other PR.

assets/src/scss/new-identity/style.scss
```diff
@@ -10,3 +10,14 @@
   --block-columns--step-number--color: var(--grey-900);
   --block-columns--card-header--background: var(--p4-dark-green-800);
 }
+
+a.external-link, a.pdf-link {
+  color: var(--grey-900) !important;
```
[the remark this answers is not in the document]
Osong-Michael: I added the `--link--color` variable, but it does not change anything, that is why i have specific styles for `pdf` and `external` links.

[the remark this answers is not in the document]
GP-Dan-Tovbein: I see it's being reflected
Screenshot 2023-03-27 at 11 55 50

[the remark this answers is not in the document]
GP-Dan-Tovbein: You'll also need to override the hover color for links. 
`--link--hover--color: var(--grey-900);`

assets/src/scss/new-identity/style.scss
```diff
@@ -33,4 +33,18 @@
   --author-block--background: var(--beige-100);
   --author-block-description-button--color: var(--gp-green-800);
   --author-block-description-button--hover--color: var(--gp-green-800);
+  --link--color: var(--grey-900);
+  --link--visited--color: var(--grey-900);
```
GP-Dan-Tovbein, line 37: Related comment https://github.com/greenpeace/planet4-master-theme/pull/1955/commits/a777d07268c0665e4f824a71e301768912a1f6e3#r1149434376

assets/src/scss/new-identity/style.scss
```diff
@@ -33,4 +33,19 @@
...
+
+a.external-link,
+a.pdf-link {
+  color: var(--grey-900) !important;
+  text-decoration: none !important;
+  transition: 0.2s cubic-bezier(0.5, 0, 0, 1);
+  border-bottom: 2px solid var(--gp-green-500);
+
+  &:hover {
+    box-shadow: inset 0 -22px var(--gp-green-100);
```
GP-Dan-Tovbein, line 49: Checking the value of `offset-y`, currently set to `-22px`, I see that a bit inconsistent if you check that on screens since the different font sizes we have (see [here](https://github.com/greenpeace/planet4-master-theme/blob/main/assets/src/scss/base/_typography.scss#L11-L19)). You can also make that test and see that on small screens there is more gap over the text, but on large screens it's almost the same height.

Can we change the `offset-y` more dynamically? Also if the user override the font-size because of `font-size: var(--text--x-large-and-up--font-size, 1.25rem);`

If you think it's a bit complicated change we can follow up it afterwards and don't block the merge! 🙏 

What do you think?

Osong-Michael: Great idea, we can look into doing that dynamically. If you have any suggestions I will love to hear them.

assets/src/scss/new-identity/style.scss
GP-Dan-Tovbein: By doing this `--link--hover--text-decoration: none;`

assets/src/scss/new-identity/style.scss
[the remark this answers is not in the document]
mleray: We need to apply these new styles to all body links, not just the ones that have `external-link` or `pdf-link` classes

mleray:
My previous comment still stands, we need to apply these changes to all body links and not just the external and PDF ones!

assets/src/scss/new-identity/style.scss
```diff
@@ -123,9 +123,12 @@ table.spreadsheet-table.is-color-gp-green {
 
 a:where(.standalone-link) {
   width: fit-content;
-  border-bottom: 2px solid var(--gp-green-500);
   font-weight: 700 !important;
-  transition: 0.3s cubic-bezier(0.5, 0, 0, 1);
+}
+
+a:where(.standalone-link, .external-link, .pdf-link) {
```
GP-Dan-Tovbein, line 129: Let's try something like this to style links without classes
```
a:where(.standalone-link, .external-link, .pdf-link),
a:not([class]) {
...
}
```

mleray: Before moving the styles to `theme.json` we used to define body link colors directly in [_typography.scss](https://github.com/greenpeace/planet4-master-theme/blob/main/assets/src/scss/base/_typography.scss#L102-L108) using just the `a` selector, so we should be able to do that now with the new identity but that might require us to add a few changes to other links that have other styles

assets/src/scss/new-identity/style.scss
```diff
@@ -123,9 +123,13 @@ table.spreadsheet-table.is-color-gp-green {
 
 a:where(.standalone-link) {
   width: fit-content;
-  border-bottom: 2px solid var(--gp-green-500);
   font-weight: 700 !important;
-  transition: 0.3s cubic-bezier(0.5, 0, 0, 1);
+}
+
+a:where(.standalone-link, .external-link, .pdf-link),
+a:not(.single-post-author a, .tag-item, .comment-link, .site-logo, .nav-link, .btn, .article-list-item-headline a, .article-list-item-author a, .article-list-item-image a) {
```
GP-Dan-Tovbein, line 130: Looks good but honestly I see it very complex to maintain it afterwards.

But If this is the only way to lead with this scenario, then we can proceed to merge it and then think about a good solution for standardizing all the links.

Looking forward what @mleray thinks?

mleray: I think that we should just use the `a` selector without all the classes, and fix the few links that have different styles such as the contextual links, because having such a complicated selector is really not ideal... It's harder to understand, and difficult to maintain for us and to override for NROs!

assets/src/scss/new-identity/style.scss
```diff
@@ -123,9 +123,12 @@ table.spreadsheet-table.is-color-gp-green {
 
 a:where(.standalone-link) {
   width: fit-content;
-  border-bottom: 2px solid var(--gp-green-500);
   font-weight: 700 !important;
-  transition: 0.3s cubic-bezier(0.5, 0, 0, 1);
+}
+
+.standalone-link, p > a {
```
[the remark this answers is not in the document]
Osong-Michael: Looking at the example page you created, one fix will be to add `.external-links` to the selector, another we can use the `post-details > a` selector, assuming all body copy links are present only within this context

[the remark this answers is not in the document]
mleray: Adding `.external-links` is definitely not the way to go, in my example the links are external but it could not be the case. If we do `.post-details > a` the ones in links won't be included, since they won't be direct children of the `.post-details` element. However we could maybe use `.post-details` 🤔 Something like `.post-details li > a`, however this will only work in Posts because other post types use different classnames (I think for Pages it's `page-content`?)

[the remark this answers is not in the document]
mleray: Maybe we should also use that for `p > a` actually, to make sure only the content is targeted and not other elements (footer, etc)

assets/src/scss/new-identity/style.scss
```diff
@@ -121,11 +121,18 @@ table.spreadsheet-table.is-color-gp-green {
...
-a:where(.standalone-link) {
+.standalone-link {
   width: fit-content;
-  border-bottom: 2px solid var(--gp-green-500);
   font-weight: 700 !important;
-  transition: 0.3s cubic-bezier(0.5, 0, 0, 1);
+}
+
+.standalone-link,
+.post-details p > a,
```
mleray, line 130: Actually, I see that for the [Lightbox](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/main/assets/src/components/Lightbox/setupLightboxForImages.js#L48) we use the `.post-content` selector rather than `.post-details` so maybe we could do that here too, for consistency?
Also, I think that maybe we can simplify this a bit by writing: 
```
.post-content,
.page-content {
  .standalone-link,
  p > a,
  li > a {
    ...
  }
}
```
What do you think? 🙂

mleray: Maybe it would also be nice to add a small comment like `// Body link` to clarify what this selector is for 🤔

mleray (approved):
Looks good to me 🎉 Just added one small comment but it's not a blocker, I'll leave it up to you 🙂

PR #1959, opened 2023-03-22
PLANET-6998: Apply new colors to comments block

GP-Dan-Tovbein:
Ref: https://jira.greenpeace.org/browse/PLANET-6998

## Description
Apply the new layout regarding the new identity project.

<!-- Ref: Please add a url to the ticket this change is addressing.

### DIFF
--- a/assets/src/scss/new-identity/style.scss
+++ b/assets/src/scss/new-identity/style.scss
@@ -15,6 +15,9 @@
   --button-primary--color: var(--grey-900);
   --button-primary--visited--color: var(--grey-900);
   --button-primary--hover--color: var(--grey-900);
+  --comment-reply-link--color: var(--grey-900);
+  --comment-reply-link--border-bottom: 2px solid var(--gp-green-500);
+  --link--hover--text-decoration: none;
   --post-tag-button--active--color: var(--grey-900);
   --post-tag-button--active--background-color: var(--gp-green-200);
   --post-tag-button--background-color: var(--gp-green-100);
--- a/assets/src/scss/new-identity/style.scss
+++ b/assets/src/scss/new-identity/style.scss
@@ -15,9 +15,6 @@
   --button-primary--color: var(--grey-900);
   --button-primary--visited--color: var(--grey-900);
   --button-primary--hover--color: var(--grey-900);
-  --comment-reply-link--color: var(--grey-900);
-  --comment-reply-link--border-bottom: 2px solid var(--gp-green-500);
-  --link--hover--text-decoration: none;
   --post-tag-button--active--color: var(--grey-900);
   --post-tag-button--active--background-color: var(--gp-green-200);
   --post-tag-button--background-color: var(--gp-green-100);
@@ -66,6 +63,7 @@
   --link--color: var(--grey-900);
   --link--visited--color: var(--grey-900);
   --link--hover--color: var(--grey-900);
+  --comment-reply-link--border-bottom: 2px solid var(--gp-green-500);
   --accordion-block-dark-style--background: var(--p4-dark-green-800);
   --accordion-block-dark-style--hover--background: var(--p4-dark-green-900);
   --accordion-block-light-style--background: var(--beige-100);
@@ -125,7 +123,7 @@ table.spreadsheet-table.is-color-gp-green {
   }
 }
 
-.standalone-link {
+a:where(.standalone-link, .comment-reply-link) {
   width: fit-content;
   font-weight: 700 !important;
 }
--- a/assets/src/scss/new-identity/style.scss
+++ b/assets/src/scss/new-identity/style.scss
@@ -125,27 +125,27 @@ table.spreadsheet-table.is-color-gp-green {
   }
 }
 
-.standalone-link,
-.comment-reply-link {
-  width: fit-content;
-  border-bottom: 2px solid var(--gp-green-500);
-  font-weight: 700 !important;
-}
-
 // Body Copy links selectors
 .post-content,
 .page-content,
 .standalone-link,
+.comment-reply-link,
 p > a,
 li > a {
-  transition: 0.2s cubic-bezier(0.5, 0, 0, 1);
   border-bottom: 2px solid var(--gp-green-500);
+  transition: 0.2s cubic-bezier(0.5, 0, 0, 1);
 
   &:hover,
   &:focus {
     box-shadow: inset 0 calc(-22px) var(--gp-green-100);
     text-decoration: none !important;
   }
+}
+
+.standalone-link,
+.comment-reply-link {
+  width: fit-content;
+  font-weight: 700 !important;
 }
 
 // Colors for transparent buttons.
--- a/assets/src/scss/new-identity/style.scss
+++ b/assets/src/scss/new-identity/style.scss
@@ -1,5 +1,16 @@
 @import "../base/tokens";
 
+@mixin shared-link-styles {
+  border-bottom: 2px solid var(--gp-green-500);
+  transition: 0.3s cubic-bezier(0.5, 0, 0, 1);
+
+  &:hover,
+  &:focus {
+    box-shadow: inset 0 calc(-22px) var(--gp-green-100);
+    text-decoration: none !important;
+  }
+}
+
 :root {
   --body--color: var(--grey-900);
   --site-footer--background: var(--p4-dark-green-800);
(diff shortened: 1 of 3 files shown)


planet-4:

:watch: 2023.03.22 18:57:44

assets/src/scss/pages/post/_comments-block.scss
```diff
@@ -55,6 +55,8 @@
     .comment-reply-link {
       float: right;
       margin-top: 0;
+      color: var(--comment-reply-link--color, var(--link--color));
```
mleray, line 58: Wouldn't these styles automatically be applied if we merge the changes for [this ticket](https://jira.greenpeace.org/browse/PLANET-6990)? 🤔

assets/src/scss/pages/post/_comments-form.scss
```diff
@@ -1,6 +1,6 @@
 .comments-block {
   .form-section .comment-respond {
-    background: $grey-05;
+    background: var(--color-background-comments_block, $grey-05);
```
mleray, line 3: Here can't you use `_--` or something like `--comments-block--` to automatically create a variable? If not, at least we should respect the same logic for the name, so start with the element name and then the CSS property 🙂

GP-Dan-Tovbein: Don't you think if it's better to start using tokens instead? `--color-background-comments_block` it's p4 token.

Also, I think the CSS variable implementation is going to be removed in the future whenever we have tokens already implemented. 
What do you think?

assets/src/scss/pages/post/_comments-block.scss
GP-Dan-Tovbein: It make sense your comment!

assets/src/scss/new-identity/style.scss
```diff
@@ -15,6 +15,9 @@
   --button-primary--color: var(--grey-900);
   --button-primary--visited--color: var(--grey-900);
   --button-primary--hover--color: var(--grey-900);
+  --comment-reply-link--color: var(--grey-900);
```
mleray, line 18: I would remove all those link-related variables here, since they should be fixed by https://github.com/greenpeace/planet4-master-theme/pull/1966

assets/src/scss/pages/post/_comments-form.scss
```diff
@@ -1,11 +1,10 @@
 .comments-block {
-  .form-section .comment-respond {
```
GP-Dan-Tovbein, line 2: I don't see necessity of looking into the `.form-section`. I think `.comment-responde` is fine.

assets/src/scss/pages/post/_comments-block.scss
GP-Dan-Tovbein: 👍 Updated!

assets/src/scss/new-identity/style.scss
```diff
@@ -62,6 +62,7 @@
   --link--color: var(--grey-900);
   --link--visited--color: var(--grey-900);
   --link--hover--color: var(--grey-900);
+  --comment-reply-link--border-bottom: 2px solid var(--gp-green-500);
```
mleray, line 65: Is this still needed since you added the `comment-reply-link` to this [selector](https://github.com/greenpeace/planet4-master-theme/pull/1959/files#diff-b418b2241107b99a37ed0291b0385d4ad8fe3762dd4d859be83bf83c070428d7R128)?

GP-Dan-Tovbein: I removed this variable and update [this line of code](https://github.com/greenpeace/planet4-master-theme/pull/1959/files#diff-b418b2241107b99a37ed0291b0385d4ad8fe3762dd4d859be83bf83c070428d7R129)

assets/src/scss/pages/post/_comments-block.scss
```diff
@@ -55,6 +55,7 @@
     .comment-reply-link {
       float: right;
       margin-top: 0;
+      border-bottom: none;
```
mleray, line 58: I don't think this is needed, since we add the border bottom only when the new identity styles are enabled? And then you can probably remove the `!important` from [here](https://github.com/greenpeace/planet4-master-theme/pull/1959/files#diff-b418b2241107b99a37ed0291b0385d4ad8fe3762dd4d859be83bf83c070428d7R129)

assets/src/scss/pages/post/_comments-form.scss
mleray: Actually, the reason why we made that difference is because the same `.comment-respond` form is used when replying to a comment, but currently in this case there's no background: see this screenshot from GPI website:

Screenshot 2023-04-21 at 08 57 19

With this new implementation, this reply form will also have a background color, I think we should check with the design team if they want to make that change:

Screenshot 2023-04-21 at 09 00 19

assets/src/scss/pages/post/_comments-block.scss
GP-Dan-Tovbein: Good catch! You're always correct 🥇

assets/src/scss/new-identity/style.scss
```diff
@@ -121,13 +124,15 @@ table.spreadsheet-table.is-color-gp-green {
   }
 }
 
-a:where(.standalone-link) {
+a:where(.standalone-link, .comment-reply-link, #cancel-comment-reply-link) {
+  color: var(--link--color) !important;
```
GP-Dan-Tovbein, line 128: I know this is redundant but we'll need to enforce to use `--link--color`. For example the `#cancel-comment-reply-link` uses a grey color.

Screenshot 2023-04-21 at 08 05 04

mleray: Maybe it's wanted though, I mean maybe the "Cancel reply" link should remain the same grey as it is now? We should check with the design team

mleray: There's a similar link style in the cookies box for the "Reject all" link:

Screenshot 2023-04-21 at 13 37 09

assets/src/scss/new-identity/style.scss
```diff
@@ -74,6 +74,9 @@
   --nav-link--visited--color: var(--grey-900);
   --nav-search-clear--background-color: var(--grey-900);
   --top-navigation--separation: 1px solid var(--grey-300);
+
+  // TODO: Review these fallback variables after we'll merge the new identity colors
+  --comments-block--comment-respond--background: var(--color-background-comments_block);
```
mleray, line 79: This variable no longer exists since you added the `.form-section` back into the selector, it's now `--comments-block--form-section--comment-respond--background` which is very long 😬 We could either use this or define another name

assets/src/scss/new-identity/style.scss
mleray: I think that we don't need the `!important` anymore since the cancel link remains grey? 🤔

assets/src/scss/new-identity/style.scss
```diff
@@ -176,3 +181,7 @@ $palette: (
     }
   }
 }
+
+.logged-in-as a {
+  color: var(--gp-green-800) !important;
```
mleray, line 186: Maybe here instead of `!important` we should write:
```
.logged-in-as a {
  color: var(--gp-green-800);

  &:hover,
  &:visited {
    color: var(--gp-green-800);
  }
}
```
I know it's longer, but it would be easier to override for NROs

assets/src/scss/new-identity/style.scss
```diff
@@ -121,13 +124,16 @@ table.spreadsheet-table.is-color-gp-green {
   }
 }
 
-a:where(.standalone-link) {
+.standalone-link,
+.comment-reply-link {
+  color: var(--link--color);
```
mleray, line 129: Actually I think that we can completely remove that line, they already use the `--link--color` variable

assets/src/scss/new-identity/style.scss
```diff
@@ -1,5 +1,16 @@
 @import "../base/tokens";
 
+@mixin shared-link-styles {
```
GP-Dan-Tovbein, line 3: @mleray I've created this `mixin` to share these styles through the other links.

GP-Dan-Tovbein: I prefer to keep it here and move to the `_mixins.scss` once the new identity is released.

What do you think?

assets/src/scss/new-identity/style.scss
```diff
@@ -122,27 +136,23 @@ table.spreadsheet-table.is-color-gp-green {
...
-
-    &:hover {
-      box-shadow: inset 0 calc(-22px) var(--gp-green-100);
-      text-decoration: none !important;
-    }
+    @include shared-link-styles;
   }
 }
 
+.standalone-link,
```
mleray, line 148: Instead of using a mixin, we could also just have kept these two under the `.post-content, .page-content` selector 🤔 I can't think of a case where they would be somewhere else?

GP-Dan-Tovbein: But the comments block is not into the page|post content

mleray (approved):
Tested on local, looks good to me 👍 🎉

PR #1961, opened 2023-03-23
Remove buttons box-shadow

mleray:
### Description

I've checked with the design team and we can already make this change regardless of the new identity.

### Testing

All buttons should no longer have a `box-shadow`.


sagarsdeshmukh (approved):
Tested on local, PR looks good to me 👍

PR #1965, opened 2023-03-29
PLANET-6994: Implement new identity colors to standalone links

GP-Dan-Tovbein:
Ref: https://jira.greenpeace.org/browse/PLANET-6994

As a part of the new identity, It's Including the new `hover` effect to some links (not all).

**IMPORTANT NOTE**
I was trying to find the best CSS pure way to define the `offset-y` (as `-22px`) through the `drop-shadow` property, without using Javascript, but I haven't find it yet. 
The problem there is that `22px`  is applied to ALL links with `font-sizes` higher than `22px` and lower. Thus, doesn't matter which is the `font-size`, the `inset` drop shadow will be always `22px` and I disagree with that 🙃. 
Of course we can do it directly with Vainilla JS, but the performance will be impacted.

### DIFF
--- a/assets/src/scss/new-identity/style.scss
+++ b/assets/src/scss/new-identity/style.scss
@@ -84,6 +84,7 @@ a:where(.standalone-link) {
 
   &:hover {
     box-shadow: inset 0 calc(-22px) var(--gp-green-100);
+    text-decoration: none !important;
   }
 }
 
--- a/assets/src/scss/new-identity/style.scss
+++ b/assets/src/scss/new-identity/style.scss
@@ -80,7 +80,7 @@ a:where(.standalone-link) {
   width: fit-content;
   border-bottom: 2px solid var(--gp-green-500);
   font-weight: 700 !important;
-  transition: 0.2s cubic-bezier(0.5, 0, 0, 1);
+  transition: 0.3s cubic-bezier(0.5, 0, 0, 1);
 
   &:hover {
     box-shadow: inset 0 calc(-22px) var(--gp-green-100);
--- a/assets/src/scss/new-identity/style.scss
+++ b/assets/src/scss/new-identity/style.scss
@@ -76,6 +76,17 @@ table.spreadsheet-table.is-color-dark-green {
   }
 }
 
+a:where(.standalone-link) {
+  width: fit-content;
+  border-bottom: 2px solid var(--gp-green-500);
+  font-weight: 700 !important;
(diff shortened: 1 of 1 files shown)


planet-4:

:watch: 2023.03.29 09:53:59

mleray (commented):
Is this new PR replacing [this one](https://github.com/greenpeace/planet4-master-theme/pull/1966/)?

assets/src/scss/new-identity/style.scss
```diff
@@ -52,6 +52,10 @@
   --block-spreadsheet-blue-header--background: #167f82;
   --block-spreadsheet-blue-odd-row--background: var(--beige-200);
   --block-spreadsheet-blue-even-row--background: var(--beige-100);
+  --link--color: var(--grey-900);
+  --link--visited--color: var(--grey-900);
+  --link--hover--color: var(--grey-900);
+  --link--hover--text-decoration: none;
```
mleray, line 58: This also affects links that **should** have an underline on hover, such as the contextual links... That means we'll need a new CSS variable for those links

mleray:
There is still the underline on hover, in addition to the new border-bottom:
Screenshot 2023-04-06 at 15 43 48

assets/src/scss/new-identity/style.scss
```diff
@@ -76,6 +76,18 @@ table.spreadsheet-table.is-color-dark-green {
...
 
+a:where(.standalone-link) {
+  width: fit-content;
+  border-bottom: 2px solid var(--gp-green-500);
+  font-weight: 700 !important;
+  transition: 0.3s cubic-bezier(0.5, 0, 0, 1);
+
+  &:hover {
+    box-shadow: inset 0 calc(-22px) var(--gp-green-100);
+    text-decoration: none !important;
```
GP-Dan-Tovbein, line 87: This is because of [this](https://github.com/greenpeace/planet4-master-theme/blob/main/assets/src/scss/base/_css-variables.scss#L9)

mleray: Since most links will have this style, I think it would be better to override the `--link--hover--text-decoration` to `none` and add new variables for the links (such as contextual links) that still need the underline on hover... But we can also think about that in the PR for body links!

mleray (approved):
This is good for now 👍 But as a follow-up, since these styles will also need to be applied to body links we will need to review the selector and probably override some of the global link variables

PR #1969, opened 2023-04-04
Convert navbar variables to CSS variables

GP-Dan-Tovbein:
## Description
As a part of https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/969 I'd suggest also to re implement the Navigation variables. 

By doing this will be more easier to maintain afterwards and to migrate to design tokens since we're are start using `CSS Variables`.

Here is the [demo page](https://www-dev.greenpeace.org/test-pluto/page-header-demo-page)


mleray (approved):
Tested on the instance, seems to work fine 👍

PR #1970, opened 2023-04-05
PLANET-7016 Apply the new colors to the Table block

Osong-Michael:
### Description: [PLANET-7016](https://jira.greenpeace.org/browse/PLANET-7016)

**Testing**

You can use this [page](https://www-dev.greenpeace.org/test-umbriel/test-new-identity-colors/) on the Umbriel instance to test the new changes for the Table block. Switching between the new identity styles and old ones should be reflected on the table.

Ref PR: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/1029

### DIFF
--- a/assets/src/scss/new-identity/style.scss
+++ b/assets/src/scss/new-identity/style.scss
@@ -42,36 +42,65 @@
   --button-secondary--hover--background: var(--p4-dark-green-800);
   --button-secondary--hover--border-color: var(--p4-dark-green-800);
   --button-secondary--hover--color: var(--white);
-  --block-spreadsheet--color: var(--grey-900);
-  --block-spreadsheet-grey-header--background: var(--grey-800);
-  --block-spreadsheet-grey-odd-row--background: #eaedee;
-  --block-spreadsheet-grey-even-row--background: var(--grey-100);
-  --block-spreadsheet-green-header--background: var(--gp-green-800);
-  --block-spreadsheet-green-odd-row--background: var(--beige-200);
-  --block-spreadsheet-green-even-row--background: var(--beige-100);
-  --block-spreadsheet-blue-header--background: #167f82;
-  --block-spreadsheet-blue-odd-row--background: var(--beige-200);
-  --block-spreadsheet-blue-even-row--background: var(--beige-100);
+  --table-footer-new--background: #dad8d1;
+  --table-header-grey--background: var(--grey-800);
+  --table-odd-row-grey--background: var(--grey-200);
+  --table-even-row-grey--background: var(--grey-100);
+  --table-footer-grey--background: var(--grey-300);
+  --table-header-green--background: var(--p4-dark-green-800);
+  --table-odd-row-green--background: var(--beige-200);
+  --table-even-row-green--background: var(--beige-100);
+  --table-footer-green--background: var(--table-footer-new--background);
+  --table-header-blue--background: #167f82;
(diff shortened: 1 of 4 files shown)


planet-4:

:watch: 2023.04.05 09:07:12

assets/src/scss/new-identity/style.scss
```diff
@@ -52,6 +52,22 @@
   --block-spreadsheet-blue-header--background: #167f82;
   --block-spreadsheet-blue-odd-row--background: var(--beige-200);
   --block-spreadsheet-blue-even-row--background: var(--beige-100);
+  --table-header-grey--background: var(--grey-800);
```
mleray, line 55: It just occurred to me that these values are the same as for the Spreadsheet block, maybe we could use the `--table` variables for both blocks? 🤔 That would simplify the code quite a bit

assets/src/scss/new-identity/style.scss
```diff
@@ -52,6 +52,22 @@
...
+  --table-footer-grey--background: var(--grey-200);
+  --table-header-green--background: var(--p4-dark-green-800);
+  --table-odd-row-green--background: var(--beige-200);
+  --table-even-row-green--background: var(--beige-100);
+  --table-footer-green--background: var(--beige-200);
+  --table-header-blue--background: #167f82;
+  --table-odd-row-blue--background: var(--beige-200);
+  --table-even-row-blue--background: var(--beige-100);
+  --table-footer-blue--background: var(--beige-200);
+  --table-header-gp-green--background: var(--gp-green-800);
```
mleray, line 67: It's more a matter of preference, but I would personally prefer to use `dark-green` rather than `gp-green` in the name, I feel it's more clear what the difference is

Osong-Michael: Actually we already have a _Dark green_ for the new identity styles, so this was more to reference the color itself so as not to cause confusion when looking at the code.

assets/src/scss/new-identity/style.scss
Osong-Michael: Yea, that is true, I will refactor and make the necessary changes

assets/src/scss/new-identity/style.scss
mleray: Well it's `green` or `dark green` right? Also for the colors' names I would use `Green` and `Dark Green` (like for the Spreadsheet block), I think for editors it's clearer... But maybe that should be discussed with Houssam, maybe he would prefer to use the names that correspond to the color palette instead 🤷‍♀️

mleray: I guess that's for him to decide, which labels he prefers, but either way we should make them consistent with the Spreadsheet ones 🙂

assets/src/scss/new-identity/style.scss
```diff
@@ -42,38 +42,67 @@
...
   --link--visited--color: var(--grey-900);
   --link--hover--color: var(--grey-900);
 }
 
 // Spreadsheet block new color.
 table.spreadsheet-table.is-color-dark-green {
-  th --block-spreadsheet-dark-green-header-- {
-    background: var(--p4-dark-green-800);
+  th {
+    background-color: var(--p4-dark-green-800);
```
mleray, line 70: Maybe here we can use the new `--table-header-gp-green--...` variables like for the Table block?

PR #1974, opened 2023-04-06
PLANET-7014 Apply new identity colors to Accordion block

mleray:
### Description

See [PLANET-7014](https://jira.greenpeace.org/browse/PLANET-7014)
These will only be applied if the new identity styles are enabled

**Related PR:** https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/1031

### Testing

You can add an Accordion block on local or on the janus test instance, making sure you enabled the new identity styles, and test out the various styles. You can also check [this page](https://www-dev.greenpeace.org/test-janus/all-the-accordions/) that I made for UAT. Make sure the block also still looks fine with the "old" styles, if the new identity is not enabled.

### DIFF
--- a/assets/src/scss/new-identity/style.scss
+++ b/assets/src/scss/new-identity/style.scss
@@ -140,3 +140,16 @@ $palette: (
     --transparent-button--active--color: #{$color};
   }
 }
+
+// Accordion block new color.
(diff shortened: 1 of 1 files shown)


sagarsdeshmukh (approved):

PR #1978, opened 2023-04-07
PLANET-7019 Remove navigation bar dark style for new identity

mleray:
### Description

See [PLANET-7019](https://jira.greenpeace.org/browse/PLANET-7019)
This will only be applied if the new identity styles are enabled. Once we merge the new identity, we'll need to run a migration to properly remove this setting from existing data.

### Testing

Either on local or on the [telesto test instance](https://www-dev.greenpeace.org/test-telesto/), make sure that the Planet 4 `Navigation > Website Navigation Style` setting is now only available in the backend when the new identity styles are disabled. When they are enabled, the light style should always be applied in the frontend.

PR #1979, opened 2023-04-07
PLANET-6996 Apply new identity colors to the navigation bar

mleray:
### Description

See [PLANET-6996](https://jira.greenpeace.org/browse/PLANET-6996)
These will only be applied if the new identity styles are enabled. This should only be merged after [this PR](https://github.com/greenpeace/planet4-master-theme/pull/1978), to make sure it's always the light style that is used, because with the dark style it looks weird 😅

### Testing

Make sure that the new identity styles setting is enabled (`Appearance > Customize > Site identity > Enable new Greenpeace visual identity`) and then you should see the new navigation bar colors. You can also test the changes on the [deimos test instance](https://www-dev.greenpeace.org/test-deimos/) where I already toggled the new identity styles.

### DIFF
--- a/assets/src/scss/layout/navbar/_search.scss
+++ b/assets/src/scss/layout/navbar/_search.scss
@@ -1,7 +1,7 @@
 .nav-search-form {
   background-color: var(--body--background-color);
   border: none;
-  border-top: 1px solid #c4c4c4;
+  border-top: var(--top-navigation--separation, 1px solid #c4c4c4);
   box-shadow: 0 2px 4px transparentize($black, 0.85);
   display: none;
   height: 56px;
@@ -102,6 +102,10 @@
     color: inherit;
     width: calc(100% - 48px);
   }
+
+  &::placeholder {
+    color: $grey-40;
+  }
 }
 
 #search_input:placeholder-shown ~ button.nav-search-clear {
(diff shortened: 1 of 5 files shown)


assets/src/scss/partials/navigation-bar-light.scss
```diff
@@ -15,14 +15,5 @@
...
 #nav-mobile-menu {
   --nav-link--color: #{$grey-40};
-  --nav-link--active--color: #{$black};
   --nav-link--opacity: 1;
 }
-
-.nav-languages-toggle {
-  @include large-and-up {
-    &::after {
-      background-color: $grey-80;
```
mleray, line 25: This is no longer needed since we now use `currentColor`

assets/src/scss/partials/navigation-bar-light.scss
```diff
@@ -15,14 +15,5 @@
 
 #nav-mobile-menu {
   --nav-link--color: #{$grey-40};
-  --nav-link--active--color: #{$black};
```
mleray, line 18: I think that this was not needed since we already set it to `black` earlier on in the file

assets/src/scss/layout/navbar/_search.scss
```diff
@@ -102,6 +102,10 @@
     color: inherit;
     width: calc(100% - 48px);
   }
+
+  &::placeholder {
```
mleray, line 106: We kept the browser default here before, but Houssam requested this color, since it's also part of our current colors (not only the new identity ones) I assume we can make the change right away

lithrel (approved):
:+1:

PR #1983, opened 2023-04-12
Adjust GF redirection message

comzeradd:
That would make it easier for editors to translate since it would appear as one line in Loco. This also gives them the option to completely blank the text if they don't wish to appear at all. 

I had to remove formatting too from the translatable part (except from `href`) to make it easier for editors.

image


GP-Dan-Tovbein (approved):
Code looks good to me! 👍

PR #1986, opened 2023-04-14
Add empty text case to cookie notice e2e test

mleray:
### Description

If the P4 settings cookies text is empty then the notice will not show. We should include this case in our test.

### Testing

On local, you can run `npx playwright test tests/e2e/cookies-banner.spec.js` with and without a cookies text defined in the P4 settings, and the test should pass in both cases.


planet-4:

:watch: 2023.04.14 07:58:08

PR #1996, opened 2023-04-21
PLANET-7132: Apply new identity colors to Split Two Columns block

GP-Dan-Tovbein:
Ref: https://jira.greenpeace.org/browse/PLANET-7132

## Description
Used the [SASS variables recently merged](https://github.com/greenpeace/planet4-design-tokens/pull/8).

You can compare the [new changes](https://www-dev.greenpeace.org/test-pandora/new-identity-layouts-links/) versus the [current behavior](https://www-dev.greenpeace.org/test-nix/explore-2/).

### DIFF
--- a/assets/src/scss/new-identity/style.scss
+++ b/assets/src/scss/new-identity/style.scss
@@ -1,13 +1,25 @@
 @import "../base/tokens";
+@import "../base/variables";
+@import "../base/mixins";
 
 @mixin shared-link-styles {
+  --link--height: 26px;
+  --transition--duration: 0.3s;
+
   border-bottom: 2px solid var(--gp-green-500);
-  transition: 0.3s cubic-bezier(0.5, 0, 0, 1);
+  transition: var(--transition--duration) cubic-bezier(0.5, 0, 0, 1);
+
+  &:hover {
+    box-shadow: inset 0 calc(var(--link--height) * -1) rgba($gp-green-500, .15);
+  }
 
   &:hover,
-  &:focus {
-    box-shadow: inset 0 calc(-22px) var(--gp-green-100);
+  &:active {
     text-decoration: none !important;
+  }
+
+  &:focus {
+    --link--focus--border-radius: 0;
   }
 }
 
@@ -156,11 +168,26 @@ table.spreadsheet-table.is-color-gp-green {
 
 // Standalone links
 .standalone-link,
-.comment-reply-link {
+.comment-reply-link,
+.split-two-column-item-link {
   @include shared-link-styles;
 
+  display: inline-block;
   width: fit-content;
+  height: var(--link--height);
   font-weight: 700 !important;
+}
+
+.split-two-column-item-link {
+  --link--color: var(--white);
+  --link--hover--color: var(--white);
+
+  text-shadow: none !important;
+
+  @include x-large-and-up {
+    --transition--duration: 0.4s;
+    --link--height: 34px;
+  }
 }
 
 // Colors for transparent buttons.


planet-4:

:watch: 2023.04.21 14:47:45

assets/src/scss/new-identity/style.scss
```diff
@@ -121,18 +121,41 @@ table.spreadsheet-table.is-color-gp-green {
   }
 }
 
-a:where(.standalone-link) {
+a:where(.standalone-link, .split-two-column-item-link) {
+  --box-shadow--offset-inset: inset;
```
GP-Dan-Tovbein, line 125: I've added this local CSS Variables to avoid duplicating properties and make it easy to follow up.

assets/src/scss/new-identity/style.scss
```diff
@@ -121,18 +121,41 @@ table.spreadsheet-table.is-color-gp-green {
...
+      var(--gp-green-100);
     text-decoration: none !important;
   }
 }
 
+.split-two-column-item-link {
+  --link--color: white;
+  --link--hover--color: white;
+
+  text-shadow: none !important;
```
mleray, line 148: Maybe we could remove this right away, I mean without waiting for the new identity? We would need to double-check with Houssam of course, but I see that we also have a `text-shadow` on some other elements of that block (the title, subtitle, etc) and I think it looks a bit weird 😅

assets/src/scss/new-identity/style.scss
```diff
@@ -121,18 +121,41 @@ table.spreadsheet-table.is-color-gp-green {
   }
 }
 
-a:where(.standalone-link) {
+a:where(.standalone-link, .split-two-column-item-link) {
```
mleray, line 124: Why don't we directly write `.standalone-link, .split-two-column-item-link` here? Is there an advantage to adding `a:where` in front? 🤔

mleray: I meant why not write directly `.standalone-link, .split-two-column-item-link { ... }` instead of `a:where(.standalone-link, .split-two-column-item-link) { ... }`, basically I'm not sure I understand why we add `a:where`

assets/src/scss/new-identity/style.scss
GP-Dan-Tovbein: I agree, but I think it should be in another PR that change

mleray: I agree 👍 Can you please create a follow-up ticket?

GP-Dan-Tovbein: Here is [PLANET-7138](https://jira.greenpeace.org/browse/PLANET-7138) !

GP-Dan-Tovbein:
> I think that the link height looks off on tablet, what do you think?
> 
> Screenshot 2023-04-26 at 10 53 33
> 
> Screenshot 2023-04-26 at 10 53 42

Hey @mleray, it's now updated!

mleray (commented):
I just noticed that the `:active` selector still has an underline and therefore looks weird (it's also the case for body links)

Screenshot 2023-04-28 at 16 39 18

assets/src/scss/new-identity/style.scss
```diff
@@ -1,12 +1,17 @@
 @import "../base/tokens";
+@import "../base/variables";
+@import "../base/mixins";
 
 @mixin shared-link-styles {
+  --link--height: 26px;
+  --box-shadow--offset-y: calc(var(--link--height) * -1);
```
mleray, line 7: Do we still need this variable? I don't see it reassigned anywhere 🤔

GP-Dan-Tovbein:
> I just noticed that the `:active` selector still has an underline and therefore looks weird (it's also the case for body links)
> 
> Screenshot 2023-04-28 at 16 39 18

I've added the `:active`too. For somehow, I don't see that on my local.

assets/src/scss/new-identity/style.scss
GP-Dan-Tovbein: It was part of the previous implementation,

assets/src/scss/new-identity/style.scss
```diff
@@ -1,12 +1,18 @@
...
+
   border-bottom: 2px solid var(--gp-green-500);
-  transition: 0.3s cubic-bezier(0.5, 0, 0, 1);
+  transition: var(--transition--duration) cubic-bezier(0.5, 0, 0, 1);
 
   &:hover,
-  &:focus {
-    box-shadow: inset 0 calc(-22px) var(--gp-green-100);
+  &:active {
+    box-shadow: inset 0 var(--box-shadow--offset-y) rgba($gp-green-500, .15);
```
GP-Dan-Tovbein, line 15: FYI, After a call with Houssam, he propose to use the same `box-shadow` to all these cases.

assets/src/scss/new-identity/style.scss
```diff
@@ -1,12 +1,16 @@
...
+  --transition--duration: 0.3s;
+
   border-bottom: 2px solid var(--gp-green-500);
-  transition: 0.3s cubic-bezier(0.5, 0, 0, 1);
+  transition: var(--transition--duration) cubic-bezier(0.5, 0, 0, 1);
 
-  &:hover,
-  &:focus {
-    box-shadow: inset 0 calc(-22px) var(--gp-green-100);
+  &:hover {
```
mleray, line 12: The `:active` selector should be added here! Right now it still has the extra underline.

Screenshot 2023-05-03 at 08 52 21

I've also just noticed that right now on focus there is a [border-radius](https://github.com/greenpeace/planet4-master-theme/blob/main/assets/src/scss/base/_typography.scss#L106) that should be applied on `:focus-visible` instead, but that could be done in a separate PR!

Screenshot 2023-05-03 at 08 50 23

GP-Dan-Tovbein: I remember I pushed this change 🤔 . I'll update it soon !

assets/src/scss/new-identity/style.scss
```diff
@@ -1,14 +1,26 @@
...
+
+  &:hover {
+    box-shadow: inset 0 calc(var(--link--height) * -1) rgba($gp-green-500, .15);
+  }
 
   &:hover,
-  &:focus {
-    box-shadow: inset 0 calc(-22px) var(--gp-green-100);
+  &:active {
     text-decoration: none !important;
```
mleray, line 18: Maybe if we remove [these lines](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/main/assets/src/styles/blocks/SplitTwoColumns.scss#L155-L159) (useless anyways because these links follow body links styles) then the `!important` is no longer needed?

assets/src/scss/new-identity/style.scss
```diff
@@ -1,14 +1,26 @@
...
   }
+
+  &:focus {
+    --link--focus--border-radius: 0;
```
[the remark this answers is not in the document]
GP-Dan-Tovbein: Since it's also used [here](https://github.com/greenpeace/planet4-master-theme/blob/8027e00a203eca3f55da7f47274ae9abf1aed936/assets/src/scss/layout/_country-selector.scss#L20). What do you think?

assets/src/scss/new-identity/style.scss
GP-Dan-Tovbein: To keep the current functionality still working as the UI expected, I would prefer to use a `CSS variable` 
```css
  &:hover {
    text-decoration: var(--link--hover--text-decoration, underline);
  }
```
and then change that local variable through the new-identity stylesheet
```css
  &:hover,
  &:active {
    --link--hover--text-decoration: none;
  }
```

Does it make sense?

GP-Dan-Tovbein: See this [other (and related) PR](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/1045)

assets/src/scss/new-identity/style.scss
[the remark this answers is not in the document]
mleray: I think that in both places it should be used in the `:focus-visible` selector instead of `:focus` 🤔 And then that should solve the problem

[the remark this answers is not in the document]
GP-Dan-Tovbein: I assume it's not only to replace to `:focus-visible` but also to re-think this part too.
```css
  button:focus {
    border-radius: var(--link--focus--border-radius, 4px);
    ...
  }

  button:focus:not(:focus-visible) {
    ...
  }
```

To be honest, I don't like to touch the current behavior of other blocks, especially when they are not part of the ticket's scope. In most of the cases I'd prefer to keep the new identity styles more cleaner with the newest changes, including overrides, and whenever we'll merge it then analyze the part of replacing the newest code with the oldest one, but it's up to you! Does it make sense?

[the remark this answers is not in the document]
mleray: Yes, you're right, we can think about this later! Maybe for now for the new identity it's fine to use `!important` once in a while.

assets/src/scss/new-identity/style.scss
```diff
@@ -1,13 +1,25 @@
...
-  transition: 0.3s cubic-bezier(0.5, 0, 0, 1);
+  transition: var(--transition--duration) cubic-bezier(0.5, 0, 0, 1);
+
+  &:hover {
+    box-shadow: inset 0 calc(var(--link--height) * -1) rgba($gp-green-500, .15);
+  }
 
   &:hover,
+  &:active {
+    --link--hover--text-decoration: none;
```
mleray, line 18: I still think here it would be simpler to just remove these overrides in the Split Two Columns styles since they are useless anyway. And then here we could just use `text-decoration`.
But if you think that's too many changes you can also revert to using `!important`, I guess that's not a big deal for now and we can figure out the refactor when merging the new identity styles...

GP-Dan-Tovbein: From my point of view, for some reasons I'd prefer to use `!important` even if it isn't the best approach.

- We'll keep the current behavior up to date and nothing could be potentially broken. If something is broken then we'll blame the `new-identity` stylesheet 😆.  
- Looking at the `!important` in the future, we'll understand that that part of the code needs to be evaluated after we'll merge the new feature.

mleray: Yeah I'm fine with that, your arguments make sense! 👍

PR #1999, opened 2023-04-26
Fix reading time translation in listing pages

mleray:
### Description

We forgot to add the possibility to translate this string here, it was reported on Slack! You can find an example page of the issue [here](https://www.greenpeace.org/brasil/blog/)

### DIFF
--- a/src/Post.php
+++ b/src/Post.php
@@ -539,7 +539,12 @@ class Post extends TimberPost
         WP_Block $block
     ): string {
         $time = ( new self($block->context['postId'] ?? null) )->reading_time();
-        return $time ? '<span class="article-list-item-readtime">' . $time . ' min read</span>' : '';
+        return $time ?
+            '<span class="article-list-item-readtime">'
+                // translators: reading time in min.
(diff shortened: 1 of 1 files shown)


src/Post.php
```diff
@@ -539,7 +539,12 @@ public static function reading_time_block(
         WP_Block $block
     ): string {
         $time = ( new self($block->context['postId'] ?? null) )->reading_time();
-        return $time ? '<span class="article-list-item-readtime">' . $time . ' min read</span>' : '';
+        return $time ?
+            '<span class="article-list-item-readtime">'
+                // translators: reading time in min.
+                . str_replace('%d', $time, __('%d min read', 'planet4-master-theme')) .
```
lithrel, line 545: using [sprintf](https://www.php.net/sprintf) might be better, as it's made for this kind of formatting
something like `sprintf(__('%d min read', 'planet4-master-theme'), $time)`

lithrel (approved):

lithrel (approved):
:+1:

PR #2031, opened 2023-06-13
Remove notifications experiment

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-6237

---

Already checked with GPI and it's not actively being used at the moment. Removing this since we haven't looked into the code for quite some time now. This will give us time to re-design and re-think the specs of this feature.

_(Added one more commit to make `phpcs` happy)_

### Testing

- Just one more look to verify there is no other code that needs to be cleaned.


lithrel (approved):
:+1:

PR #2058, opened 2023-07-05
Fix link style rule

lithrel:
Links styles don't apply if a link is included in an inline tag like `<em>` or `<strong>`.
This is due to a child combinator `p > a` that doesn't allow for `p > em > a`. The declaration also misses `.post-content > article` selector for lists.

Example: https://planet4.greenpeace.org/story/16394/v2-105-106/

### DIFF
--- a/assets/src/scss/new-identity/style.scss
+++ b/assets/src/scss/new-identity/style.scss
@@ -286,8 +286,9 @@ table.spreadsheet-table.is-color-gp-green {
 
 // Body links
 .post-content,
+.post-content > article,
 .page-content {
-  p > a,
+  p:not(.article-list-item-meta) a,
   > ul li a,
   > ol li a {
     @include shared-link-styles;


mleray (commented):
On local on my homepage I see that the author links in the Articles block now have the new link styles, I don't think that's what we want?

Screenshot 2023-07-06 at 09 38 39

lithrel:
> On local on my homepage I see that the author links in the Articles block now have the new link styles, I don't think that's what we want?

I'm a bit confused as to how to solve this; everything will be in .page-content anyway, so should we add more specific rules for the author link in the Articles block or is there a more global way ?

mleray:
> I'm a bit confused as to how to solve this; everything will be in .page-content anyway, so should we add more specific rules for the author link in the Articles block or is there a more global way ?

`p :not(.article-list-item-author) a` seems to solve the problem, but it's not ideal 😕 We could do that for now, and maybe come up with something better when we release the new styles for good?

mleray (approved):
Looks good to me 👍

PR #2062, opened 2023-07-10
PLANET-7118 Update Media Archive import flow

mleray:
### Description

See [PLANET-7118](https://jira.greenpeace.org/browse/PLANET-7118) & [PLANET-7116](https://jira.greenpeace.org/browse/PLANET-7116)
This is part of the Media Archive redesign

### Testing

Either on local or on the [saturn test instance](https://www-dev.greenpeace.org/test-saturn/wp-admin/admin.php?page=media-picker), you can check in all screen sizes the process of importing an image and make sure that it follows the [designs](https://www.figma.com/file/cnd1hDu5MnXD2kmGxfwH4N/Media-Archive-Wireframes-(Final-Beta)?node-id=220%3A8208&mode=dev) (`Importing image (flow)` section) 
I've also implemented in this PR the changes from the `Edit option` section, so you can test that too 🙂

PR #2068, opened 2023-07-12
PLANET-7115: Implement media library multi search options

GP-Dan-Tovbein:
Ref: https://jira.greenpeace.org/browse/PLANET-7115

## Description
This implementation includes the new ticket's functionality which allows to the end-user to search by multiple options instead of a single one. You can search not only `texts` but also `SystemIdentifiers` which is a nice-to-have feature not part of the ticket's scope.

In addition to the multi search options this PR includes a big refactor of the base `ArchivePicker` code and some more UI/UX things that we've been discussing with the design team in the meanwhile of this implementation.
- Implement the [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) to prevent multi requests at the same time
- Implement `context` in a replacement of the `useImages` hook. By having a context, we'll be able to share a single instance to all the context's children. In addition to that, we won't need to implement tricky stuffs anymore, for example `timeouts` to update state vars. And the most important thing by using a `context` is to avoid passing multiple `props` within children which in some case is really difficult to maintain.
- Implement `reducer` to have a shared state by using the `dispatcher` and avoid using a lot the the tricky and asynchronous `useState` hook.
- Encourage the use of `useEffect` for fetching or updating data.
- Integrate the `useMemo` and `useCallback` to memorize data and prevent "re-renders".
- Move the list logic to an isolated component to be easier to maintain.

All of these stuffs will improve the component's performance, also make more easier for maintenance and better for scalability.

## Testing

**Heads up!**
Before we merge this, make user if #2062 is still working as expected.


planet-4:

:watch: 2023.07.14 16:47:36

admin/css/picker.css
```diff
@@ -138,12 +144,124 @@
...
+.close-icon {
+  background: url("../../images/black-cross.svg");
+  cursor: pointer;
+  height: 8px;
+  width: 8px;
+  background-size: contain;
+}
+
 .picker-sidebar .edit-image {
+  display: flex;
```
mleray, line 156: It seems all of these new properties are not needed for this button, if I remove them nothing changes 🤔 (except for the margin-top and the color changes)

admin/css/picker.css
```diff
@@ -138,12 +144,124 @@
...
+  margin: 0;
+  box-sizing: border-box;
+  overflow: hidden;
+}
+
+.multiple-search-form {
+  display: flex;
+  margin-top: 8px;
+  margin-bottom: 8px;
+  align-items: flex-start;
```
mleray, line 185: Isn't this the default value?

admin/css/picker.css
```diff
@@ -4,18 +4,23 @@
 :root {
   --sidebar-width-large: 450px;
   --sidebar-width-tablet: 400px;
+  --button-color: #3858e9;
```
mleray, line 7: I know this is part of the designs, but I personally think it looks weird, WordPress buttons in the rest of the interface look different (this color is for the Gutenberg editor only if I'm not mistaken)

mleray (commented):
This is what the input currently looks like when focused, the border looks a bit weird no?

Screenshot 2023-07-24 at 10 50 40

assets/src/js/Components/ArchivePicker.js
```diff
@@ -1,125 +1,274 @@
...
+          <h3>API error:</h3>
+          <p> {error.message} </p>
         </div>
       )}
-    </div>
-  </Fragment>;
-};
 
-export default ArchivePicker;
+      <div className={classNames('image-picker', {'open-sidebar': Object.keys(selectedImages).length})}>
```
mleray, line 244: Maybe you could create a `numberOfSelectedImages` variable or something like that instead of duplicating several times the `Object.keys(selectedImages).length` line? I think it would be a bit easier to read

GP-Dan-Tovbein: Totally agree!

admin/css/picker.css
mleray: Did you ask Taylor about this @GP-Dan-Tovbein ?

mleray (commented):
It seems it's limited to 6 search terms right? In this case it's a bit strange because the `Search` button is still available but clicking on it does nothing, maybe we should disable it in this situation?

src/MediaArchive/Rest.php
```diff
@@ -40,13 +40,34 @@ public static function register_endpoints(): void
...
+
             if ($search_text) {
                 // todo: avoid repetition of default Mediatype:Image param (maybe value object for query?).
-                $params['query'] = '(text:' . $search_text . ') and (Mediatype:Image)';
+
+                $system_identifier = [];
+                $text = [];
+
+                foreach ($search_text as $txt) {
+                    if (preg_match('/^GP/', $txt)) {
```
mleray, line 51: I'm not sure what should happen if we leave the image extension? 🤔  So for example if the user types `GP0STWY0M.jpg` instead of just `GP0STWY0M`? Right now the first one doesn't work, but maybe that's what we want

GP-Dan-Tovbein: Looking at the summary of this ticket, the idea is to search only by IDs (System Identifiers) and each of them starts with `GP`. If you type "GP0STWY0M.jpg" it will be searched by `text` by the API and you'll get a bunch of pictures since is matching with `GP` or `jpg`.

In addition to that, I've also implemented the possibility to search by `text`. That's why the query is now more complex.

mleray: If I understand the regex used here, you check if the `$text` starts with `GP` which is still the case for `GP0STWY0M.jpg`, so it will look for an ID, right? At least on local if I search for this (with the image extension) nothing shows

GP-Dan-Tovbein: As far as I was [looking here](https://docs.google.com/document/d/1qM0H2rbSuUqA66Yvd4UwYwP6M2T51lWm95r1dEWOPNA/edit), there is no way to include the file extension through the search. You can use `Multimedia` but for files. So, that is why I'm parsing the extension and remove it from the text.

mleray (approved):
Looks good to me 👍 Great job 👏

PR #2079, opened 2023-07-27
PLANET-7159 Add help tooltip to Media Archive

mleray:
### Description

See [PLANET-7159](https://jira.greenpeace.org/browse/PLANET-7159)

### Testing

You can test the changes by going to the Media Archive either on local or on the [telesto test instance](https://www-dev.greenpeace.org/test-telesto/wp-admin/upload.php?page=media-picker), and make sure the help tooltip looks and behaves as expected in all screen sizes.

### DIFF
--- a/admin/css/picker.css
+++ b/admin/css/picker.css
@@ -117,6 +117,7 @@
   background: white;
   width: 80%;
   font-size: 12px;
+  line-height: 1.5;
   border-radius: 4px;
   padding: 16px;
   margin-bottom: 8px;
@@ -359,6 +360,21 @@
 @media (min-width: 1200px) {
   .picker-sidebar {
     width: var(--sidebar-width-large);
+  }
+
+  .image-picker .help span {
+    line-height: 32px;
+    width: 32px;
+    height: 32px;
+    font-size: 18px;
+  }
+
+  .image-picker .help:hover .tooltip {
+    font-size: 14px;
+  }
+
+  .image-picker .help:hover .tooltip:after {
+    right: 8px;
   }
 }
 
--- a/admin/css/picker.css
+++ b/admin/css/picker.css
@@ -22,6 +22,11 @@
   display: grid;
   grid-template-columns: repeat(1, 1fr);
   gap: 0 16px;
+  position: relative;
+}
+
+.image-picker {
+  position: relative;
 }
 
 .image-picker li img {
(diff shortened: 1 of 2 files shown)


planet-4:

:watch: 2023.07.27 15:50:58

assets/src/js/Components/ArchivePicker.js
```diff
@@ -246,6 +258,20 @@ export default function ArchivePicker() {
       <div className={classNames('image-picker', {'open-sidebar': selectedImagesAmount > 0})}>
         <ArchivePickerList />
 
+        {images.length > 0 && (
```
GP-Dan-Tovbein, line 261: Checking `images.length` is enough!

assets/src/js/Components/ArchivePicker.js
```diff
@@ -246,6 +258,20 @@ export default function ArchivePicker() {
...
+          <div className="help">
+            <div
+              className="tooltip"
+              dangerouslySetInnerHTML={{
+                __html: __(
```
GP-Dan-Tovbein, line 266: I'm not sure if I understand why you use `dangerouslySetInnerHTML` instead of just adding content into the `div`.

assets/src/js/Components/ArchivePicker.js
mleray: Nope, if I do that a `0` shows when the page is loading

assets/src/js/Components/ArchivePicker.js
mleray: If I don't use `dangerouslySetInnerHTML` the HTML tags are shown (probably because it's in a translation)

assets/src/js/Components/ArchivePicker.js
GP-Dan-Tovbein: Just a comment, but if`images.length` is equal to `0`, `0` will be rendered since you're evaluating with `&&`. Whenever you check for arrays lengths, is better to use ternary conditions `? :`, and return `null` instead.

admin/css/picker.css
```diff
@@ -82,6 +87,62 @@
   opacity: 100%;
 }
 
+.image-picker.open-sidebar .help {
+  display: none;
```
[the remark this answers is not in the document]
mleray: That's necessary to hide it on mobile when the sidebar is open, because we have `z-index: 1`... However I changed the implementation a bit since and it seems that this `z-index` value is no longer needed, so I will update it!

assets/src/js/Components/ArchivePicker.js
mleray: I'm not sure I get why this is better than directly checking `images.length > 0`? Then you don't need the ternary 🤔

admin/css/picker.css
```diff
@@ -82,6 +87,62 @@
...
+  margin-bottom: 8px;
+  color: black;
+}
+
+.image-picker .help:hover .tooltip:after {
+  content: "";
+  position: absolute;
+  bottom: -14px;
+  right: 4px;
+  border-top: 8px solid white;
```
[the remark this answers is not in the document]
mleray: You're right! Updated 🙂

assets/src/js/Components/ArchivePicker.js
GP-Dan-Tovbein: To avoid getting `0` you can just replace to `!!images.length && ...` to cast it to `false`. This how the [short circuits works](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_precedence#short-circuiting).

GP-Dan-Tovbein: But take in mind this is not a required change 🙏

mleray: Updated!

GP-Dan-Tovbein (approved):
Code looks good to me!

Just a comment, I've tested on desktop and the icon is really difficult to see. I think we should talk to the design team and review its layout.

Screenshot 2023-08-02 at 07 08 02

PR #2084, opened 2023-08-01
PLANET-7117: Implment Bulk Select / Bulk Import in the media archive

GP-Dan-Tovbein:
Ref: [PLANET-7117](https://jira.greenpeace.org/browse/PLANET-7117)

# Description
Implement bulk select and bulk upload feature.
- Allow users to multi-select images by clicking on `Shift` key.  For example, if you click on index `4` and then on index `10`, all the pictures in between will be selected.
- Parallelize requests
- Allow users to multi-select images by clicking on `Alt` key.
- Moved the whole logic to `Components/ArchivePicker` for a better consistency.
- Add spinners to each image when it's being processed.
- Improve performance stunning
- Improve layout
Among other stuffs..

Also update the UI according to the newest updates. See on [Figma](https://www.figma.com/file/cnd1hDu5MnXD2kmGxfwH4N/Media-Archive-Wireframes-(Final-Beta)?type=design&node-id=220-7510&mode=design&t=dA1jMYtrxMLCX802-4). 

# Testing
Go to `Media > Archive` and click on the `Bulk Select` button.


planet-4:

:watch: 2023.08.01 02:49:39

mleray (commented):
I think there should be some space between these 2 buttons?

Screenshot 2023-08-01 at 11 09 39

Also maybe between the search input and the button on mobile

Screenshot 2023-08-01 at 11 20 46

assets/src/js/Components/archivePicker/ArchivePickerToolbar.js
```diff
@@ -0,0 +1,52 @@
...
+            }}
+            type="button"
+            className="button"
+          >{__('Cancel', 'planet4-master-theme-backend')}</button>
+        )}
+
+        {bulkSelect ?
+          <button
+            onClick={async () => {
+              if (window.confirm(`You are about to import [${Object.keys(selectedImages).length}] photos to the media library. 'Cancel' to stop, 'OK' to import.`)) { // eslint-disable-line no-alert
```
mleray, line 34: Why do we add `[]` around the amount of selected images here? I think it looks weird 🤔 
Also maybe we should add a verb to the last line, like `Press 'Cancel' to stop, 'OK' to import.`

Screenshot 2023-08-01 at 11 13 41

mleray (commented):
There's no indication that the bulk import is happening, maybe we should add something? Like a `Processing` text somewhere? Or at least we could disable all other actions, because right now the `Bulk import` button for example is still available then...

Also, we should clear the selected images once the importing is done, because when testing I've selected another image right after and it said I had selected 3 of them because it was still counting the 2 previously selected ones (I hope that makes sense 😅)

assets/src/js/Components/archivePicker/ArchivePickerList.js
```diff
@@ -47,18 +48,32 @@ export default function ArchivePickerList() {
...
                 }}
                 role="presentation"
               />
+              {
+                bulkSelect && !wordpress_id && (
+                  <div
+                    role="button"
+                    aria-hidden="true"
+                    className={classNames('bulk-select-checkbox', {'is-checked': selectedImages[image.id]})}
+                    onClick={() => {
```
mleray, line 71: I guess it's gonna be for the design team to decide, but maybe this should be implemented when clicking on the image rather than specifically on the checkbox?

mleray (commented):
It's a tiny detail, but maybe when bulk selecting we should remove the pointer cursor on the images that were already added to the library? Since we cannot do anything with them in this case, it could be confusing

assets/src/js/Components/archivePicker/ArchivePickerToolbar.js
GP-Dan-Tovbein: These 2 different questions are [answered here](https://www.figma.com/file/cnd1hDu5MnXD2kmGxfwH4N/Media-Archive-Wireframes-(Final-Beta)?type=design&node-id=220-7573&mode=dev).

About the `[]`, I copied the same text from the design. Do you think we should change it?

And regarding the modal, if you look at the bottom of design "Notes / Bahaviour", it says 
> "This is the native popup (in the bulk select/delete option in the media library), so I left that as is for now. I think we can revisit this at a later stage."   

What do you think?

assets/src/js/Components/archivePicker/ArchivePickerList.js
GP-Dan-Tovbein: It makes more sense!

assets/src/js/Components/archivePicker/ArchivePickerToolbar.js
mleray: Yeah I'm fine with leaving is as it, it's not a big deal 😄

mleray (commented):
We can select multiple images using the `Shift` key even when the bulk select isn't activated, in which case the selected images are just displayed in the sidebar:

Screenshot 2023-08-16 at 09 47 29

Is that expected behaviour?

admin/css/picker.css
```diff
@@ -313,11 +430,12 @@
   }
 
   .image-picker.open-sidebar {
-    width: calc(100% - var(--sidebar-width-tablet) - 52px);
+    width: 100%;
```
mleray, line 433: Probably because of that change, now when the sidebar is open we no longer see the help tooltip

GP-Dan-Tovbein: I've just pushed a new change which modify the way of scaling the list. The help tooltip is now visible every time.

GP-Dan-Tovbein:
> We can select multiple images using the `Shift` key even when the bulk select isn't activated, in which case the selected images are just displayed in the sidebar:
> Screenshot 2023-08-16 at 09 47 29
> 
> Is that expected behaviour? Seems a bit weird to me because no actions are possible with the images then

Before this implementation, you were able to multi select by clicking only the `Alt` key. For a better UX it's also possible to multi select by clicking on `Shift` key which does not refer to the `Bulk Select` feature.

admin/css/picker.css
mleray: I still don't see the help tooltip on mobile, and on tablet when the sidebar is open 😕

mleray: On large screens (not XL) it's also a bit cut off:

Screenshot 2023-08-16 at 13 27 44

mleray:
> Before this implementation, you were able to multi select by clicking only the `Alt` key. For a better UX it's also possible to multi select by clicking on `Shift` key which does not refer to the `Bulk Select` feature.

I see, I guess I'm just not sure how useful this is? Because as I mentioned in my first comment, if you select multiple images this way you can't do anything with them, they just show in the sidebar 🤔

GP-Dan-Tovbein:
> > Before this implementation, you were able to multi select by clicking only the `Alt` key. For a better UX it's also possible to multi select by clicking on `Shift` key which does not refer to the `Bulk Select` feature.
> 
> I see, I guess I'm just not sure how useful this is? Because as I mentioned in my first comment, if you select multiple images this way you can't do anything with them, they just show in the sidebar 🤔

I see your point and completely agree with your thoughts!
From my perspective, I don't see the importance of having the sidebar with multiple images at all, since as you mentioned before, is basically to see a bit bigger the image. However, if you add "Upload All" afterwards that will make more sense.

mleray (commented):
I think we're almost there 💪 2 small things I noticed while testing:
- the `Cancel` and `Bulk upload` buttons are still available while uploading images, maybe they should be disabled then
- the `Bulk upload` button is enabled even if no image is selected, maybe it should be disabled until it's no longer the case

Screenshot 2023-08-17 at 15 39 15

admin/css/picker.css
mleray: It's now a bit off, it should be on top of the last image but instead it's outside of the list

Screenshot 2023-08-18 at 13 41 26

mleray (approved):
Looks good to me, apart from a small comment on the help tooltip positioning that is not a blocker 👍

PR #2094, opened 2023-08-09
PLANET-7177 Move Default P4 Post Type setting under Settings > Writing

mleray:
### Description

See [PLANET-7177](https://jira.greenpeace.org/browse/PLANET-7177)

### Testing

The `Default P4 Post Type` setting should now be in `Settings > Writing`, you can make sure it still works as expected by creating a new post without any post type.

### DIFF
--- a/src/Migrations/M021MigrateDefaultPostType.php
+++ b/src/Migrations/M021MigrateDefaultPostType.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace P4\MasterTheme\Migrations;
+
+use P4\MasterTheme\MigrationRecord;
+use P4\MasterTheme\MigrationScript;
+use P4\MasterTheme\Settings\DefaultPostType;
+
+class M021MigrateDefaultPostType extends MigrationScript
+{
+    /**
+     * Perform the actual migration.
+     *
+     * @param MigrationRecord $record Information on the execution, can be used to add logs.
+     * phpcs:disable SlevomatCodingStandard.Functions.UnusedParameter -- interface implementation
+     */
+    public static function execute(MigrationRecord $record): void
+    {
+        $value = planet4_get_option('default_p4_pagetype');
+
+        $options = get_option('planet4_options');
+        unset($options['default_p4_pagetype']);
+        update_option('planet4_options', $options);
+
+        if (!$value) {
+            return;
+        }
+
+        update_option(DefaultPostType::KEY, $value);
+    }
+    // phpcs:enable SlevomatCodingStandard.Functions.UnusedParameter
+}
--- a/src/Settings/DefaultPostType.php
+++ b/src/Settings/DefaultPostType.php
@@ -1,0 +1,53 @@
+<?php
+
+namespace P4\MasterTheme\Settings;
+
+class DefaultPostType
+{
+    public const KEY = 'planet4_default_post_type';
+
+    public function __construct()
+    {
+        $this->hooks();
+    }
+
+    public function hooks(): void
+    {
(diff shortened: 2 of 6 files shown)


planet-4:

:watch: 2023.08.09 08:37:47

planet-4:

:watch: 2023.08.10 09:20:46

src/Migrations/M021MigrateDefaultPostType.php
```diff
@@ -0,0 +1,32 @@
...
+     *
+     * @param MigrationRecord $record Information on the execution, can be used to add logs.
+     * phpcs:disable SlevomatCodingStandard.Functions.UnusedParameter -- interface implementation
+     */
+    public static function execute(MigrationRecord $record): void
+    {
+        $value = get_option('default_p4_pagetype');
+
+        $options = get_option('planet4_options');
+        unset($options['default_p4_pagetype']);
```
sagarsdeshmukh, line 22: Hi @mleray , The old default p4 page type value is saved in `planet4_options` (i.e ``$options['default_p4_pagetype']), we are reseting old value that look fine, but on line # 29  we have to update the same value in new key, right? 🤔  (`update_option(DefaultPostType::KEY, $options['default_p4_pagetype']);`)

mleray: Yeah, it's saved in the `$value` variable from line 19, no?

mleray: By the time we get to line 29 `$options['default_p4_pagetype'])` is now unset so we can't use that

sagarsdeshmukh: Yes, we need to move line 29 before line 22 or save `default_p4_pagetype` into other var.

The `get_options()` function search the values for `option name`. 
Here in our case `get_option('default_p4_pagetype')` & `get_option('planet4_options')` return different values based on the option name we pass as a argument.
https://developer.wordpress.org/reference/functions/get_option/

mleray: Ah okay, from a quick test on local it seems that on line 19 I should use `planet4_get_option` instead of `get_option`, is that what you mean?

mleray: Similar to what we do in this [migration](https://github.com/greenpeace/planet4-master-theme/blob/main/src/Migrations/M019MigrateReadingTime.php)

sagarsdeshmukh: That should work, the `planet4_get_option()` is a wrapper function on `get_option()` wp function. 🙂 
```
function planet4_get_option(string $key = '', $default = null)
{
    $options = get_option('planet4_options');

    return $options[ $key ] ?? $default;
}
```

src/Settings/DefaultPostType.php
```diff
@@ -0,0 +1,53 @@
...
+                'hide_empty' => 0,
+                'orderby' => 'name',
+                'selected' => self::get_option(),
+                'name' => self::KEY,
+                'taxonomy' => 'p4-page-type',
+            ]
+        );
+    }
+
+    public static function get_option(): int
```
sagarsdeshmukh, line 49: It would be nice, if we add a function doc.

src/Migrations/M021MigrateDefaultPostType.php
mleray: Updated!

sagarsdeshmukh (approved):
👍

PR #2112, opened 2023-09-05
PLANET-7158 Editor Version of Media Archive

Osong-Michael:
**Ref: [PLANET-7158](https://jira.greenpeace.org/browse/PLANET-7158)**

New implementation for the Media Archive on the editor.

**Testing:**

- Locally, run `npm run build` when you have pulled in the changes for this branch.
- The functionality should still be the same, but slight changes in the UI.

You can also test on the assigned test instance.

_So far this implementation works for just the **Image and Happy Point blocks**. In the Editor, you can add a new image using the new Media Archive implementation or even replace the existing image with a new one. At the moment changes are not saved, so make sure to update or publish your changes to a post/page._

### DIFF
--- a/admin/css/archive-picker.css
+++ b/admin/css/archive-picker.css
@@ -19,8 +19,32 @@
   display: none;
 }
 
+.media-archive-info {
+  text-align: center;
+  margin-top: 100px;
+}
+
 .archive-picker {
   width: 100%;
+  display: flex;
+  height: 94vh;
+  gap: 10px;
+}
+
+.archive-picker-main {
+  position: relative;
+  flex: 100%;
+}
+
+.archive-picker-sidebar {
+  display: none;
+}
+
+.archive-picker-sidebar-open {
+  display: block;
+  height: 100%;
+  flex: 36%;
+  overflow-y: auto;
 }
 
 .archive-picker-search {
@@ -34,10 +58,6 @@
   display: grid;
   grid-template-columns: repeat(1, 1fr);
   gap: 0 16px;
-  position: relative;
-}
-
-.image-picker {
   position: relative;
 }
 
@@ -115,22 +135,20 @@
   background-repeat: no-repeat;
 }
 
-.image-picker li img {
+.archive-picker-main li img {
   margin: 0;
   border: 2px solid transparent;
   box-sizing: border-box;
   width: 100%;
   object-fit: cover;
   cursor: pointer;
-  height: 220px;
 }
 
-.image-picker li {
-  display: inline-block;
+.archive-picker-main li {
   position: relative;
 }
 
-.image-picker .added-to-library {
+.archive-picker-main .added-to-library {
   display: block;
   position: absolute;
   top: 2px;
@@ -141,7 +159,7 @@
   cursor: pointer;
 }
 
-.image-picker .added-to-library:before {
+.archive-picker-main .added-to-library:before {
   content: "";
   display: block;
   background: url("../../images/added-to-library.svg");
@@ -152,11 +170,11 @@
   left: 2px;
 }
 
-.image-picker .added-to-library span {
+.archive-picker-main .added-to-library span {
   display: none;
 }
 
-.image-picker .added-to-library:hover span {
+.archive-picker-main .added-to-library:hover span {
   display: block;
   margin-left: 25px;
   font-size: 10px;
@@ -164,15 +182,15 @@
   line-height: 25px;
 }
 
-.image-picker.open-sidebar li img {
+.archive-picker-main.is-open li img {
   opacity: 20%;
 }
 
-.image-picker.open-sidebar li img:hover {
+.archive-picker-main.is-open li img:hover {
   opacity: 100%;
 }
 
-.archive-picker.open-sidebar .help {
+.archive-picker.is-open .help {
   display: none;
 }
 
@@ -180,13 +198,13 @@
   opacity: 100% !important;
 }
 
-.image-picker.open-sidebar li img.picker-selected {
+.archive-picker-main.is-open li img.picker-selected {
   border-color: #1772a5;
   opacity: 100%;
 }
 
-.image-picker .help {
-  position: fixed;
+.archive-picker-main .help {
+  position: absolute;
   bottom: 8px;
   right: 8px;
   display: flex;
@@ -195,7 +213,7 @@
   z-index: 1;
 }
 
-.image-picker .help span {
+.archive-picker-main .help span {
   background: white;
   border-radius: 50%;
   text-align: center;
@@ -207,11 +225,11 @@
   box-shadow: 0 2px 4px rgba(0, 0, 0, 0.25);
 }
 
-.image-picker .help .tooltip {
+.archive-picker-main .help .tooltip {
   display: none;
 }
 
-.image-picker .help:hover .tooltip {
+.archive-picker-main .help:hover .tooltip {
   display: block;
   position: relative;
   background: white;
@@ -224,7 +242,7 @@
   color: black;
 }
 
-.image-picker .help:hover .tooltip:after {
+.archive-picker-main .help:hover .tooltip:after {
   content: "";
   position: absolute;
   bottom: -14px;
@@ -233,7 +251,7 @@
   border-top-color: white;
 }
 
-.image-picker .help:hover .tooltip a {
+.archive-picker-main .help:hover .tooltip a {
   color: black;
 }
 
@@ -250,22 +268,20 @@
 }
 
 .picker-sidebar {
-  position: fixed;
-  top: 0;
-  bottom: 0;
-  right: 0;
   background: white;
   padding: 64px 20px 0 20px;
   border-left: 2px solid black;
   overflow-y: auto;
   display: flex;
   flex-direction: column;
+  height: 100%;
 }
 
 .picker-sidebar img {
   object-fit: cover;
   width: 100%;
   margin-bottom: 10px;
+  height: 270px;
 }
 
 .picker-sidebar-header {
@@ -308,6 +324,9 @@
   min-height: 16px;
   width: 15px;
   background-repeat: no-repeat;
+  position: absolute;
+  top: 10px;
+  right: 20px;
 }
 
 .close-icon {
@@ -384,8 +403,8 @@
 
 .multiple-search-wrapper-input input:focus {
   outline-width: 0 !important;
-  border-bottom-color: transparent;
-  box-shadow: none;
+  border-bottom-color: transparent !important;
+  box-shadow: none !important;
 }
 
 .multiple-search-nav {
@@ -444,45 +463,33 @@
     padding-left: 20px;
   }
 
-  .archive-picker.open-sidebar {
-    width: calc(100% - var(--sidebar-width-large) + 8px);
-  }
-
-  .archive-picker.open-sidebar .help {
+  .archive-picker.is-open .help {
     display: flex;
     right: var(--sidebar-width-large);
   }
 
-  .archive-picker.open-sidebar .multiple-search-form {
+  .archive-picker.is-open .multiple-search-form {
     width: 100%;
     margin-top: 16px;
   }
 
-  .archive-picker.open-sidebar .nav-bulk-select {
+  .archive-picker.is-open .nav-bulk-select {
     order: 0;
-  }
-
-  .picker-sidebar {
-    width: var(--sidebar-width-tablet);
   }
 
   .picker-list {
     grid-template-columns: repeat(3, 1fr);
   }
 
-  .image-picker.open-sidebar {
-    width: 100%;
-  }
-
-  .image-picker.open-sidebar .picker-list {
+  .archive-picker-main.is-open .picker-list {
     grid-template-columns: repeat(1, 1fr);
   }
 
-  .image-picker li img {
+  .archive-picker-main li img {
     height: 160px;
   }
 
-  .image-picker .help:hover .tooltip {
+  .archive-picker-main .help:hover .tooltip {
     max-width: 330px;
   }
 
@@ -500,7 +507,7 @@
   }
 
   .multiple-search-wrapper-input {
-    min-width: auto;
+    flex: 100%;
     margin: 0;
   }
 
@@ -562,60 +569,52 @@
     margin-left: auto;
   }
 
-  .image-picker.open-sidebar .picker-list {
+  .archive-picker-main.is-open .picker-list {
     grid-template-columns: repeat(1, 1fr);
   }
 }
 
 @media (min-width: 1200px) {
-  .archive-picker.open-sidebar .multiple-search-form {
+  .archive-picker.is-open .multiple-search-form {
     width: 100%;
   }
 
-  .archive-picker.open-sidebar .multiple-search-form .multiple-search {
+  .archive-picker.is-open .multiple-search-form .multiple-search {
     max-width: 100%;
   }
 
-  .archive-picker.open-sidebar {
-    width: calc(100% - var(--sidebar-width-large) - 42px);
-  }
-
-  .archive-picker.open-sidebar .help {
+  .archive-picker.is-open .help {
     display: flex;
     right: calc(var(--sidebar-width-large) + 50px);
   }
 
-  .picker-sidebar {
-    width: var(--sidebar-width-large);
-  }
-
-  .image-picker .help span {
+  .archive-picker-main .help span {
     line-height: 32px;
     width: 32px;
     height: 32px;
     font-size: 18px;
   }
 
-  .image-picker .help:hover .tooltip {
+  .archive-picker-main .help:hover .tooltip {
     font-size: 14px;
   }
 
-  .image-picker .help:hover .tooltip:after {
+  .archive-picker-main .help:hover .tooltip:after {
     right: 8px;
   }
 
-  .image-picker.open-sidebar .picker-list {
+  .archive-picker-main.is-open .picker-list {
     grid-template-columns: repeat(2, 1fr);
   }
 }
 
 @media (min-width: 1440px) {
-  .archive-picker.open-sidebar .multiple-search-form {
+  .archive-picker-main.is-open .multiple-search-form {
     width: fit-content;
     margin-top: 0;
   }
 
-  .archive-picker.open-sidebar .nav-bulk-select {
+  .archive-picker.is-open .nav-bulk-select {
     order: 1;
   }
 
@@ -623,7 +622,7 @@
     grid-template-columns: repeat(5, 1fr);
   }
 
-  .image-picker.open-sidebar .picker-list {
+  .archive-picker-main.is-open .picker-list {
     grid-template-columns: repeat(4, 1fr);
   }
 
@@ -639,5 +638,12 @@
     display: flex;
     height: 100%;
     border-top: none;
+    flex: 40%;
+  }
+}
+
+@media (min-width: 1600px) {
+  .picker-list {
+    grid-template-columns: repeat(6, 1fr);
   }
 }
--- a/admin/css/archive-picker.css
+++ b/admin/css/archive-picker.css
@@ -43,8 +43,13 @@
 .archive-picker-sidebar-open {
   display: block;
   height: 100%;
-  flex: 36%;
   overflow-y: auto;
+  flex: 70%;
+}
+
+.archive-picker-sidebar-open::-webkit-scrollbar,
+.picker-sidebar-fields::-webkit-scrollbar {
+  display: none;
 }
 
 .archive-picker-search {
@@ -59,6 +64,7 @@
   grid-template-columns: repeat(1, 1fr);
   gap: 0 16px;
   position: relative;
+  margin-top: 54px;
 }
 
 .picker-list li {
@@ -89,7 +95,7 @@
   flex-wrap: wrap;
   justify-content: space-between;
   align-items: center;
-  margin-top: 32px;
+  margin-top: 50px;
   margin-bottom: 16px;
   width: 100%;
 }
@@ -305,7 +311,6 @@
   width: fit-content;
   align-self: center;
   margin-bottom: 24px;
-  margin-top: 16px;
 }
 
 .archive-picker-loading {
@@ -327,6 +332,7 @@
   position: absolute;
   top: 10px;
   right: 20px;
+  text-indent: -9999px; /* hide button text on admin page  */
 }
 
 .close-icon {
@@ -450,9 +456,17 @@
   line-height: 150%;
 }
 
-@media (min-width: 576px) {
+@media (max-width: 767px) {
   .picker-list {
     grid-template-columns: repeat(2, 1fr);
+  }
+
+  .archive-picker-main.is-open .picker-list {
+    grid-template-columns: repeat(1, 1fr);
+  }
+
+  .archive-picker-sidebar-open {
+    flex: 100%;
   }
 }
 
@@ -479,10 +493,6 @@
 
   .picker-list {
     grid-template-columns: repeat(3, 1fr);
-  }
-
-  .archive-picker-main.is-open .picker-list {
-    grid-template-columns: repeat(1, 1fr);
   }
 
   .archive-picker-main li img {
@@ -570,7 +580,7 @@
   }
 
   .archive-picker-main.is-open .picker-list {
-    grid-template-columns: repeat(1, 1fr);
+    grid-template-columns: repeat(2, 1fr);
   }
 }
 
@@ -606,7 +616,23 @@
   .archive-picker-main.is-open .picker-list {
     grid-template-columns: repeat(2, 1fr);
   }
+
+  .archive-picker-sidebar-open {
+    flex: 50%;
+  }
 }
+
+@media (max-width: 1230px) {
+  .archive-picker-main.is-open .archive-picker-title {
+    width: 100%;
+  }
+}
+
+/* @media (max-width: 1350px) {
+  .archive-picker-sidebar-open {
+    flex: 70%;
+  }
+} */
 
 @media (min-width: 1440px) {
   .archive-picker-main.is-open .multiple-search-form {
@@ -640,10 +666,21 @@
     border-top: none;
     flex: 40%;
   }
+
+  .archive-picker-sidebar-open {
+    flex: 40%;
+  }
 }
 
 @media (min-width: 1600px) {
   .picker-list {
     grid-template-columns: repeat(6, 1fr);
+  }
+}
+
+
+@media (max-width: 600px) {
+  .picker-sidebar-header .close-sidebar {
+    top: 50px;
   }
 }
--- a/assets/src/js/Components/ArchivePicker/ArchivePicker.js
+++ b/assets/src/js/Components/ArchivePicker/ArchivePicker.js
@@ -15,6 +15,9 @@ const timeout = delay => {
 };
 const acceptedBlockTypes = ['core/image', 'planet4-blocks/happypoint'];
 
+export const EDITOR_VIEW = 'editor';
+export const ADMIN_VIEW = 'admin';
+
 export const ACTIONS = {
   CLOSE_SIDEBAR: 'closeSidebar',
   BULK_SELECT_ENABLE: 'bulkSelectEnable',
@@ -222,7 +225,7 @@ const reducer = (state, action) => {
   }
 };
 
-export default function ArchivePicker({mediaView, adminView}) {
+export default function ArchivePicker({view = ADMIN_VIEW}) {
   const [{
     loading,
     loaded,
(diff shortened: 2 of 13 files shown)


planet-4:

:watch: 2023.09.05 20:34:46

assets/src/js/Components/ArchivePicker/ArchivePickerList.js
```diff
@@ -131,9 +132,17 @@ export default function ArchivePickerList() {
                 role="presentation"
               />
               {wordpress_id && (
-                <div className="added-to-library">
-                  <span>{__('Added to Media Library', 'planet4-master-theme-backend')}</span>
-                </div>
+                <>
+                  {(currentBlockImageId === wordpress_id) ? (
```
GP-Dan-Tovbein, line 136: What do you think if you just create only one `added-to-media-library` and the implement the logic only to the text?

assets/src/js/Components/ArchivePicker/ArchivePickerList.js
```diff
@@ -131,9 +132,17 @@ export default function ArchivePickerList() {
...
+                    <div className="added-to-library">
```
GP-Dan-Tovbein, line 137:
```
<div className="added-to-library">
                      <span>
  {(currentBlockImageId === wordpress_id) ? (
{__('Added to Media Library & Post', 'planet4-master-theme-backend')}
) : (
{__('Added to Media Library', 'planet4-master-theme-backend')}
)
</span>
                    </div>
```

admin/css/picker.css
```diff
@@ -21,6 +21,25 @@
...
 .archive-picker {
   width: 100%;
+  display: flex;
+  height: 94vh;
+  gap: 10px;
+}
+
+.archive-picker-main {
+  position: relative;
+  flex: 100%;
```
GP-Dan-Tovbein, line 31: I assume you want to implement `flex-basis` here, right? If yes, I'd prefer to directly do `flex-basis: 30%` instead of `flex: 30%`.

admin/css/picker.css
```diff
@@ -21,6 +21,25 @@
...
+}
+
+.archive-picker-sidebar {
+  display: none;
+}
+
+.archive-picker-sidebar-open {
+  display: block;
+  height: 100%;
+  flex: 36%;
```
GP-Dan-Tovbein, line 41: Same here

admin/css/picker.css
Osong-Michael: The `flex` is just shorthand that incorporates `flex-shrink`, `flex-grow` and `flex-basis`, so it makes the code cleaner in my opinion, what do you think?

GP-Dan-Tovbein:
@Osong-Michael Have you tried to convert the picker `CSS` stylesheet to a `SASS`?

Osong-Michael:
> @Osong-Michael Have you tried to convert the picker `CSS` stylesheet to a `SASS`?

Not yet, when I tried earlier I had more issues than I could deal with, so I wanted to finish the feature first the convert the styles later.

GP-Dan-Tovbein:
> > @Osong-Michael Have you tried to convert the picker `CSS` stylesheet to a `SASS`?
> 
> Not yet, when I tried earlier I had more issues than I could deal with, so I wanted to finish the feature first then convert the styles later.

Sounds logic! And what if we rename the `picker.css` to `archive-picker.css` then? I see it more consistent with the component name. What do you think?

GP-Dan-Tovbein:
Is the [titan](https://www-dev.greenpeace.org/test-titan) the assigned instance or should I test within another one? Since as far as I saw, it's been assigned but no label is being added.

Osong-Michael:
> Is the [titan](https://www-dev.greenpeace.org/test-titan) the assigned instance or should I test within another one? Since as far as I saw, it's been assigned but no label is being added.

Titan is the assigned instance, you can test on it.

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -308,51 +392,104 @@ export default function ArchivePicker() {
...
+              </div>
+
+              <div className={classNames('archive-picker-sidebar', {'archive-picker-sidebar-open': selectedImages.length > 0 && !bulkSelect})}>
+                <div className="picker-sidebar">
+                  {selectedImages.length === 1 ? <SingleSidebar image={selectedImages[0]} mediaView={mediaView} /> : <MultiSidebar />}
+                </div>
+              </div>
+            </section>
+          ) : (
+            <h3 className="media-archive-info">Use Media Library Tab for this Block!</h3>
```
sagarsdeshmukh, line 489: The `Use Media Library Tab for this Block!` string can be translatable.

assets/src/scss/editorOverrides.scss
```diff
@@ -280,3 +280,104 @@ input.describe[type=text][data-setting=caption] {
 .customize-control[id*="new_identity_styles"] label {
   display: inline;
 }
+
+// Media Acrhive Style Override
```
sagarsdeshmukh, line 284: Looks like a typo 🙂

assets/src/js/media_archive_editor_view.js
```diff
@@ -0,0 +1,68 @@
...
+
+      var State = wp.media.controller.State.extend({
+        insert: function() {
+          this.frame.close();
+        }
+      });
+
+      this.states.add([
+        new State({
+          id: 'mediaArchive1',
```
sagarsdeshmukh, line 20: The `id` name `mediaArchive1` is intentional or we can use just `mediaArchive`? 🤔

assets/src/js/media_archive_editor_view.js
```diff
@@ -0,0 +1,68 @@
...
+import ArchivePicker from '../js/Components/ArchivePicker/ArchivePicker';
+
+var frame = wp.media.view.MediaFrame.Select;
+
+const mediaArchiveTab = () => {
+  wp.media.view.MediaFrame.Select = frame.extend({
+    initialize: function() {
+      frame.prototype.initialize.apply(this, arguments);
+
+      var State = wp.media.controller.State.extend({
```
GP-Dan-Tovbein, line 12: Can we use `const` instead?

assets/src/js/Components/ArchivePicker/blockUpdateFunctions.js
```diff
@@ -0,0 +1,44 @@
...
+    originalContent: div.innerHTML,
+  };
+
+  return updatedAttributes;
+};
+
+export const updateHappyPointAttributes = id => {
+  // Need to update only the id property for the Happy Point block
+  const updatedAttributes = {attributes: {id}};
+  return updatedAttributes;
```
GP-Dan-Tovbein, line 43: I think you can just do this instead
```
return {attributes: {id}}
```

assets/src/js/Components/ArchivePicker/blockUpdateFunctions.js
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  const updatedAttributes = {
+    attributes: {
+      id: image.id,
+      url: image.source_url,
+      caption: image.caption.raw,
+      alt: image.alt_text,
+    },
+
```
GP-Dan-Tovbein, line 33: Could we remove this empty line?

assets/src/js/media_archive_editor_view.js
```diff
@@ -0,0 +1,68 @@
...
+
+    browseRouter: function(routerView) {
+      routerView.set({
+        upload: {
+          text: wp.media.view.l10n.uploadFilesTitle,
+          priority: 20
+        },
+        mediaArchive1: {
+          text: 'Media Archive',
+          priority: 30
```
GP-Dan-Tovbein, line 37: missing `,`

assets/src/js/media_archive_editor_view.js
```diff
@@ -0,0 +1,68 @@
...
+      ]);
+
+      this.on('content:render:mediaArchive1', this.renderMediaArchive, this);
+    },
+
+    browseRouter: function(routerView) {
+      routerView.set({
+        upload: {
+          text: wp.media.view.l10n.uploadFilesTitle,
+          priority: 20
```
GP-Dan-Tovbein, line 33: missing ,

assets/src/js/media_archive_editor_view.js
```diff
@@ -0,0 +1,68 @@
...
+        },
+        mediaArchive1: {
+          text: 'Media Archive',
+          priority: 30
+        },
+        browse: {
+          text: wp.media.view.l10n.mediaLibraryTitle,
+          priority: 40
```
GP-Dan-Tovbein, line 41: Same

assets/src/js/media_archive_editor_view.js
```diff
@@ -0,0 +1,68 @@
...
+      });
+    },
+
+    renderMediaArchive: function() {
+      var MediaArchiveContent = wp.Backbone.View.extend({
+        tagName: 'div',
+        className: 'media-archive-content',
+        template: _.template("<div id='user-info'></div>"),
+        render() {
+          var domeNode = this.$("#user-info").prevObject.get(0);
```
GP-Dan-Tovbein, line 52: Is intentional `domeNode` or did you want to use `domNode`?

assets/src/scss/editorOverrides.scss
```diff
@@ -280,3 +280,104 @@ input.describe[type=text][data-setting=caption] {
...
+.media-archive-content {
```
GP-Dan-Tovbein, line 285: Can we use `p4 variables` here?

assets/src/js/media_archive_editor_view.js
```diff
@@ -0,0 +1,68 @@
+/* eslint-disable */
```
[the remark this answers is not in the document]
Osong-Michael: Eslint throws errors on certain implementations that when I changed made the code not to run.

src/AdminAssets.php
```diff
@@ -14,6 +16,10 @@ class AdminAssets
      */
     public static function enqueue_js(): void
     {
+        if (MediaArchive::is_active()) {
```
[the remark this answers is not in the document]
Osong-Michael: I need help understanding how the JS is loaded here because if I add this check, the Media Archive does not load for the admin panel, this code `const currentBlock = wp.data.select('core/block-editor').getSelectedBlock();` causes the component to throw an error, if I wrap that code in a `useEffect` to display only in the editor, strangely enough it returns undefined, any thoughts or ideas on why this may be?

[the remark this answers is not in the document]
comzeradd: We can do that in the next iteration when we remove the feature flag.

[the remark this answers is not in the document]
sagarsdeshmukh: I had a DM with Osong on this comment, we can drop this suggestion, As we are using the same component for both the admin dashboard and the editor.

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -308,51 +392,104 @@ export default function ArchivePicker() {
...
+              </div>
+
+              <div className={classNames('archive-picker-sidebar', {'archive-picker-sidebar-open': selectedImages.length > 0 && !bulkSelect})}>
+                <div className="picker-sidebar">
+                  {selectedImages.length === 1 ? <SingleSidebar image={selectedImages[0]} mediaView={mediaView} /> : <MultiSidebar />}
+                </div>
+              </div>
+            </section>
+          ) : (
+            <h3 className="media-archive-info">{__('Use Media Library Tab for this Block!', 'planet4-master-theme-backend')}</h3>
```
comzeradd, line 489: Tiny suggestion: using a paragraph tag here instead of a heading looks fine too

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -185,7 +215,7 @@ const reducer = (state, action) => {
   }
 };
 
-export default function ArchivePicker() {
+export default function ArchivePicker({mediaView, adminView}) {
```
GP-Dan-Tovbein, line 218: What do you think if we move these two props into the context?

GP-Dan-Tovbein: In addition to that, instead of using 2 different props, we can just use one like `view` which accepts `media|admin`

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -233,7 +265,50 @@ export default function ArchivePicker() {
     }
   }, [loading, loaded, images, error, searchText, pageNumber, dispatch]);
 
-  const includeInWp = async (ids = []) => {
+  const currentBlock = wp.data.select('core/block-editor').getSelectedBlock();
+
+  const getImageDets = useCallback(async id => {
```
GP-Dan-Tovbein, line 270: I'd prefer to rename to `getImageDetails` instead which is more understandable !

assets/src/js/media_archive_editor_view.js
```diff
@@ -0,0 +1,65 @@
...
+
+      this.states.add([
+        new State({
+          id: 'mediaArchive',
+          search: false,
+          title: 'Media Archive',
+        }),
+      ]);
+
+      this.on('content:render:mediaArchive1', this.renderMediaArchive, this);
```
GP-Dan-Tovbein, line 25: Is that intentional the `mediaArchive1` ?

Osong-Michael: No that's an error, should say `mediaArchive` instead, fixing now.

assets/src/js/Components/ArchivePicker/blockUpdateFunctions.js
```diff
@@ -0,0 +1,42 @@
...
+    } else {
+      // If there is no content then create block content.
+      const blockContent = `
+        <figure class="wp-block-image size-large">
+          <img src=${image.source_url} alt=${image.alt_text} class="wp-image-${image.id}"/>
+          <figcaption class="wp-element-caption">
+            ${image.caption.raw}
+          </figcaption>
+        </figure>`;
+      div.innerHTML = blockContent;
```
GP-Dan-Tovbein, line 22: What if you define it directly instead of create a `const`? 

```javascript
 div.innerHTML = `
        <figure class="wp-block-image size-large">
          <img src=${image.source_url} alt=${image.alt_text} class="wp-image-${image.id}"/>
          <figcaption class="wp-element-caption">
            ${image.caption.raw}
          </figcaption>
        </figure>`;
```

assets/src/js/media_archive_editor_view.js
```diff
@@ -0,0 +1,65 @@
...
+      this.on('content:render:mediaArchive', this.renderMediaArchive, this);
+    },
+
+    browseRouter(routerView) {
+      routerView.set({
+        upload: {
+          text: wp.media.view.l10n.uploadFilesTitle,
+          priority: 20,
+        },
+        mediaArchive1: {
```
GP-Dan-Tovbein, line 34: is it intentional `mediaArchive1`?

assets/src/scss/editorOverrides.scss
```diff
@@ -280,3 +281,104 @@ input.describe[type=text][data-setting=caption] {
...
+    gap: 0;
+  }
+
+  #import-to-library,
+  #import-to-library:disabled {
+    display: none;
+  }
+
+  .multiple-search {
+    margin-left: 20px;
```
GP-Dan-Tovbein, line 303: I have a small doubt with this.. don't we need to implement `RTL` here?

assets/src/js/media_archive_editor_view.js
Osong-Michael: An error, my plan originally was to create two different pop ups, but does not seem like it is needed at the moment.

assets/src/scss/editorOverrides.scss
Osong-Michael: What is RTL @GP-Dan-Tovbein ?

comzeradd: I assume Right-to-Left. So maybe `margin-inline-start` could work better here.

assets/src/js/media_archive_editor_view.js
```diff
@@ -0,0 +1,65 @@
...
+      });
+    },
+
+    renderMediaArchive() {
+      const MediaArchiveContent = wp.Backbone.View.extend({
+        tagName: 'div',
+        className: 'media-archive-content',
+        template: _.template('<div id=\'user-info\'></div>'), // eslint-disable-line no-undef
+        render() {
+          const domNode = this.$('#user-info').prevObject.get(0);
```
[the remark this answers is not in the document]
Osong-Michael: I am not totally sure but I think it is a `Backbone JS `thing where using the normal `document` object returns nothing on my end, maybe you have more experience with Backbone but all the resources I found used `jQuery`.

assets/src/scss/editorOverrides.scss
GP-Dan-Tovbein: Yep! You can test it out by modifying the `direction` (`document.body.style.direction = "rtl"` or `ltr`)  through the browser inspector.

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -5,10 +5,15 @@ import ArchivePickerList from './ArchivePickerList';
 import SingleSidebar from './SingleSidebar';
 import MultiSidebar from './MultiSidebar';
 import ArchivePickerToolbar from './ArchivePickerToolbar';
+import {updateImageBlockAttributes, updateHappyPointAttributes} from './blockUpdateFunctions';
 
 const {apiFetch, url, i18n} = wp;
 const {addQueryArgs} = url;
 const {__} = i18n;
+const timeout = delay => {
```
GP-Dan-Tovbein, line 13: @Osong-Michael I'm not sure if I understand why we need to implement this `timeout` 🙏

Osong-Michael: So for one, when requesting image details back from `WP` when the images are first added to the Media Library, `WP` will sometime return undefined even though the image is in the Media Library, some resources I found online tell you to run the function in `async` mode with a `timeout` so the details can be fetched, the best way I found to execute this is run the fetch the first time, if we get undefined, wait a few seconds then run it again, I also made the timeout reusable so i can use it in a few different places too.

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -26,6 +31,9 @@ export const ACTIONS = {
   REMOVE_ERROR: 'removeError',
   SEARCH: 'search',
   SET_ERROR: 'setError',
+  ADDING_IMAGE_TO_POST: 'addingImageToPost',
```
GP-Dan-Tovbein, line 34: Is it possible if we change to `ADD_IMAGE_TO_POST` instead?

assets/src/js/Components/ArchivePicker/ArchivePicker.js
GP-Dan-Tovbein: Oh I see. It's like a `sleep()`, am I correct?

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -233,7 +265,50 @@ export default function ArchivePicker() {
...
+        const updatedAttributes = updateHappyPointAttributes(id);
+        await wp.data.dispatch('core/block-editor').updateBlock(currentBlock.clientId, updatedAttributes);
+      }
+
+      dispatch({type: ACTIONS.ADDED_IMAGE_TO_POST});
+      await timeout(1000);
+      dispatch({type: ACTIONS.CLOSE_SIDEBAR});
+      document.querySelector('.media-modal-close').click();
+    } catch (err) {
+      throw err;
```
GP-Dan-Tovbein, line 307: Shouldn't we dispatch an error if it fails? Consider if it's fine to re-use `SET_ERROR` OR `PROCESS_IMAGES_ERROR` OR create a new one like `ADD_IMAGE_TO_POST_ERROR`

GP-Dan-Tovbein:
@Osong-Michael I've noticed this this 2 different issues

1. The Spinner positioning issue 
Screenshot 2023-09-21 at 08 56 20

2. When you click on any image, then is impossible to close the sidebar since the X button close the modal
Screenshot 2023-09-21 at 08 55 52

Osong-Michael:
> @Osong-Michael I've noticed this this 2 different issues
> 
> 1. The Spinner positioning issue
>    Screenshot 2023-09-21 at 08 56 20
> 2. When you click on any image, then is impossible to close the sidebar since the X button close the modal
>    Screenshot 2023-09-21 at 08 55 52

@GP-Dan-Tovbein , for the sidebar, that is based on the Figma design as seen [here](https://www.figma.com/file/cnd1hDu5MnXD2kmGxfwH4N/Media-Archive-Wireframes-(Final-Beta)?type=design&node-id=0-1&mode=design&t=4aRpp5ELfez5kkU5-0).

Let me look into the positioning issue for the Spinner.

assets/src/js/media_archive_editor_view.js
```diff
@@ -0,0 +1,63 @@
...
+      ]);
+
+      this.on('content:render:mediaArchive', this.renderMediaArchive, this);
+    },
+
+    browseRouter(routerView) {
+      routerView.set({
+        upload: {
+          text: wp.media.view.l10n.uploadFilesTitle,
+          priority: 20,
```
GP-Dan-Tovbein, line 32: I think just defining 1,2 or 3 it's fine

Osong-Michael: Yes it would be if i was sure of the behaviour when the view is rendered. I couldn't find anything specific but most of the resources and code snippets suggested this, stating possible overriding default behaviour issues or rendering and re-rendering issues and this being like a safe value.

assets/src/js/media_archive_editor_view.js
```diff
@@ -0,0 +1,63 @@
...
+        insert() {
+          this.frame.close();
+        },
+      });
+
+      this.states.add([
+        new State({
+          id: 'mediaArchive',
+          search: false,
+          title: 'Media Archive',
```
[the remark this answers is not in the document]
Osong-Michael: Not for this.

assets/src/js/media_archive_editor_view.js
```diff
@@ -0,0 +1,63 @@
...
+    },
+
+    browseRouter(routerView) {
+      routerView.set({
+        upload: {
+          text: wp.media.view.l10n.uploadFilesTitle,
+          priority: 20,
+        },
+        mediaArchive: {
+          text: 'Media Archive',
```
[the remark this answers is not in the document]
Osong-Michael: Yes, I will make this translatable

admin/css/picker.css
GP-Dan-Tovbein: I agree that "yes" in some cases but some others "no". 
Just to clarify my point, if you want to apply all that properties then it's OK to use a shorthand and then use `flex`, however, if your decision is to apply ONLY `flex-basis` or `flex-grow` then I'd prefer to define separately.

assets/src/js/media_archive_editor_view.js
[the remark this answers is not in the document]
GP-Dan-Tovbein: If you you think that it will be feasible to use vanilla Javascript to find a selector, then it would be great. If not, it would be fine too.

admin/css/archive-picker.css
```diff
@@ -115,22 +135,20 @@
...
   box-sizing: border-box;
   width: 100%;
   object-fit: cover;
   cursor: pointer;
-  height: 220px;
 }
 
-.image-picker li {
-  display: inline-block;
+.archive-picker-main li {
```
GP-Dan-Tovbein, line 147: Looking at [this](https://github.com/greenpeace/planet4-master-theme/pull/2112/files#diff-61826b7e198488e0d8bd10ca420f37ebf03d0241dff3417b92560675f2142451R151), I think that only the `.archive-picker-main` should be positioned to `relative`. Not necessary the `li`.

assets/src/js/Components/ArchivePicker/ArchivePicker.js
GP-Dan-Tovbein: @Osong-Michael, have you checked this?
```javascript
export const MEDIA = 'media';
export const ADMIN = 'admin

export default function ArchivePicker({view}) {
...
}
...
{(image && view === ADMIN) && (
```
Less props we pass, easier is to scale it.

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -177,6 +189,31 @@ const reducer = (state, action) => {
...
+    };
+  }
+  case ACTIONS.ADDED_IMAGE_TO_POST: {
+    return {
+      ...state,
+      processing: false,
+      imageAdded: true,
+    };
+  }
+  case ACTIONS.ADD_IMAGE_TO_POST_ERROR: {
```
GP-Dan-Tovbein, line 205: Just to be curious, is for any reason you create a new specific action error for this instead of using the [SET_ERROR](https://github.com/greenpeace/planet4-master-theme/pull/2112/files#diff-2d5f38a9e36d2b751a99282786e7adfec59d239898030fce76b89ddbbeb6a426R169)?

assets/src/js/Components/ArchivePicker/ArchivePicker.js
Osong-Michael: Oh I see what you are saying, in that case we don't need the extra variables then, we could just use the `view` prop and check if it is _'admin'_ or _'editor'_ and call the component like `< ArchivePicker view="admin" />`.  What do you think @GP-Dan-Tovbein ?

GP-Dan-Tovbein: One more thing.. you can store the `view` prop to the `reducer` and then is shared to all nested components (which uses the `Context`). And finally, you'll avoid to pass the value like you do [here](https://github.com/greenpeace/planet4-master-theme/pull/2112/files#diff-2d5f38a9e36d2b751a99282786e7adfec59d239898030fce76b89ddbbeb6a426R318), but pass as a dependency.

For instance,
Instead of 
```javascript
const includeInWp = async (ids = [], isMediaView = false) => { ... }
```
Do this
```javascript
const includeInWp = useCallback(async (ids = []) => {}, [view]);
```

assets/src/js/Components/ArchivePicker/ArchivePicker.js
Osong-Michael: Yes, your _SET_ERROR_ case does a lot of things which I am not fully aware of, plus I needed something simple I could track, and since the `error` state was available and not used in the `SideBar` component, I created a simple case to use that instead so I can also use it to display errors for the Media Archive.

admin/css/archive-picker.css
Osong-Michael: I think the `li` is `position: relative` so the _added-to-library_  `div` can be `position: absolute` within the `li`

Osong-Michael: I split the whole component into 2

- _archive-picker-main_, houses the search bar and other form elements, the image list, help tooltip and the Spinner component.
- _archive-picker-sidebar_ houses the Sidebar component only.

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -233,7 +275,50 @@ export default function ArchivePicker() {
...
+      dispatch({type: ACTIONS.ADDED_IMAGE_TO_POST});
+      await timeout(1000);
+      dispatch({type: ACTIONS.CLOSE_SIDEBAR});
+      document.querySelector('.media-modal-close').click();
+    } catch (err) {
+      dispatch({type: ACTIONS.ADD_IMAGE_TO_POST_ERROR, payload: err});
+    }
+  };
+
+  const includeInWp = async (ids = [], viewProp) => {
```
Osong-Michael, line 321: Hey @GP-Dan-Tovbein , I had to change this back because the `useCallback` you added was not firing when the button was clicked because the values in its dependency array didn't change, so this now works well.

assets/src/js/Components/ArchivePicker/ArchivePickerList.js
```diff
@@ -131,9 +132,15 @@ export default function ArchivePickerList() {
                 role="presentation"
               />
               {wordpress_id && (
-                <div className="added-to-library">
-                  <span>{__('Added to Media Library', 'planet4-master-theme-backend')}</span>
-                </div>
+                <>
```
GP-Dan-Tovbein, line 135: Don't need `<>`  here

GP-Dan-Tovbein (approved):
Code looks good to me! EXCELLENT job! 👏 

I left only a comment for a small change 🙏

assets/src/js/Components/ArchivePicker/SingleSidebar.js
```diff
@@ -54,19 +60,28 @@ export default function SingleSidebar({image}) {
...
+          )}
         </div>
         <button
           className="close-sidebar"
           onClick={() => {
             dispatch({type: ACTIONS.CLOSE_SIDEBAR});
           }}
-        />
+        >
+          Close
```
mleray, line 83: Don't we need to add a translation here?

PR #2128, opened 2023-10-12
PLANET-7227 Enable new Media Archive implementation

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-7227

---

### Testing
- This should enable the new Media Archive implementation by default under _Media > Archive_.
- No relevant option under _Planet 4 > Features_ anymore.


comzeradd:
> Do we need to do something about the GPI Media Library plugin?

Yes, we should remove the plugin at the same time. I opened https://github.com/greenpeace/planet4-base/pull/264 for that.

PR #2131, opened 2023-10-16
PLANET-7250 Add Actions list to posts report

mleray:
### Description

See [PLANET-7250](https://jira.greenpeace.org/browse/PLANET-7250)

**Note:** SonarCloud is complaining but in this case the duplicated code makes sense 😬 we could try refactoring the code to avoid this issue, but it should be done in a separate PR

### Testing

Make sure to test it with the new IA enabled but also when it's disabled, to make sure it doesn't cause any issues then!


posts-report.php
```diff
@@ -8,6 +8,9 @@
 
 use Timber\Timber;
 
+$context = Timber::get_context();
+$context['new_ia'] = !empty(planet4_get_option('new_ia'));
```
lithrel, line 12: Doesn't seem to work for me, as the Timber::render on l.14 is not called with $context, and new_ia is null in the template. This works on my local:
```php
$context['spinner_gif'] = get_template_directory_uri() . '/images/wpspin_light-2x.gif';
Timber::render([ 'posts_report.twig' ], $context);
```

mleray: That's weird, it used to work for me on local but indeed not anymore 😅 I see that it doesn't work on the test instance either... Code updated 👍

lithrel (approved):
:+1:

PR #2132, opened 2023-10-16
Fix GF client side script path

lithrel:
path assets/src is not available in production

Screenshot from 2023-10-16 11-58-16

## Fix
- adding script to build system 
- changing path to assets/build


mleray (approved):
Looks good to me 👍

PR #2137, opened 2023-10-20
PLANET-7214 Build a Posts List block using Query Loop

Osong-Michael:
**Ref: [PLANET-7214](https://jira.greenpeace.org/browse/PLANET-7214)**

**Testing:**


src/Loader.php
```diff
@@ -157,6 +160,32 @@ public static function add_blocks(): void
         new Blocks\GuestBook();//NOSONAR
     }
 
+      /**
+     * Load editorIndex file for beta blocks.
+     */
+    public static function add_beta_blocks(): void
+    {
+        if (!BetaBlocks::is_active() && !AllowAllBlocks::is_active()) {
```
comzeradd, line 168: I think it makes more sense to check for Planet4Blocks flag instead of AllBlocks (in addition to Beta)

Osong-Michael: We need the `All Blocks` Feature enabled so we have access to the `Query Loop Block` which the `Posts List block` extends, if not, it does not show up in the editor.

Now that I think about it, is it redundant to have the check for `AllBlocks` here, because it still won't show up if that is not enabled but `Beta Blocks` is 🤔 ?

assets/src/scss/blocks/PostsListBlock.scss
```diff
@@ -0,0 +1,143 @@
...
+  .wp-block-group.is-layout-flow {
+    flex-basis: 70%;
+  }
+
+  .wp-block-post-title {
+    font-weight: 700;
+    color: var(--grey-900);
+    font-family: var(--font-family-tertiary);
+    font-size: 1.375rem;
+    margin-top: 5px;
```
GP-Dan-Tovbein, line 35: Can we use spacing variables? i.e. `$sp-1x` which is equal to `6px` instead of `5px`. The we have all the spacing values already standardized.

assets/src/scss/blocks/PostsListBlock.scss
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  .wp-block-post-excerpt {
+    margin-top: -15px;
```
GP-Dan-Tovbein, line 39: Same here!

assets/src/scss/blocks/PostsListBlock.scss
```diff
@@ -0,0 +1,143 @@
+.posts-list-query {
+  .wp-block-post-template.is-layout-flow li {
+    width: 100% !important;
+  }
+
+  .wp-block-columns.is-layout-flex {
+    gap: 20px;
```
GP-Dan-Tovbein, line 7: Same here!

GP-Dan-Tovbein (commented):
General commens! 
1. Can we use spacing variables?
2. Can we check the places where we use `font-size|weight` or `line-height` could be replaced by tokens?

assets/src/blocks/PostsList/PostsListBlock.js
```diff
@@ -0,0 +1,68 @@
...
+    },
+    scope: ['inserter'],
+    innerBlocks: [
+      ['core/heading', {content: 'Related Posts'}],
+      ['core/paragraph', {
+        placeholder: __('Enter description', 'planet4-blocks-backend'),
+        style: {
+          spacing: {
+            margin: {
+              top: '25px',
```
GP-Dan-Tovbein, line 42: Why don't we use spacing values?

assets/src/scss/blocks/PostsListBlock.scss
```diff
@@ -0,0 +1,143 @@
...
+
+    img {
+      @include medium-and-up {
+        height: 180px;
+      }
+
+      @include large-and-up {
+        height: 210px;
+      }
+      width: 100% !important;
```
GP-Dan-Tovbein, line 21: Could we move the next properties to line 14?

assets/src/scss/blocks/PostsListBlock.scss
```diff
@@ -0,0 +1,143 @@
...
+      object-fit: cover;
+    }
+  }
+
+  .wp-block-group.is-layout-flow {
+    flex-basis: 70%;
+  }
+
+  .wp-block-post-title {
+    font-weight: $font-weight-bold;
```
GP-Dan-Tovbein, line 31: You can use tokens instead!

assets/src/scss/blocks/PostsListBlock.scss
```diff
@@ -0,0 +1,143 @@
...
+    color: var(--grey-900);
+    font-family: var(--font-family-tertiary);
+    font-size: $font-size-l--font-family-primary;
```
GP-Dan-Tovbein, line 34: I think it's better to use `CSS variables` instead!

assets/src/scss/blocks/PostsListBlock.scss
```diff
@@ -0,0 +1,143 @@
...
+  .is-layout-flow > .is-layout-flow {
+    display: flex;
+    gap: $sp-x;
+    font-size: calc($font-size-xxs--font-family-primary - 2px);
+    color: var(--grey-600);
+    font-family: var(--font-family-paragraph-secondary);
+  }
+
+  .wp-block-post-author-name {
+    font-size: calc($font-size-xxs--font-family-primary - 2px);
```
GP-Dan-Tovbein, line 75: This looks hard to follow up!

assets/src/scss/blocks/PostsListBlock.scss
```diff
@@ -0,0 +1,143 @@
...
+    &::after {
+      content: ".";
+      font-size: calc($font-size-m--font-family-primary + 5px);
+      margin-top: calc(2px - $sp-1);
+      margin-left: $sp-x;
+    }
+  }
+
+  .wp-block-post-terms {
+    margin-top: calc(3px - $sp-1);
```
GP-Dan-Tovbein, line 94: Same as [here](https://github.com/greenpeace/planet4-master-theme/pull/2137/files#r1379129875)

assets/src/scss/editorOverrides.scss
```diff
@@ -415,3 +418,48 @@ input.describe[type=text][data-setting=caption] {
     }
   }
 }
+
+// Posts List Block Override
+.posts-list-query {
+  .wp-block-post-template.is-layout-flow li {
+    height: 240px;
+    margin-bottom: calc($sp-1 + 2px);
```
GP-Dan-Tovbein, line 426: Same as [here](https://github.com/greenpeace/planet4-master-theme/pull/2137/files#r1379129875)

src/Loader.php
```diff
@@ -157,6 +159,42 @@ public static function add_blocks(): void
         new Blocks\GuestBook();//NOSONAR
     }
 
+    /**
+     * Load editorIndex file and add core Query Loop Block for Beta Blocks.
+     */
+    public static function add_beta_blocks(): void
+    {
+        if (!BetaBlocks::is_active()) {
```
GP-Dan-Tovbein, line 167: I think we can move this statement to [here](https://github.com/greenpeace/planet4-master-theme/pull/2137/files#diff-1314d7908e6167b9fbe3d52c6f0c3cadf76637c75eca9c9c36d66106c579edd3R62). And we'll avoid to invoke unnecessary this function.

assets/src/blocks/PostsList/PostsListBlock.js
```diff
@@ -0,0 +1,68 @@
+export const POSTS_LIST_BLOCK = 'planet4-blocks/posts-list-block';
+
+export const registerPostsListBlock = () => {
+  const {registerBlockVariation} = wp.blocks;
+  const {__} = wp.i18n;
+
+  return registerBlockVariation('core/query', {
+    name: POSTS_LIST_BLOCK,
+    title: 'P4 Posts List Block',
```
GP-Dan-Tovbein, line 9: Don't we need to translate this text?

assets/src/scss/blocks/PostsListBlock.scss
Osong-Michael: That is a token from [here](https://github.com/greenpeace/planet4-master-theme/blob/main/assets/src/scss/base/_tokens.scss#L114), I can use the css variable instead if that is what you mean.

assets/src/scss/blocks/PostsListBlock.scss
Osong-Michael: Based on the tokens we have, we don't have a token for `14px`, so I used the `calc()` to subtract `2px` from the token we have for `16px` (based of your recommendation to use tokens). I could add the values in comments to make it clearer.

assets/src/scss/blocks/PostsListBlock.scss
Osong-Michael: I am using [spacing values](https://github.com/greenpeace/planet4-master-theme/blob/main/assets/src/scss/base/_variables.scss#L14) here, just that when I need specific units I need to calculate them, like above I need a negative value to move the element up.

assets/src/blocks/PostsList/PostsListBlock.js
comzeradd: I think we don't offer translation for our block names, just to make things less confusing on Handbook documentation.

comzeradd: Can't we just use "Posts List" as a name?

Osong-Michael: I changed the name because WP offers a variation of the Query Loop which is also called Posts List, so we may need to change ours or add `Block` at the end to differentiate the two.

Screenshot 2023-11-02 at 09 42 04

assets/src/blocks/PostsList/PostsListBlock.js
GP-Dan-Tovbein: For instance `24px`

assets/src/blocks/PostsList/PostsListBlock.js
```diff
@@ -0,0 +1,68 @@
...
+      ['core/paragraph', {
+        placeholder: __('Enter description', 'planet4-blocks-backend'),
+        style: {
+          spacing: {
+            margin: {
+              top: '25px',
+              bottom: '35px',
+            },
+          },
+        },
```
GP-Dan-Tovbein, line 46: And here `36px`

GP-Dan-Tovbein: But i'd suggest to validate with Houssam first!

GP-Dan-Tovbein (commented):
Code looks good to me!

General comment! From my point of view, I disagree to use `calc()` with `subract` or `add` to `margin` or `padding` since it's really hard to understand the assigned value.
In that cases I'd prefer to follow up with the design team first and then define static values, such as `32px`. It should be always multiple of `4`.

assets/src/scss/style.scss
```diff
@@ -79,6 +79,9 @@ Text Domain: planet4-master-theme
 // CSS variables
 @import "base/css-variables";
 
+// Posts List Block
+@import "blocks/PostsListBlock";
```
mleray, line 83: Why do we not add this to the [blocks style file](https://github.com/greenpeace/planet4-master-theme/blob/main/assets/src/blocks/styles/blocks.scss) instead?

Osong-Michael: Oh my bad, I had started working on this before Sagar merged the code to move blocks to the master theme, so let me change that.

mleray (commented):
We can change this description via the `description` attribute, maybe we should? It's a bit vague right now since it's the default

Screenshot 2023-11-13 at 10 28 05

assets/src/scss/style.scss
mleray: Maybe it actually makes more sense to do it this way 🤔 I'm not sure we need a separate style file for blocks, also it seems strange to have some styles outside of the `scss` folder

Osong-Michael:
> PLANET-7214


> We can change this description via the `description` attribute, maybe we should? It's a bit vague right now since it's the default
> 
> Screenshot 2023-11-13 at 10 28 05

Tagging @comzeradd to see if he has any suggestions.

comzeradd:
> We can change this description via the `description` attribute, maybe we should?

We can adapt the one suggested by the design team for the Articles block:

Posts List is the place in Planet 4 that the latest articles, press releases and publications can be found.

mleray (commented):
I think that we should remove most of the sidebar options, that might be confusing. According to the ticket requirements we only need to be able to filter by taxonomies, so adding `allowedControls: ['taxQuery']` should do the trick 🙂 See [docs](https://developer.wordpress.org/block-editor/how-to-guides/block-tutorial/extending-the-query-loop-block/#disabling-irrelevant-or-unsupported-query-controls) for more details

mleray (commented):
I see in the ticket requirements that it should be possible to change the amount of posts, I've added something similar in my [Actions List PR](https://github.com/greenpeace/planet4-master-theme/pull/2149/files#diff-aabb623e32776e19ec89d1c093007def03e230b12831fa51b20ebcfa735a5513) if you want inspiration (although that PR is still a draft so if you have better ideas of how to implement this, please let me know!)

src/Loader.php
mleray: I think that here we should also check that the blocks plugin is active, by adding `!Planet4Blocks::is_active()`

Osong-Michael:
> I see in the ticket requirements that it should be possible to change the amount of posts, I've added something similar in my [Actions List PR](https://github.com/greenpeace/planet4-master-theme/pull/2149/files#diff-aabb623e32776e19ec89d1c093007def03e230b12831fa51b20ebcfa735a5513) if you want inspiration (although that PR is still a draft so if you have better ideas of how to implement this, please let me know!)

Yes, I discussed that with Nikos while you were away and we decided to leave that option in the toolbar above the block since we may soon allow Editors to use the _**Query Loop Block**_ and thats's where the option to change the number of posts is.

src/Loader.php
Osong-Michael: > I think that here we should also check that the blocks plugin is active, by adding `!Planet4Blocks::is_active()` (like we do in the `add_blocks` function)

I am not sure, maybe @comzeradd can shed more light on this, but my understanding was this implementation was to exist despite us having to move the _Blocks_ to the theme folder, so the _Beta Blocks_ should be implemented even without the Planet4Blocks feature

comzeradd: For now I think it's fine to just check for Beta blocks. Since this feature flag only exists on dev sites there is no danger for someone accidentally enabling this block in production.

Once we are ready to actual get this block live then we can substitute the Beta check with the `Planet4Blocks` feature flag.

comzeradd:
> Yes, I discussed that with Nikos while you were away and we decided to leave that option in the toolbar above the block

Yes, we still need to [limit to 3](https://github.com/greenpeace/planet4-master-theme/pull/2137/files#diff-ce4b283c7f18ddd76b48c753fd7426e195c1eae9a2305cf9fdeb56bdbf30782aR22) by default and allow editors to change that but since QLB offers that functionality out of the box in its toolbar we can use just that instead of replicating it in the sidebar.

src/Loader.php
mleray: Maybe we can directly add this code to the `add_blocks` function? Instead of creating a new `add_beta_blocks`? 🤔

mleray: Also, we already enqueue this script [in the MasterBlocks file](https://github.com/greenpeace/planet4-master-theme/blob/1200cf93530ee9a9e0a85569644ac74617971d2f/src/MasterBlocks.php#L51), should we remove that then if we add it here?

assets/src/blocks/PostsList/PostsListBlock.js
```diff
@@ -0,0 +1,70 @@
+export const POSTS_LIST_BLOCK = 'planet4-blocks/posts-list-block';
+
+export const registerPostsListBlock = () => {
+  const {registerBlockVariation} = wp.blocks;
+  const {__} = wp.i18n;
+
+  return registerBlockVariation('core/query', {
```
mleray, line 7: If you add `category: 'planet4-blocks-beta'` then it will show in the editor in the blocks list along with our other beta blocks, which I believe is what we want?

Screenshot 2023-11-16 at 11 00 26

mleray: (not sure if we should add `(beta`) to the name 🤔)

src/Loader.php
Osong-Michael: Hey @mleray, thanks for the feedback.
So just a few points to see what you think. The reason for creating the `add_beta_blocks` function is so that in the future if we have to change things we can just isolate that particular code and integrate it into the `add_blocks`, depending on the route we choose. Also, it is easier to manage the script for the Beta blocks since we are enqueueing just that script while the `MasterBlock` file does a few other things.

mleray: But won't we be loading the file twice then? 🤔

Osong-Michael: I think the file is only loaded when the `Planet4Blocks` feature is enabled, which by default is not since that will throw an error because of the GuestBook block

mleray: Actually I think most of our sites have it enabled, no? Doesn't it mean that our blocks plugin is enabled?

Osong-Michael: Let me double check again.

Osong-Michael: My bad, just realised that. I will have the code updated asap

assets/src/blocks/PostsList/PostsListBlock.js
```diff
@@ -0,0 +1,71 @@
+const POSTS_LIST_BLOCK = 'planet4-blocks/posts-list-block';
+
+export const registerPostsListBlock = () => {
+  const {registerBlockVariation} = wp.blocks;
+  const {__} = wp.i18n;
+
+  return registerBlockVariation('core/query', {
+    name: POSTS_LIST_BLOCK,
+    title: 'Posts List',
```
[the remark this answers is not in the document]
comzeradd: I think it’s ok for now since it’s Beta. I’ll open a follow up ticket to investigate hiding the native QL block variations

mleray (commented):
I think we need to add a limit to the amount of lines for the headline, I believe it should be 2?

Screenshot 2023-11-22 at 13 48 58

assets/src/blocks/PostsList/PostsListBlock.js
```diff
@@ -0,0 +1,71 @@
...
+          ['core/group', {}, [
+            ['core/post-terms', {
+              term: 'category',
+              separator: ' ',
+            }],
+            ['core/post-title', {isLink: true, linkTarget: '_blank'}],
+            ['core/post-excerpt'],
+            ['core/group', {className: 'posts-list-meta'}, [
+              ['core/post-author-name', {isLink: true, linkTarget: '_blank'}],
+              ['core/post-date'],
```
mleray, line 63: Do we need to add the reading time after this, similar to what we do for [listing pages](https://github.com/greenpeace/planet4-master-theme/blob/05346596a9e9255c5df800429c559550b68d6800/parts/query-grid.html#L29)?

Osong-Michael: Hey @mleray and @comzeradd, so I was looking into this and after a conversation with @lithrel, this [block](https://github.com/greenpeace/planet4-master-theme/blob/main/functions.php#L237) is not registered on the frontend so it throws an error and crashed the editor. We could create another ticket to register the block to the frontend and add to the Posts List block(so as to maintain this ticket's scope) or do we do it in this ticket?

comzeradd: I can create a follow up ticket for the "reading time" block to be added in the Posts List

assets/src/blocks/PostsList/index.js
```diff
@@ -0,0 +1,71 @@
...
+      }],
+      ['core/post-template', {}, [
+        ['core/columns', {}, [
+          ['core/post-featured-image', {isLink: true, linkTarget: '_blank'}],
+          ['core/group', {}, [
+            ['core/post-terms', {
+              term: 'category',
+              separator: ' ',
+            }],
+            ['core/post-title', {isLink: true, linkTarget: '_blank'}],
```
mleray, line 59: I'm wondering if these links should actually have `_blank` target? The Articles block doesn't do that 🤔 I did the same thing for the Actions List, and that might have been a mistake since the Covers block also doesn't do that 😬

comzeradd: Ah yes, I don't think these links are supposed to open in a new tab in either of these blocks :)

Osong-Michael: I will update the Actions List block too

assets/src/scss/editorOverrides.scss
```diff
@@ -415,3 +417,48 @@ input.describe[type=text][data-setting=caption] {
     }
   }
 }
+
+// Posts List Block Override
+.posts-list-query {
```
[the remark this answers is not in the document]
Osong-Michael: @mleray, yes we do, some elements appear in the editor that we do not want to show to editors, so I hide them using css, also the styling is a bit different in the editor than on the page, so need this to make it similar.

mleray (approved):
Looks good to me 👍 Thanks for your patience Osong! 🙏

PR #2149, opened 2023-11-13
PLANET-7251 Build an Actions List block

mleray:
### Description

See [PLANET-7251](https://jira.greenpeace.org/browse/PLANET-7251)

- [x] Post Type: fetches only Actions, except if the "New IA" feature flag is off, then it should fetch Act page children pages
- [x] Display 3 Actions by default
- [x] This block should be available when `Planet 4 > Features > Beta blocks` is enabled
- [x] The implement code should live in the master-theme
- [x] Block name: `Actions List`

Sidebar sections/options:
- Settings (via the native toolbar)
  - [x] Actions count
- Filters (via the native sidebar, using the `Taxonomies` filters)
  - [x] Categories 
  - [x] Tags

### Testing

You can try out the new block on local (making sure the beta blocks are enabled) or on [this page](https://www-dev.greenpeace.org/test-venus/the-new-actions-list-block/) for example. Make sure to test it with and without the new IA enabled, as well as with the various possible settings (Actions count and filters)


planet-4:

:watch: 2023.11.13 09:37:39

mleray:
There is a problem with this implementation, that if we toggle the new IA existing blocks will not be updated, only newly created ones will show the correct pages... I'm not sure there's a solution for that 😕 @comzeradd maybe we should only enable this block when the new IA is enabled?

assets/src/blocks/ActionsList/ActionsListSidebar.js
```diff
@@ -0,0 +1,41 @@
...
+  const withActionsListOptions = BlockEdit => props => {
+    const {attributes, name, setAttributes} = props;
+    const isActionsList = name === 'core/query' && attributes.namespace === BLOCK_NAME;
+
+    return isActionsList ? (
+      <>
+        <InspectorControls>
+          <PanelBody title={__('Planet 4 Options')}>
+            <TextControl
+              label={__('Actions count', 'planet4-blocks-backend')}
```
comzeradd, line 19: I think we can skip that completely. If we document the block properly we can guide editors to use the toolbar section to adjust the number of Actions to show.

mleray:
> Can you maybe test one thing: if you enable "New IA" and then convert the Act children pages to Actions using the Post Type Switcher plugin, does it work for the Query Loop block?

@comzeradd yes it works, the converted page does show up in the Actions List then!

Osong-Michael (approved):
👍 looks good to me

PR #2154, opened 2023-11-23
PLANET-7273 Add toggle button to show listing pages list or grid view

mleray:
### Description

See [PLANET-7273](https://jira.greenpeace.org/browse/PLANET-7273)

This commit also removes the `ListingPageGridView` feature flag and associated code.


planet-4:

:watch: 2023.11.23 11:30:58

index.php
```diff
@@ -44,13 +43,12 @@
...
 
     if (!empty(planet4_get_option('new_ia'))) {
-        $view = ListingPageGridView::is_active() ? 'grid' : 'list';
-
-        $query_template = file_get_contents(get_template_directory() . "/parts/query-$view.html");
-
-        $content = do_blocks($query_template);
-
-        $context['query_loop'] = $content;
+        $grid_template = file_get_contents(get_template_directory() . "/parts/query-grid.html");
```
GP-Dan-Tovbein, line 46: What do you think if we create a static PHP class to get all these code which is being repeated also in other listing pages?

mleray: I thought about it, I agree it's not ideal having duplicated code like this... But I think it should be handled in another PR, we could create a follow-up maintenance ticket

GP-Dan-Tovbein (commented):
Code looks good to me, however, I have a small doubt about the behavior.

I tested the shared demo page and noticed that both views (grid and list) are rendered into a page even if you don't switch the layout. From my point of view and thinking about the performance stunning, it could be great if we avoid that behavior and try to render just one template ([here](https://github.com/greenpeace/planet4-master-theme/pull/2154/files#diff-215a8c8a1fb314669a4d20de3ce3abe72340f7322ffa3e24584378cf0b3b41afR23)) and then magic might be happen within CSS and JS.

In addition to that, maybe I'm wrong but I don't see any significant difference between  both templates (grid and list).

What do you think?

mleray:
> From my point of view and thinking about the performance stunning, it could be great if we avoid that behavior and try to render just one template ([here](https://github.com/greenpeace/planet4-master-theme/pull/2154/files#diff-215a8c8a1fb314669a4d20de3ce3abe72340f7322ffa3e24584378cf0b3b41afR23)) and then magic might be happen within CSS and JS.

I 100% agree with that, I'll look into improving it!
 
> In addition to that, maybe I'm wrong but I don't see any significant difference between both templates (grid and list).

That's something that should be discussed with the design team 😬 I don't know who requested this feature in the first place

comzeradd:
> > In addition to that, maybe I'm wrong but I don't see any significant difference between both templates (grid and list).
> 
> That's something that should be discussed with the design team 😬 I don't know who requested this feature in the first place

I think that @GP-Dan-Tovbein doesn't talk about the design output but more about the code of the [two partials](https://github.com/greenpeace/planet4-master-theme/tree/listing-pages-view-switch/parts)? If the template code are almost identical we could potentially use the same for both and adjust the two different UI on css only.

mleray:
> I think that @GP-Dan-Tovbein doesn't talk about the design output but more about the code of the [two partials](https://github.com/greenpeace/planet4-master-theme/tree/listing-pages-view-switch/parts)? If the template code are almost identical we could potentially use the same for both and adjust the two different UI on css only.

The templates are quite a bit different actually, for example they don't use the same Query Loop arguments and I don't see how we can conditionally change that via CSS/JS 🤔

assets/src/scss/pages/_listing-page.scss
```diff
@@ -196,6 +196,11 @@
   // Only applied to "grid" layout
   &.wp-block-query--grid {
     .wp-block-post-template {
+      display: flex;
+      flex-direction: row;
+      flex-wrap: wrap;
+      gap: 1.25em;
```
GP-Dan-Tovbein, line 202: Shouldn't we use spacing values here?

GP-Dan-Tovbein (approved):
Code looks good to me! Well done !

I've only one left comment but not affecting the merge.

assets/src/scss/pages/_listing-page.scss
mleray: I used the values from WordPress Query Loop block

index.php
mleray: Created a [ticket](https://jira.greenpeace.org/browse/PLANET-7365) for this

PR #2166, opened 2023-12-07
Capture errors on google client instantiation

lithrel:
Ref: https://sentry.greenpeace.org/organizations/greenpeace-org/issues/230/

Capture errors on Google libs usage to avoid crashing instance and blocking post editing.

This is currently deployed on Japan instance and handles error properly.
https://sentry.greenpeace.org/organizations/greenpeace-org/discover/results/?field=title&field=event.type&field=timestamp&field=url&field=handled&name=All+Events&query=undefined+method&sort=-timestamp&statsPeriod=24h&yAxis=count%28%29

### DIFF
--- a/src/AnalyticsValues.php
+++ b/src/AnalyticsValues.php
@@ -173,14 +173,21 @@ final class AnalyticsValues
 
             return self::from_cache_array($cache);
         }
-        $instance = self::using_google();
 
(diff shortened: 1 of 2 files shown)


comzeradd (approved):
Looks good. You just have a typo in your commit message 😁

PR #2168, opened 2023-12-08
Fix Gravity Forms confirmation text test

mleray:
### Description

The generic name previously used was sometimes considered spam and therefore would not behave as expected


lithrel (approved):
:grin: :snowflake: :+1:

PR #2179, opened 2023-12-18
Guestbook: Change package name of class

GP-Dan-Tovbein:
No ticket needed

# Description
We do not need to use P4GBKS to blocks into the theme


sagarsdeshmukh (approved):
👍

PR #2185, opened 2024-01-04
PLANET-7367: Add manual override to Posts List block

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-7367

> Follow up from [PLANET-7214](https://jira.greenpeace.org/browse/PLANET-7214) to meet feature parity with the Articles block. We need an option for editors to able to select which specific posts they want to be displayed.

- import of component PostSelector from plugin
- use of a custom filter `postIn` and handle of value in hooks
  - I tried to use the `include` filter available on rest API, but it generates some issues with various requests (maybe mixing up some cache), coming probably from [this piece of code](https://github.com/WordPress/gutenberg/blob/e2bda4036db44339aa56d0c85fa3d75edd79cd81/packages/core-data/src/queried-data/get-query-parts.js#L82-L91)

Screenshot from 2024-01-05 16-58-36

## Test

- Activate Core blocks and All blocks in _planet4 > Features_
- Add a _Posts List_ block to a new page
- Use the _Manual Override_ option to display specific posts in the block
  - Check in the editor and on the frontend that those articles are present and in order

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -165,6 +165,7 @@ final class Loader
             return;
         }
 
+        Blocks\PostsList::registerHooks();
         add_filter(
             'allowed_block_types_all',
             function ($allowed_block_types) {
--- a/assets/src/block-editor/PostSelector/PostSelector.js
+++ b/assets/src/block-editor/PostSelector/PostSelector.js
@@ -1,0 +1,115 @@
+import {FormTokenField} from '@wordpress/components';
+import {dispatch, useSelect} from '@wordpress/data';
+const {__} = wp.i18n;
+
+// Allows to query a custom endpoint with select('core') tools
+dispatch('core').addEntities([{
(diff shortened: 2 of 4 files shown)


planet-4:

:watch: 2024.01.04 21:18:05

src/Loader.php
```diff
@@ -160,6 +160,7 @@ public static function add_blocks(): void
         new Blocks\Accordion();//NOSONAR
         new Blocks\Cookies();//NOSONAR
         new Blocks\Counter();//NOSONAR
+        Blocks\PostsList::registerHooks();
```
Osong-Michael, line 163: Shouldn't we have this code run after the Beta Blocks check, since it affects the Posts List block only?

lithrel: Good point, fixed :+1:

Osong-Michael (approved):
👍

PR #2187, opened 2024-01-08
PLANET-7135: Fix country selector tab index

GP-Dan-Tovbein:
Ref: https://jira.greenpeace.org/browse/PLANET-7135

[Demo page](https://www-dev.greenpeace.org/test-jupiter)

# Description
- Fixed the issue but also Improved SEO on the country selector component. 
- Additionally, I've also Implemented a new script which is in charge to add/remove dynamically the `tabindex` attribute depending on the country selector state (`open|close`).

## Testing
1. Scroll down to the footer
2. Click on it footer
3. Then press the `Tab` key.

Within the previous behavior ([see the international dev site](https://www-dev.greenpeace.org/international/)),  the end-user is not able to navigate within page's links NOR social links.
Within the current fix, the user can navigate through ALL the footer's links, including sub-links, even if the list is opened or closed.

https://github.com/greenpeace/planet4-master-theme/assets/77975803/1128f149-22c8-474e-ab60-f4c2731a7804

### DIFF
--- a/assets/src/js/country_selector.js
+++ b/assets/src/js/country_selector.js
@@ -14,7 +14,7 @@ export const setupCountrySelector = () => {
             // This check applies not only to the parent link but also to children
             // In some cases, the country has also different link's languages
             if (countryItem.classList.contains('country-group')) {
-              countryItem.setAttribute('tabIndex', isOpen ? 0 : -1);
+              countryItem.setAttribute('tabIndex', -1);
 
               for (const link of countryItem.querySelectorAll('li')) {
                 link.setAttribute('tabIndex', isOpen ? 0 : -1);
--- a/assets/src/js/country_selector.js
+++ b/assets/src/js/country_selector.js
@@ -1,0 +1,31 @@
+// The script basically add/remove the tabindex attribute within the country selector
+// It helps to improve SEO
+export const setupCountrySelector = () => {
+  const countries = document.querySelectorAll('.countries-list .countries > li');
+
+  // Create a new MutationObserver with a country selector
+  new MutationObserver(
+    mutationList => {
+      for (const mutation of mutationList) {
+        if (mutation.target.className.includes('footer-country-selector') && mutation.attributeName === 'class') {
+          const isOpen = mutation.target.className.includes('open');
+
+          for (const countryItem of countries) {
+            // This check applies not only to the parent link but also to children
+            // In some cases, the country has also different link's languages
+            if (countryItem.classList.contains('country-group')) {
+              countryItem.setAttribute('tabIndex', isOpen ? 0 : -1);
+
+              for (const link of countryItem.querySelectorAll('li')) {
+                link.setAttribute('tabIndex', isOpen ? 0 : -1);
+              }
+            }
+          }
+        }
+      }
+    }
+  ).observe(
+    document.querySelector('.footer-country-selector'),
+    {attributes: true, childList: true, subtree: true}
+  );
+};
--- a/assets/src/js/country_selector.js
+++ b/assets/src/js/country_selector.js
@@ -1,31 +1,50 @@
 // The script basically add/remove the tabindex attribute within the country selector
 // It helps to improve SEO
 export const setupCountrySelector = () => {
-  const countries = document.querySelectorAll('.countries-list .countries > li');
+  const targetNode = document.querySelector('#country-selector');
+  const countriesList = document.querySelector('.countries-list');
+  const footerMenu = document.querySelector('.footer-menu');
+  const footerSocialMedia = document.querySelector('.footer-social-media');
+  const largeAndUpScreens = '(min-width: 992px)';
 
-  // Create a new MutationObserver with a country selector
-  new MutationObserver(
+  const switchOrders = (evt = null) => {
+    const event = evt ? evt : window.matchMedia(largeAndUpScreens);
+    if (event.matches) {
+      footerMenu.parentNode.insertBefore(footerMenu, footerMenu.parentNode.firstChild);
+    } else {
+      footerSocialMedia.parentNode.insertBefore(footerSocialMedia, footerSocialMedia.parentNode.firstChild);
+    }
+  };
+
+  // Create an observer instance linked to the callback function
+  const observer = new MutationObserver(
+    // Callback function to execute when mutations are observed
     mutationList => {
+      // Set default visibility
       for (const mutation of mutationList) {
-        if (mutation.target.className.includes('footer-country-selector') && mutation.attributeName === 'class') {
-          const isOpen = mutation.target.className.includes('open');
-
-          for (const countryItem of countries) {
-            // This check applies not only to the parent link but also to children
-            // In some cases, the country has also different link's languages
-            if (countryItem.classList.contains('country-group')) {
-              countryItem.setAttribute('tabIndex', -1);
-
-              for (const link of countryItem.querySelectorAll('li')) {
-                link.setAttribute('tabIndex', isOpen ? 0 : -1);
-              }
-            }
+        if (mutation.attributeName === 'class') {
+          if (targetNode.classList.contains('open')) {
+            countriesList.style.visibility = 'visible';
+          } else {
+            setTimeout(() => {
+              countriesList.style.visibility = 'hidden';
+            }, 1000); // Follow the .countries-list transition time
           }
         }
       }
     }
-  ).observe(
-    document.querySelector('.footer-country-selector'),
+  );
+
+  // Start observing the target node for configured mutations
+  observer.observe(
+    targetNode,
     {attributes: true, childList: true, subtree: true}
   );
+
+  window.matchMedia(largeAndUpScreens).addEventListener('change', event => {
+    switchOrders(event);
+  });
+
+  // Set default order
+  switchOrders();
 };
--- a/assets/src/js/country_selector.js
+++ b/assets/src/js/country_selector.js
@@ -1,12 +1,27 @@
-// The script basically add/remove the tabindex attribute within the country selector
-// It helps to improve SEO
+/**
+ * The script basically add/remove the tabindex attribute within the country selector
+ * It helps to improve SEO
+ * @return {void}
+ */
 export const setupCountrySelector = () => {
   const targetNode = document.querySelector('#country-selector');
   const countriesList = document.querySelector('.countries-list');
   const footerMenu = document.querySelector('.footer-menu');
   const footerSocialMedia = document.querySelector('.footer-social-media');
-  const largeAndUpScreens = '(min-width: 992px)';
 
+  if (!targetNode || !countriesList || !footerMenu || !footerSocialMedia) {
+    return;
+  }
+
+  const largeAndUpScreens = '(min-width: 992px)';
+  const countriesListTransitionDuration = (
+    parseFloat(getComputedStyle(document.querySelector('.countries-list')).transitionDuration) * 1000
+  ) || 500;
+
+  /**
+   * This funtion re-order the elements from the footer
+   * @param {*} evt
+   */
   const switchOrders = (evt = null) => {
     const event = evt ? evt : window.matchMedia(largeAndUpScreens);
     if (event.matches) {
@@ -28,7 +43,7 @@ export const setupCountrySelector = () => {
           } else {
             setTimeout(() => {
               countriesList.style.visibility = 'hidden';
-            }, 1000); // Follow the .countries-list transition time
+            }, countriesListTransitionDuration);
           }
         }
       }
@@ -36,10 +51,7 @@ export const setupCountrySelector = () => {
   );
 
   // Start observing the target node for configured mutations
-  observer.observe(
-    targetNode,
-    {attributes: true, childList: true, subtree: true}
-  );
+  observer.observe(targetNode, {attributes: true, childList: true, subtree: true});
 
   window.matchMedia(largeAndUpScreens).addEventListener('change', event => {
     switchOrders(event);
--- a/assets/src/js/country_selector.js
+++ b/assets/src/js/country_selector.js
@@ -15,7 +15,7 @@ export const setupCountrySelector = () => {
 
   const largeAndUpScreens = '(min-width: 992px)';
   const countriesListTransitionDuration = (
-    parseFloat(getComputedStyle(document.querySelector('.countries-list')).transitionDuration) * 1000
+    parseFloat(getComputedStyle(countriesList).transitionDuration) * 1000
   ) || 500;
 
   /**
@@ -25,9 +25,9 @@ export const setupCountrySelector = () => {
   const switchOrders = (evt = null) => {
     const event = evt ? evt : window.matchMedia(largeAndUpScreens);
     if (event.matches) {
-      footerMenu.parentNode.insertBefore(footerMenu, footerMenu.parentNode.firstChild);
+      footerMenu.parentNode.prepend(footerMenu);
     } else {
-      footerSocialMedia.parentNode.insertBefore(footerSocialMedia, footerSocialMedia.parentNode.firstChild);
+      footerSocialMedia.parentNode.prepend(footerSocialMedia);
     }
   };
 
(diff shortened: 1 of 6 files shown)


planet-4:

:watch: 2024.01.08 18:15:58

templates/footer_country_selector.twig
```diff
@@ -29,7 +29,7 @@
             <ul aria-label="{{ __('Sites starting with the letter .' ~ initial ~ '.', 'planet4-master-theme') }}" role="list">
                 {% for country in countries %}
                 {% set main_lang, lang_count = country.lang|first, country.lang|length %}
-                <li>
+                <li role="listitem">
```
GP-Dan-Tovbein, line 32: Based on the [Mozilla's best practices page for lists](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listitem_role#best_practices)

templates/footer_country_selector.twig
```diff
@@ -39,7 +39,7 @@
                     {% if lang_count > 1  %}
                     <ul class="lang-list" role="list">
                         {% for k, lang in country.lang %}
-                        <li>
+                        <li role="listitem">
```
GP-Dan-Tovbein, line 42: Same here

templates/footer.twig
```diff
@@ -4,15 +4,12 @@
         {% include 'footer_country_selector.twig' %}
     </div>
     {% endif %}
-    <div class="container d-flex flex-column flex-lg-row-reverse">
```
GP-Dan-Tovbein, line 7: Remove the `flex-lg-row-reverse` bootstrap's class which was wrongly [merged a while ago](https://github.com/greenpeace/planet4-master-theme/commit/31b4b18359006498dd64dd405b5472e16e69facc#diff-9ccc5c1e49803dab74b7d5fb3b391bb54c4521688a0f01b887e1d082a4759e60R2).

assets/src/js/country_selector.js
```diff
@@ -0,0 +1,39 @@
...
+  // Create a new MutationObserver with a country selector
+  new MutationObserver(
+    mutationList => {
+      for (const mutation of mutationList) {
+        if (mutation.target.className.includes('footer-country-selector') && mutation.attributeName === 'class') {
+          const isOpen = mutation.target.className.includes('open');
+
+          for (const countryItem of countries) {
+            // This check applies not only to the parent link but also to children
+            // In some cases, the country has also different link's languages
```
GP-Dan-Tovbein, line 15: Add the `tabindex` to ALL elements

assets/src/js/country_selector.js
```diff
@@ -0,0 +1,39 @@
...
+                countryItem.setAttribute('tabIndex', 0);
+              } else {
+                countryItem.removeAttribute('tabIndex');
+              }
+
+              for (const link of countryItem.querySelectorAll('li')) {
+                if (isOpen) {
+                  link.setAttribute('tabIndex', 0);
+                } else {
+                  link.removeAttribute('tabIndex');
```
GP-Dan-Tovbein, line 27: Remove the `tabindex` when the country selector is closed

templates/footer.twig
mleray: I think this was intentional, so that on mobile the social media links are shown first? (see [GPI's website](https://www.greenpeace.org/international/) for example)

GP-Dan-Tovbein: I agree! (also that is my code 🙃)

I've decided to change `flex-column flex-lg-row-reverse` with `flex-column-reverse flex-lg-row` to get the tab feature properly working on desktop.

mleray (commented):
I know it's not the scope of this problem, but it's likely a tiny fix and it's related to the country selector too... I think that there's been an issue when merging the new identity, the border bottom is different when the selector is closed and when it's open:

- Closed: 
Screenshot 2024-01-12 at 11 46 24

- Open:
Screenshot 2024-01-12 at 11 46 33

I think in both cases it should be the first one (more transparent)

mleray:
Also, when I test on the phobos instance I get a different behaviour than what you show in your screen recording:

https://github.com/greenpeace/planet4-master-theme/assets/6949075/8c001f3b-2bee-41cc-a304-b35c8f99bd07

It seems I get to the groups before getting to the actual links? Maybe it's a browser thing? I'm using Chrome

mleray:
I still see the same behaviour on the phobos test instance 🤔 Also, I re-read the ticket description and I think that the issue was misunderstood! It was actually working fine when the country selector was open, the problem was when the country selector is closed, then it still goes through all the country links before getting to the other sections of the footer... Does that make sense? I still see it happening on the phobos test instance

templates/footer.twig
```diff
@@ -4,10 +4,7 @@
...
-        {% if (social_overrides is defined and social_overrides|length) or footer_social_menu|length %}
-            {% include 'footer_social_media.twig' %}
-        {% endif %}
+    <div class="container d-flex flex-column flex-lg-row">
```
GP-Dan-Tovbein, line 7: I've removed the `*-reverse` class since is not compatible with a18y. Why? Because even if the element is re-arranged, that beaviour doesn't modify the `focus` tab. [The re-arrangement might be added programatically](https://github.com/greenpeace/planet4-master-theme/pull/2187/files#diff-7413612e8358cee4efb121d4eece663de1227d3f1caf111b09506a840a8a4dc3R10-R49).

GP-Dan-Tovbein:
@mleray I've updated this PR.

If you notuce a tricky implementation is because of https://developer.mozilla.org/en-US/docs/Web/CSS/flex-direction#accessibility

> Using the flex-direction property with values of row-reverse or column-reverse will create a disconnect between the visual presentation of content and DOM order. This will adversely affect users experiencing low vision navigating with the aid of assistive technology such as a screen reader. If the visual (CSS) order is important, then screen reader users will not have access to the correct reading order.

mleray:
@GP-Dan-Tovbein the demo page is not up to date

templates/footer_social_media.twig
```diff
@@ -29,12 +30,14 @@
             {% set name = class_name %}
           {% endif %}
         {% endfor %}
-        <li>
+        <li tabindex="-1">
           <a
+          tabindex="0"
```
mleray, line 35: Indentation seems off here

GP-Dan-Tovbein:
> @GP-Dan-Tovbein the demo page is not up to date

I didn't update the description before 🤦 Thanks @mleray !

assets/src/js/country_selector.js
```diff
@@ -0,0 +1,50 @@
...
+              countriesList.style.visibility = 'hidden';
+            }, 1000); // Follow the .countries-list transition time
+          }
+        }
+      }
+    }
+  );
+
+  // Start observing the target node for configured mutations
+  observer.observe(
```
mardelnet, line 39: The `MutationObserver` is never disconnected. Couldn't that lead to memory leaks?

assets/src/js/country_selector.js
```diff
@@ -0,0 +1,50 @@
...
+    {attributes: true, childList: true, subtree: true}
+  );
+
+  window.matchMedia(largeAndUpScreens).addEventListener('change', event => {
+    switchOrders(event);
+  });
+
+  // Set default order
+  switchOrders();
+};
```
mardelnet, line 50: Shouldn't we wait for the DOM to be fully loaded?

assets/src/js/country_selector.js
mardelnet, line 1: Can you please add JSDocs to all the functions to improve clarity?

assets/src/js/country_selector.js
```diff
@@ -0,0 +1,50 @@
...
+    mutationList => {
+      // Set default visibility
+      for (const mutation of mutationList) {
+        if (mutation.attributeName === 'class') {
+          if (targetNode.classList.contains('open')) {
+            countriesList.style.visibility = 'visible';
+          } else {
+            setTimeout(() => {
+              countriesList.style.visibility = 'hidden';
+            }, 1000); // Follow the .countries-list transition time
```
mardelnet, line 31: I haven't tested this, but I think we can obtain this value dynamically:

```js
const transitionTime = parseFloat(getComputedStyle(countriesList).transitionDuration) * 1000 || 1000;
```

assets/src/js/country_selector.js
```diff
@@ -0,0 +1,50 @@
...
+export const setupCountrySelector = () => {
+  const targetNode = document.querySelector('#country-selector');
+  const countriesList = document.querySelector('.countries-list');
+  const footerMenu = document.querySelector('.footer-menu');
+  const footerSocialMedia = document.querySelector('.footer-social-media');
+  const largeAndUpScreens = '(min-width: 992px)';
+
+  const switchOrders = (evt = null) => {
+    const event = evt ? evt : window.matchMedia(largeAndUpScreens);
+    if (event.matches) {
```
mardelnet, line 12: The `insertBefore` call always executes regardless of the current order, even if the elements are already in the correct position:
```suggestion
    if (event.matches && footerMenu !== footerMenu.parentNode.firstChild) {
```

assets/src/js/country_selector.js
```diff
@@ -0,0 +1,50 @@
+// The script basically add/remove the tabindex attribute within the country selector
+// It helps to improve SEO
+export const setupCountrySelector = () => {
+  const targetNode = document.querySelector('#country-selector');
+  const countriesList = document.querySelector('.countries-list');
+  const footerMenu = document.querySelector('.footer-menu');
+  const footerSocialMedia = document.querySelector('.footer-social-media');
```
mardelnet, line 7: Should we add checks to ensure the elements exist before interacting with them?

mardelnet (commented):
Hey @dantovbein, in my local environment, the order of the columns is different from that of the test instance. Does the same happen to you?

Screenshot from 2024-11-18 12-20-31

mleray:
> Hey @dantovbein, in my local environment, the order of the columns is different from that of the test instance. Does the same happen to you?

@mardelnet maybe you need to clear your Twig cache? 🤔 `npx wp-env run cli wp timber clear_cache`

GP-Dan-Tovbein:
> > Hey @dantovbein, in my local environment, the order of the columns is different from that of the test instance. Does the same happen to you?
> 
> @mardelnet maybe you need to clear your Twig cache? 🤔 `npx wp-env run cli wp timber clear_cache`

Yep! Exactly. Every time you update a twig template, you'll need clear the cache too.

assets/src/js/country_selector.js
GP-Dan-Tovbein: I don't see neccesary in this case since the footer is part of the index page. And the `window.matchMedia` is invoked whenever it resized.

assets/src/js/country_selector.js
GP-Dan-Tovbein: Sure!

assets/src/js/country_selector.js
GP-Dan-Tovbein: It's a good feedback! I think that in this case the country selector might be observed every time during the execution. If you stop the observer, then the country selector wouldn't be never modified.

assets/src/js/country_selector.js
GP-Dan-Tovbein: Sorry! I'm not sure if I understand correctly your suggestion 🙏 .

When the event matches, the `footerMenu` and `footerMenu.parentNode.firstChild` are never the same. This part is invoked right only the window size is being resized and following this `min-width: 992px`, and not every time you resize even if the size is `1024px` or `1600px`.

assets/src/js/country_selector.js
```diff
@@ -0,0 +1,62 @@
...
+
+  if (!targetNode || !countriesList || !footerMenu || !footerSocialMedia) {
+    return;
+  }
+
+  const largeAndUpScreens = '(min-width: 992px)';
+  const countriesListTransitionDuration = (
+    parseFloat(getComputedStyle(document.querySelector('.countries-list')).transitionDuration) * 1000
```
mardelnet, line 18:
```suggestion
    parseFloat(getComputedStyle(countriesList).transitionDuration) * 1000
```

templates/footer.twig
mardelnet: @dantovbein 
What about using the CSS [order](https://developer.mozilla.org/en-US/docs/Web/CSS/order) property here?
I think we could use it and avoid depending on the `switchOrders` function and a new listener that is always active.

assets/src/js/country_selector.js
```diff
@@ -0,0 +1,62 @@
...
+              countriesList.style.visibility = 'hidden';
+            }, countriesListTransitionDuration);
+          }
+        }
+      }
+    }
+  );
+
+  // Start observing the target node for configured mutations
+  observer.observe(targetNode, {attributes: true, childList: true, subtree: true});
```
mardelnet, line 54: Do we need a mutation observer?
Is it not enough with an `onClick` listener?

assets/src/js/country_selector.js
```diff
@@ -0,0 +1,62 @@
...
+  ) || 500;
+
+  /**
+   * This funtion re-order the elements from the footer
+   * @param {*} evt
+   */
+  const switchOrders = (evt = null) => {
+    const event = evt ? evt : window.matchMedia(largeAndUpScreens);
+    if (event.matches) {
+      footerMenu.parentNode.insertBefore(footerMenu, footerMenu.parentNode.firstChild);
```
mardelnet, line 28: Maybe this way is easier to read?
```suggestion
      footerMenu.parentNode.prepend(footerMenu);
```

assets/src/js/country_selector.js
```diff
@@ -0,0 +1,62 @@
...
+    } else {
+      footerSocialMedia.parentNode.insertBefore(footerSocialMedia, footerSocialMedia.parentNode.firstChild);
```
mardelnet, line 30: Maybe this way is easier to read?
```suggestion
      footerSocialMedia.parentNode.prepend(footerSocialMedia);
```

assets/src/js/country_selector.js
GP-Dan-Tovbein: Good catch!

assets/src/js/country_selector.js
GP-Dan-Tovbein: Have you tested this suggestion?

mardelnet: Yes. It worked on my end.

assets/src/js/country_selector.js
GP-Dan-Tovbein: It could be potentially another option. In that case you might handle 2 different listeners. One for the open button, another for the close button and make sure to not propagate. If you think that that is a better approach I can work on that and re-implement the logic. What do you think?

mardelnet: From the performance perspective, I think 2 `onClick` listeners are better than an observer continuously monitoring changes, but I am not sure. I also think that listeners are easier to read, understand, and maintain. Whatever you choose to do will be okay with me.

GP-Dan-Tovbein: Yep, I see your point! But being honest, I don't see a (potential) performance issue within this implementation. But of course we can follow up this implementation in the future during the `performance and page speed` project.

assets/src/js/country_selector.js
```diff
@@ -0,0 +1,62 @@
+/**
+ * The script basically add/remove the tabindex attribute within the country selector
+ * It helps to improve SEO
```
mardelnet, line 3: I think the script description is not accurate enough (for instance, the script does nothing related to `tabindex`).
IMO, it would be important to mention why we are switching positions between the footer elements, otherwise, the purpose of this code may be hard to understand in the future.

mardelnet (approved):
Great job, @dantovbein

assets/src/js/country_selector.js
GP-Dan-Tovbein: Oh sorry 🙏 ! I've noticed this after I merge the PR. Here is another #2456  which follows your suggestion.

PR #2196, opened 2024-01-17
Include wide-blocks stylesheet into master theme

GP-Dan-Tovbein:
[No ticket needed]

# Description
Include `.block-wide` and `.alignfull` CSS Classes. It should works when the plugin is disabled.

### DIFF
--- a/assets/src/scss/style.scss
+++ b/assets/src/scss/style.scss
@@ -50,6 +50,7 @@ Text Domain: planet4-master-theme
 @import "layout/skewed-overlay";
 @import "layout/tables";
 @import "layout/featured-action";
+@import "layout/wide-blocks";
 
 // Pages
 @import "pages/page";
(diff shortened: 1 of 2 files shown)


planet-4:

:watch: 2024.01.17 12:41:19

assets/src/scss/style.scss
```diff
@@ -50,6 +50,7 @@ Text Domain: planet4-master-theme
 @import "layout/skewed-overlay";
 @import "layout/tables";
 @import "layout/featured-action";
+@import "layout/wide-blocks";
```
mleray, line 53: Importing this here means that it will always be loaded, right? Shouldn't it be loaded only when the "Planet 4 all blocks" feature is enabled, to avoid duplications with the plugin styles? 🤔 Maybe then we should import it in `blocks.scss` instead for example

mleray (approved):
👍

PR #2207, opened 2024-01-31
PLANET-7348 Migrate Page Header options to Page Header block pattern

Osong-Michael:
**Ref: [PLANET-7348](https://jira.greenpeace.org/browse/PLANET-7348)**

**Testing**

You can compare certain pages on the test instance to those of the [DefaultContent](https://www-dev.greenpeace.org/defaultcontent/) instance to see the changes

For example: 
 - [Privacy and cookies](https://www-dev.greenpeace.org/defaultcontent/privacy-and-cookies/) (Default Content Instance)
 - [Privacy and cookies](https://www-dev.greenpeace.org/test-oberon/privacy-and-cookies/) (Test instance)

### DIFF
--- a/src/Migrations/M028MovePageHeaderSideBarOptions.php
+++ b/src/Migrations/M028MovePageHeaderSideBarOptions.php
@@ -1,0 +1,236 @@
+<?php
+
+namespace P4\MasterTheme\Migrations;
+
+use P4\MasterTheme\MigrationRecord;
+use P4\MasterTheme\MigrationScript;
+
+/**
+ * Add default for Posts page.
+ */
+class M028MovePageHeaderSideBarOptions extends MigrationScript
+{
+    /**
+     * Perform the actual migration.
+     *
+     * @param MigrationRecord $record Information on the execution, can be used to add logs.
+     * phpcs:disable SlevomatCodingStandard.Functions.UnusedParameter -- interface implementation
+     */
+    protected static function execute(MigrationRecord $record): void
+    {
+        $args = array(
+            'post_type' => 'page',
+            'posts_per_page' => -1,
+        );
+
+        $pages = get_posts($args);
+
+        foreach ($pages as $page) {
+            $page_meta_data = get_post_meta($page->ID);
+
+            $page_header_title = $page_meta_data['p4_title'][0] ?? '';
+            $page_header_image = $page_meta_data['background_image'][0] ?? '';
+            $page_header_image_id = $page_meta_data['background_image_id'][0] ?? '';
+            $page_header_description = $page_meta_data['p4_description'][0] ?? '';
+            $page_header_button_title = $page_meta_data['p4_button_title'][0] ?? '';
+            $page_header_button_link = $page_meta_data['p4_button_link'][0] ?? '';
+            $page_header_button_link_checkbox = $page_meta_data['p4_button_link_checkbox'][0] ?? '';
+            $page_header_hide_page_title = $page_meta_data['p4_hide_page_title_checkbox'][0] ?? '';
+
+            $page_template = get_page_template_slug($page->ID);
+
+            // If Sitemap Page, update on Page Title.
+            if ('page-templates/sitemap.php' === $page_template) {
+                if (empty($page_header_title)) {
+                    continue;
+                }
+
+                $updated_args = array(
+                    'ID' => $page->ID,
+                    'post_title' => $page_header_title,
+                );
+
+                self::update_meta_options($page_meta_data, $page_header_title, $page);
+
+                wp_update_post($updated_args);
+
+                continue;
+            }
+
+            // phpcs:disable Generic.Files.LineLength.MaxExceeded
+            if (!empty($page_header_image_id) && !empty($page_header_title)) {
+                // Build Page header Pattern
+                $content = '
+                <!-- wp:planet4-block-templates/page-header {"mediaPosition":"right"} -->
+                <!-- wp:group {"align":"full"} -->
+                <div class="wp-block-group alignfull">
+                <!-- wp:group {"className":"container"} -->
+                <div class="wp-block-group container">
+                <!-- wp:media-text {"align":"full","mediaPosition":"right","mediaId": ' . $page_header_image_id . ',"mediaLink":"","mediaType":"image","imageFill":false,"className":"is-pattern-p4-page-header is-style-parallax"} -->
+                <div class="wp-block-media-text alignfull has-media-on-the-right is-stacked-on-mobile is-pattern-p4-page-header is-style-parallax">
+
+                <div class="wp-block-media-text__content"><!-- wp:group -->
+                    <div class="wp-block-group">
+                    <!-- wp:heading {"level":1,"placeholder":"Enter title","backgroundColor":"white"} -->
+                        <h1 class="wp-block-heading has-white-background-color has-background">
+                            ' . $page_header_title . '
+                        </h1>
+                    <!-- /wp:heading --></div>
+                <!-- /wp:group -->
+
+                ' . (!empty($page_header_description) ?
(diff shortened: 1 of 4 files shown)


planet-4:

:watch: 2024.01.31 13:16:23

lithrel:
I think we're missing a step from the ticket:
> Currently we may have some checks in place to hide the page title when the header title is defined. If that's the case we should also enable the Hide page title option during the migration.  

(for reference, [wp title is currently discarded when p4_title is set](https://github.com/greenpeace/planet4-master-theme/blob/650bfbd12468521e00618b991bb2b3692833e630/index.php#L33-L35))

While playing with the script on international db, I get some pages that end up with 2 titles ([this one](https://www.greenpeace.org/international/explore/energy/russian-military-threat-ukraine-nuclear-reactors-facilities-map/) for example); from what I understand, I think the group block should have a H1 tag instead of H2, and `Hide page title` option should be activated

src/Migrations/M028MovePageHeaderSideBarOptions.php
```diff
@@ -0,0 +1,222 @@
...
+                    empty($page_header_image_id)
+                    && empty($page_header_title)
+                    && empty($page_header_description)
+                    && empty($page_header_button_title)
+                ) {
+                    continue;
+                }
+
+                $content = '
+                    <!-- wp:group {"layout":{"type":"flex","orientation":"vertical"}} -->
```
lithrel, line 148: missing a `</div>` with l.178 to close the group if there's no page header button link; I think you should open the group div independent of `!empty($page_header_title)`, so 
```
                $content = '
                    <!-- wp:group {"layout":{"type":"flex","orientation":"vertical"}} -->
                    <div class="wp-block-group">
                        ' . (!empty($page_header_title) ?
```
and close it with the closing comment, so it's easier to follow
```
                    </div>
                    <!-- /wp:group -->
                ';
```

src/Migrations/M028MovePageHeaderSideBarOptions.php
```diff
@@ -0,0 +1,222 @@
...
+                        ' . (!empty($page_header_title) ?
+                            '<div class="wp-block-group"><!-- wp:heading -->
+                                <h1 class="wp-block-heading">' . $page_header_title . '</h1>
```
lithrel, line 151: [default level](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/heading/block.json#L22) for wp-heading is `2`, if you want to use `h1` you have to specify it with `<!-- wp:heading {"level":1} -->`
Side question, do we want the class `page-header-title` for these titles to match the current layout ?

src/Migrations/M028MovePageHeaderSideBarOptions.php
```diff
@@ -0,0 +1,222 @@
...
+                                ' . $page_header_button_title . '
+                            </a>
+                        </div>
+                    <!-- /wp:buttons -->'
+                : '') . '
+
+                </div>
+                <figure class="wp-block-media-text__media">
+                <img
+                    src=' . $page_header_image . '
```
lithrel, line 87: missing quotes around src value `src="' . $page_header_image . '"`

lithrel (changes requested):

src/Migrations/M028MovePageHeaderSideBarOptions.php
```diff
@@ -0,0 +1,222 @@
...
+                                <div class="wp-block-buttons"><!-- wp:button -->
+                                <div class="wp-block-button is-style-cta">
+                                    <a class="wp-block-button__link wp-element-button"
+                                        href="' . $page_header_button_link . '"
+                                        target=' . ($page_header_button_link_checkbox ? "_blank" : "") . '
+                                        rel="noreferrer noopener"
+                                    >
+                                        ' . $page_header_button_title . '
+                                    </a>
+                            </div>
```
comzeradd, line 174: I think that `</div>` should be moved a few lines lower, next to `<!-- /wp:group -->`. This is probably why the footer looks wrong in the Terms page.

src/Migrations/M028MovePageHeaderSideBarOptions.php
Osong-Michael: I will replaces the class then so we have a consistent layout.

src/Migrations/M028MovePageHeaderSideBarOptions.php
```diff
@@ -0,0 +1,288 @@
...
+            $page_header_image = $page_meta_data['background_image'][0] ?? '';
+            $page_header_image_id = $page_meta_data['background_image_id'][0] ?? '';
+            $page_header_description = $page_meta_data['p4_description'][0] ?? '';
+            $page_header_button_title = $page_meta_data['p4_button_title'][0] ?? '';
+            $page_header_button_link = $page_meta_data['p4_button_link'][0] ?? '';
+            $page_header_button_link_checkbox = $page_meta_data['p4_button_link_checkbox'][0] ?? '';
+            $page_header_hide_page_title = $page_meta_data['p4_hide_page_title_checkbox'][0] ?? '';
+
+            // If Sitemap Page, update on Page Title.
+            if ($page->post_title === 'Sitemap') {
```
comzeradd, line 41: I think it would safer to check post_meta and look for `_wp_page_template` and whether it uses `page-templates/sitemap.php`. Some NROs may have a different page title  in the Sitemap page

src/Migrations/M028MovePageHeaderSideBarOptions.php
```diff
@@ -0,0 +1,288 @@
...
+
+                self::update_meta_options($page_meta_data, $page_header_title, $page);
+
+                wp_update_post($updated_args);
+
+                continue;
+            }
+
+            // Terms page, change page title update description and/or button
+            if ($page->post_title === 'Terms') {
```
comzeradd, line 59: The "Terms" page is not a special page, it's just a page using `page-templates/evergreen.php`. So NROs will probably have other pages using that template.

I think the solution you are applying here should be applied on all pages of the second scenario (when we can't use the Page Header block template). So in practice we use `page_header_title` to replace the `post_title` and not add a heading block at all. That way we won't need to adjust any top margin css code.

src/Migrations/M028MovePageHeaderSideBarOptions.php
```diff
@@ -0,0 +1,236 @@
...
+                    && empty($page_header_button_title)
+                ) {
+                    continue;
+                }
+
+                $content = '
+                    <!-- wp:group {"layout":{"type":"flex","orientation":"vertical"}} -->
+                    <div class="wp-block-group">
+                    ' . (!empty($page_header_description) ?
+                            '<!-- wp:paragraph -->
```
comzeradd, line 151: On the test instance (haven't test this locally yet) I see the whole group block being "under recovery" in the editor (eg. in the "Terms" page). When trying to recover it fails and the content is lost.

Screenshot from 2024-03-06 15-13-40


Maybe it's because of the double `<p>` markup? I think if you just have `<!-- wp:paragraph -->` you may not need to add a `<p>` as well. I tried re-creating the group block with a paragraph block. This how the two markups look on the code editor (the one on the top is the one from this migration that is broken in the visual editor).

Screenshot from 2024-03-06 15-13-40

Osong-Michael: Oh I see, but what if the description does not have the `<p>`,  will that still render correctly?

Maybe we could strip the outer `p` tag in the description if it has one so that way there are no duplicates.

comzeradd: Ah yes, that sounds like a much safer option if that extra `<p>` is coming from the description field. The field is not WYSIWYG so we could strip html tags in general.

lithrel (approved):
:+1:

PR #2217, opened 2024-02-15
Make pre-publish error message more prominent

comzeradd:
Used the same color options as [forms validation messages](https://github.com/greenpeace/planet4-master-theme/blob/v1.281.0/assets/src/scss/layout/_gravity-forms.scss#L106-L107).

### Testing
To view the message try two use cases:
1. Create a new page, add a Carousel Header and try to publish
2. Create a new post, leave the title empty and try to publish


lithrel (approved):
:+1:

PR #2223, opened 2024-03-04
PLANET-7344 Make featured image mandatory for publishing

mleray:
### Description

See [PLANET-7344](https://jira.greenpeace.org/browse/PLANET-7344)

I didn't add a WP alert similar to what we do [here](https://github.com/greenpeace/planet4-master-theme/blob/main/src/MasterSite.php#L430) for the title, because I couldn't find a way to properly check whether or not a featured image was set at this point, so the alert was always shown no matter what. From what I've tested the check from the plugin is enough to disable the update/publish button when no image is set, only issue then is that there is no explanation when it's an update but that's also the case for the title 🤷‍♀️ 

**Related PR:** https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/1190

### Testing

You should no longer be able to publish or update a Page/Post/Action without setting a featured image or adding an image in the content.

### DIFF
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -442,17 +442,15 @@ class MasterSite extends TimberSite
      */
     public static function require_post_featured_image(array $data, array $postarr): ?array
     {
-        $thumbnail = (get_the_post_thumbnail() ?: null)
-            ?? $postarr['meta_input']['_thumbnail_id']
-            ?? null;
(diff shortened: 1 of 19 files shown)


planet-4:

:watch: 2024.03.11 14:18:21

lithrel (approved):
:+1:

PR #2231, opened 2024-03-20
PLANET-7420 Actions List: make the whole card clickable

Osong-Michael:
**Ref: [PLANET-7420](https://jira.greenpeace.org/browse/PLANET-7420)**

**Testing:** 

Actions List cards are now clickable and redirect to page/post

### DIFF
--- a/assets/src/js/actions_list_clickable_cards.js
+++ b/assets/src/js/actions_list_clickable_cards.js
@@ -1,0 +1,20 @@
+export const setupClickabelActionsListCards = () => {
+  const liElements = document.querySelectorAll('.actions-list ul li:not(.carousel-li)');
+
+  liElements.forEach(li => {
+    const linkElement = li.querySelector('.wp-block-read-more');
+    if (linkElement) {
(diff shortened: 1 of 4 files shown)


mleray:
I've tested it on the instance, I'm wondering how this does in terms of accessibility since it's not technically a link 🤔  If I hover on a link, I see the url appear in the bottom left corner of my screen, indicating that I will navigate to this page. But with the current implementation this doesn't happen, which might be confusing? If I use tabs, only the image/title/button are links and therefore selectable, not the whole card.

You can compare the 2 behaviours with the Covers block on [this page](https://www-dev.greenpeace.org/test-nix/covers-block/) for example and the Actions List on [this one](https://www-dev.greenpeace.org/test-nix/actions-lists/).

Osong-Michael:
> this one

I wasn't sure either, but I figured since this was a beta block and not the final implementation, we will need to i

sagarsdeshmukh (approved):
PR looks good to me 👍

PR #2237, opened 2024-03-26
PLANET-7430: Elastic search not returning correct results

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-7430

### Sort by date doesn't work on ajax request
- Wrongly applied sanitize+decode on query string to obtain sort field

Taking the filter first and then sanitizing it works.

### First request and ajax request have different structures
- Ajax requests go through `admin-ajax.php`, which triggers a different behavior from ElasticPress

Setting filter [`ep_is_integrated_request`](https://10up.github.io/ElasticPress/ep_is_integrated_request.html) to `true` fixes most behaviors, including the usual double nested `query.function_score.query.function_score`.

## Test

Use the query log page of ElasticPress to debug queries (`wp-admin/admin.php?page=ep-query-log`)
"Search, public" query body and "Search, ajax" query body should have the same structure (apart from aggregation data).

Using `international` DB:
- searching for "waste colonialism", the article "#PlasticsTreaty" should appear in second page
- sorting results by "Newest" should work properly, even with "Load more" results

### DIFF
--- a/src/ElasticSearch.php
+++ b/src/ElasticSearch.php
@@ -3,7 +3,6 @@
 namespace P4\MasterTheme;
 
 use UnexpectedValueException;
-use ElasticPress\ElasticSearch as ElasticPressCient;
 
 /**
  * Class P4\MasterTheme\ElasticSearch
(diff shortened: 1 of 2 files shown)


mleray (approved):
Tested on the instance, seems to fix the reported issues 👍

PR #2251, opened 2024-04-09
PLANET-7492 Make sure blocks are not registered twice

mleray:
### Description

See [PLANET-7492](https://jira.greenpeace.org/browse/PLANET-7492)
The blocks registered via the `editorIndex` file were sometimes registered twice:

Screenshot 2024-04-09 at 15 20 01

### Testing

The console warnings should no longer be there when opening the editor. There is still one for a block pattern being registered twice, but Osong is looking into it since he's working on moving another pattern at the moment.

### DIFF
--- a/assets/src/blocks/Media/MediaBlock.js
+++ b/assets/src/blocks/Media/MediaBlock.js
@@ -17,7 +17,13 @@ export const lacksAttributes = attributes => {
 };
 
 export const registerMediaBlock = () => {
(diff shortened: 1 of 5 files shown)


lithrel (approved):
:+1: 

We're going to need the same for [block-templates](https://github.com/greenpeace/planet4-master-theme/blob/main/assets/src/block-templates/register.js#L35-L38), maybe we could make a reusable function for that, with block name as arg

PR #2254, opened 2024-04-12
PLANET-7458 Fix Actions List and Posts List carousel layout in RTL sites

mleray:
### Description

See [PLANET-7458](https://jira.greenpeace.org/browse/PLANET-7458)
It seems removing the `flex` display does the trick

### Testing

Add a Posts List or Actions List to a page with the carousel layout, then in the frontend you can manually add `dir="rtl"` to the html element. On `main` branch the block should be blank but on this branch it should display as expected.

On the janus instance I've created [this page](https://www-dev.greenpeace.org/test-janus/the-list-blocks/) with both blocks that you can check as well.


lithrel (approved):
:+1:

PR #2261, opened 2024-04-18
Add setting for custom Tag Manager scripts URL

stduerre:
Ref: https://github.com/greenpeace/planet4/issues/187

Adds an additional setting in the Analytics section of P4 settings for a custom domain name to load Tag Manager scripts from. The domain is then used in two places where the tag manager scripts are loaded. When the setting is left empty, the default domain of Tag Manager is used.

### DIFF
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -655,6 +655,8 @@ class MasterSite extends TimberSite
         // or enforce_cookies_policy setting is not enabled.
         $context['enforce_cookies_policy'] = isset($options['enforce_cookies_policy']) ? true : false;
         $context['google_tag_value'] = $options['google_tag_manager_identifier'] ?? '';
+        $context['google_tag_domain'] = !empty($options['google_tag_manager_domain']) ?
+            $options['google_tag_manager_domain'] : 'www.googletagmanager.com';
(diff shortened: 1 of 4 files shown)


src/MasterSite.php
```diff
@@ -655,6 +655,7 @@ function ($item) {
         // or enforce_cookies_policy setting is not enabled.
         $context['enforce_cookies_policy'] = isset($options['enforce_cookies_policy']) ? true : false;
         $context['google_tag_value'] = $options['google_tag_manager_identifier'] ?? '';
+        $context['google_tag_domain'] = !empty($options['google_tag_manager_domain']) ? $options['google_tag_manager_domain'] : 'www.googletagmanager.com';
```
comzeradd, line 658: Can you break up this line for readability? `phpcs` linting tests fail:

```
Line exceeds maximum limit of 120 characters; contains 155 characters
```

comzeradd (changes requested):
PR looks good 👍🏻 

It just needs a minor fix to make `phpcs` happy.

src/MasterSite.php
stduerre: Done!

comzeradd (approved):
☕

PR #2262, opened 2024-04-19
PLANET-7199 Fix Image block caption/credits

mleray:
### Description

See [PLANET-7199](https://jira.greenpeace.org/browse/PLANET-7199)
They would sometimes appear in two separate lines, because the previous implementation added a new `figcaption` element with the credits

### Testing

You can compare the possible caption/credits combinations on these 2 pages:
- [Broken](https://www-dev.greenpeace.org/test-rhea/images/)
- [Fixed](https://www-dev.greenpeace.org/test-jupiter/images/)

### DIFF
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -1504,25 +1504,19 @@ class MasterSite extends TimberSite
             $image_credit = ' ©' . $image_credit;
         }
 
-        $caption_open = '<figcaption>';
-        $caption_close = '</figcaption>';
-        $caption = explode($caption_open, $content, 2)[1] ?? '';
+        $caption = wp_get_attachment_caption($image_id) ?? '';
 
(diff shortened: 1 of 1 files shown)


src/MasterSite.php
```diff
@@ -1504,25 +1504,19 @@ public function p4_core_image_block_render(array $attributes, string $content):
...
-        } else {
-            $content = str_replace(
-                '</figure>',
-                $caption_open . esc_attr($image_credit) . $caption_close . '</figure>',
-                $content
-            );
-        }
-
-        return $content;
+        return str_replace(
```
mleray, line 1513: I'm not 100% sure about what to do when there is no caption but there are credits, currently we show the credits in the frontend but not in the editor... Hopefully during UAT Taylor will let us know but for now I've replicated our implementation

PR #2263, opened 2024-04-19
Fix p4 variables from Cookies

GP-Dan-Tovbein:
No ticket needed

# Description
The issue is raised whenever the P4 Gutenberg plugin is deactivated.
- Also force to set as `Boolean` instead of `undefined`, for a better typing.

## Testing
- Just run the site with the p4 Gutenberg plugin is deactivated

### DIFF
--- a/assets/src/js/cookies.js
+++ b/assets/src/js/cookies.js
@@ -1,8 +1,8 @@
 /* global dataLayer */
 export const setupCookies = () => {
   window.dataLayer = window.dataLayer || [];
-  const ENABLE_ANALYTICAL_COOKIES = window.p4bk_vars.enable_analytical_cookies;
-  const ENABLE_GOOGLE_CONSENT_MODE = window.p4bk_vars.enable_google_consent_mode;
+  const ENABLE_ANALYTICAL_COOKIES = window.p4_vars.enable_analytical_cookies || false;
+  const ENABLE_GOOGLE_CONSENT_MODE = window.p4_vars.enable_google_consent_mode || false;
 
   // Possible cookie values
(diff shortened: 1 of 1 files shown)


assets/src/js/cookies.js
```diff
@@ -1,8 +1,8 @@
 /* global dataLayer */
 export const setupCookies = () => {
   window.dataLayer = window.dataLayer || [];
-  const ENABLE_ANALYTICAL_COOKIES = window.p4bk_vars.enable_analytical_cookies;
-  const ENABLE_GOOGLE_CONSENT_MODE = window.p4bk_vars.enable_google_consent_mode;
+  const ENABLE_ANALYTICAL_COOKIES = window.p4_vars.enable_analytical_cookies || false;
```
mleray, line 4: What happens if the `Activate all blocks` feature is disabled? I think we don't set these variables then 🤔

GP-Dan-Tovbein: Gotcha! 
But my point is whenever we finish the moving from plugin to the theme, I assume we're not going to use `p4bk_vars` anymore. Am I missing something else?

planet-4:

:watch: 2024.04.19 10:14:46

assets/src/js/cookies.js
mleray: Indeed we should not use them anymore then

GP-Dan-Tovbein: But we'll need to replace and use `p4_vars` instead, am I correct?

GP-Dan-Tovbein:
Follow up this PR after we move all blocks from the plugin to the them.

PR #2272, opened 2024-04-25
PLANET-7503: Fix selector to exclude submenu header in submenu block

stduerre:
The h2 of the submenu block has ancestors that match :not(.submenu-block), that's why it was included in the results. By only looking for direct ancestors that match :not(.submenu-block) it is excluded as intended.

Ref: https://jira.greenpeace.org/browse/PLANET-7503


GP-Dan-Tovbein (approved):
Code looks good to me! 
Tested locally and seems to be working as expected 👏 

Thanks @stduerre

PR #2282, opened 2024-05-06
PLANET-7263 Media Archive: add in editor option for P4 blocks

Osong-Michael:
Ref: https://jira.greenpeace.org/browse/PLANET-7263

**Testing:** 

- Added option to use _Greenpeace Media_ to upload images for the `core/media-text` block.
- All blocks using `core/image` or `core/media-text` for images now have access to the Greenpeace Media
- `planet4-blocks/carousel-header` also has access to the Greenpeace Media.

### DIFF
--- a/admin/css/archive-picker.css
+++ b/admin/css/archive-picker.css
@@ -332,6 +332,7 @@
   position: absolute;
   top: 0;
   right: 0;
+  background-color: transparent;
 }
 
 #archive-picker-root .picker-sidebar-header .close-sidebar {
@@ -340,7 +341,7 @@
 
 .media-modal-content .picker-sidebar-header .close-sidebar {
   right: 35px;
-  background: url("../../images/arrow-right.svg");
+  background-image: url("../../images/arrow-right.svg");
 }
 
 .multiple-search-item .delete-icon {
--- a/assets/src/blocks/CarouselHeader/CarouselHeaderBlock.js
+++ b/assets/src/blocks/CarouselHeader/CarouselHeaderBlock.js
@@ -36,6 +36,10 @@ const attributes = {
       return {isValid, messages};
     },
   },
+  currentImageIndex: {
+    type: 'integer',
+    default: 0,
+  },
 };
 
 export const registerCarouselHeaderBlock = () =>
--- a/assets/src/js/Components/ArchivePicker/ArchivePicker.js
+++ b/assets/src/js/Components/ArchivePicker/ArchivePicker.js
@@ -5,7 +5,12 @@ import ArchivePickerList from './ArchivePickerList';
 import SingleSidebar from './SingleSidebar';
 import MultiSidebar from './MultiSidebar';
 import ArchivePickerToolbar from './ArchivePickerToolbar';
-import {updateImageBlockAttributes, updateHappyPointAttributes} from './blockUpdateFunctions';
+import {
+  updateImageBlockAttributes,
+  updateHappyPointAttributes,
+  updateMediaAndTextAttributes,
+  updateCarouselBlockAttributes,
+} from './blockUpdateFunctions';
 
 const {apiFetch, url, i18n} = wp;
 const {addQueryArgs} = url;
@@ -13,7 +18,7 @@ const {__} = i18n;
 const timeout = delay => {
   return new Promise(resolve => setTimeout(resolve, delay));
 };
-const acceptedBlockTypes = ['core/image', 'planet4-blocks/happypoint'];
+const acceptedBlockTypes = ['core/image', 'planet4-blocks/happypoint', 'core/media-text', 'planet4-blocks/carousel-header'];
 
 export const EDITOR_VIEW = 'editor';
 export const ADMIN_VIEW = 'admin';
@@ -293,18 +298,23 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
     }
   }, []);
 
+  const processImageForBlock = async (imageID, updateAttributeFunction) => {
+    const uploadedImage = await getImageDetails(imageID);
+    const updatedAttributes = updateAttributeFunction(uploadedImage, currentBlock);
+    await wp.data.dispatch('core/block-editor').updateBlock(currentBlock.clientId, updatedAttributes);
+  };
+
   const processImageToAddToEditor = async id => {
     dispatch({type: ACTIONS.ADD_IMAGE_TO_POST});
     try {
       if (currentBlock.name === 'core/image') {
-        // Get current Image details from WP
-        const uploadedImage = await getImageDetails(id);
-        // Get Updated Atrributes
-        const updatedAttributes = updateImageBlockAttributes(uploadedImage, currentBlock);
-        await wp.data.dispatch('core/block-editor').updateBlock(currentBlock.clientId, updatedAttributes);
-      }
-
-      if (currentBlock.name === 'planet4-blocks/happypoint') {
+        await processImageForBlock(id, updateImageBlockAttributes);
+      } else if (currentBlock.name === 'planet4-blocks/carousel-header') {
+        await processImageForBlock(id, updateCarouselBlockAttributes);
+      } else if (currentBlock.name === 'core/media-text') {
+        await processImageForBlock(id, updateMediaAndTextAttributes);
+      } else {
+        // Happypoint Block
         const updatedAttributes = updateHappyPointAttributes(id);
         await wp.data.dispatch('core/block-editor').updateBlock(currentBlock.clientId, updatedAttributes);
       }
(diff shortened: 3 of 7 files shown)


planet-4:

:watch: 2024.05.06 15:42:27

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -309,6 +314,18 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
         await wp.data.dispatch('core/block-editor').updateBlock(currentBlock.clientId, updatedAttributes);
       }
 
+      if (currentBlock.name === 'core/media-text') {
```
GP-Dan-Tovbein, line 317: I noticed both if-else are basically doing the same. Then I think we can do this 👇 instead
```suggestion
      if (currentBlock.name === 'core/media-text' || || currentBlock.name === 'planet4-blocks/carousel-header') {
        const uploadedImage = await getImageDetails(id);
        const updatedAttributes = updateCarouselBlockAttributes(uploadedImage, currentBlock);
        await wp.data.dispatch('core/block-editor').updateBlock(currentBlock.clientId, updatedAttributes);
      }
      }
```

What do you think?

assets/src/js/Components/ArchivePicker/blockUpdateFunctions.js
```diff
@@ -39,3 +39,59 @@ export const updateHappyPointAttributes = id => {
...
+    div.innerHTML = currentBlock.originalContent;
+    div.querySelector('img').setAttribute('class', `wp-image-${image.id}`);
+    div.querySelector('img').src = image.source_url;
+  } else {
+    div.innerHTML = `
+    <div class="wp-block-media-text is-stacked-on-mobile">
+      <figure class="wp-block-media-text__media">
+        <img src=${image.source_url} alt=${image.alt_text} class="wp-image-${image.id} size-full"/>
+      </figure>
+      <div class="wp-block-media-text__content">
```
GP-Dan-Tovbein, line 56: Is it necessary this empty `div`?

assets/src/js/Components/ArchivePicker/blockUpdateFunctions.js
```diff
@@ -39,3 +39,59 @@ export const updateHappyPointAttributes = id => {
...
+      mediaUrl: image.source_url,
+    },
+    originalContent: div.innerHTML,
+  };
+
+  return updatedAttributes;
+};
+
+export const updateCarouselBlockAttributes = (image, currentBlock) => {
+  const imageArray = currentBlock.attributes.slides;
```
GP-Dan-Tovbein, line 74: Regarding naming convention, I think we can change to `slides` instead

assets/src/js/Components/ArchivePicker/blockUpdateFunctions.js
```diff
@@ -39,3 +39,59 @@ export const updateHappyPointAttributes = id => {
...
+  ${full.source_url} ${full.width}w,
+  `;
+
+  imageArray[currentImageIndex].image = image.id;
+  imageArray[currentImageIndex].image_url = image.source_url;
+  imageArray[currentImageIndex].image_srcset = imageSrcSet;
+
+  const updatedAttributes = {
+    attributes: {
+      slides: imageArray,
```
GP-Dan-Tovbein, line 91: And then just `slides,`

assets/src/js/Components/ArchivePicker/blockUpdateFunctions.js
```diff
@@ -39,3 +39,59 @@ export const updateHappyPointAttributes = id => {
...
+    },
+  };
+
+  return updatedAttributes;
```
GP-Dan-Tovbein, line 95: We can return directly the `attributes` instead of defining a `const`

assets/src/js/Components/ArchivePicker/ArchivePicker.js
Osong-Michael: Well they are not doing the same thing, if you look they both call different functions that updated the attributes differently.

assets/src/js/Components/ArchivePicker/blockUpdateFunctions.js
Osong-Michael: In my testing it was necessary to make sure we are passing the element properly. I use it to make sure the HTML is rendered properly and we can change specific attributes of the HTML element.

assets/src/js/Components/ArchivePicker/blockUpdateFunctions.js
Osong-Michael: This is for the dispatch method that updates the block, you have to use a certain structure to pass to the function.

assets/src/js/Components/ArchivePicker/ArchivePicker.js
mleray: It seems that at least two of these lines of code are repeated so maybe there's a way to remove that duplication?
```
const uploadedImage = await getImageDetails(id);
```
and
```
await wp.data.dispatch('core/block-editor').updateBlock(currentBlock.clientId, updatedAttributes);
```

assets/src/blocks/CarouselHeader/CarouselHeaderBlock.js
```diff
@@ -36,6 +36,10 @@ const attributes = {
       return {isValid, messages};
     },
   },
+  currentImageIndex: {
```
mleray, line 39: Don't we need to add this new attribute in the PHP definition of the block as well?

mleray (changes requested):
On the test instance it doesn't seem to work for the Carousel Header block? I've tried on the [homepage](https://www-dev.greenpeace.org/test-oberon/) to change one of the slide images and even though I can see the `Greenpeace Media` tab, I couldn't actually select an image from it to be used in the block but maybe I'm missing something 🤔 

I also noticed that there's this arrow instead of the 'close' button but maybe that's normal?
Screenshot 2024-05-21 at 16 15 23

Osong-Michael:
> admin

Oh for the image it was a design change requested by Taylor. 

About the functionality, I just realised I made the changes in the Theme repo where we moved the blocks, I suspect the functionality is being pulled from the blocks repo so I will update the block there too and link to this PR.

admin/css/archive-picker.css
```diff
@@ -341,6 +341,7 @@
 .media-modal-content .picker-sidebar-header .close-sidebar {
   right: 35px;
   background: url("../../images/arrow-right.svg");
+  background-repeat: no-repeat;
```
mleray, line 344: I'm aware that I'm nitpicking a bit, sorry about that 😅 but if you inspect in the editor you can see that the `background-repeat` property is actually already set somewhere else in the file, it's just overridden by `background`. So what I would prefer as a fix would be using the `background-image` and `background-color` properties, see below:

- Current:
Screenshot 2024-05-23 at 10 37 41

- Suggested:
Screenshot 2024-05-23 at 10 38 33

Let me know what you think 🙂

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -309,6 +314,18 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
...
+        const uploadedImage = await getImageDetails(id);
+        const updatedAttributes = updateMediaAndTextAttributes(uploadedImage, currentBlock);
+        await wp.data.dispatch('core/block-editor').updateBlock(currentBlock.clientId, updatedAttributes);
+      }
+
+      if (currentBlock.name === 'planet4-blocks/carousel-header') {
```
mleray, line 323: I think this should be a `else if` statement instead of just `if`

assets/src/js/Components/ArchivePicker/blockUpdateFunctions.js
mleray: I think what Dan meant is that you could directly write
```
return {
  attributes: {
    slides,
  },
};
```
instead of creating the `updatedAttributes` constant

assets/src/js/Components/ArchivePicker/ArchivePicker.js
mleray: It seems it would be possible to write something like:
```
const updatedAttributes = currentBlock.name === 'core/media-text' ? updateMediaAndTextAttributes(uploadedImage, currentBlock) : updateCarouselBlockAttributes(uploadedImage, currentBlock);
```

assets/src/js/Components/ArchivePicker/ArchivePicker.js
mleray: Note that this comment becomes irrelevant if the code is combined to remove duplications as suggested above

mleray: @Osong-Michael following the previous refactor, I think these should be `else if` statements instead of just `if`

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -293,22 +298,32 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
     }
   }, []);
 
+  const processImageForBlock = async (imageID, updateAttributeFunction) => {
```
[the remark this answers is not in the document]
Osong-Michael: Yes, but we would have to prevent the `getImageDetails()` method from running which in that case means we will need to run another check within the function just for the Happypoint block. So my thinking was run the check once and use the function for only blocks that follow same steps.

What do you think?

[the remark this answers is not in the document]
mleray: Yeah I see your point, makes sense 🙂

mleray:
I've done another round of tests on this [page](https://www-dev.greenpeace.org/test-oberon/about-us-2/), and I've been able to select an image from the Greenpeace Media for the Media & Text block in the editor, however in the frontend it still shows the old one and also if I try to click "Replace" the old one still shows as selected... Am I missing something? 
However for the other blocks it seems to work as expected 👍

Osong-Michael:
> page

checking that now...

mleray (approved):
Looks good to me 👍

PR #2297, opened 2024-05-29
Add filter hook to allow changing the formSubmission event parameters

stduerre:
Adds a filter hook in both places the formSubmission event is pushed to dataLayer. The filter enables the modification of the data that's sent with the event.

Ref: https://github.com/greenpeace/planet4/issues/188

For testing, make sure the formSubmission event is still correctly sent when a form is submitted.

### DIFF
--- a/src/GravityFormsExtensions.php
+++ b/src/GravityFormsExtensions.php
@@ -342,19 +342,28 @@ class GravityFormsExtensions
             }
         }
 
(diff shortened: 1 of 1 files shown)


sagarsdeshmukh (approved):
The PR tested on local, The "planet4_datalayer_form_submission" filter works as expected 💯

PR #2301, opened 2024-05-31
PLANET-7362 Write a Greenpeace Media e2e test for importing images

mleray:
### Description

See [PLANET-7362](https://jira.greenpeace.org/browse/PLANET-7362)

**Test Steps**

1. Login to the Admin Panel.
2. Navigate to Media > Greenpeace Media.
3. Search for a term (eg. "Rainbow Warrior").
4. Pick the first image and import to the Library.
5. Assert that after some time the Processing... message is replaced with a link to the imported image.
6. Search for an image identifier (eg. GP0STXWME) and assert there is one result displayed.
7. Append one more image identified (eg. GP0STXWZH) and assert that now are two images displayed.
8. Click on "Bulk Select" and select both images.
9. Click on "Bulk Upload".
10. Assert that after some time the Processing 2 images message is hidden and the search box is shown again.
11. Remove the imported images from the Library.

### DIFF
--- a/tests/e2e/greenpeace-media.spec.js
+++ b/tests/e2e/greenpeace-media.spec.js
@@ -5,8 +5,7 @@ import {
   waitForLibraryLoad,
 } from './tools/lib/media-library.js';
 
-const IMAGE_1 = 'GP0STXWME';
-const IMAGE_2 = 'GP0STXWZH';
+const TEST_IMAGES = ['GP0STXWME', 'GP0STXWZH'];
 const MEDIA_LIBRARY_PAGE = './wp-admin/admin.php?page=media-picker';
 
 test.useAdminLoggedIn();
@@ -49,13 +48,13 @@ test.describe('Greenpeace Media tests', () => {
 
     // Search for 2 specific images.
     const images = page.locator('.picker-list > li');
-    await searchImages(page, IMAGE_1);
-    await expect(images).toHaveCount(1);
-    await searchImages(page, IMAGE_2);
-    await expect(images).toHaveCount(2);
-    await page.getByRole('button', {name: 'Bulk Select'}).click();
+    for (let index = 0; index < 2; index++) {
+      await searchImages(page, TEST_IMAGES[index]);
+      await expect(images).toHaveCount(index + 1);
+    }
 
     // Select the 2 images.
+    await page.getByRole('button', {name: 'Bulk Select'}).click();
     for (let index = 0; index < 2; index++) {
       await images.nth(index).click();
     }
@@ -69,7 +68,9 @@ test.describe('Greenpeace Media tests', () => {
     // Delete the images from the library.
     await page.reload();
     await waitForLibraryLoad(page);
-    await deleteImageFromLibrary(page, IMAGE_1);
-    await deleteImageFromLibrary(page, IMAGE_2);
+    await deleteImageFromLibrary(page, TEST_IMAGES[0]);
+    await page.goto(MEDIA_LIBRARY_PAGE);
+    await waitForLibraryLoad(page);
+    await deleteImageFromLibrary(page, TEST_IMAGES[1]);
   });
 });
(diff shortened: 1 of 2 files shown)


tests/e2e/greenpeace-media.spec.js
```diff
@@ -0,0 +1,75 @@
...
+    // Delete the image from the library.
+    await deleteImageFromLibrary(page);
+  });
+
+  test('bulk select and upload two images using identifiers', async ({page}) => {
+    // Make sure to close any alerts.
+    page.on('dialog', dialog => dialog.accept());
+
+    // Go to the Media Library and wait for the page to load.
+    await page.goto('./wp-admin/admin.php?page=media-picker');
```
mardelnet, line 46: Maybe `./wp-admin/admin.php?page=media-picker` can be also a constant value, since we are using it twice.

tests/e2e/greenpeace-media.spec.js
```diff
@@ -0,0 +1,75 @@
...
+    await expect(image).toHaveClass('is-selected');
+    const sidebar = page.locator('.archive-picker-sidebar');
+    await expect(sidebar).toBeVisible();
+
+    // Import the image.
+    await sidebar.getByRole('button', {name: 'Import to Library'}).click();
+    await expect(sidebar.getByText('Processing...')).toBeVisible();
+    await expect(sidebar.getByText('Added to Library')).toBeVisible();
+
+    // Delete the image from the library.
```
mardelnet, line 37: Shouldn't we check that the image also exists in `/wp-admin/upload.php` before deleting it?

tests/e2e/tools/lib/media-library.js
```diff
@@ -0,0 +1,59 @@
+import {expect} from './test-utils.js';
+
+/**
+ * Searches for images in the Library based on a search term or image id.
+ *
+ * @param {Object} page       - The page object for interacting with the browser.
+ * @param {string} searchTerm - The term to be searched, can be an image id.
+ */
+async function searchImages(page, searchTerm) {
```
mardelnet, line 9: What if no images are found?
Maybe this could be a problem on line 31 when trying to delete an image.

tests/e2e/tools/lib/media-library.js
```diff
@@ -0,0 +1,59 @@
...
+  const sidebar = page.locator('.archive-picker-sidebar');
+  if (imageId) {
+    // Search for the image.
+    await searchImages(page, imageId);
+    const image = page.locator('.picker-list > li');
+
+    // Select the image and open its link via the sidebar.
+    await image.click();
+    await expect(sidebar).toBeVisible();
+  }
```
mardelnet, line 33: What about using a guard clause here?
If no `imageId` is provided, I think we should not execute anything else.

```suggestion
  if (!imageId) return;
  
  // Search for the image.
    await searchImages(page, imageId);
    const image = page.locator('.picker-list > li');

    // Select the image and open its link via the sidebar.
    await image.click();
    await expect(sidebar).toBeVisible();
```

mardelnet (commented):
Good job, @mleray 
I have some comments for you:

tests/e2e/tools/lib/media-library.js
mleray: Actually with the way it is implemented and used now, we should still execute the following lines, because in the first test the image is already selected hence why we don't provide the image id and don't need to search for it... I hope that makes sense 😅

tests/e2e/greenpeace-media.spec.js
mleray: Actually for deleting an image, I had another possible implementation in mind, in which we could get the WordPress image id (from the data attributes) and go directly to the associated media page to delete it. I didn't go for it because it felt less "end to end" but what do you think?

tests/e2e/tools/lib/media-library.js
mardelnet: It makes sense!
In that case, should we turn `imageId` into an optional parameter and explain in the JSDocs that in some cases passing it is not necessary?

tests/e2e/greenpeace-media.spec.js
mardelnet: I like the idea :)

mardelnet (approved):
Great job, @mleray

PR #2304, opened 2024-06-05
PLANET-7381 Hide Page title by default when a pattern layout is being used

Osong-Michael:
**Ref: https://jira.greenpeace.org/browse/PLANET-7381**

**Testing:**

When creating a new page using a pattern layout, the `Hide Page Title` option is enabled when the page is being published.


functions.php
```diff
@@ -490,3 +490,32 @@ function ($canonical_url, $post) {
...
+add_action(
+    'transition_post_status',
+    function ($new_status, $old_status, $post): void {
+        // Check if the new status is 'publish' and the old status is not 'publish'
+        if ($new_status !== 'publish' || $old_status === 'publish') {
+            return;
+        }
+
+        // Check if the post type is 'page'
+        if ($post->post_type !== 'page') {
```
mleray, line 504: This `if` statement could be merged with the previous one

mleray: The ticket also specifies that `Besides pages, this should also be applied in Actions and Campaigns.` so I guess a more accurate check would be that it's not a Post?

Osong-Michael:
> This implementation updates the setting on publish and not on page creation as the ticket says, is that because it wasn't possible on creation?

The issue with page creation is once the new page auto saves, the hooks to check if the page is new all return `false`, but since the `publish` status is not updated till after the page has been published, it is easier to update the settings during that stage.

mleray (approved):
Tested on the instance, looks good to me 👍

functions.php
```diff
@@ -490,3 +490,28 @@ function ($canonical_url, $post) {
         return $options;
     });
 }
+
+// Enable Hide Page title by default when Pattern Layout is used
+add_action(
+    'transition_post_status',
+    function ($new_status, $old_status, $post): void {
+        // Check if the new status is 'publish' and the old status is not 'publish'
+        // And it is a Page
```
mleray, line 499: This comment should be updated

mleray:
Maybe we should add a comment in the ticket to explain that instead of page creation this will be applied on page publishing? Could be confusing otherwise 🤔

PR #2308, opened 2024-06-13
Generate and save a hashed gp_user_id on form submissions

stduerre:
The hashed `gp_user_id` is saved to dataLayer and a cookie also called `gp_user_id`. In order to respect user consent and flexibility for different privacy laws, additional settings for default values for Google Consent Mode are also added.

Ref: [Please add a url to the ticket this change is addressing.
](https://github.com/greenpeace/planet4/issues/189)

Instructions for testing:
- It's best to delete all cookies and not click anything in the cookie box before starting.
- In WP Admin go to _Planet4_ -> _Cookies_ and make sure _Enable Google Consent Mode_ is activated.
- Test the 4 following settings if the default values are set correctly in dataLayer (note that these can be overwritten from Tag Manager, so it's best to block it for this part of the testing)
- Submit a form that contains an email address in a field
- If `analytics_storage` is set to `granted` in dataLayer, a cookie should be set containing the hashed email address and dataLayer should also contain the value in a parameter called `gpUserId`
- The hash needs to be the same as [generated by this function](https://gist.github.com/stduerre/f98ed2cd4ac540da1c3389b674a64b90).

Variations:
- It should work with both an email or any other form field (most likely normal text input) in the form, as long as the email address is the only content of the form field. The email field is preferred over other fields.
- `analytics_storage` can be changed by the selection in the cookie box (depending on settings). The current value needs to be respected when saving the `gp_user_id` cookie.

### DIFF
--- a/assets/src/js/cookies.js
+++ b/assets/src/js/cookies.js
@@ -10,7 +10,7 @@ export const setupCookies = () => {
   const NECESSARY_ANALYTICAL = '3';
   const NECESSARY_ANALYTICAL_MARKETING = '4';
 
-  const ALL_COOKIES = ENABLE_ANALYTICAL_COOKIES ? NECESSARY_ANALYTICAL_MARKETING : NECESSARY_MARKETING;
+  const ALL_COOKIES = NECESSARY_ANALYTICAL_MARKETING;
 
   function gtag() {
     dataLayer.push(arguments);
@@ -95,7 +95,7 @@ export const setupCookies = () => {
         ad_storage: 'granted',
         ad_user_data: 'granted',
         ad_personalization: 'granted',
-        ...ENABLE_ANALYTICAL_COOKIES && {analytics_storage: 'granted'},
+        analytics_storage: 'granted',
       });
     }
 
@@ -156,12 +156,20 @@ export const setupCookies = () => {
 
       // Update ad storage and analytics storage if Google Consent Mode is enabled
       if (ENABLE_GOOGLE_CONSENT_MODE) {
-        updateGoogleConsent({
-          ad_storage: marketingCookiesChecked ? 'granted' : 'denied',
-          ad_user_data: marketingCookiesChecked ? 'granted' : 'denied',
-          ad_personalization: marketingCookiesChecked ? 'granted' : 'denied',
-          ...ENABLE_ANALYTICAL_COOKIES && {analytics_storage: analyticalCookiesChecked ? 'granted' : 'denied'},
-        });
+        if (ENABLE_ANALYTICAL_COOKIES) {
+          updateGoogleConsent({
+            ad_storage: marketingCookiesChecked ? 'granted' : 'denied',
+            ad_user_data: marketingCookiesChecked ? 'granted' : 'denied',
+            ad_personalization: marketingCookiesChecked ? 'granted' : 'denied',
+            analytics_storage: analyticalCookiesChecked ? 'granted' : 'denied',
+          });
+        } else {
+          updateGoogleConsent({
+            ad_storage: marketingCookiesChecked ? 'granted' : 'denied',
+            ad_user_data: marketingCookiesChecked ? 'granted' : 'denied',
+            ad_personalization: marketingCookiesChecked ? 'granted' : 'denied',
+          });
+        }
       }
 
       hideCookiesBox();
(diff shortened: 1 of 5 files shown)


assets/src/js/cookies.js
```diff
@@ -156,12 +156,21 @@ export const setupCookies = () => {
 
       // Update ad storage and analytics storage if Google Consent Mode is enabled
       if (ENABLE_GOOGLE_CONSENT_MODE) {
-        updateGoogleConsent({
-          ad_storage: marketingCookiesChecked ? 'granted' : 'denied',
-          ad_user_data: marketingCookiesChecked ? 'granted' : 'denied',
-          ad_personalization: marketingCookiesChecked ? 'granted' : 'denied',
-          ...ENABLE_ANALYTICAL_COOKIES && {analytics_storage: analyticalCookiesChecked ? 'granted' : 'denied'},
-        });
+        if (ENABLE_ANALYTICAL_COOKIES) {
```
stduerre, line 159: When analytical cookies functionality is disabled, we simply don't update `analytics_storage` and it will stay on the default value that was set before the cookie box was clicked. Before the change, it was always updated to `denied` if the functionality is disabled.

assets/src/js/cookies.js
```diff
@@ -10,7 +10,7 @@ export const setupCookies = () => {
   const NECESSARY_ANALYTICAL = '3';
   const NECESSARY_ANALYTICAL_MARKETING = '4';
 
-  const ALL_COOKIES = ENABLE_ANALYTICAL_COOKIES ? NECESSARY_ANALYTICAL_MARKETING : NECESSARY_MARKETING;
```
sagarsdeshmukh, line 13: Hi Stefan, Could you please elaborate on why we dropped this condition? If a user accepts all cookies and ENABLE_ANALYTICAL_COOKIES is unset from the admin settings, the Greenpeace cookie was set to 2 (i.e., NECESSARY_MARKETING), but now it's set to 4 (i.e., NECESSARY_ANALYTICAL_MARKETING).

assets/src/js/cookies.js
```diff
@@ -95,7 +95,7 @@ export const setupCookies = () => {
         ad_storage: 'granted',
         ad_user_data: 'granted',
         ad_personalization: 'granted',
-        ...ENABLE_ANALYTICAL_COOKIES && {analytics_storage: 'granted'},
```
sagarsdeshmukh, line 98: Same here, 
image
The analytics _storage set granted by default will disrespect the above admin setting.

assets/src/js/cookies.js
```diff
@@ -156,12 +156,20 @@ export const setupCookies = () => {
 
       // Update ad storage and analytics storage if Google Consent Mode is enabled
       if (ENABLE_GOOGLE_CONSENT_MODE) {
-        updateGoogleConsent({
```
sagarsdeshmukh, line 159: Don't both code snippets produce the same results? 🤔

assets/src/js/cookies.js
stduerre: See comment for line 13, it's the same situation. Disabling the setting for analytics cookies in the cookies box does not mean they are not allowed to use. Those two things are decoupled now

assets/src/js/cookies.js
sagarsdeshmukh: Thanks for the detailed info, I got your point 👍

assets/src/js/cookies.js
stduerre: No, because there's a difference between the value for `analytics_storage` not being set at all or `denied`. Not set means the user didn't make a choice, `denied` means they actively denied.
The old code assumed `denied` even if the user wasn't presented with a choice (ENABLE_ANALYTICAL_COOKIES  ==  false). The revised code will not set it if no choice was made.

Let's take Switzerland as an example. We can load analytics with no active consent, but we might want to give people an opportunity to opt out. So for us, the setup in tag manager would look like this:
- `analytics_storage` == `denied`: don't load analytics
- `analytics_storage` == `granted`: load analytics
- **`analytics_storage`  is not set: load analytics**

That's different from a country that needs active consent, they would not load analytics if `analytics_storage` is not set.

Basically, there are 3 possible values, not 2: `granted`, `denied` and NOT SET.

sagarsdeshmukh (approved):
The PR has been tested on the [test-titan](https://www-dev.greenpeace.org/test-titan/test-gravity-form-2/) instance and everything looks great. Excellent work 🥇 , Stefan!

stduerre:
Is there anything I need to do about the failing Sonar Cloud Code Analysis? The code duplication it detects is the configuration array for the WordPress admin settings. Not sure how that could be done in a better way.

PR #2309, opened 2024-06-14
PLANET-7531 Posts/Actions blocks: Show message when no posts found

Osong-Michael:
**Ref: https://jira.greenpeace.org/browse/PLANET-7531**

**Testing:** 

I added the `core/query-no-results` block to the Posts Lists and Actions List blocks. This gives Editors the freedom to add text or other blocks to display in case the category used to filter the posts returns no posts.

I also added a function to run in the front end to remove the text or block added by the `core/query-no-results` in case the title and description of the Posts/Actions List block are not set.

You can test this locally or on the assigned instance.

### DIFF
--- a/assets/src/blocks/ActionsList/index.js
+++ b/assets/src/blocks/ActionsList/index.js
@@ -61,6 +61,9 @@ export const registerActionsListBlock = () => {
           },
         },
       }],
+      ['core/query-no-results', {}, [
+        ['core/paragraph', {content: __('No posts found. (This default text can be edited)', 'planet4-blocks-backend')}],
+      ]],
       ['core/post-template', {lock: {move: true, remove: true}}, [
         ['core/post-featured-image', {isLink: true}],
         ['core/group', {}, [
--- a/assets/src/js/app.js
+++ b/assets/src/js/app.js
@@ -9,6 +9,7 @@ import {setupExternalLinks} from './external_links';
 import {setupListingPages} from './listing_pages';
 import {setupQueryLoopCarousel} from './query_loop_carousel';
 import {setupClickabelActionsListCards} from './actions_list_clickable_cards';
+import {removeNoPostText} from './query-no-posts';
 
 function requireAll(r) {
   r.keys().forEach(r);
@@ -24,4 +25,5 @@ setupSearch();
 setupExternalLinks();
 setupListingPages();
 setupQueryLoopCarousel();
+removeNoPostText();
 setupClickabelActionsListCards();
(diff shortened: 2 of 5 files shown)


planet-4:

:watch: 2024.06.14 12:21:45

assets/src/js/app.js
```diff
@@ -9,6 +9,7 @@ import {setupExternalLinks} from './external_links';
 import {setupListingPages} from './listing_pages';
 import {setupQueryLoopCarousel} from './query_loop_carousel';
 import {setupClickabelActionsListCards} from './actions_list_clickable_cards';
+import {removeNoPostText} from './query-no-posts';
```
GP-Dan-Tovbein, line 12: What do you think if we move this function into [the listing pages setup file](https://github.com/greenpeace/planet4-master-theme/blob/15a9ccf5fdb89933dbda704fb081d2e3337ab786/assets/src/js/listing_pages.js)?

GP-Dan-Tovbein: Oh sorry! I shared the wrong file 🙏  I meant into the [Query Loop extension](https://github.com/greenpeace/planet4-master-theme/blob/c1c59752ca53594edb804b59d29a0327bac89b83/assets/src/block-editor/QueryLoopBlockExtension/index.js).
Do you think it's better?

GP-Dan-Tovbein (commented):
Code looks good to me and works as expected!
However, I've left a new comment and I'm now looking forward to your feedback!

If you think is a good change, I will wait for new updates. If no, I will accept and then merge it!

assets/src/js/app.js
Osong-Michael: Oh yea great suggestion, will update the code asap

assets/src/blocks/ActionsList/index.js
```diff
@@ -61,6 +61,9 @@ export const registerActionsListBlock = () => {
           },
         },
       }],
+      ['core/query-no-results', {}, [
+        ['core/paragraph', {content: __('No posts found.(This default text can be edited)', 'planet4-blocks-backend')}],
```
comzeradd, line 65: Minor tweak: can you add a space before the opening parenthesis?

assets/src/js/app.js
GP-Dan-Tovbein: Got it! Another suggestion and looking forward your point of view.
I was thinking something like this 👇 

```
register_block_type( 'core/query', array(
    'render_callback' => function (array $attributes, $content, $block) {
       // Here override the rendered content if the data is empty 
        return $content;
    },
) );
```

Does it make sense for you?

Osong-Michael: It may work 🤔, but what about the case where editors may want to render some custom text or block since they know that is available with the core block, how do we handle that?

comzeradd (approved):

PR #2319, opened 2024-07-11
PLANET-7525 Media Import does not work with Featured Image

Osong-Michael:
**Ref: https://jira.greenpeace.org/browse/PLANET-7525**

**Testing:**

Adding Featured Image should now work with Greenpeace Media.

### DIFF
--- a/assets/src/js/Components/ArchivePicker/ArchivePicker.js
+++ b/assets/src/js/Components/ArchivePicker/ArchivePicker.js
@@ -253,6 +253,7 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
   }, dispatch] = useReducer(reducer, initialState);
   const [abortController, setAbortController] = useState(null);
   const [selectedImagesIds, setSelectedImagesIds] = useState([]);
+  const [isUserSelectingFeaturedImage, setIsUserSelectingFeaturedImage] = useState(false);
 
   const fetch = useCallback(async () => {
     dispatch({type: ACTIONS.FETCH_IMAGES});
@@ -324,13 +325,13 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
       dispatch({type: ACTIONS.ADDED_IMAGE_TO_POST});
       await timeout(1000);
       dispatch({type: ACTIONS.CLOSE_SIDEBAR});
-      document.querySelector('.media-modal-close').click();
+      document.querySelector('.media-modal-close .media-modal-icon').click();
     } catch (err) {
       dispatch({type: ACTIONS.ADD_IMAGE_TO_POST_ERROR, payload: err});
     }
   };
 
-  const includeInWp = async (ids = [], viewProp) => {
+  const includeInWp = async (ids = [], viewProp, isFeaturedImage = false) => {
     dispatch({type: ACTIONS.PROCESS_IMAGES, payload: {selection: ids}});
 
     Promise.all(ids.map(id => {
@@ -344,9 +345,16 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
       });
     })).then(result => {
       const payload = result.flat();
+      const imgId = payload[0].wordpress_id;
+
       if (viewProp === EDITOR_VIEW) {
-        processImageToAddToEditor(payload[0].wordpress_id);
+        if (!isFeaturedImage) {
+          processImageToAddToEditor(imgId);
+        } else {
+          setCustomFeaturedImage(imgId);
+        }
       }
+
       dispatch({type: ACTIONS.PROCESSED_IMAGES, payload: {images: payload}});
     }).catch(err => {
       dispatch({type: 'SET_ERROR', payload: {
@@ -355,6 +363,23 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
       }});
     });
   };
+
+  const setCustomFeaturedImage = async imgId => {
+    await wp.data.dispatch('core/editor').editPost({featured_media: imgId});
+    dispatch({type: ACTIONS.CLOSE_SIDEBAR});
+    document.querySelector('.media-modal-close .media-modal-icon').click();
+  };
+
+  // Prevent multiple instances of the Media Upload Modal.
+  const mediaModalPopup = document.querySelectorAll('.media-modal');
+
+  useEffect(() => {
+    if (mediaModalPopup.length > 1) {
+      const modals = Array.from(mediaModalPopup);
+      const modalToRemove = modals[0];
+      modalToRemove.parentNode.removeChild(modalToRemove);
+    }
+  }, [mediaModalPopup]);
 
   useEffect(() => {
     if (loaded) {
@@ -400,6 +425,51 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
     }
   }, [currentBlock]);
 
+  // Check whether the archive picker modal was opened from the Featured Image button.
+  // Only execute this code in the Editor.
+  useEffect(() => {
+    if (view !== EDITOR_VIEW) {
+      return;
+    }
+
+    const mediaModal = wp.media.frame || wp.media.featuredImage.frame();
+    // Sets the isUserSelectingFeaturedImage to either true or false
+    setIsUserSelectingFeaturedImage(mediaModal?.options?.state === 'featured-image');
+  }, [view]);
+
+  // Hide the modal toolbar when the Greenpeace Media tab is active.
+  // The toolbar is not used on this tab, since the options are
+  // in the tab sidebar.
+  useEffect(() => {
+    const targetElement = document.querySelector('#menu-item-mediaArchive');
+    const toolBar = document.querySelector('.media-toolbar');
+
+    if (!targetElement || !toolBar) {
+      return;
+    }
+
+    if (targetElement.className === 'media-menu-item active') {
+      toolBar.style.display = 'none';
+    }
+
+    const classChangeCallback = mutationsList => {
+      mutationsList.forEach(mutation => {
+        if (mutation.type === 'attributes' && mutation.attributeName === 'class') {
+          if (targetElement.className === 'media-menu-item active') {
+            toolBar.style.display = 'none';
+          } else {
+            toolBar.style.display = 'block';
+          }
+        }
+      });
+    };
+
+    const observer = new MutationObserver(classChangeCallback);
+    observer.observe(targetElement, {attributes: true});
+
+    return () => observer.disconnect();
+  }, []);
+
   return useMemo(() => (
     <Context.Provider
       value={{
@@ -422,6 +492,8 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
         processedIds,
         dispatch,
         imageAdded,
+        setCustomFeaturedImage,
+        isUserSelectingFeaturedImage,
         currentBlockImageId,
         view,
       }}
@@ -469,7 +541,7 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
 
       {view === EDITOR_VIEW && (
         <>
-          {(acceptedBlockTypes.includes(currentBlock.name)) ? (
+          {(isUserSelectingFeaturedImage || acceptedBlockTypes.includes(currentBlock?.name)) ? (
             <section className="archive-picker">
               <div className={classNames('archive-picker-main', {'is-open': selectedImages.length > 0 && !bulkSelect})}>
                 <ArchivePickerToolbar />
--- a/assets/src/js/Components/ArchivePicker/SingleSidebar.js
+++ b/assets/src/js/Components/ArchivePicker/SingleSidebar.js
@@ -24,6 +24,8 @@ export default function SingleSidebar({image}) {
     images,
     imageAdded,
     processImageToAddToEditor,
+    setCustomFeaturedImage,
+    isUserSelectingFeaturedImage,
     currentBlockImageId,
     view,
   } = useArchivePickerContext();
@@ -111,7 +113,25 @@ export default function SingleSidebar({image}) {
           )}
           {view === EDITOR_VIEW && (
             <>
-              {!image.wordpress_id ? (
+              {isUserSelectingFeaturedImage && image.wordpress_id && (
+                <button
+                  disabled={!!processing}
+                  className="button sidebar-action"
+                  onClick={async () => await setCustomFeaturedImage(image.wordpress_id)}
+                >
+                  {__('Set Featured Image', 'planet4-master-theme-backend')}
+                </button>
+              )}
+              {isUserSelectingFeaturedImage && !image.wordpress_id && (
+                <button
+                  disabled={!!processing}
+                  className="button sidebar-action"
+                  onClick={async () => await includeInWp([image.id], view, true)}
+                >
+                  {__('Import to Library & Set Featured Image', 'planet4-master-theme-backend')}
+                </button>
+              )}
+              {!isUserSelectingFeaturedImage && !image.wordpress_id && (
                 <button
                   disabled={!!processing}
                   className="button sidebar-action"
@@ -119,7 +139,8 @@ export default function SingleSidebar({image}) {
                 >
                   {__('Import to Library & Post', 'planet4-master-theme-backend')}
                 </button>
-              ) : (
+              )}
+              {!isUserSelectingFeaturedImage && image.wordpress_id && (
                 <button
                   disabled={!!processing}
                   className="button sidebar-action"
--- a/assets/src/js/Components/ArchivePicker/blockUpdateFunctions.js
+++ b/assets/src/js/Components/ArchivePicker/blockUpdateFunctions.js
@@ -58,6 +58,7 @@ export const updateMediaAndTextAttributes = (image, currentBlock) => {
       mediaAlt: image.alt_text,
       mediaId: image.id,
       mediaType: 'image',
+      useFeaturedImage: false,
     },
     originalContent: div.innerHTML,
   };
--- a/assets/src/js/media_archive.js
+++ b/assets/src/js/media_archive.js
@@ -1,6 +1,6 @@
+import {createRoot} from 'react-dom/client';
 import ArchivePicker from './Components/ArchivePicker/ArchivePicker';
 
-const {createRoot} = wp.element;
 const rootElement = document.getElementById('archive-picker-root');
 
 createRoot(rootElement).render(<ArchivePicker />);
--- a/assets/src/js/media_archive_editor_view.js
+++ b/assets/src/js/media_archive_editor_view.js
@@ -1,7 +1,7 @@
+import {createRoot} from 'react-dom/client';
 import ArchivePicker, {EDITOR_VIEW} from '../js/Components/ArchivePicker/ArchivePicker';
 
 const {__} = wp.i18n;
-const {createRoot} = wp.element;
 const frame = wp.media.view.MediaFrame.Select;
 
 (function() {


planet-4:

:watch: 2024.07.11 08:24:48

mardelnet (commented):
Hey @Osong-Michael 
Please, correct me if I'm wrong but I think that adding a featured image is mandatory for posts. So, based on that, why don't we render the image picker if the user is editing a post? Something like this:

Replace line 283:
```js
const currentBlock = wp.data.select('core/block-editor').getSelectedBlock();
```

With:
```js
const wpData = wp.data
const currentBlock = wpData.select('core/block-editor').getSelectedBlock();
const currentPostType = wpData.select('core/editor').getCurrentPostType();
```

Add before line 406:
```js
const isPost = currentPostType === "post";
const hasAcceptedBlockTypes = acceptedBlockTypes.includes(currentBlock?.name);
```

And finally replace line 470:
```js
{(acceptedBlockTypes.includes(currentBlock.name)) ? (
```

With:
```js
{( isPost || hasAcceptedBlockTypes ) ? (
```

Osong-Michael:
> Hey @Osong-Michael Please, correct me if I'm wrong but I think that adding a featured image is mandatory for posts. So, based on that, why don't we render the image picker if the user is editing a post? Something like this:
> 
> Replace line 283:
> 
> ```js
> const currentBlock = wp.data.select('core/block-editor').getSelectedBlock();
> ```
> 
> With:
> 
> ```js
> const wpData = wp.data
> const currentBlock = wpData.select('core/block-editor').getSelectedBlock();
> const currentPostType = wpData.select('core/editor').getCurrentPostType();
> ```
> 
> Add before line 406:
> 
> ```js
> const isPost = currentPostType === "post";
> const hasAcceptedBlockTypes = acceptedBlockTypes.includes(currentBlock?.name);
> ```
> 
> And finally replace line 470:
> 
> ```js
> {(acceptedBlockTypes.includes(currentBlock.name)) ? (
> ```
> 
> With:
> 
> ```js
> {( isPost || hasAcceptedBlockTypes ) ? (
> ```

It is a great suggestion, but we still need to make sure we know when the user is setting the Featured Image, so we need to tie into the event that is triggered when that button is clicked so we can update the `feature_image` attribute for post.

My idea was having a variable we can track when that even is triggered so we can use it to update that attribute based on the image selected.

mardelnet:
> It is a great suggestion, but we still need to make sure we know when the user is setting the Featured Image, so we need to tie into the event that is triggered when that button is clicked so we can update the `feature_image` attribute for post.

I don't understand why we have to wait for the Featured Image to be clicked...

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -398,6 +399,30 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
     }
   }, [currentBlock]);
 
+  useEffect(() => {
+    const sideBarElements = document.getElementsByClassName('components-panel__body');
```
GP-Dan-Tovbein, line 403: Can we use `querySelector` instead?

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -398,6 +399,30 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
...
+
+    for (let i = 0; i < sideBarElements.length; i++) {
+      const featuredImageDiv = sideBarElements[i].querySelector('.editor-post-featured-image');
+      if (featuredImageDiv) {
+        const featuredImageDivCtn = featuredImageDiv.querySelector('.editor-post-featured-image__container');
```
GP-Dan-Tovbein, line 408: Why don't directly do something like this?

```suggestion
        const featuredImageDivCtn = featuredImageDiv.querySelector('.editor-post-featured-image__container .editor-post-featured-image__toggle');
```

mardelnet (commented):
Hey @Osong-Michael 
I don't see this working correctly yet.
When I click on the Media Library tab of the Featured Image modal, then the GP Library is no longer accessible.

[AwesomeScreenshot-9_16_2024,2_14_17PM.webm](https://github.com/user-attachments/assets/37e2f659-9d72-46e1-ad3b-76096c888574)

mleray:
The Greenpeace Media page crashes with the following console error:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'frame')
    at media_archive.js?ver=1726652942:2:28702
    at ri (react-dom.min.js?ver=18.3.1:2:93198)
    at Ss (react-dom.min.js?ver=18.3.1:2:113320)
    at react-dom.min.js?ver=18.3.1:2:109924
    at S (react-dom.min.js?ver=18.3.1:2:130522)
    at MessagePort.T (react-dom.min.js?ver=18.3.1:2:131054)
```

Osong-Michael:
> The [Greenpeace Media page](https://www-dev.greenpeace.org/test-janus/wp-admin/upload.php?page=media-picker) crashes with the following console error:
> 
> ```
> Uncaught TypeError: Cannot read properties of undefined (reading 'frame')
>     at media_archive.js?ver=1726652942:2:28702
>     at ri (react-dom.min.js?ver=18.3.1:2:93198)
>     at Ss (react-dom.min.js?ver=18.3.1:2:113320)
>     at react-dom.min.js?ver=18.3.1:2:109924
>     at S (react-dom.min.js?ver=18.3.1:2:130522)
>     at MessagePort.T (react-dom.min.js?ver=18.3.1:2:131054)
> ```
> 
> This explains why (some of) the e2e tests are failing.

Fixed now

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -400,6 +414,66 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
...
+  useEffect(() => {
+    const targetElement = document.querySelector('#menu-item-mediaArchive');
+
+    if (!targetElement) {
+      return;
+    }
+
+    const toolBars = document.querySelectorAll('.media-frame-toolbar');
+
+    if (!toolBars) {
```
mleray, line 443: Maybe this check could be combined with the previous one?

```
if (!targetElement || !toolbars) {
      return;
}
```

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -400,6 +414,66 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
...
+      return;
+    }
+
+    if (targetElement.className === 'media-menu-item active') {
+      toolBars.forEach(element => {
```
mleray, line 448: This can be shortened to 
```
toolBars.forEach(element => element.style.display = 'none');`
```

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -400,6 +414,66 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
...
+        element.style.display = 'none';
+      });
+    }
+
+    const classChangeCallback = mutationsList => {
+      mutationsList.forEach(mutation => {
+        if (mutation.type === 'attributes' && mutation.attributeName === 'class') {
+          if (targetElement.className === 'media-menu-item active') {
```
mleray, line 456: This all `if` statement could be shortened to
```
toolBars.forEach(element => element.style.display = targetElement.className === 'media-menu-item active' ? 'none' : 'block');
```

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -356,6 +364,12 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
     });
   };
 
+  const setCustomFeaturedImage = async imgId => {
+    await wp.data.dispatch('core/editor').editPost({featured_media: imgId});
+    dispatch({type: ACTIONS.CLOSE_SIDEBAR});
+    document.querySelector('.media-modal-close').click();
```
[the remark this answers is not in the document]
Osong-Michael: This behaviour should work for all instances, but I have witnessed situations too where it does not work so I will try something different and see

assets/src/js/Components/ArchivePicker/ArchivePicker.js
Osong-Michael: For some reason if I shorten this, the toolbar no longer appears

mleray: That's weird 🤔 But it's not a big deal, if it breaks we can leave it as is

assets/src/js/Components/ArchivePicker/ArchivePicker.js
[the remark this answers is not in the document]
mleray: This still doesn't work as expected for the Media & Text block 😕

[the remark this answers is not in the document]
Osong-Michael: Funny thing is, I haven't changed the implementation yet 😅. [This](https://github.com/greenpeace/planet4-master-theme/blob/main/assets/src/js/Components/ArchivePicker/ArchivePicker.js#L316) is the implantation for the Media and Text block, same as Image Block and Featured Image.

mleray:
@Osong-Michael If I select an image for the featured image from the Greenpeace Media, it works fine 🎉 However if I then want to change it and click "Replace", that image is the only one I see in the Media Library tab:

Screenshot 2024-10-02 at 11 58 02

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -400,6 +421,53 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
     }
   }, [currentBlock]);
 
+  // Check whether the archive picker modal was opened from the Featured Image button.
+  // Only execute this code in the Editor.
+  useEffect(() => {
+    if (view === EDITOR_VIEW) {
```
mleray, line 427: I think it would be better to write 
```
if (view !== EDITOR_VIEW) {
  return;
}
```

mleray:
We should add the `Set featured image` button back at the bottom of the modal when switching tabs, currently if I open the `Greenpeace Media` tab and then open the `Media Library` one the button remains hidden...

Screenshot 2024-10-03 at 10 07 07

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -356,6 +368,15 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
     });
   };
 
+  const setCustomFeaturedImage = async imgId => {
+    await wp.data.dispatch('core/editor').editPost({featured_media: imgId});
+    dispatch({type: ACTIONS.CLOSE_SIDEBAR});
+    const closeBtn = document.querySelector('.media-modal-close');
+    if (closeBtn) {
+      closeBtn.click();
```
mleray, line 376: From my testing this doesn't work, for me the modal only closes when setting a featured image that wasn't already imported to the library (so clicking on `Import to Library & Set Featured Image` instead of the `Set Featured Image`)

assets/src/js/Components/ArchivePicker/ArchivePicker.js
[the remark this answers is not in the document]
mleray: What I see now for the Media & Text block is that the Greenpeace Media sidebar closes, but not the whole modal. Same for the Image block.

Also, when clicking on `Import to Library & Post` for a brief moment I see an `Add image to post` button, which is confusing because after a few seconds it disappears and an `Added!` message shows. So there's actually no need for that button.

Screenshot 2024-10-03 at 10 22 30

mleray:
If there is already a set featured image and I click on `Replace`, the Greenpeace Media tab crashes and is blank, in the console I see this error:

Screenshot 2024-10-03 at 10 27 35

Osong-Michael:
> We should add the `Set featured image` button back at the bottom of the modal when switching tabs, currently if I open the `Greenpeace Media` tab and then open the `Media Library` one the button remains hidden... I see the same issue with the modal blocks for blocks where the `Select` button remains hidden too. It comes back only when closing the modal and opening it again.
> 
> Screenshot 2024-10-03 at 10 07 07

This is interesting because in me digging into the issue, I found something interesting. When you first load the page and open the modal using the featured image button, everything works as expected. Once you close the modal and open again then what happens is, the active class remains on the Media Library tab if you use javascript selectors to get back the element, but if you just inspect the page, the active class is on the GP Media tab, it is interesting because I have no clue how that happens.

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -324,13 +325,13 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
       dispatch({type: ACTIONS.ADDED_IMAGE_TO_POST});
       await timeout(1000);
       dispatch({type: ACTIONS.CLOSE_SIDEBAR});
-      document.querySelector('.media-modal-close').click();
+      document.querySelector('.media-modal-icon').click();
```
mleray, line 328: I don't think that this selector is specific enough 🤔 

Screenshot 2024-10-07 at 15 08 11

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -356,6 +364,23 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
     });
   };
 
+  const setCustomFeaturedImage = async imgId => {
+    await wp.data.dispatch('core/editor').editPost({featured_media: imgId});
+    dispatch({type: ACTIONS.CLOSE_SIDEBAR});
+    document.querySelector('.media-modal-icon').click();
```
mleray, line 370: Same remark as previous comment on similar code, this selector is not specific enough

mleray:
Is there already a ticket to implement the Greenpeace Media for the OpenGraph image override?

Screenshot 2024-10-08 at 13 45 27

Osong-Michael:
> Is there already a ticket to implement the Greenpeace Media for the OpenGraph image override?
> 
> Screenshot 2024-10-08 at 13 45 27

No idea, maybe @comzeradd can respond to this.

mleray:
> We should add the `Set featured image` button back at the bottom of the modal when switching tabs, currently if I open the `Greenpeace Media` tab and then open the `Media Library` one the button remains hidden... I see the same issue with the modal blocks for blocks where the `Select` button remains hidden too. It comes back only when closing the modal and opening it again.

@Osong-Michael I still see this happening when an image has already been selected from the Greenpeace Media, if I then click "Replace" the first tab opening is the Greenpeace Media so the button remains hidden ([example page](https://www-dev.greenpeace.org/test-janus/wp-admin/post.php?post=9&action=edit)).

assets/src/js/Components/ArchivePicker/ArchivePicker.js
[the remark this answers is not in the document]
mleray: @Osong-Michael I still see this

mleray:
@Osong-Michael is this still WIP (according to the label)?

Osong-Michael:
> @Osong-Michael is this still WIP (according to the label)?

Hey @mleray funny thing is , I remember changing that to Review, but suprised it was still WIP 😅

mleray:
I've noticed that once I've selected the `Use featured image` option (here for a Media & Text block) and then try to replace it with an image from the Greenpeace Media, it doesn't work... Unless I manually deselect the `Use featured image` option first. If I decide to use an image from the Media Library it works though. It might not be related to these changes though, maybe it's already a bug with the current implementation 🤔 

Screenshot 2024-10-15 at 11 31 21

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -356,6 +364,23 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
...
+    await wp.data.dispatch('core/editor').editPost({featured_media: imgId});
+    dispatch({type: ACTIONS.CLOSE_SIDEBAR});
+    document.querySelector('.media-modal-close .media-modal-icon').click();
+  };
+
+  // Prevent multiple instances of the Media Upload Modal.
+  const mediaModalPopup = document.querySelectorAll('.media-modal');
+
+  useEffect(() => {
+    if (mediaModalPopup && mediaModalPopup.length > 1) {
```
GP-Dan-Tovbein, line 377: I think `media.length` is ok. `document.querySelectorAll` always returns `[]` if no selector is found.

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -356,6 +364,23 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
...
+      const modals = Array.from(mediaModalPopup);
```
GP-Dan-Tovbein, line 378: Thinking loud.. is `[...mediaModalPopup]` a better approach?

Osong-Michael: `mediaModalPopup` is a NodeList we are converting to an array, so hence the approach above, I think 🤔

mardelnet:
@Osong-Michael 
According to our previous conversation on Slack, I tested using the Featured Image in the Media and Text block and it worked correctly, both locally and in the testing environment.

assets/src/js/Components/ArchivePicker/SingleSidebar.js
```diff
@@ -111,15 +113,34 @@ export default function SingleSidebar({image}) {
           )}
           {view === EDITOR_VIEW && (
             <>
-              {!image.wordpress_id ? (
+              {isUserSelectingFeaturedImage && image.wordpress_id && (
```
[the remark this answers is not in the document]
mleray: Or at least we could create a constant for the props that they all have in common (e.g. `disabled` and `className`)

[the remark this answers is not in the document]
Osong-Michael: I agree, reason I created a follow up ticket to refactor this components and look into other things.

assets/src/js/Components/ArchivePicker/blockUpdateFunctions.js
```diff
@@ -58,6 +58,7 @@ export const updateMediaAndTextAttributes = (image, currentBlock) => {
       mediaAlt: image.alt_text,
       mediaId: image.id,
       mediaType: 'image',
+      useFeaturedImage: false,
```
GP-Dan-Tovbein, line 61: Is it complex if we rename it? Basically because, as a React convention, `use` prefix is used only for `hooks`.

assets/src/js/Components/ArchivePicker/SingleSidebar.js
```diff
@@ -111,15 +113,34 @@ export default function SingleSidebar({image}) {
...
                   disabled={!!processing}
                   className="button sidebar-action"
                   onClick={async () => await includeInWp([image.id], view)}
                 >
                   {__('Import to Library & Post', 'planet4-master-theme-backend')}
                 </button>
-              ) : (
+              )}
+              {!isUserSelectingFeaturedImage && image.wordpress_id && (
                 <button
```
GP-Dan-Tovbein, line 144: This if condition can be also added into the above condition wrapped by `Fragments`.

assets/src/js/Components/ArchivePicker/blockUpdateFunctions.js
Osong-Michael: That is how the WP attribute is named.

assets/src/js/Components/ArchivePicker/SingleSidebar.js
Osong-Michael: Yea, like I said earlier, I created a follow up [ticket](https://jira.greenpeace.org/browse/PLANET-7641) to refactor this components.

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -324,13 +325,13 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
       dispatch({type: ACTIONS.ADDED_IMAGE_TO_POST});
       await timeout(1000);
       dispatch({type: ACTIONS.CLOSE_SIDEBAR});
-      document.querySelector('.media-modal-close').click();
+      document.querySelector('.media-modal-close .media-modal-icon').click();
```
GP-Dan-Tovbein, line 328: I've tested without `.media-modal-icon` and seems to be working fine. Could you check it, please?

Osong-Michael: Yes it works, but @mleray wanted it to be more specific. The class could be present in some other element.

assets/src/js/Components/ArchivePicker/ArchivePicker.js
```diff
@@ -356,6 +364,23 @@ export default function ArchivePicker({view = ADMIN_VIEW}) {
...
   };
 
+  const setCustomFeaturedImage = async imgId => {
+    await wp.data.dispatch('core/editor').editPost({featured_media: imgId});
+    dispatch({type: ACTIONS.CLOSE_SIDEBAR});
+    document.querySelector('.media-modal-close .media-modal-icon').click();
+  };
+
+  // Prevent multiple instances of the Media Upload Modal.
+  const mediaModalPopup = document.querySelectorAll('.media-modal');
```
GP-Dan-Tovbein, line 374: I'd try to move the `mediaModalPopup` as a state hook, and declare it within the initial use effect hook. Then you'll convert the code more Reactive. Also will avoid to re-deeclare in every render.

For example
```
const mediaModalPopup = document.querySelectorAll('.media-modal');
```
Replace with
```
useEffect(() => {
if(document.querySelectorAll('.media-modal')) {
   setMediaModalPopup(document.querySelectorAll('.media-modal'));
}, []);
```

GP-Dan-Tovbein: What do you think?

GP-Dan-Tovbein (approved):
Tested and seems to be working as expected!

Nice job @Osong-Michael 🥳

PR #2322, opened 2024-07-15
PoC for preloading responsive images for blocks

Osong-Michael:
Ref: PoC for [Preloading big images](https://web.dev/articles/preload-responsive-images) (eg. Carousel Header, Page Header pattern) while using imagesrcset.

Some implementations propose using `<link rel="preload" as="image" href="images/example-medium.jpg" media="(max-width: 1023px)">`, while others suggest  replacing the `media` with a `sizes` attribute instead. 

**Testing**

If you disable caching on the Network tab and enable 3G, then you reload the page there are two visible scenarios
- The first case, which is the normal way our websites currently load is, you see the images load at the same time the page shows, so for a split second the Carousel Header has an empty image.
- Second case, with the current implementation, when you reload the page, the first image in the Carousel Header block is being pre-loaded as the page loads, meaning by the time the page becomes visible, the image is present.

This implementation is good for blocks or scenarios where we control what images are being displayed.

### DIFF
--- a/assets/src/blocks/CarouselHeader/SlideBackground.js
+++ b/assets/src/blocks/CarouselHeader/SlideBackground.js
@@ -1,12 +1,44 @@
+function parseWidths(srcset) {
+  const entries = srcset.split(',').map(entry => entry.trim());
+  const widths = entries.map(entry => {
+    const width = entry.split(' ')[1];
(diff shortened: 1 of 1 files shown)


planet-4:

:watch: 2024.07.15 09:41:10

PR #2369, opened 2024-09-05
 PLANET-7606 Fix Cookies block crashing in the frontend

mleray:
### Description

See [PLANET-7606](https://jira.greenpeace.org/browse/PLANET-7606)

In this PR I also try separating the backend and frontend files, I feel it's less confusing that way and more consistent with what we do with other blocks, but let me know what you think 🙂 

### Testing

- [Broken version](https://www-dev.greenpeace.org/test-pluto/privacy-and-cookies/)
- [Fixed version](https://www-dev.greenpeace.org/test-nix/privacy-and-cookies/)

### DIFF
--- a/assets/src/blocks/Cookies/CookiesScript.js
+++ b/assets/src/blocks/Cookies/CookiesScript.js
@@ -1,9 +1,8 @@
 import {createRoot} from 'react-dom/client';
-import {BLOCK_NAME} from './CookiesBlock';
 import {CookiesFrontend} from './CookiesFrontend';
 
 // Fallback for non migrated content. Remove after migration.
-document.querySelectorAll(`[data-render="${BLOCK_NAME}"]`).forEach(
+document.querySelectorAll('[data-render="planet4-blocks/cookies"]').forEach(
   blockNode => {
     const attributes = JSON.parse(blockNode.dataset.attributes);
     const rootElement = createRoot(blockNode);
--- a/assets/src/blocks/Cookies/CookiesBlock.js
+++ b/assets/src/blocks/Cookies/CookiesBlock.js
@@ -1,11 +1,9 @@
 import {CookiesEditor} from './CookiesEditor.js';
 
-export const BLOCK_NAME = 'planet4-blocks/cookies';
-
 export const registerCookiesBlock = () => {
   const {registerBlockType} = wp.blocks;
 
-  registerBlockType(BLOCK_NAME, {
+  registerBlockType('planet4-blocks/cookies', {
     title: 'Cookies',
     icon: 'welcome-view-site',
     category: 'planet4-blocks',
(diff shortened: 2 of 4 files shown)


planet-4:

:watch: 2024.09.05 15:51:35

assets/src/blocks/Cookies/CookiesScript.js
```diff
@@ -1,9 +1,8 @@
 import {createRoot} from 'react-dom/client';
-import {BLOCK_NAME} from './CookiesBlock';
 import {CookiesFrontend} from './CookiesFrontend';
 
 // Fallback for non migrated content. Remove after migration.
-document.querySelectorAll(`[data-render="${BLOCK_NAME}"]`).forEach(
+document.querySelectorAll('[data-render="planet4-blocks/cookies"]').forEach(
```
mleray, line 5: This is the actual fix, because importing something from the `CookiesBlock` also imported the `CookiesEditor` component and caused the problem.

mardelnet (commented):
Thanks for this, @mleray 
I still see some issues with this block.
For instance, the "Other Cookies" checkbox cannot be checked/unchecked in the front.
By trying to do it, an error appears in the browser console:
```js 
Uncaught ReferenceError: capabilities is not defined
CookiesScript.js
``` 
Also, it is not entirely clear to me why we have to replace the BLOCK_NAME constant with the actual block name. What about creating a specific file to contain the constants (or is it an overkill)?

mleray:
> Thanks for this, @mleray I still see some issues with this block. For instance, the "Other Cookies" checkbox cannot be checked/unchecked in the front. By trying to do it, an error appears in the browser console:

Good catch @mardelnet! I've messaged Sagar about this because I see it's related to changes he made previously, not sure why it no longer works 🤔 

> Also, it is not entirely clear to me why we have to replace the `BLOCK_NAME` constant with the actual block name. What about creating a specific file to contain the constants (or is it an overkill)?

This is because if we import `BLOCK_NAME` from the CookiesBlock file, it also reads the rest of the file and therefore also imports `CookiesEditor` which itself uses the `Tooltip` component and this isn't accessible in the frontend, only in the backend... We could indeed create a separate file, but for only one variable that is used twice I'd argue it's unnecessary!

mardelnet (approved):

mardelnet:
> Hi @mardelnet , Could you please check now?

Hey @sagarsdeshmukh, have you pushed your changes?
I don't see any difference in this PR.

mleray:
> Hey @sagarsdeshmukh, have you pushed your changes? I don't see any difference in this PR.

Hey @mardelnet, there were no code changes added because the issue was with the "Analytical cookies" fields being empty in the settings... Now that @sagarsdeshmukh filled them up it works as expected 🙂 We've opened a [follow-up ticket](https://jira.greenpeace.org/browse/PLANET-7607) to look into this edge case.

PR #2370, opened 2024-09-06
PLANET-7592 Re-implement related articles with Query Loop block

Osong-Michael:
**Ref: https://jira.greenpeace.org/browse/PLANET-7592**

**Testing:**

On local or test instance, the "_Related Articles_" section is now rendered using the Query loop block.

UI wise, it still should be similar.

### DIFF
--- a/assets/src/js/remove_related_section_no_posts.js
+++ b/assets/src/js/remove_related_section_no_posts.js
@@ -1,0 +1,8 @@
+export const removeRelatedPostsSection = () => {
+  // Remove Related Posts section when Query loop returns no posts.
+  const section = document.querySelector('.post-articles-block');
+
+  if (section && !section.querySelector('.p4-query-loop .wp-block-post-template')) {
+    section.style.display = 'none';
+  }
+};
--- a/functions.php
+++ b/functions.php
@@ -249,6 +249,7 @@ function register_more_blocks(): void
         ]
     );
     // Like the core block but with an appropriate sizes attribute.
+     // phpcs:disable Generic.Files.LineLength.MaxExceeded
     register_block_type(
         'p4/post-featured-image',
         [
@@ -264,7 +265,7 @@ function register_more_blocks(): void
                     // For example, it could already access displayLayout from Query block to know how many columns are
                     // being rendered. If it then also knows the flex gap and container width, it should have all needed
                     // info to support a large amount of cases.
-                    [ 'sizes' => '(min-width: 1600px) 389px, (min-width: 1200px) 335px, (min-width: 1000px) 281px, (min-width: 780px) 209px, (min-width: 580px) 516px, calc(100vw - 24px)' ] // phpcs:ignore Generic.Files.LineLength.MaxExceeded
+                    [ 'sizes' => '(min-width: 1600px) 389px, (min-width: 1200px) 335px, (min-width: 1000px) 281px, (min-width: 780px) 209px, (min-width: 580px) 516px, calc(100vw - 24px)' ]
                 );
 
                 return "<a href='$post_link'>$featured_image</a>";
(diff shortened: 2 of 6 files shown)


planet-4:

:watch: 2024.09.06 12:50:04

single.php
```diff
@@ -73,16 +73,56 @@
...
+        ],
+        'namespace' => 'planet4-blocks/posts-list',
     ];
-
-    $post->articles = '<!-- wp:planet4-blocks/articles '
-        . wp_json_encode($block_attributes, JSON_UNESCAPED_SLASHES)
-        . ' /-->';
+    // phpcs:disable Generic.Files.LineLength.MaxExceeded
+    $post->articles = '<!-- wp:query ' . wp_json_encode($block_attributes, JSON_UNESCAPED_SLASHES) . ' -->
+        <div class="wp-block-query posts-list p4-query-loop is-custom-layout-list"><!-- wp:group {"layout":{"type":"flex","justifyContent":"space-between"}} -->
```
mleray, line 95: There should be a way to avoid re-writing all this 🤔

Osong-Michael: I hope so, because the Query Loop block needs a layout, unlike the articles block that already had one defined layout.

comzeradd: Is it possible to re-use the [listing pages partial](https://github.com/greenpeace/planet4-master-theme/blob/main/parts/query-listing-page.html)?

Osong-Michael: I am not totally sure, but since we want the "_Related Articles_" section, means we should be able to render the partial based on the attributes passed 🤔

mleray: I've opened a [PoC PR](https://github.com/greenpeace/planet4-master-theme/pull/2375) that should allow us to just write `<!-- wp:planet4-blocks/posts-list/-->`

mleray: Well that PoC PR didn't work in the end 😬 But @Osong-Michael could you please look into what Pedro suggested in this [comment](https://github.com/greenpeace/planet4-master-theme/pull/2375#discussion_r1781107675)?

mleray:
@Osong-Michael the corresponding e2e test will need to be updated if we go with this implementation!

tests/e2e/blocks/related-articles.spec.js
```diff
@@ -32,10 +32,10 @@ test('Test Related Articles block', async ({page, admin, editor}) => {
   await page.waitForTimeout(1000); // letting metabox post query finish
 
   await page.goto(postUrl);
-  const relatedSection = await page.locator('[data-render="planet4-blocks/articles"]');
+  const relatedSection = await page.locator('.p4-query-loop');
```
mleray, line 35: Technically we could have another Query Loop block in the post, no? I'm wondering if we need to refine this test to make sure that we are targeting the one for the Related Articles section 🤔

functions.php
```diff
@@ -264,18 +265,75 @@ function register_more_blocks(): void
...
             },
             'uses_context' => [ 'postId' ],
         ]
     );
+    // Block displays related posts using the Query Loop block
+    register_block_type(
+        'p4/related-posts',
+        [
+            'attributes' => [
+                'queryAttributes' => [
```
mleray, line 281: Is it not `query_attributes`? If I look [here](https://github.com/greenpeace/planet4-master-theme/pull/2370/files#diff-8e039a17b2aadb723e862bced701672b6e20a79e19ca6426b49084fc76d52624R96)

mleray (commented):
I'm guessing we should either hide the entire "Related Posts" section or add a "No posts found" block when the query returns no posts (see [here](https://www-dev.greenpeace.org/test-saturn/publication/260/mauris-quis-dictum-magna/) for example)

assets/src/js/remove_related_section_no_posts.js
```diff
@@ -0,0 +1,8 @@
+export const removeRelatedPostsSection = () => {
+  // Remove Related Posts sectio when Query loop returns no posts.
```
mleray, line 2: Small typo here, `sectio` => `section`

mleray (approved):
Looks good to me 👍

PR #2375, opened 2024-09-11
Test new Posts List implementation

mleray:
### Description

To replace the "Related Articles" post setting we would like to use our new Posts List block, but currently in order to achieve this the code is quite long ([see PR](https://github.com/greenpeace/planet4-master-theme/pull/2370)). It would be nice to find a shorter way to write it.

Instead of a block variation, we could create it as a separate block. This would allow us to easily use it in our code, for example we should be able to write `<!-- wp:planet4-blocks/posts-list-test /-->` (this can be tested in the backend via the code editor). It seems there is no way to do this with block variations.

For now I haven't looked into the Query Loop extensions (sidebar options) that we have and how to make them work with this new version, but I guess it should be rather straightforward.

We should also be able to add attributes to this block if needed via the `block.json` file.

### DIFF
--- a/assets/src/blocks/PostsListTest/index.js
+++ b/assets/src/blocks/PostsListTest/index.js
@@ -1,0 +1,14 @@
+import metadata from './block.json';
+import template from './template';
+
+const {registerBlockType} = wp.blocks;
+const {useBlockProps, InnerBlocks} = wp.blockEditor;
+
+export const registerPostsListTest = () => registerBlockType(metadata, {
+  edit: props => (
+    <div {...useBlockProps()}>
+      {wp.element.createElement(InnerBlocks, {template: template(props.attributes ?? {})})}
+    </div>
+  ),
+  save: () => wp.element.createElement(wp.blockEditor.InnerBlocks.Content, {}),
+});
--- a/assets/src/blocks/PostsListTest/template.js
+++ b/assets/src/blocks/PostsListTest/template.js
@@ -1,0 +1,95 @@
+export const POSTS_LIST_BLOCK_NAME = 'planet4-blocks/posts-list';
+export const POSTS_LISTS_LAYOUT_TYPES = [
+  {label: 'List', value: 'default', columnCount: 3},
+  {label: 'Grid', value: 'grid', columnCount: 4},
+  {label: 'Carousel', value: 'flex', columnCount: 8},
+];
+
+const {__} = wp.i18n;
+
+const newsPageLink = window.p4_vars.news_page_link;
+
+const seeAllLink = ['core/navigation-link', {...!newsPageLink ? {className: 'd-none'} : {
+  url: newsPageLink,
+  label: __('See all stories', 'planet4-blocks-backend'),
+  className: 'see-all-link',
+}}];
+
+const template = () => ([
+  ['core/query', {
+    className: 'posts-list p4-query-loop is-custom-layout-list',
+    query: {
+      perPage: 3,
+      pages: 0,
+      offset: 0,
+      postType: 'post',
+      order: 'desc',
+      orderBy: 'date',
+      author: '',
+      search: '',
+      exclude: [],
+      sticky: '',
+      inherit: false,
+      postIn: [],
+      hasPassword: false,
+    },
+    layout: {
+      type: 'default',
+      columnCount: 3,
+    },
+  }, [
+    ['core/group', {layout: {type: 'flex', justifyContent: 'space-between'}}, [
+      ['core/heading', {lock: {move: true}, content: __('Related Posts', 'planet4-blocks-backend')}],
+      seeAllLink,
+    ]],
+    ['core/paragraph', {
+      lock: {move: true},
+      placeholder: __('Enter description', 'planet4-blocks-backend'),
+      style: {
+        spacing: {
+          margin: {
+            top: '24px',
+            bottom: '36px',
+          },
+        },
+      },
+    }],
+    ['core/query-no-results', {}, [
+      ['core/paragraph', {content: __('No posts found. (This default text can be edited)', 'planet4-blocks-backend')}],
+    ]],
+    ['core/post-template', {lock: {move: true, remove: true}}, [
+      ['core/columns', {}, [
+        ['core/post-featured-image', {isLink: true}],
+        ['core/group', {}, [
+          ['core/group', {layout: {type: 'flex'}}, [
+            ['core/post-terms', {
+              term: 'category',
+              separator: ' | ',
+            }],
+            ['core/post-terms', {
+              term: 'post_tag',
+              separator: ' ',
+            }],
+          ]],
+          ['core/post-title', {isLink: true}],
+          ['core/post-excerpt'],
+          ['core/group', {className: 'posts-list-meta'}, [
+            ['core/post-author-name', {isLink: true}],
+            ['core/post-date'],
+          ]],
+        ]],
+      ]],
+    ]],
+    ['core/buttons', {
+      className: 'carousel-controls',
+      lock: {move: true},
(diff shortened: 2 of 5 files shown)


planet-4:

:watch: 2024.09.11 09:32:46

mleray:
**Update on this implementation:** it doesn't work for the Related Articles 😢 It works in the code editor though, so it seems that the issue is in the frontend rendering

single.php
```diff
@@ -63,26 +63,7 @@
...
-        'tags' => $tag_id_array,
-        'categories' => $category_id_array,
-        'article_heading' => __('Related Articles', 'planet4-blocks'),
-        'read_more_text' => __('Load more', 'planet4-blocks'),
-    ];
-
-    $post->articles = '<!-- wp:planet4-blocks/articles '
-        . wp_json_encode($block_attributes, JSON_UNESCAPED_SLASHES)
-        . ' /-->';
+    $post->articles = '<!-- wp:planet4-blocks/posts-list-test /-->';
```
mardelnet, line 66: Can't we just pass here a query block code with the "post-list" variation?
Something like:

<details>
<summary>Click to expand!</summary>

``` 
<!-- wp:query {"queryId":4,"query":{"perPage":3,"pages":0,"offset":0,"postType":"post","order":"desc","orderBy":"date","author":"","search":"","exclude":[1113],"sticky":"","inherit":false,"postIn":[],"hasPassword":false},"namespace":"planet4-blocks/posts-list","className":"posts-list p4-query-loop is-custom-layout-list","layout":{"type":"default","columnCount":3}} -->
<div class="wp-block-query posts-list p4-query-loop is-custom-layout-list"><!-- wp:group {"layout":{"type":"flex","justifyContent":"space-between"}} -->
<div class="wp-block-group"><!-- wp:heading {"lock":{"move":true}} -->
<h2 class="wp-block-heading">Related Posts</h2>
<!-- /wp:heading -->

<!-- wp:navigation-link {"label":"See all stories","url":"http://www.planet4.test/news-stories/","className":"see-all-link"} /--></div>
<!-- /wp:group -->

<!-- wp:paragraph {"placeholder":"Enter description","lock":{"move":true},"style":{"spacing":{"margin":{"top":"24px","bottom":"36px"}}}} -->
<p style="margin-top:24px;margin-bottom:36px"></p>
<!-- /wp:paragraph -->

<!-- wp:query-no-results -->
<!-- wp:paragraph -->
<p>No posts found. (This default text can be edited)</p>
<!-- /wp:paragraph -->
<!-- /wp:query-no-results -->

<!-- wp:post-template {"lock":{"move":true,"remove":true}} -->
<!-- wp:columns -->
<div class="wp-block-columns"><!-- wp:post-featured-image {"isLink":true} /-->

<!-- wp:group -->
<div class="wp-block-group"><!-- wp:group {"layout":{"type":"flex"}} -->
<div class="wp-block-group"><!-- wp:post-terms {"term":"category","separator":" | "} /-->

<!-- wp:post-terms {"term":"post_tag","separator":" "} /--></div>
<!-- /wp:group -->

<!-- wp:post-title {"isLink":true} /-->

<!-- wp:post-excerpt /-->

<!-- wp:group {"className":"posts-list-meta"} -->
<div class="wp-block-group posts-list-meta"><!-- wp:post-author-name {"isLink":true} /-->

<!-- wp:post-date /--></div>
<!-- /wp:group --></div>
<!-- /wp:group --></div>
<!-- /wp:columns -->
<!-- /wp:post-template -->

<!-- wp:buttons {"lock":{"move":true},"className":"carousel-controls","layout":{"type":"flex","justifyContent":"space-between","orientation":"horizontal","flexWrap":"nowrap"}} -->
<div class="wp-block-buttons carousel-controls"><!-- wp:button {"className":"carousel-control-prev"} -->
<div class="wp-block-button carousel-control-prev"><a class="wp-block-button__link wp-element-button">Prev</a></div>
<!-- /wp:button -->

<!-- wp:button {"className":"carousel-control-next"} -->
<div class="wp-block-button carousel-control-next"><a class="wp-block-button__link wp-element-button">Next</a></div>
<!-- /wp:button --></div>
<!-- /wp:buttons -->

<!-- wp:navigation-link {"label":"See all stories","url":"http://www.planet4.test/news-stories/","className":"see-all-link"} /--></div>
<!-- /wp:query -->
```

mardelnet: Right, it's a lot of code. I'll keep digging...

mardelnet: Any way to import a JSON file into both the `single.php` and the assets/src/blocks/PostsListTest/template.js` to avoid repeating the block attributes and inner blocks?

Osong-Michael: @mardelnet, @mleray @sagarsdeshmukh, is it possible to have a partial that takes arguments/params that way we could modify this [listing page](https://github.com/greenpeace/planet4-master-theme/blob/main/parts/query-listing-page.html) or create one just for the Related Articles.

mleray: Hm 🤔 This would still be duplicated code indeed, but it would make it more readable so I would say it's better!

Osong-Michael: Sorry just looking into this now @mleray and @mardelnet. Question, does the new template part also accept params for how to display the post lists block, because the data needed to display each block is different depending on the attributes.

assets/src/blocks/PostsListTest/index.js
```diff
@@ -0,0 +1,14 @@
...
+const {registerBlockType} = wp.blocks;
+const {useBlockProps, InnerBlocks} = wp.blockEditor;
+
+export const registerPostsListTest = () => registerBlockType(metadata, {
+  edit: props => (
+    <div {...useBlockProps()}>
+      {wp.element.createElement(InnerBlocks, {template: template(props.attributes ?? {})})}
+    </div>
+  ),
+  save: () => wp.element.createElement(wp.blockEditor.InnerBlocks.Content, {}),
```
GP-Dan-Tovbein, line 13: I see you are already destructured `InnerBlocks`

assets/src/blocks/PostsListTest/template.js
```diff
@@ -0,0 +1,95 @@
+export const POSTS_LIST_BLOCK_NAME = 'planet4-blocks/posts-list';
+export const POSTS_LISTS_LAYOUT_TYPES = [
+  {label: 'List', value: 'default', columnCount: 3},
+  {label: 'Grid', value: 'grid', columnCount: 4},
+  {label: 'Carousel', value: 'flex', columnCount: 8},
+];
+
+const {__} = wp.i18n;
```
GP-Dan-Tovbein, line 8: What do you think if we keep dependencies to the top of the file and then "custom" `consts`?

assets/src/blocks/PostsListTest/template.js
```diff
@@ -0,0 +1,95 @@
...
+            bottom: '36px',
+          },
+        },
+      },
+    }],
+    ['core/query-no-results', {}, [
+      ['core/paragraph', {content: __('No posts found. (This default text can be edited)', 'planet4-blocks-backend')}],
+    ]],
+    ['core/post-template', {lock: {move: true, remove: true}}, [
+      ['core/columns', {}, [
```
GP-Dan-Tovbein, line 61: Thinking loud.. do we really need to use `columns` here? Wouldn't be better if we use a `group`?

assets/src/blocks/PostsListTest/template.js
mleray: Hey Dan 👋 You probably missed it, but we kinda abandoned this implementation idea because it doesn't work for the Related Articles 😢 It works in the code editor though, so it seems that the issue is in the frontend rendering

single.php
mardelnet: @Osong-Michael 
I think we cannot pass parameters to a template part. Our options are:
- Turn the template into a custom block that accepts attributes.
- Create a filter to modify the content of the template part (for example, [this one](https://developer.wordpress.org/reference/hooks/render_block_this-name/))

PR #2394, opened 2024-10-02
PLANET-7108 Fix double registration of share functions

mleray:
### Description

See [PLANET-7108](https://jira.greenpeace.org/browse/PLANET-7108)

This would happen in posts with a Gravity Forms block since both the post and the block use the share buttons template

### Testing


planet-4:

:watch: 2024.10.02 12:55:08

templates/blocks/share_buttons.twig
```diff
@@ -1,14 +1,14 @@
 {% block javascripts %}
     <script>
-            window.addEventListener('DOMContentLoaded', () => {
-                // If the native share functionality is available on the device, we show the corresponding share buttons.
-                if (!navigator.share) {
-                    return;
-                }
+        window.addEventListener('DOMContentLoaded', () => {
```
mleray, line 3: The indentation has also been fixed

mardelnet (dismissed):
Well done, @mleray 
BTW, should we check if `typeof dataLayer !== 'undefined'` ?

mleray:
@mardelnet I'm asking for your feedback again because after testing some more I realised there were some issues left, in particular when clicking on the icons to share we would get an error message that `dataLayerPush` isn't defined (even though it worked properly 🤔)
I've added these functions to the `window` object and now it seems all good, but let me know what you think 🙏

mardelnet:
> @mardelnet I'm asking for your feedback again because after testing some more I realised there were some issues left, in particular when clicking on the icons to share we would get an error message that `dataLayerPush` isn't defined (even though it worked properly 🤔) I've added these functions to the `window` object and now it seems all good, but let me know what you think 🙏

How can I test/replicate the new errors you've found? @mleray

mleray:
> How can I test/replicate the new errors you've found? @mleray

@mardelnet yeah sorry it's not great, I should have pushed a new commit instead of overriding the previous one. Do you want me to push again as 2 separate commits? Then at least on local you can see the 2 behaviours.

mardelnet:
> @mardelnet yeah sorry it's not great, I should have pushed a new commit instead of overriding the previous one. Do you want me to push again as 2 separate commits? Then at least on local you can see the 2 behaviours.

No, it's fine. Just let me know how I can see errors with the share icons.

mleray:
> No, it's fine. Just let me know how I can see errors with the share icons.

@mardelnet they were just console errors!

templates/blocks/share_buttons.twig
```diff
@@ -1,23 +1,28 @@
...
-                // If the native share functionality is available on the device, we show the corresponding share buttons.
-                if (!navigator.share) {
-                    return;
-                }
-
-                const nativeShareButtons = document.querySelectorAll('.share-buttons .share-btn.native');
-                nativeShareButtons.forEach(nativeShareButton => nativeShareButton.style.display = 'block');
-            });
+        window.addEventListener('DOMContentLoaded', () => {
+            if (typeof dataLayer === undefined) {
```
mardelnet, line 4: I am not 100% sure about this, but I think this will always evaluate to false because `typeof` returns a string, so the correct way to check if `dataLayer` is undefined would be:
```js
if (typeof dataLayer === 'undefined') {
    return;
}
```

templates/blocks/share_buttons.twig
```diff
@@ -1,23 +1,28 @@
 {% block javascripts %}
     <script>
```
mardelnet, line 2: This is an idea for another ticket/PR:
IMO, in terms of code separation and maintainability, this JS logic should not be part of the Twig template.
It should exist on a separate file and be included using the `wp_enqueue_script` or similar, which also has benefits in terms of security, caching, minification, etc...

templates/blocks/share_buttons.twig
mleray: Funny that you say that, because I just had this exact same issue in another PR of mine, and I think you're right! I'll update the code

templates/blocks/share_buttons.twig
mleray: I agree! Can you please create a ticket for it? 🙏

templates/blocks/share_buttons.twig
mleray: Although the share functions are still working so that confuses me 🤔

mardelnet (approved):

templates/blocks/share_buttons.twig
mardelnet: Done! https://jira.greenpeace.org/browse/PLANET-7625

PR #2397, opened 2024-10-04
PLANET-7607 Fix Cookies block frontend bug

mleray:
### Description

See [PLANET-7607](https://jira.greenpeace.org/browse/PLANET-7607)

It doesn't always work depending on analytical cookies settings, because the capabilities object is not defined

### Testing

Before this fix, when enabling the analytical cookies but not filling the corresponding text fields, the block would not work properly in the frontend. You can check the following instances to test:

- [Broken version](https://www-dev.greenpeace.org/test-pluto/privacy-and-cookies/)
- [Fixed version](https://www-dev.greenpeace.org/test-nix/privacy-and-cookies/)

### DIFF
--- a/admin/js/options.js
+++ b/admin/js/options.js
@@ -7,24 +7,42 @@ spamButtons.forEach(box => {
 // Show/hide old special pages when enabling or disabling the new IA.
 document.addEventListener('DOMContentLoaded', () => {
   const newIASetting = document.querySelector('#new_ia');
+  const analyticalCookiesCheckbox = document.querySelector('#enable_analytical_cookies');
 
-  if (!newIASetting) {
+  if (!newIASetting && !analyticalCookiesCheckbox) {
     return;
   }
 
-  const actPageDropdown = document.querySelector('[data-fieldtype="act_page_dropdown"]');
-  const explorePageDropdown = document.querySelector('[data-fieldtype="explore_page_dropdown"]');
-  const issuesParentCategoryDropdown = document.querySelector('[data-fieldtype="category_select_taxonomy"]');
+  if (newIASetting) {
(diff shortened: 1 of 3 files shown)


planet-4:

:watch: 2024.10.04 09:32:31

mleray:
> Thanks, Maud for fixing this bug, can we also make it easy for the admin user not to miss analytics fields, when enabling it by changing the following

@sagarsdeshmukh maybe we could hide/show the analytical cookies text fields based on the checkbox 🤔

comzeradd:
Yeah, both changes make sense. The checkbox should be above the relevant text fields. And both of them should be hidden if the option is not enabled.

mleray:
@sagarsdeshmukh can you please have a look? I've updated the PR as suggested by previous comments.
@comzeradd I'm not sure if this would require a FLAG, but maybe documentation needs to be updated.

admin/js/options.js
```diff
@@ -7,24 +7,42 @@ spamButtons.forEach(box => {
...
-  explorePageDropdown.classList.toggle('hidden', newIASetting.checked);
-  issuesParentCategoryDropdown.classList.toggle('hidden', newIASetting.checked);
-
-  newIASetting.addEventListener('change', event => {
-    const {checked} = event.currentTarget;
-    actPageDropdown.classList.toggle('hidden', checked);
-    explorePageDropdown.classList.toggle('hidden', checked);
-    issuesParentCategoryDropdown.classList.toggle('hidden', checked);
-  });
+  if (newIASetting) {
```
mleray, line 16: I was wondering if this is still needed? 🤔

sagarsdeshmukh (approved):
The PR tested on local and test instance. Looks good to me. Nice work 👍

PR #2357, opened 2024-08-23
PLANET-7521: Migrate Media Block to Audio/Video blocks

mardelnet:
Ref: https://jira.greenpeace.org/browse/PLANET-7521

<hr>

**DESCRIPTION:**
This PR migrates the Media blocks to Audio, Embed or Video blocks.
It also removes the Media blocks from our codebase to prevent users from using them after the migration.

<hr>

**RELATED PRs:**
- https://github.com/greenpeace/planet4-master-theme/pull/2367
- https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/1236

<hr>

**NOTES:**
- As Video blocks connected to YouTube, Vimeo, or Soundcloud elements are automatically turned into a YouTube, Vimeo, or Soundcloud block (which are Embed blocks), this PR migrates those kinds of elements to Embed blocks directly.
- The title of the Media blocks is turned into heading blocks.
- The description of the Media blocks is turned into paragraph blocks.
- The Video poster image of the Media blocks is turned into the poster of the video migrated block.

<hr>

**TESTING INSTRUCTIONS:**
1. Run this PR in your local environment.
2. Identify a post using Media blocks or create a new one with them (at the end of this description, you'll find Media blocks in code format that can be copied and pasted into the post editor).
3. Run the migrator command: `npx wp-env run cli wp p4-run-activator`
4. Check whether the migration worked as expected.
5. To reset the environment database back to the latest default content and repeat the tests, run: `npm run db:reset`. After resetting the database, you'll find some posts using Media blocks, such as the 952, 914, 907, 904, 866 and 810.

<hr>

```
[...]

PR #2399, opened 2024-10-04
Add styles to embed elements

mardelnet:
**DESCRIPTION:**

This PR re-adds the styles removed [here](https://github.com/greenpeace/planet4-master-theme/pull/2357/files) from the `Media.scss` file.
By removing the `Media.scss` file many embed elements (such as Youtube or Vimeo iframes) lost their styles, especially the full-width property.
Additionally, the scss code was refactored to avoid as much duplicate code as possible.

- Example without styles: https://www-dev.greenpeace.org/gutenberg/media-block-test-page/
- Example with styles: https://www-dev.greenpeace.org/test-iocaste/media-block-test-page/

### DIFF
--- a/assets/src/scss/base/_mixins.scss
+++ b/assets/src/scss/base/_mixins.scss
@@ -99,6 +99,15 @@
   left: 0;
 }
 
+// Move an element at the top-left corner of its nearest ancestor and set its width and height to 100%.
+@mixin absolute-position {
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+}
+
 // Build a linear gradient with a direction and any number of color stops.
 @mixin linear-gradient($direction, $color-stops...) {
   background: nth(nth($color-stops, 1), 1); /* Old browsers */
--- a/assets/src/scss/core-overrides/embed.scss
+++ b/assets/src/scss/core-overrides/embed.scss
@@ -1,0 +1,131 @@
+// This combination of height: 0, and padding-bottom: 56.25%
+// is to force the aspect ratio of the video to 16:9, based
+// on the width. See: https://css-tricks.com/fluid-width-video/
+@mixin embed-wrapper {
+    position: relative;
+    overflow: hidden;
+    max-width: 100%;
+    max-height: 100%;
+    padding-bottom: 56.25%;
+    height: 0;
+}
+
+@mixin absolute-position {
+    position: absolute;
+    top: 0;
+    left: 0;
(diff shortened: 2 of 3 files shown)


planet-4:

:watch: 2024.10.04 12:02:05

mleray:
It seems strange to me that the Soundcloud embed isn't full width like the rest of them, no?

Screenshot 2024-10-08 at 13 38 06

mardelnet:
> It seems strange to me that the Soundcloud embed isn't full width like the rest of them, no? 🤔

Yes, it is strange.
I just copied the styles as they were.
This should be something to discuss with the design team.

assets/src/scss/core-overrides/embed.scss
```diff
@@ -0,0 +1,132 @@
...
+@mixin embed-wrapper {
+  position: relative;
+  overflow: hidden;
+  max-width: 100%;
+  max-height: 100%;
+  padding-bottom: 56.25%;
+  height: 0;
+}
+
+@mixin absolute-position {
```
mleray, line 13: We might need this mixin for other blocks 🤔 Maybe we could move it to the `_mixins.scss` file?

mleray: I actually see we have this [fill-container](https://github.com/greenpeace/planet4-master-theme/blob/main/assets/src/scss/base/_mixins.scss#L93-L100) mixin in that file, doesn't it do the same thing?

mleray: Okay, thanks for testing! Do you think it makes sense to move this one to the mixins file or not?

assets/src/scss/base/_mixins.scss
```diff
@@ -99,6 +99,15 @@
   left: 0;
 }
 
+// Move an element at the top-left corner of its nearest ancestor and set its width and height to 100%.
+@mixin absolute-position {
+  position: absolute;
+  top: 0;
+  left: 0;
```
mleray, line 106: I'm wondering if for RTL sites this should be `right` instead?

mleray: Although it probably doesn't matter since it's full width/height

mardelnet: It will always be full width and height :)

mleray (approved):
Looks good to me 👍 

It would be good to ask the design team why the Soundcloud embed isn't full width and if we want to change that, but it should be in a follow-up PR 🙂

PR #2349, opened 2024-08-14
PLANET-7489 Migrate Campaign Covers to P4 Columns block

sagarsdeshmukh:
Add a migration script which transform campaign covers block into planet4 columns block

Ref. https://jira.greenpeace.org/browse/PLANET-7489

**Testing:**
- On local dev env, run `npx wp-env run cli wp p4-run-activator` to run the migration script.

[Ref.  ](https://app.gitbook.com/o/-LMm4Q4AuKcwl38JYrxF/s/-M15KrJzoMvhbv4NcO9o/tech/data-migrations)

### Post deploy script execution on test instance -
https://app.circleci.com/pipelines/github/greenpeace/planet4-test-mars/734/workflows/e33ae2e2-ee71-473b-a1b2-b3c3e8fcd1e9/jobs/2351/parallel-runs/0/steps/0-108

```
Running the script: 01-run-p4-activator.sh

Run P4_Activator...
Parsing post 810
Migration successful
Parsing post 69
Migration successful
Parsing post 60
Migration successful
```

**Comparison of same posts on [mars-test](https://www-dev.greenpeace.org/test-mars/wp-admin) (with migration) & [titan-test](https://www-dev.greenpeace.org/test-titan/wp-admin) (without migration) -**

1. Please check the "Covers block" > "Campaign covers style" on page

https://www-dev.greenpeace.org/test-mars/testing-all-blocks-evergreen-template/

Similar page on test-titan instance
https://www-dev.greenpeace.org/test-titan/testing-all-blocks-evergreen-template/#covers-block-2

2.
https://www-dev.greenpeace.org/test-mars/get-informed/people/#:~:text=at%20eleifend%20ullamcorper.-,Our%20campaigns,-%23Consumption

https://www-dev.greenpeace.org/test-titan/get-informed/people/#:~:text=at%20eleifend%20ullamcorper.-,Our%20campaigns,-%23Climate

3.
https://www-dev.greenpeace.org/test-mars/get-informed/energy/

https://www-dev.greenpeace.org/test-titan/get-informed/energy/

PR #1253, opened 2020-12-09
Planet 5705 - Update form inputs UI

mleray:
### Description
- [PLANET-5705](https://jira.greenpeace.org/browse/PLANET-5705): update form input styles (border, fonts, colors)
- [PLANET-5512](https://jira.greenpeace.org/browse/PLANET-5512): add label/placeholder animation

### Related PRs
- styleguide: https://github.com/greenpeace/planet4-styleguide/pull/87
- gutenberg-blocks: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/454

PR #2402, opened 2024-10-10
PLANET-7629 Move Covers block to the master theme

mleray:
### Description

See [PLANET-7629](https://jira.greenpeace.org/browse/PLANET-7629)

It has been removed from the plugin at the same time ([#1253](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/1253)). For now I kept the Campaigns style, I think it would be easier to remove it in a follow-up PR (and after the [migration](https://github.com/greenpeace/planet4-master-theme/pull/2349) has been merged).

### Testing

The Covers block should behave as expected, both new and existing blocks with the different styles. [This page](https://www-dev.greenpeace.org/test-tavros/all-the-covers/) for example has all of them so you can use it for testing!


planet-4:

:watch: 2024.10.10 09:03:42

assets/src/blocks/Covers/CoversEditorScript.js
```diff
@@ -0,0 +1,3 @@
+import {registerCoversBlock} from './CoversBlock';
+
+registerCoversBlock();
```
mardelnet, line 3: Can't we just move the code of the `registerCoversBlock` function to this file?

src/Loader.php
```diff
@@ -211,6 +211,7 @@ public static function add_blocks(): void
         new Blocks\TableOfContents();//NOSONAR
         new Blocks\TakeActionBoxout();//NOSONAR
         new Blocks\Timeline();//NOSONAR
+        new Blocks\Covers();//NOSONAR
```
mardelnet, line 214: This is super nicky picky, but it is not alphabetically ordered 😆

functions.php
mardelnet, line 1: This is an idea for another PR:
It seems to me that such an important piece of code like this one in charge of registering the API endpoints should be part of a separate class and not just a hook in the middle of the `functions.php` file.

mardelnet (commented):
Nice job, @mleray 
Let me give you some initial comments while I complete the review:

The following elements were not brought from the plugin, don't we need them?
- public/images/climate/covers-block.svg
- public/images/oil/covers-block.svg
- tests/e2e/covers-campaign.spec.js

assets/src/blocks/Covers/CoversEditorScript.js
mardelnet: Forget this comment, it does not make sense to introduce changes to a block that will be migrated and removed soon.

functions.php
```diff
@@ -102,6 +102,7 @@ function (): void {
         Api\Search::register_endpoint();
         Api\Settings::register_endpoint();
         Api\ENForm::register_endpoint();
+        Api\Covers::register_endpoint();
```
mardelnet, line 105: This is an idea for another PR:
It seems to me that such an important piece of code like this one in charge of registering the API endpoints should be part of a separate class and not just a hook in the middle of the `functions.php` file.

mleray:
> The following elements were not brought from the plugin, don't we need them?
> 
> * public/images/climate/covers-block.svg
> * public/images/oil/covers-block.svg
> * tests/e2e/covers-campaign.spec.js

The first two are not used anymore, so no. The third one I didn't move because at first I thought I would also retire the Campaigns style in this PR, but since now I think it would be best to handle in a follow-up PR I could bring it back... But also it will be removed soon anyway, so what do you think?

src/Loader.php
mleray: Not nit picky, we should keep it ordered! Updated 👍

mardelnet:
> The first two are not used anymore, so no we don't need them. The third one I didn't move because at first I thought I would also retire the Campaigns style in this PR, but since now I think it would be best to handle in a follow-up PR I could bring it back... But also it will be removed soon anyway, so what do you think?

Let's bring it now so we don't forget about it later.

mardelnet (approved):
LGTM! Good job, @mleray

functions.php
mardelnet: No comments here so I assume you disagree 😆  That's okay

mleray: Sorry no I don't disagree haha! I don't have a strong opinion either way.

PR #2415, opened 2024-10-21
PLANET-7637 Planet 4 Bug Report: Gallery Block - Upload image issue

Osong-Michael:
**Ref: https://jira.greenpeace.org/browse/PLANET-7637**

**Testing:**

When uploading images through the _Upload Tab_ in the pop up, there is no validation for images so non-supported images get uploaded but don't show up. With this fix there is an error messages that prompts the user of the image type being uploaded. Also the block validation error no longer exists for the _Carousel Header_ block.

_To test the Image Validation_
- In the editor add the image block.
- Choose the media library
- Go to the upload tab and click to upload an image.
- Only supported images(png, jpg, gif and ico) are allowed to be uploaded, other image types throw and error.

To test the Carousel Block Validation, try editing a page like the Homepage that already has the Carousel block. The validation error should no longer be present in the console.

### DIFF
--- a/assets/src/blocks/CarouselHeader/StaticCaption.js
+++ b/assets/src/blocks/CarouselHeader/StaticCaption.js
@@ -21,7 +21,7 @@ export const StaticCaption = ({slide}) => (
             data-ga-category="Carousel Header"
             data-ga-action="Call to Action"
             data-ga-label={slide.index}
-            {...slide.link_url_new_tab && {rel: 'noreferrer', target: '_blank'}}
+            {...slide.link_url_new_tab && {rel: 'noreferrer noopener', target: '_blank'}}
           >
             <span>
               {slide.link_text}
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -135,6 +135,7 @@ class MasterSite extends TimberSite
         add_action('pre_insert_term', [$this, 'disallow_insert_term'], 1, 2);
         add_filter('wp_dropdown_users_args', [$this, 'filter_authors'], 10, 1);
         add_filter('wp_image_editors', [$this, 'allowedEditors']);
+        add_filter('wp_handle_upload_prefilter', [$this, 'image_type_validation']);
         add_filter('jpeg_quality', fn () => 60);
         add_filter('http_request_timeout', fn () => 10);
         add_action('after_setup_theme', [$this, 'add_image_sizes']);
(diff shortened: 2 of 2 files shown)


planet-4:

:watch: 2024.10.21 13:18:57

mleray:
@Osong-Michael can you please add step-by-step testing instructions?

mleray:
@Osong-Michael I was able to add an MP3 file

Screenshot 2024-10-21 at 16 47 35

Osong-Michael:
> @Osong-Michael I was able to add an MP3 file in an Image block on [this page](https://www-dev.greenpeace.org/test-tavros/wp-admin/post.php?post=234&action=edit)
> 
> Screenshot 2024-10-21 at 16 47 35

Yea, I just realised that happens actually. You can add any file type using this method 🤔. I guess WP uses the same default upload scheme for all files, because it works similarly in the admin panel, you can upload any file using the upload functionality. Should look into it more.

Osong-Michael:
> > @Osong-Michael I was able to add an MP3 file in an Image block on [this page](https://www-dev.greenpeace.org/test-tavros/wp-admin/post.php?post=234&action=edit)
> > Screenshot 2024-10-21 at 16 47 35
> 
> Yea, I just realised that happens actually. You can add any file type using this method 🤔. I guess WP uses the same default upload scheme for all files, because it works similarly in the admin panel, you can upload any file using the upload functionality. Should look into it more.

There is actually two behaviours to the upload mechanism. If you click on the upload button directly on the Image Block, it filters the options to only accept images that match the extensions WP uses, but if you go the the Media Library tab, then Upload tab, that seems to accept any file type so long as it is accepted by WP.

mleray:
> There is actually two behaviours to the upload mechanism. If you click on the upload button directly on the Image Block, it filters the options to only accept images that match the extensions WP uses, but if you go the the Media Library tab, then Upload tab, that seems to accept any file type so long as it is accepted by WP.

@Osong-Michael what's next then?

Osong-Michael:
> > There is actually two behaviours to the upload mechanism. If you click on the upload button directly on the Image Block, it filters the options to only accept images that match the extensions WP uses, but if you go the the Media Library tab, then Upload tab, that seems to accept any file type so long as it is accepted by WP.
> 
> @Osong-Michael what's next then?

I don't know for sure, maybe @comzeradd may have some feedback regarding this.

comzeradd:
Have we confirmed that the bug happens when unsupported files are being used? Testing on [their dev site](https://www-dev.greenpeace.org/india/en/) I can replace the image with a new image (even upload a new one) but if I click "Save" and refresh the editor I still get a "This block contains unexpected or invalid content." message.

Osong-Michael:
> Have we confirmed that the bug happens when unsupported files are being used? Testing on [their dev site](https://www-dev.greenpeace.org/india/en/) I can replace the image with a new image (even upload a new one) but if I click "Save" and refresh the editor I still get a "This block contains unexpected or invalid content." message.

The validation error is fixed in this PR, just since the ticket title mentioned the upload issue I was looking into it too and found the issues above.

comzeradd:
> The validation error is fixed in this PR, just since the ticket title mentioned the upload issue I was looking into it too and found the issues above.

Ah that makes sense. For the scope of this ticket, I think it's fine to fix the validation issue and then add a sanity check on image types.

To @mleray's point, the validation only checks if an image is within the supported file types. So users indeed can still upload an mp3 file while editing an Image block. But it's probably outside the scope of this PR to look into that too. I think at a later point we can see if we want to introduce per block restrictions. For instance, it doesn't make much sense to use an `.ico` on the Carousel Header either.

mleray (approved):
Looks good to me 👍

PR #2188, opened 2024-01-08
PLANET-7393: Update webpack to v5

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-7393

Update Webpack to match Gutenberg and be able to use up-to-date packages.

[Update dev log](https://gist.github.com/lithrel/c1001e68d907b293f8b239dbf16f6fbe)

## Updated
- webpack 4.40.2 -> 5.89.0 -> 5.95
- @wordpress/scripts 3.3.0 -> 26.19.0 // matching gutenberg
- ajv x -> 8.7.1 // matching gutenberg, required to fix version after packages update
- postcss-loader 3.0.0 -> 6.2.1 // matching gutenberg
- sass 1.57.1 -> 1.69.7 -> 1.77.7 // last version
- sass-loader 8.0.0 -> 13.3.3 -> 16.0.2 // last version
- css-loader 3.6.0 -> 6.8.1 // last version
- mini-css-extract-plugin 0.8.2 -> 2.7.6 // last version
- terser-webpack-plugin 2.2.1 -> 5.3.9 // last version

## Added
- postcss 8.4.16 // required by postcss-loader
- ~@wordpress/element 5.25.0 // explicit dependency after removal of defaultConfig~
- ~@wordpress/block-editor 12.16.0 // explicit dependency after removal of defaultConfig~
- css-minimizer-webpack-plugin 5.0.1 // recommended to replace optimize-css-assets-webpack-plugin
- ~babel-loader 8.2.3 // explicit dependency after removal of defaultConfig~ https://github.com/greenpeace/planet4-master-theme/pull/2344

## Removed
- webpack-fix-style-only-entries // not working with webpack 5, not maintained
- optimize-css-assets-webpack-plugin // not working with webpack 5
- remove-files-webpack-plugin // not needed

## Changed
- specified `mode` and `devtool`  for webpack config
- CI resource class to create zip file is now ~~`large`~~ `xlarge`
  webpack 5 seems to take more time and ressources, job was crashing randomly 

## Test

```console
> npm install --legacy-peer-deps
> npm run build
> npm start
```

## Related PRs (might be merged before Webpack 5)
- https://github.com/greenpeace/planet4-master-theme/pull/2343
- https://github.com/greenpeace/planet4-master-theme/pull/2417
- https://github.com/greenpeace/planet4-master-theme/pull/2418
- https://github.com/greenpeace/planet4-master-theme/pull/2419
- https://github.com/greenpeace/planet4-master-theme/pull/2427

PR #2417, opened 2024-10-21
Fix Carousel Header layout

GP-Dan-Tovbein:
In preparation of #2188 we need to split fixes and improvents in small PRs, and then easy to review and merge.

# About the issue
Is not currently possible to click on controls. That's why the related spec is failing.

### DIFF
--- a/assets/src/scss/blocks/CarouselHeader/CarouselHeaderStyle.scss
+++ b/assets/src/scss/blocks/CarouselHeader/CarouselHeaderStyle.scss
@@ -359,6 +359,7 @@ $medium-image-height: 600px;
       width: 50%;
       position: absolute;
       opacity: 0;
+      pointer-events: none;
 
       html[dir="rtl"] & {
         background-position: bottom left;


mleray (commented):
@GP-Dan-Tovbein can you please clarify what the issue is? Because I'm testing on the [Gutenberg instance homepage](https://www-dev.greenpeace.org/gutenberg/) and I can use the indicators just fine 🤔

planet-4:

:watch: 2024.10.21 19:17:19

mardelnet (commented):
Hey @GP-Dan-Tovbein 
I don't completely understand how changing the `pointer-events` property can fix/affect the carousel buttons. 
If the problem is the buttons cannot be clicked, shouldn't we use for the overlay a `z-index smaller than the buttons', instead?

GP-Dan-Tovbein:
> Hey @GP-Dan-Tovbein I don't completely understand how changing the `pointer-events` property can fix/affect the carousel buttons. If the problem is the buttons cannot be clicked, shouldn't we use for the overlay a `z-index` smaller than the buttons', instead?

The reason is that there is an overlay which is on top of buttons and the buttons cannot be clicked. So, disabling any pointer events to overlay will fix the issue. Regarding `z-index`, is not always a good practice honestly.

mardelnet (approved):

PR #2363, opened 2024-09-02
PLANET-7578: Migrate Content Covers Blocks to Posts List blocks

mardelnet:
Ref: https://jira.greenpeace.org/browse/PLANET-7578

<hr>

**DESCRIPTION:**
This PR migrates the Content Covers blocks to Posts List blocks.

<hr>

**TESTING INSTRUCTIONS:**
1. Run this PR in your local environment.
2. Add a Content Cover block to a post or page. it won't work on the backend because of this PR's changes but it's good enough to test the migration script. You can also directly copy/paste this in the code editor:

```
<!-- wp:planet4-blocks/covers {"title":"Content Cover #0\u003cbr\u003eThis is the title","description":"This is the content cover description.\u003cbr\u003eLayout: grid\u003cbr\u003eRows: 1\u003cbr\u003eTags: -\u003cbr\u003ePost types: -\u003cbr\u003eManual override: 3 pages","posts":[256,810,1113],"readMoreText":"More and more"} -->
<div class="wp-block-planet4-blocks-covers"><div data-hydrate="planet4-blocks/covers" data-attributes="{&quot;cover_type&quot;:&quot;content&quot;,&quot;initialRowsLimit&quot;:1,&quot;title&quot;:&quot;Content Cover #0&lt;br&gt;This is the title&quot;,&quot;description&quot;:&quot;This is the content cover description.&lt;br&gt;Layout: grid&lt;br&gt;Rows: 1&lt;br&gt;Tags: -&lt;br&gt;Post types: -&lt;br&gt;Manual override: 3 pages&quot;,&quot;version&quot;:2,&quot;tags&quot;:[],&quot;post_types&quot;:[],&quot;posts&quot;:[256,810,1113],&quot;layout&quot;:&quot;grid&quot;,&quot;isExample&quot;:false,&quot;readMoreText&quot;:&quot;More and more&quot;,&quot;covers&quot;:[]}"></div></div>
<!-- /wp:planet4-blocks/covers -->

<!-- wp:planet4-blocks/covers {"title":"Content Cover #1\u003cbr\u003eThis is the title","description":"This is the content cover description.\u003cbr\u003eLayout: grid\u003cbr\u003eRows: 1\u003cbr\u003eTags: renewables, consumption\u003cbr\u003ePost types: press release","tags":[7,8],"post_types":[14],"readMoreText":"More and more"} -->
[...]

PR #2418, opened 2024-10-21
Fix Covers block

GP-Dan-Tovbein:
In preparation of #2188 we need to split fixes and improvents in small PRs, and then easy to review and merge.

# About the issue
The covers attributes are being used within CoversV3 but the initialization is right after the import.
```
import {coversV3} from './deprecated/coversV3';
```

```
export const attributes = {
  cover_type: {
    type: 'string',
    default: 'content',
  },
```
 
Screen Shot 2024-10-17 at 07 03 01


mleray (commented):
Why create a new file instead of adding these to `CoversConstants`? Also, the frontend tests are failing

planet-4:

:watch: 2024.10.21 14:17:42

mardelnet (commented):
These changes look good to me, however, before approving them I'd like to know how to replicate the issue because in my local environment, I don't see the error when using the Covers blocks.

mleray:
> These changes look good to me, however, before approving them I'd like to know how to replicate the issue because in my local environment, I don't see the error when using the Covers blocks.

@mardelnet I believe it's because the error is due to the Webpack/Node upgrade

GP-Dan-Tovbein:
> These changes look good to me, however, before approving them I'd like to know how to replicate the issue because in my local environment, I don't see the error when using the Covers blocks.

@mardelnet, @mleray has already [answered](https://github.com/greenpeace/planet4-master-theme/pull/2418#issuecomment-2429430590) your question

mardelnet:
> @mardelnet, @mleray has already [answered](https://github.com/greenpeace/planet4-master-theme/pull/2418#issuecomment-2429430590) your question

Thanks @GP-Dan-Tovbein
Let me know once the `__` function is defined and the rest of the adjustments to pass the frontend tests are done and I'll approve this PR.
@mleray I think the version was already defined [here](https://github.com/greenpeace/planet4-master-theme/blob/469295d0ba8d37289b409fb97da759bcf4ce91a6/assets/src/blocks/Covers/CoversConstants.js#L47), so there's no need for an additional constant value.

mleray:
@mardelnet you're right, I guess we don't need it. However we still need the `__` function, and the indentation should be fixed!

assets/src/blocks/Covers/CoversConstants.js
```diff
@@ -12,3 +14,53 @@ export const COVERS_LAYOUTS = {
 export const CAROUSEL_LAYOUT_COVERS_LIMIT = 12;
 
 export const BLOCK_NAME = 'planet4-blocks/covers';
+
+export const COMMON_ATTRIBUTES = {
```
GP-Dan-Tovbein, line 18: This part of the code has been modified since Sonar reported `"Duplicated Lines (%) on New Code
76,3%"`.

assets/src/blocks/Covers/CoversBlock.js
```diff
@@ -1,67 +1,15 @@
...
 import {coversV3} from './deprecated/coversV3';
-import {BLOCK_NAME, COVERS_TYPES, COVERS_LAYOUTS} from './CoversConstants';
 import {example} from './example.js';
 import {getStyleLabel} from '../../functions/getStyleLabel';
 import {CoversFrontend} from './CoversFrontend.js';
-import {renderToString} from 'react-dom/server';
 
 const {__} = wp.i18n;
 
-export const VERSION = 2;
```
GP-Dan-Tovbein, line 13: Unnecesery declaration

GP-Dan-Tovbein:
@mardelnet all checks passed! Also made some code changes to pass Sonar.

assets/src/blocks/Covers/CoversConstants.js
mleray: I find it weird that we have to split these attributes, sometimes we can ignore SonarCloud and I would do that here, what do you think?

mardelnet: I agree with @mleray, especially considering that we want to get rid of Cover blocks soon. For me, it does not make sense to introduce many changes to them.

assets/src/blocks/Covers/deprecated/coversV3.js
```diff
@@ -1,8 +1,15 @@
...
+
+const {__} = wp.i18n;
 
 export const coversV3 = {
-  attributes,
+  attributes: {
+    ...ATTRIBUTES,
+    readMoreText: {
+      type: 'string',
+      default: __('Load more', 'planet4-blocks'),
```
mardelnet, line 11: Can we keep using "Read more"?
As we want to get rid of Cover blocks soon it doesn't make sense to introduce changes to the block.

assets/src/blocks/Covers/CoversConstants.js
mardelnet: Generally speaking, as you say @GP-Dan-Tovbein  we always have to avoid duplicated code. 
However, IMO we should change this block as minimum as possible as we'll eliminate it soon.

assets/src/blocks/Covers/deprecated/coversV3.js
GP-Dan-Tovbein: We'll need to make this change since it required by its spec.

mardelnet: Do you mean the e2e test `tests/e2e/tools/lib/covers.js`?

GP-Dan-Tovbein: Yes

assets/src/blocks/Covers/CoversConstants.js
GP-Dan-Tovbein: @mardelnet just cuoriosly, what do you mean about "soon"? Are we closer to that date or is a plan for the future? From my point of view.. if we are no closer to the date, I don't think if making tiny changes will affect it. We might always to keep this and all blocks working stable, even if we're going to remove them afterwards.

But again, it's a point of view! If tou think it's fine to revert that change I'll do it! 
What do you both think?

mardelnet: Fair enough asking the meaning of "soon".
The answer is: I don't really know.
I just know that we already have 2 migration tickets ([here](https://github.com/greenpeace/planet4-master-theme/pull/2363) and [here](https://github.com/greenpeace/planet4-master-theme/pull/2349)) for the Cover block, but realistically speaking, completing the migration could take several weeks.

assets/src/blocks/Covers/deprecated/coversV3.js
GP-Dan-Tovbein: To be honest, I'm not sure why it is failing the issue since the [released file is set with "Read More"](https://github.com/greenpeace/planet4-master-theme/blob/main/tests/e2e/tools/lib/covers.js#L109)

GP-Dan-Tovbein: I know what took another value. Fixing now! Thanks for following up on this 🙏

tests/e2e/tools/lib/covers.js
```diff
@@ -106,7 +106,7 @@ async function checkCoversBlock(page, style) {
   // Check Load more button
   const loadMoreBtn = await page.$('button.load-more-btn');
   if (loadMoreBtn) {
-    await expect(page.locator('button.load-more-btn')).toHaveText('Read more');
+    await expect(page.locator('button.load-more-btn')).toHaveText('Load more');
```
GP-Dan-Tovbein, line 109: @mardelnet I think that we should change this text to make the specs working again, since this PR is blocking other PRs. We are only touching its spec, not the block functionality!

See what is going on https://output.circle-artifacts.com/output/job/608801f4-f131-49ac-9e4b-f2e9d05393f6/artifacts/0/usr/artifacts/playwright/report/index.html#?testId=62dc1fe6c3ba2c74379e-0a5f1c3f20abc1629ec0 .

In addition we can follow up specs whenever the Cover's migration is finished.

assets/src/blocks/Covers/deprecated/coversV3.js
GP-Dan-Tovbein: FYI, is rolled back the change again, since we need this change to pass specs.

tests/e2e/tools/lib/covers.js
mleray: I don't understand why we change this, the button text is changed [here](https://github.com/greenpeace/planet4-master-theme/blob/3d089f9746865effc3cb401dd679a4fabb83edfc/tests/e2e/tools/lib/covers.js#L69) and the point of the test is to check that this attribute `readMoreText` is working as expected if changed

mleray: If the test fails without this change of text, it means the `readMoreText` behaviour is broken

GP-Dan-Tovbein: I suspect it takes the Covers's v2. Let me re-check it again.

mleray (approved):
Tested on the instance, all looks good 👍

PR #2422, opened 2024-10-23
PLANET-7614 Add categories filter to News & Stories page

mleray:
### Description

See [PLANET-7614](https://jira.greenpeace.org/browse/PLANET-7614)

### Testing

You can check the [News & Stories page](https://www-dev.greenpeace.org/test-oberon/news-stories/) on the oberon instance. Please also make sure that the other listing pages (tag, category, author...) are not affected!

### DIFF
--- a/assets/src/js/listing_pages.js
+++ b/assets/src/js/listing_pages.js
@@ -1,5 +1,3 @@
-const {__} = wp.i18n;
-
 export const setupListingPages = () => {
   // Setup behaviour for list/grid toggle.
   const listViewToggle = document.querySelector('.list-view-toggle');
@@ -44,12 +42,4 @@ export const setupListingPages = () => {
 
   // Category filter.
   document.getElementById(CATEGORY_FILTER).onchange = updateFilter;
-
-  // Add 'No posts found' text when needed.
-  if (!listingPageContent.querySelector('.wp-block-post-template')) {
-    const noPostsFound = document.createElement('p');
-    noPostsFound.classList.add('listing-page-no-posts-found');
-    noPostsFound.innerHTML = __('No posts found!', 'planet4-master-theme');
-    listingPageContent.appendChild(noPostsFound);
-  }
 };
--- a/assets/src/js/listing_pages.js
+++ b/assets/src/js/listing_pages.js
@@ -1,3 +1,5 @@
+const {__} = wp.i18n;
+
 export const setupListingPages = () => {
   // Setup behaviour for list/grid toggle.
   const listViewToggle = document.querySelector('.list-view-toggle');
@@ -41,4 +43,12 @@ export const setupListingPages = () => {
 
   // Category filter.
   document.getElementById(CATEGORY_FILTER).onchange = updateFilter;
+
+  // Add 'No posts found' text when needed.
+  if (!listingPageContent.querySelector('.wp-block-post-template')) {
+    const noPostsFound = document.createElement('p');
+    noPostsFound.classList.add('listing-page-no-posts-found');
+    noPostsFound.innerHTML = __('No posts found!', 'planet4-master-theme');
+    listingPageContent.appendChild(noPostsFound);
+  }
 };
--- a/assets/src/js/listing_pages.js
+++ b/assets/src/js/listing_pages.js
@@ -32,13 +32,14 @@ export const setupListingPages = () => {
 
   const updateFilter = event => {
     const {target: {id, value}} = event;
-    const urlParams = new URLSearchParams(window.location.search);
+    const newUrl = new URL(window.location.href.split('/page/')[0]);
+
     if (value) {
-      urlParams.set(id, value);
+      newUrl.searchParams.set(id, value);
     } else {
-      urlParams.delete(id);
+      newUrl.searchParams.delete(id);
     }
-    window.location.search = urlParams;
+    window.location.href = newUrl.href;
   };
 
   // Category filter.
--- a/assets/src/js/listing_pages.js
+++ b/assets/src/js/listing_pages.js
@@ -1,4 +1,5 @@
 export const setupListingPages = () => {
+  // Setup behaviour for list/grid toggle.
   const listViewToggle = document.querySelector('.list-view-toggle');
   const gridViewToggle = document.querySelector('.grid-view-toggle');
 
@@ -17,4 +18,27 @@ export const setupListingPages = () => {
 
   listViewToggle.onclick = switchViews;
   gridViewToggle.onclick = switchViews;
+
+  // Setup filters for the News & Stories page.
+  const filters = document.querySelector('.listing-page-filters');
+  if (!filters) {
+    return;
+  }
+
+  // Functions and constants.
+  const CATEGORY_FILTER = 'category';
+
+  const updateFilter = event => {
+    const {target: {id, value}} = event;
+    const urlParams = new URLSearchParams(window.location.search);
+    if (value) {
+      urlParams.set(id, value);
+    } else {
+      urlParams.delete(id);
+    }
+    window.location.search = urlParams;
+  };
+
+  // Category filter.
+  document.getElementById(CATEGORY_FILTER).onchange = updateFilter;
 };
--- a/assets/src/scss/pages/_listing-page.scss
+++ b/assets/src/scss/pages/_listing-page.scss
@@ -376,3 +376,9 @@
     }
   }
 }
+
+.listing-page-no-posts-found {
+  text-align: center;
+  font-size: var(--font-size-m--font-family-primary);
+  padding-top: $sp-4;
+}
--- a/assets/src/scss/pages/_listing-page.scss
+++ b/assets/src/scss/pages/_listing-page.scss
@@ -376,9 +376,3 @@
     }
   }
 }
-
-.listing-page-no-posts-found {
-  text-align: center;
-  font-size: var(--font-size-m--font-family-primary);
-  padding-top: $sp-4;
-}
--- a/assets/src/scss/pages/_listing-page.scss
+++ b/assets/src/scss/pages/_listing-page.scss
@@ -364,3 +364,15 @@
     }
   }
 }
+
+.listing-page-filters {
+  width: 100%;
+
+  .listing-page-select {
+    width: 220px;
+
+    @include mobile-only {
+      width: 100%;
+    }
+  }
+}
--- a/functions.php
+++ b/functions.php
@@ -415,7 +415,12 @@ add_action(
         if (!$query->is_main_query() || is_admin() || !is_home()) {
             return;
         }
-        $category_filter = isset($_GET['category']) ? [$_GET['category']] : [];
+        $category_slug = isset($_GET['category']) ? $_GET['category'] : '';
+        if (!empty($category_slug)) {
+            $category = get_category_by_slug($category_slug);
+            $category_id = $category->term_id;
+        }
+        $category_filter = empty($category_slug) ? [] : [$category_id];
         $query->set('category__in', $category_filter);
     }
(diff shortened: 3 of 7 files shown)


planet-4:

:watch: 2024.10.24 09:28:40

assets/src/js/listing_pages.js
```diff
@@ -17,4 +18,27 @@ export const setupListingPages = () => {
...
+  }
+
+  // Functions and constants.
+  const CATEGORY_FILTER = 'category';
+
+  const updateFilter = event => {
+    const {target: {id, value}} = event;
+    const urlParams = new URLSearchParams(window.location.search);
+    if (value) {
+      urlParams.set(id, value);
```
mardelnet, line 35: Should we pass the category slug instead of the ID as the parameter value?

mleray: To make it more readable?

mleray: It would be a bit more complicated to set up the query later on, since the `category__in` parameter uses the ids

templates/listing-page-filters.twig
```diff
@@ -0,0 +1,17 @@
+<div class="listing-page-filters">
+    <label for="categories" class="visually-hidden">
+        {{ __('Filter by category', 'planet4-master-theme' ) }}
+    </label>
+    <select class="form-select listing-page-select" id="category">
```
mardelnet, line 5: Should we use a more specific ID just in case something else is named as "category" in the future?

templates/listing-page.twig
```diff
@@ -1,7 +1,11 @@
 <div class="page-content container" id="content">
     <div class="wp-block-separator is-style-wide listing-page-separator"></div>
     <div class="d-flex align-items-center justify-content-between listing-page-title">
-        <h2>{{ listing_page_title ?? __( 'All articles', 'planet4-master-theme' ) }}</h2>
+        {% if fn('is_home') %}
```
mardelnet, line 4: Can you explain to me this condition?
In which scenario can I see the `else` statement as true?

assets/src/js/listing_pages.js
mardelnet: > To make it more readable?

Yes. Also to make it more SEO friendly and to follow the [UK example](https://www.greenpeace.org.uk/latest-updates/?topic=air-pollution&entry_type) as stated in the ticket description.

templates/listing-page.twig
mleray: This is a bit confusing on WP's part, but [is_home()](https://developer.wordpress.org/reference/functions/is_home/) what we can use to check that this is in fact the Posts Page and not another listing page (tag, category, etc) We use the same check [here](https://github.com/greenpeace/planet4-master-theme/blob/main/src/ListingPage.php#L47)

templates/listing-page-filters.twig
mleray: We can, I just don't see what else we could add there since this page is literally just a list of posts, but if you are worried about it I don't mind renaming it `category-filter` for example

assets/src/js/listing_pages.js
```diff
@@ -17,4 +18,27 @@ export const setupListingPages = () => {
...
+    } else {
+      urlParams.delete(id);
+    }
+    window.location.search = urlParams;
+  };
+
+  // Category filter.
+  document.getElementById(CATEGORY_FILTER).onchange = updateFilter;
```
mardelnet, line 43: Should we check whether the element exist?

```suggestion
  const categoryElement = document.getElementById(CATEGORY_FILTER);
  if (categoryElement) {
    categoryElement.onchange = updateFilter;
  }
```

assets/src/js/listing_pages.js
mleray: Okay, I'll look into it

assets/src/scss/pages/_listing-page.scss
```diff
@@ -364,3 +364,15 @@
     }
   }
 }
+
+.listing-page-filters {
+  width: 100%;
+
+  .listing-page-select {
+    width: 220px;
```
mardelnet, line 372: If the selected option value is longer than 220px, the arrow icon is covered:

News-Stories-Greenpeace

src/ListingPage.php
```diff
@@ -47,6 +47,9 @@ private function update_context(): void
         $this->context['page_category'] = is_home() ? 'News' : 'Listing Page';
         $this->context['og_type'] = isset($this->context['author']) ? 'profile' : 'website';
 
+        // Filters (News & Stories page only)
+        $this->set_filters();
```
mardelnet, line 51: This is only a personal suggestion:

```suggestion
        $this->context['current_category'] = is_home() && isset($_GET['category']) && $_GET['category'] ? $_GET['category'] : '';
```

templates/listing-page-filters.twig
mardelnet: Don't worry, let's keep "category"

assets/src/js/listing_pages.js
mleray: Hm, we could, I didn't do that because [earlier](https://github.com/greenpeace/planet4-master-theme/pull/2422/files/015960f79578fba419a4e8fe8c71bc7d236eddf6#diff-dbed45bed73d24a7bc0816904ff6f7a067c5aad976a8b2baab355047778346dbR23) I check that the filters div is there

templates/listing-page-filters.twig
```diff
@@ -0,0 +1,17 @@
+<div class="listing-page-filters">
+    <label for="categories" class="visually-hidden">
```
mardelnet, line 2: You set `for="categories"`, but the `<select>` has `id="category"`. 
These IDs should match for accessibility.

functions.php
```diff
@@ -349,6 +349,24 @@ function ($urls, $post_id) {
...
+// Right now only "category" is available.
+add_action(
+    'pre_get_posts',
+    function ($query): void {
+        if (!$query->is_main_query() || is_admin() || !is_home()) {
+            return;
+        }
+        $category_slug = isset($_GET['category']) ? $_GET['category'] : '';
+        if (!empty($category_slug)) {
+            $category = get_category_by_slug($category_slug);
```
mardelnet, line 362: If `get_category_by_slug($category_slug)` doesn’t find a matching category, it will return false, which will cause an error when you try to access `$category->term_id`.

functions.php
```diff
@@ -349,6 +349,24 @@ function ($urls, $post_id) {
...
+            $category_id = $category->term_id;
+        }
+        $category_filter = empty($category_slug) ? [] : [$category_id];
```
mardelnet, line 365: You don’t need to initialize `$category_filter` if the category is empty. You can directly set it within the if statement to improve readability.

functions.php
```diff
@@ -349,6 +349,24 @@ function ($urls, $post_id) {
     2
 );
 
+// Add filters to the News & Stories listing page.
+// Right now only "category" is available.
+add_action(
```
mardelnet, line 354: As this function is specific to the News & Stories listing page, should we move it to the `ListingPage` class?

assets/src/js/listing_pages.js
```diff
@@ -1,4 +1,5 @@
 export const setupListingPages = () => {
+  // Setup behaviour for list/grid toggle.
```
mardelnet, line 2: Thinking loud: should we wait for the DOM to be loaded before executing the code within the `setupListingPages` function?

templates/listing-page.twig
mardelnet: Ohh, I see, thanks!
Maybe adding a comment explaining it will be good enough.

functions.php
mleray: I've tried that but it wasn't applied properly then, probably a priority issue?

functions.php
mleray: Done

functions.php
mleray: Updated

assets/src/scss/pages/_listing-page.scss
mleray: Good catch, fixed!

assets/src/js/listing_pages.js
mleray: We can, I tried adding it actually and it didn't seem to change anything, I know we do it for some of these functions and for others we don't (by "these functions" I mean the ones similar to this one, that we add via `app.js`)

mardelnet (commented):
@mleray 
There's a problem when you try to filter paginated posts.
For example, go to [this page](https://www-dev.greenpeace.org/test-oberon/news-stories/page/2/) and use the category filter.
You'll end up on a [404 page](https://www-dev.greenpeace.org/test-oberon/news-stories/page/2/?category=greenpeace).

mleray:
> @mleray There's a problem when you try to filter paginated posts. 

@mardelnet good catch 🤔 I guess we should reset the pagination when filtering. On it!

Osong-Michael (commented):
Quick question @mleray, could we have like a default message for when no posts match a particular category like in this [example](https://www-dev.greenpeace.org/test-oberon/news-stories/?category=greenpeace)?

mleray:
> Quick question @mleray, could we have like a default message for when no posts match a particular category like in this [example](https://www-dev.greenpeace.org/test-oberon/news-stories/?category=greenpeace)?

@Osong-Michael this is weird, categories with no corresponding posts should not show in the dropdown (see PR description) [This one](https://www-dev.greenpeace.org/test-oberon/category/test/) is an example of that. The category you linked (`Greenpeace`) used to have a related post I think, did you remove it for testing purposes?

mleray:
@Osong-Michael I added a `no posts found` text, ideally we would use the core block for this, but I couldn't figure out how to use translations in the html template 😕

mleray:
@mardelnet pagination issue should be solved now!

assets/src/js/listing_pages.js
```diff
@@ -17,4 +20,36 @@ export const setupListingPages = () => {
...
+      newUrl.searchParams.delete(id);
+    }
+    window.location.href = newUrl.href;
+  };
+
+  // Category filter.
+  document.getElementById(CATEGORY_FILTER).onchange = updateFilter;
+
+  // Add 'No posts found' text when needed.
+  if (!listingPageContent.querySelector('.wp-block-post-template')) {
```
mardelnet, line 49: I'm sorry @mleray, but I am not a big fan of this solution.
Cannot we handle this on the server side?
Maybe we can modify the `get_categories` function used in the `listing-page-filters.twig` file to get only categories from posts, not pages. What do you think?

mleray: You're right, updated!

mardelnet: Now, if I try to access a category that exists but has no associated posts, I see an empty list.
Check it: https://www-dev.greenpeace.org/test-oberon/news-stories/?category=greenpeace

mardelnet: But the rest of the categories behave differently:
https://www-dev.greenpeace.org/test-oberon/news-stories/?category=fake-category

mleray: Hm 🤔 What do you think we should do in this case? Add the "No posts found" message back?

mardelnet: For consistency, I'd keep the same behavior as a nonexistent category: show all the posts and select the "All topics" option in the dropdown.

mleray: That might be tricky, but I'll look into it

mleray: Updated!

mardelnet (approved):
Great job, @mleray

PR #2423, opened 2024-10-23
PLANET-7643: Media Library - Show restrictions on imported media

mardelnet:
Ref: https://jira.greenpeace.org/browse/PLANET-7643

<hr>

**DESCRIPTION:**
This PR fetches restrictions from GP Media API and shows them on the import sidebar before importing the file and also on the Media Library while browsing the media file.

Greenpeace-Media-‹-Greenpeace-—-WordPress(1)

### DIFF
--- a/admin/css/archive-picker.css
+++ b/admin/css/archive-picker.css
@@ -667,3 +667,9 @@
     grid-template-columns: repeat(6, 1fr);
   }
 }
+
+.compat-field-restrictions_text .field,
+.restrictions-warning dd {
+  color: #D43B57; /* var(--red-500); */
+  font-weight: bold;
+}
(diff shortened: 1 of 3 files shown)


comzeradd:
Maybe we could use a more prominent color to draw attention to the restrictions, as it was in the original implementation. Looking at the Design System, we have a [Semantic Red](https://p4-designsystem.greenpeace.org/05f6e9516/p/228275-colors/t/page-228275-72178708-34bf41-28) color we could use.

mardelnet:
> Maybe we could use a more prominent color to draw attention to the restrictions, as it was in the original implementation. Looking at the Design System, we have a [Semantic Red](https://p4-designsystem.greenpeace.org/05f6e9516/p/228275-colors/t/page-228275-72178708-34bf41-28) color we could use.

Good idea. Done!

admin/css/archive-picker.css
```diff
@@ -667,3 +667,9 @@
     grid-template-columns: repeat(6, 1fr);
   }
 }
+
+.compat-field-restrictions_text .field,
+.restrictions-warning dd {
+  color: #D43B57; /* var(--red-500); */
+  font-weight: bold;
+}
```
mleray, line 675: Looks like there's an empty line missing at the end here

mardelnet: Fixed!

mleray (approved):
Looks good to me 👍

PR #2419, opened 2024-10-21
Fix e2e specs

GP-Dan-Tovbein:
In preparation of #2188 we need to split fixes and improvents in small PRs, and then easy to review and merge.

# About the issue
The specs are not passing in Webpack 5 since the `editor.canvas` is `undefined`. It's going to be replaced by `page`.

PR #2343, opened 2024-08-06
PLANET-7530: Upgrade to node 18

GP-Dan-Tovbein:
Ref: https://jira.greenpeace.org/browse/PLANET-7530

# Description
Upgrade to LTS Node version including compatible NPM

Initially testing v18 and see what is going on.

PR #2427, opened 2024-10-28
Add new e2e utility function

GP-Dan-Tovbein:
In preparation of:
- https://github.com/greenpeace/planet4-master-theme/pull/2419 

# Description
This function will help to insert new block into page using the block inserter. The function is not being used by any spec for now.


planet-4:

:watch: 2024.10.28 21:00:14

tests/e2e/tools/lib/editor.js
```diff
@@ -64,10 +66,36 @@ async function removeAllPostTypes({editor}) {
...
+/**
+ * Insert new block into page using the block inserter
+ *
+ * @param {Object} page             - The page object used to interact with the page.
+ * @param {Object} params           - Additional block's parameters.
+ * @param {Object} params.blockName - The name of the block.
+ * @param {Object} params.namespace - The namespace to search if it is needed.
+ * @return {Promise<void>}          - Playwright Locator
+ */
+const searchAndInsertBlock = async (page, {blockName, namespace = ''}) => {
```
GP-Dan-Tovbein, line 78: Is going to be used like this
https://github.com/greenpeace/planet4-master-theme/pull/2419/files#diff-63f0e6680f538b7f8f083348e387fc2b82c29a4f4ff0502c69d21648ef760a46R11

tests/e2e/tools/lib/editor.js
```diff
@@ -1,3 +1,5 @@
+import {expect, Locator} from '@playwright/test'; // eslint-disable-line no-unused-vars
```
mleray, line 1: Why do we need to import `Locator`? I don't see it used

tests/e2e/tools/lib/editor.js
mleray: Why do we need to add this function in a separate PR? Why not in the PR you refer to here?

tests/e2e/tools/lib/editor.js
```diff
@@ -64,10 +66,36 @@ async function removeAllPostTypes({editor}) {
...
+ * Insert new block into page using the block inserter
+ *
+ * @param {Object} page             - The page object used to interact with the page.
+ * @param {Object} params           - Additional block's parameters.
+ * @param {Object} params.blockName - The name of the block.
+ * @param {Object} params.namespace - The namespace to search if it is needed.
+ * @return {Promise<void>}          - Playwright Locator
+ */
+const searchAndInsertBlock = async (page, {blockName, namespace = ''}) => {
+  // Can be also used the aria-label="Toggle block inserter"
```
mleray, line 79: As mentioned in this [comment](https://github.com/greenpeace/planet4-master-theme/pull/2419#discussion_r1810824342) when needing to add several blocks in a row for a test, it seems that the `Toggle block inserter` version is the only one that works so we should maybe use it instead?

tests/e2e/tools/lib/editor.js
GP-Dan-Tovbein: It's used [here](https://github.com/greenpeace/planet4-master-theme/blob/6dafd13dbdc1a92c7f1bcaa4a4b39d53b66d0e51/tests/e2e/tools/lib/editor.js#L6). It's going to be required when we'll update webpack and related deps. Eslint is going to be more strict.

tests/e2e/tools/lib/editor.js
GP-Dan-Tovbein: Since I need to fix the specs fom [this other PR](https://github.com/greenpeace/planet4-master-theme/pull/2343).

tests/e2e/tools/lib/editor.js
GP-Dan-Tovbein: I am not going to fix the specs on tha PR, just make them pass and finally make sure that Node 18 is potentially ready to be merged.

tests/e2e/tools/lib/editor.js
mleray, line 78: In our other functions, we set the first parameter as an object (so here it would be `{editor}` for example) but the other parameters not. Maybe for consistency we should do the same here?

mleray: I was thinking `({page}, blockName, namespace = '')`, what do you think? Also the docs should be updated accordingly

GP-Dan-Tovbein: I'd prefer to use `{blockName, namespace = ''}` as `params` and be more consistent with the rest of functions.

mleray: The rest of the functions don't use an object as second parameter 🤔

GP-Dan-Tovbein: Oh you're correct!

tests/e2e/tools/lib/editor.js
```diff
@@ -64,10 +66,35 @@ async function removeAllPostTypes({editor}) {
   }
 }
 
+/**
+ * Insert new block into page using the block inserter
+ *
+ * @param {{Page}} page
+ * @param {Object} blockName - The name of the block.
```
mleray, line 73: These need to be updated, they are now strings and not objects

mleray (approved):
Looks good to me 👍

PR #2428, opened 2024-10-29
PLANET-7446: Replace lightbox functionality

mardelnet:
Ref:
- https://jira.greenpeace.org/browse/PLANET-7446
- https://jira.greenpeace.org/browse/PLANET-7510

<hr>

**DESCRIPTION:**
This PR is a proof of concept that attempts to remove the custom lightbox functionality and use the native lightbox as a replacement. This PR implements the native lightbox feature in the core image block and the Planet4 gallery block. When expanded, the native lightbox shows the image caption, however, I could not replicate the navigation between images that the custom lightbox offers.

<hr>

**TESTING:**
- https://www-dev.greenpeace.org/test-janus/planet-7446-replace-lightbox-functionality/
- Note: disabling the P4 plugin is recommended, as it contains logic of the gallery block and the custom lightbox.

PR #2430, opened 2024-10-30
PLANET-7446: Force no lightbox

mardelnet:
Ref: https://jira.greenpeace.org/browse/PLANET-7446

<hr>

**DESCRIPTION:**
This PR allows editors to remove the lightbox that is added by default to images in the core image block, the gallery block, the media and text block, and inline images of the paragraph block. In order to do that, the `force-no-lightbox` class has to be added to the corresponding block.

Edit-Page-“Test-Lightbox”-‹-Greenpeace-—-WordPress

<hr>

**TESTING:**
https://www-dev.greenpeace.org/test-tavros/test-lightbox/

<hr>

**RELATED PR:**
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/1259


planet-4:

:watch: 2024.10.30 14:34:51

mleray (commented):
I see that the cursor behaviour is not consistent in all these use cases, sometimes it's `pointer` when there's no lightbox (Gallery for example) or it's not `pointer` when needed (inline images for example)

mleray:
But also this PR would not be needed if we end up going with your [native implementation](https://github.com/greenpeace/planet4-master-theme/pull/2428), right?

mardelnet:
> I see that the cursor behaviour is not consistent in all these use cases, sometimes it's `pointer` when there's no lightbox (Gallery for example) or it's not `pointer` when needed (inline images for example)

Fixed!

mardelnet:
> But also this PR would not be needed if we end up going with your [native implementation](https://github.com/greenpeace/planet4-master-theme/pull/2428), right?

The design team decided not to proceed with the native lightbox since it doesn't implement navigation between slides.

mleray (commented):
Images no longer show full size in the lightbox, they are rendered there at the same size at they are in the page itself... I've added different images to your [page](https://www-dev.greenpeace.org/test-tavros/test-lightbox/) for testing

mleray (commented):
This PR doesn't address the ticket's main issue, which is that we don't respect the "Expand on click" setting. We use the `force-no-caption` class instead. 
If possible we should respect that setting, if not maybe there is a way to remove it using `theme.json`?

mardelnet:
> Images no longer show full size in the lightbox, they are rendered there at the same size at they are in the page itself... I've added different images to your [page](https://www-dev.greenpeace.org/test-tavros/test-lightbox/) for testing

It seems that this problem is unrelated to this PR, as you can see in [this instance](https://www-dev.greenpeace.org/test-rhea/another-test-page/). Apparently, when the image of an Image block is expanded it doesn't use the original size, but the one selected by the editor. Maybe this can be fixed in a follow-up PR.

mardelnet:
> This PR doesn't address the ticket's main issue, which is that we don't respect the "Expand on click" setting. We use the `force-no-caption` class instead. If possible we should respect that setting, if not maybe there is a way to remove it using `theme.json`?

@comzeradd 
This PR addresses many different changes since the solution seems more complex than we initially thought from the ticket description. In a previous conversation with @mleray we agreed on the following items, but before moving on with them and merging the changes, we would like to know your opinion.

In summary, this PR:
- Receives all the logic of the custom lightbox that was removed from the P4 plugin ([check this PR](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/1259)).
- Allows editors to remove the lightbox that is added by default to images in the core image block, the gallery block, the media and text block, and inline images of the paragraph block by adding the `force-no-lightbox` class to the corresponding block (check this PR description for additional details).
- Removes the native "Expand on click" button WordPress adds only to the Image block, in order to use only one solution (adding the `force-no-lightbox`) and avoid confusion among the web editors.

In a follow-up PR, we want to implement a checkbox (or similar) to the blocks sidebar to let editors decide whether to expand images on click, instead of manually adding the `force-no-lightbox` class.
Any thoughts?

mleray:
> It seems that this problem is unrelated to this PR, as you can see in [this instance](https://www-dev.greenpeace.org/test-rhea/another-test-page/). Apparently, when the image of an Image block is expanded it doesn't use the original size, but the one selected by the editor. Maybe this can be fixed in a follow-up PR.

@mardelnet can you please open a bug ticket for this? 🙏 Pretty sure it used to work 🤔

comzeradd:
> In a follow-up PR, we want to implement a checkbox (or similar) to the blocks sidebar to let editors decide whether to expand images on click, instead of manually adding the `force-no-lightbox` class.

Yes, as discussed, let's remove the native toggle and just add the `force-no-lightbox` css class for now.

theme.json
```diff
@@ -179,7 +179,13 @@
...
           ]
         }
-      }
+      },
+      "core/image": {
+				"lightbox": {
+          "enabled": false,
+					"allowEditing": false
+				}
+			}
```
comzeradd, line 188: Looks like you are adding tabs here for indendation 😁 Check out [editorconfig](https://greenpeace-gpi.slack.com/archives/C0160MX64AG/p1729754027541789)

mleray (approved):
Looks good to me, except for Nikos' comment about the indentation in `theme.json` 👍 Tested on the instance and all works as expected 👏

mardelnet:
> > It seems that this problem is unrelated to this PR, as you can see in [this instance](https://www-dev.greenpeace.org/test-rhea/another-test-page/). Apparently, when the image of an Image block is expanded it doesn't use the original size, but the one selected by the editor. Maybe this can be fixed in a follow-up PR.
> 
> @mardelnet can you please open a bug ticket for this? 🙏 Pretty sure it used to work 🤔

Done @mleray 
https://jira.greenpeace.org/browse/PLANET-7649

theme.json
mardelnet: Fixed!

PR #2431, opened 2024-11-04
PLANET-7563 Apply New Link Style to lists in Posts

mleray:
### Description

See latest comments on [PLANET-7563](https://jira.greenpeace.org/browse/PLANET-7563)

This was only working for other post types

### Testing

On [this post](https://www-dev.greenpeace.org/test-deimos/press/1109/duis-posuere-4/) for example, the list link should now have the same styles as our other links.

### DIFF
--- a/assets/src/scss/pages/post/_article-content.scss
+++ b/assets/src/scss/pages/post/_article-content.scss
@@ -101,7 +101,8 @@
 
   p:not(.article-list-item-meta) a,
   > article > ul li a,
-  > article > ol li a {
+  > article > ol li a,
+  > article > li a {
     @include shared-link-styles;
   }
 
--- a/assets/src/scss/pages/post/_article-content.scss
+++ b/assets/src/scss/pages/post/_article-content.scss
@@ -100,8 +100,8 @@
   }
 
   p:not(.article-list-item-meta) a,
-  > ul li a,
-  > ol li a {
+  > article > ul li a,
+  > article > ol li a {
     @include shared-link-styles;
   }
 


assets/src/scss/pages/post/_article-content.scss
```diff
@@ -100,8 +100,8 @@
   }
 
   p:not(.article-list-item-meta) a,
-  > ul li a,
-  > ol li a {
+  ul li a,
```
mleray, line 103: In Posts lists aren't direct children of the `post-content` div, because they're in an `article` element

planet-4:

:watch: 2024.11.04 14:14:22

assets/src/scss/pages/post/_article-content.scss
```diff
@@ -100,8 +100,8 @@
   }
 
   p:not(.article-list-item-meta) a,
-  > ul li a,
-  > ol li a {
+  > article > ul li a,
```
mardelnet, line 103: I wonder if we should remove the `ul` and `ol` from the selector.
In the bizarre scenario where editors add `<li>` elements using the HTML block instead of the List block, the styles are lost ([check here](https://www-dev.greenpeace.org/test-deimos/press/1109/duis-posuere-4/)).
I think, but not sure, that the selector could be even more simple and just be:  `.post-content li a`

mleray: Unfortunately that wouldn't work because we have some blocks that use lists (such as the Articles block, or the new Posts/Actions List) and with this simpler selector then their links would also get the new styles and we don't want that

mleray: I could add `> article > li a`, even though it's an edge case

mardelnet: Works for me!

mleray: Updated!

mardelnet (approved):

PR #2432, opened 2024-11-05
PLANET-6532 Adding a bottom navigation block for pages

Osong-Michael:
**Ref: https://jira.greenpeace.org/browse/PLANET-6532**

**Testing:**

On the [test-instance](https://www-dev.greenpeace.org/test-saturn/get-informed/) or local, when on a page that is part of the Navigation Menu, you should find the block at the bottom just before the footer.

Also, to test _RTL_ locally, you can edit the [base.twig](https://github.com/greenpeace/planet4-master-theme/blob/main/templates/base.twig) file and add `dir="rtl"` to the body tag. Make sure to run `npx wp-env run cli wp timber clear_cache` to see the changes in the front end.


planet-4:

:watch: 2024.11.19 09:04:18

assets/src/scss/pages/_page.scss
```diff
@@ -121,3 +121,16 @@
...
+.bottom-navigation {
+  height: 70px;
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  font-weight: 700;
+  font-family: var(--font-family-heading);
+
+  a:hover {
+    text-decoration: none;
```
mleray, line 134: Is that a request from the design team? Seems weird to me that we want no hover effects, especially after the accessibility work we've done 🤔

mleray (commented):
On this [page](https://www-dev.greenpeace.org/test-saturn/take-action/) for example there is only a next item, not a previous one, in that case shouldn't it be on the right?

functions.php
```diff
@@ -347,6 +354,70 @@ function render_related_posts_block(array $attributes): string
...
+ * @param array  $attributes Array of dynamic attributes to render section.
+ *
+ * @return string HTML markup for front end.
+ */
+// phpcs:ignore SlevomatCodingStandard.Functions.UnusedParameter.UnusedParameter
+function render_navigation_block(array $attributes): string
+{
+    global $post;
+
+    $menu_items = wp_get_nav_menu_items('Navigation Bar Menu');
```
mleray, line 369: I'm wondering if this might cause issues because this string can be translated 🤔 I see that for the navigation bar [this is what we use](https://github.com/greenpeace/planet4-master-theme/blob/main/src/MasterSite.php#L654-L655) to get the items

mleray (commented):
Not sure if it's related to your changes but the [News & Stories page](https://www-dev.greenpeace.org/test-saturn/news-stories/) is broken on the instance

functions.php
```diff
@@ -347,6 +354,70 @@ function render_related_posts_block(array $attributes): string
...
+
+    // Check if the current page is in the menu
+    $current_item = null;
+    foreach ($menu_items as $item) {
+        if ((int)$item->object_id === (int)$post->ID) {
```
mleray, line 374: I think there should be a space after `(int)`

functions.php
```diff
@@ -347,6 +354,70 @@ function render_related_posts_block(array $attributes): string
...
+        }
+        if ($next_item) {
+            $output .= '<a href="' . esc_url($next_item->url) . '">' . esc_html($next_item->title) . '<span> &rarr;</span></a>';
+        }
+    }
+
+    // For child pages, only show link to the parent
+    if ($current_item->menu_item_parent !== 0) {
+        $parent_item = array_filter($menu_items, function ($item) use ($current_item) {
+            return (int)$item->ID === (int)$current_item->menu_item_parent;
```
mleray, line 409: Again, I would add a space after `(int)`

mleray:
In RTL sites if there is only a next item and not a previous one, the arrow remains on the right, but I think it should be moved to the left in that case?

Screenshot 2024-12-04 at 10 32 52

assets/src/scss/pages/_page.scss
```diff
@@ -121,3 +121,57 @@
...
+  }
+
+  &-next {
+    margin-left: auto;
+
+    &::after {
+      content: "\2192";
+      font-weight: bold;
+      margin-left: 7px;
```
mleray, line 143: Can we use variables here? Maybe `$sp-1` which is 8px?

mleray: Also, to make it simpler to accommodate for RTL sites too, you can use `margin-inline-end` or `margin-inline-start`

assets/src/scss/pages/_page.scss
```diff
@@ -121,3 +121,62 @@
...
+    text-decoration: none;
+  }
+
+  &-next {
+    margin-inline-start: auto;
+
+    &::after {
+      content: "\2192";
+      font-weight: var(--font-weight-bold);
+      margin-left: $sp-1;
```
mleray, line 143: Can use `margin-inline-start` here

assets/src/scss/pages/_page.scss
```diff
@@ -121,3 +121,62 @@
...
+html[dir="rtl"] {
+  .bottom-navigation {
+    &-next {
+      margin-inline-start: 0;
+      margin-inline-end: auto;
+
+      &::after {
+        display: inline-block;
+        transform: scaleX(-1);
+        margin-inline-end: $sp-1;
```
mleray, line 165: This should no longer be necessary 🤔 That's the whole point of using `margin-inline-...`

Osong-Michael: The arrows are one direction only, need to flip them when the site changes from LTR to RTL

mleray: Yes I understand that, but that's the transform, the margin should remain the same if you use `margin-inline-end` and `margin-inline-start`

mleray: Btw instead of `scaleX` maybe we should use `rotate`? Seems more clear to me.

assets/src/scss/pages/_page.scss
```diff
@@ -121,3 +121,69 @@
...
+  a:hover {
+    text-decoration: none;
+  }
+
+  &-next {
+    margin-inline-start: auto;
+
+    .bottom-navigation-arrow {
+      background-image: url("../../images/arrow_forward.svg");
+      margin-inline-start: 5px;
```
mleray, line 143: Could we use `$sp-x` (4px) here?

assets/src/scss/pages/_page.scss
```diff
@@ -121,3 +121,69 @@
...
+  height: 10px;
+  width: 20px;
+  background-position: center;
+  background-repeat: no-repeat;
+  margin-bottom: 2px;
+}
+
+html[dir="rtl"] {
+  .bottom-navigation-arrow {
+    display: inline-block;
```
mleray, line 181: This is can be removed, we add it whether it's RTL or not

Osong-Michael:
> I still see some issues on small screens:
> 
> * When cutting the text, there is no ellipsis at the end, maybe you could use our new [clamp-text mixin](https://github.com/greenpeace/planet4-master-theme/blob/main/assets/src/scss/base/_mixins.scss#L313)?
> 
> Screenshot 2024-12-05 at 16 15 44
> * In really small screens (< 400px wide) it looks off:
> 
> Screenshot 2024-12-05 at 16 16 01

Did you run the server, because it doesn't look like the styles got applied.

assets/src/scss/pages/_page.scss
```diff
@@ -121,3 +121,80 @@
...
+      margin-inline-end: $sp-x;
+    }
+  }
+
+  @include small-and-less {
+    a {
+      width: 50%;
+    }
+
+    .bottom-navigation-link-text {
```
mleray, line 165: This doesn't work for sub-items, for example on the test instance for the children of `The issues we work on`

assets/src/scss/pages/_page.scss
```diff
@@ -121,3 +121,79 @@
...
+  font-family: var(--font-family-primary);
+  font-size: var(--font-size-xxs--font-family-primary);
+
+  a {
+    display: flex;
+    align-items: center;
+  }
+
+  a:hover {
+    text-decoration: none;
```
GP-Dan-Tovbein, line 139: @Osong-Michael what about [this comment](https://github.com/greenpeace/planet4-master-theme/pull/2432/files#r1865601305)?

Osong-Michael: @GP-Dan-Tovbein, Houssam asked to remove the underline from the block. To leave it as per the designs

GP-Dan-Tovbein (approved):
Code looks good to me! It looks SO nice 👏 

I have also followed up with Houssam and I decided to create [this tiny PR](https://github.com/greenpeace/planet4-master-theme/pull/2481) in the meanwhile!

Screen Shot 2024-12-16 at 16 04 28

PR #2435, opened 2024-11-06
PLANET-7489: Migrate Campaign Covers Blocks to Columns blocks

mardelnet:
Ref: https://jira.greenpeace.org/browse/PLANET-7489

<hr>

**DESCRIPTION:**
This PR migrates the Campaign Covers blocks to Columns blocks.

<hr>

**TESTING INSTRUCTIONS:**
1. Run this PR in your local environment.
2. Identify a post using a Campaign content block.
3. Run the migrator command: `npx wp-env run cli wp p4-run-activator`
4. Check whether the migration worked as expected.
5. To reset the environment database back to the latest default content and repeat the tests, run: `npm run db:reset`.

PR #2433, opened 2024-11-05
PLANET-7445: Migrate Split Two Columns blocks to Columns blocks

mardelnet:
Ref: https://jira.greenpeace.org/browse/PLANET-7445

<hr>

**DESCRIPTION:**
This PR migrates the Split Two Columns blocks to Columns blocks.

<hr>

**RELATED PR:**
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/1261

<hr>

**TESTING INSTRUCTIONS:**
1. Run this PR in your local environment.
2. Add a Split Two Columns block to a post or page. it won't work on the backend because of this PR's changes but it's good enough to test the migration script. You can also directly copy/paste this in the code editor:

```
<!-- wp:planet4-blocks/split-two-columns {"title":"This is issue name","issue_description":"And this is the description","issue_link_text":"A test link","issue_link_path":"https://www.greenpeace.org/","select_tag":6,"tag_name":"Climate","tag_description":"Our climate is sick but we have the power to heal it and ourselves, and create a greener more equal world.\r\rExtreme weather events, sea level rise and droughts are all symptoms of a global climate on the brink of breakdown.\r\rWe are at a crossroads. If we act now, as a global community, to keep temperature rise below catastrophic levels, we have a chance to re-stabilise our delicate climate.\r\rWe ca","tag_link":"http://www.planet4.test/tag/climate/","button_text":"Get involved","button_link":"http://www.planet4.test/tag/climate/","tag_image_id":344,"edited":{"title":true,"issue_description":true,"issue_link_text":true,"tag_description":false,"button_text":false,"issue_image_id":false,"tag_image_id":false}} -->
[...]

PR #2436, opened 2024-11-06
PLANET-7523 Remove Twitter from Social Media block and rename it Meta

mleray:
### Description

See [PLANET-7523](https://jira.greenpeace.org/browse/PLANET-7523)

This also includes cleaning up the code of all previous Twitter/X customisations since we will now use the Embed block for these instead. The corresponding migration is included in this PR too.

**Note:** it seems X urls were not working anyways 😬 see [here](https://www-dev.greenpeace.org/gutenberg/twitter-x-embeds/) for example

### Testing

**For the block changes:**

Make sure the block is now called `Meta` and it has no more mentions of Twitter/X in its options.

**For the migration:**

1. Run this PR in your local environment.
2. Add a Meta block to a page using a Twitter/X post url ([example](https://x.com/Greenpeace/status/1842512562383913248)), it won't embed the block anymore because of this PR's changes but it's good enough to test the migration script.

You can also directly copy/paste this in the code editor:

```
<!-- wp:planet4-blocks/social-media {"title":"Twitter","description":"A Twitter feed","social_media_url":"https://twitter.com/Greenpeace"} -->
<section class="block social-media-block wp-block-planet4-blocks-social-media"><header><h2 class="page-section-header">Twitter</h2></header><p class="page-section-description">A Twitter feed</p></section>
<!-- /wp:planet4-blocks/social-media -->

<!-- wp:planet4-blocks/social-media {"title":"X","description":"A X post","social_media_url":"https://x.com/Greenpeace/status/1842512562383913248"} -->
<section class="block social-media-block wp-block-planet4-blocks-social-media"><header><h2 class="page-section-header">X</h2></header><p class="page-section-description">A X post</p></section>
<!-- /wp:planet4-blocks/social-media -->
```

4. Run the migrator command: `npx wp-env run cli wp p4-run-activator`
5. Check whether the migration worked as expected. You might need to run `npx wp-env run cli wp embed cache clear <post-id>` to make sure that the new Embeds render in the frontend.
6. To reset the environment database back to the latest default content and repeat the tests, run: `npm run db:reset`.

### DIFF
--- a/src/Migrations/M033MigrateSocialMediaTwitterBlockToEmbedBlock.php
+++ b/src/Migrations/M033MigrateSocialMediaTwitterBlockToEmbedBlock.php
@@ -59,6 +59,7 @@ class M033MigrateSocialMediaTwitterBlockToEmbedBlock extends MigrationScript
      * we want to migrate.
      *
      * @param array $block - A parsed Social Media block.
+     *
      * @return bool - Whether or not the block should be migrated.
      */
     private static function check_is_valid_block(array $block): bool
@@ -105,6 +106,7 @@ class M033MigrateSocialMediaTwitterBlockToEmbedBlock extends MigrationScript
      * and a final block (Embed) for the Twitter/X embed.
      *
      * @param array $block_attrs - The initial Social Media block attributes.
+     *
      * @return array - The transformed block.
      */
     private static function transform_block(array $block_attrs): array
@@ -134,47 +136,18 @@ class M033MigrateSocialMediaTwitterBlockToEmbedBlock extends MigrationScript
             $block_description,
         ) : null;
 
-        $block = [];
-        $block['blockName'] = Utils\Constants::BLOCK_GROUP;
-        $block['attrs']['metadata']['name'] = 'Twitter/X Group';
+        $attrs = [];
+        $attrs['metadata']['name'] = 'Twitter/X Group';
 
-        $block['innerBlocks'] = [];
-        $block['innerBlocks'][0] = $title;
-        $block['innerBlocks'][1] = $description;
-        $block['innerBlocks'][3] = Utils\Functions::create_embed_block(
+        $inner_blocks = [];
+        $inner_blocks[0] = $title;
+        $inner_blocks[1] = $description;
+        $inner_blocks[3] = Utils\Functions::create_embed_block(
             $social_media_url,
             'rich',
             Utils\Constants::TWITTER,
         );
 
-        // IMPORTANT: DO NOT MODIFY THIS FORMAT!
-        $block['innerHTML'] =
-        '<div class="wp-block-group">
-
-
-
-
-
-        </div>';
-
-        // IMPORTANT: DO NOT MODIFY THIS FORMAT!
-        $block['innerContent'] = array (
-            0 => '
-        <div class="wp-block-group">',
-            1 => null,
-            2 => '
-        ',
-            3 => null,
-            4 => '
-        ',
-            5 => null,
-            6 => '
-        ',
-            7 => null,
-            8 => '</div>
-        ',
-        );
-
-        return $block;
+        return Utils\Functions::create_group_block($inner_blocks, $attrs);
     }
 }
--- a/src/Migrations/M033MigrateSocialMediaTwitterBlockToEmbedBlock.php
+++ b/src/Migrations/M033MigrateSocialMediaTwitterBlockToEmbedBlock.php
@@ -34,7 +34,7 @@ class M033MigrateSocialMediaTwitterBlockToEmbedBlock extends MigrationScript
     protected static function execute(MigrationRecord $record): void
     {
         try {
-            // Get the list of posts using Media blocks.
+            // Get the list of posts using Social Media blocks.
             $posts = Utils\Functions::get_posts_using_specific_block(
                 Utils\Constants::BLOCK_SOCIAL_MEDIA,
                 Utils\Constants::ALL_POST_TYPES
@@ -60,12 +60,12 @@ class M033MigrateSocialMediaTwitterBlockToEmbedBlock extends MigrationScript
 
                 $current_post_id = $post->ID; // Store the current post ID
 
-                echo 'Parsing post ', $post->ID, "\n"; // phpcs:ignore
+                echo 'Parsing post ', $current_post_id, "\n"; // phpcs:ignore
 
                 $blocks = $parser->parse($post->post_content);
 
                 if (!is_array($blocks)) {
-                    throw new \Exception("Invalid block structure for post #" . $post->ID);
+                    throw new \Exception("Invalid block structure for post #" . $current_post_id);
                 }
 
                 foreach ($blocks as &$block) {
@@ -90,7 +90,7 @@ class M033MigrateSocialMediaTwitterBlockToEmbedBlock extends MigrationScript
                     }
 
                     // If not, we get the social media URL.
-                    $social_media_url = self::get_social_media_url($block);
+                    $social_media_url = $block['attrs']['social_media_url'] ?? null;
 
                     // If there's no social media url, we do nothing.
                     if (!$social_media_url) {
@@ -132,7 +132,7 @@ class M033MigrateSocialMediaTwitterBlockToEmbedBlock extends MigrationScript
                 }
 
                 $post_update = array(
-                    'ID' => $post->ID,
+                    'ID' => $current_post_id,
                     'post_content' => $new_content,
                 );
 
@@ -140,7 +140,7 @@ class M033MigrateSocialMediaTwitterBlockToEmbedBlock extends MigrationScript
                 $result = wp_update_post($post_update);
 
                 if ($result === 0) {
-                    throw new \Exception("There was an error trying to update the post #" . $post->ID);
+                    throw new \Exception("There was an error trying to update the post #" . $current_post_id);
                 }
 
                 echo "Migration successful\n";
@@ -151,21 +151,7 @@ class M033MigrateSocialMediaTwitterBlockToEmbedBlock extends MigrationScript
             echo $e->getMessage(), "\n";
         }
     }
-
     // phpcs:enable SlevomatCodingStandard.Functions.UnusedParameter
-
-    /**
-     * Get the social media URL from a Social Media block.
-     *
-     * @param array $block - A parsed Media block.
-     * @return string - The social media URL.
-     */
-    private static function get_social_media_url(array $block): mixed
-    {
-        $social_media_url = $block['attrs']['social_media_url'] ?? null;
-
-        return $social_media_url;
-    }
 
     /**
      * Get the source of a Social Media block.
@@ -202,7 +188,7 @@ class M033MigrateSocialMediaTwitterBlockToEmbedBlock extends MigrationScript
 
         $block = [];
         $block['blockName'] = Utils\Constants::BLOCK_GROUP;
-        $block['attrs']['metadata']['name'] = 'Social Media Group';
+        $block['attrs']['metadata']['name'] = 'Twitter/X Group';
 
         $block['innerBlocks'] = [];
         $block['innerBlocks'][0] = $block_title ? self::get_heading_block($block_title) : null;
@@ -241,7 +227,7 @@ class M033MigrateSocialMediaTwitterBlockToEmbedBlock extends MigrationScript
     }
 
     /**
-     * Transform a Media block into an embed block.
+     * Transform a Social Media block into an embed block.
      *
      * @param array $block - A parsed Social Media block.
      * @param string $social_media_url - The Social Media URL.
--- a/src/Migrations/M033MigrateSocialMediaTwitterBlockToEmbedBlock.php
+++ b/src/Migrations/M033MigrateSocialMediaTwitterBlockToEmbedBlock.php
@@ -107,10 +107,14 @@ class M033MigrateSocialMediaTwitterBlockToEmbedBlock extends MigrationScript
                         continue;
                     }
 
-                    // If the source is Twitter, we want to change it to X.
-                    // This avoids an extra redirection.
-                    if ($source === self::SOURCE_TWITTER) {
-                        $social_media_url = str_replace(self::SOURCE_TWITTER, self::SOURCE_X, $social_media_url);
+                    // If the source is X, we want to change it to Twitter.
+                    // Links from x.com don't work in the Embed block at the moment.
+                    if ($source === self::SOURCE_X) {
+                        $social_media_url = str_replace(
+                            self::SOURCE_X . '.com',
+                            self::SOURCE_TWITTER . '.com',
+                            $social_media_url,
+                        );
                     }
 
                     // Transform the Social Media block into an Embed block.
--- a/src/Migrations/M033MigrateSocialMediaTwitterBlockToEmbedBlock.php
+++ b/src/Migrations/M033MigrateSocialMediaTwitterBlockToEmbedBlock.php
@@ -6,25 +6,12 @@ namespace P4\MasterTheme\Migrations;
 
 use P4\MasterTheme\MigrationRecord;
 use P4\MasterTheme\MigrationScript;
-use WP_Block_Parser;
 
 /**
  * Migrate Social Media blocks with Twitter/X urls to Embed blocks.
  */
 class M033MigrateSocialMediaTwitterBlockToEmbedBlock extends MigrationScript
 {
-    private const SOURCE_INSTAGRAM = 'instagram';
-    private const SOURCE_FACEBOOK = 'facebook';
-    private const SOURCE_TWITTER = 'twitter';
-    private const SOURCE_X = 'x';
-
-    private const SOURCE_PATTERNS = [
-        self::SOURCE_INSTAGRAM => '/^(?:(?:http|https):\/\/)?(?:www.)?(?:instagram.com|instagr.am|instagr.com)\/.+$/i',
-        self::SOURCE_FACEBOOK => '/^(?:(?:http|https)?:\/\/)?(?:www\.)?(mbasic.facebook|m\.facebook|facebook|fb)\.(com|me)\/.+$/i',
-        self::SOURCE_TWITTER => '/^(https?\:\/\/)?(www\.)?(twitter\.com)\/.+$/i',
-        self::SOURCE_X => '/^(https?\:\/\/)?(www\.)?x\.com\/.+$/i',
-    ];
-
     /**
      * Perform the actual migration.
      *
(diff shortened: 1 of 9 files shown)


templates/blocks/social-media.twig
mleray, line 1: We previously forgot to move this file to the theme

src/Migrations/M032MigrateSocialMediaTwitterBlockToEmbedBlock.php
```diff
@@ -0,0 +1,360 @@
...
+     * Perform the actual migration.
+     *
+     * @param MigrationRecord $record Information on the execution, can be used to add logs.
+     * phpcs:disable SlevomatCodingStandard.Functions.UnusedParameter -- interface implementation
+     */
+    protected static function execute(MigrationRecord $record): void
+    {
+        try {
+            // Get the list of posts using Media blocks.
+            $posts = Utils\Functions::get_posts_using_specific_block(
```
mardelnet, line 38: This function has to be adjusted to include all post statuses.
Check it out here: https://github.com/greenpeace/planet4-master-theme/blob/a98281c2d697f301c6aa32bdecabde76e69b1e56/src/Migrations/M033MigrateCampaignCoversToP4ColumnsBlock.php#L26

src/Migrations/M032MigrateSocialMediaTwitterBlockToEmbedBlock.php
```diff
@@ -0,0 +1,360 @@
...
+    /**
+     * Perform the actual migration.
+     *
+     * @param MigrationRecord $record Information on the execution, can be used to add logs.
+     * phpcs:disable SlevomatCodingStandard.Functions.UnusedParameter -- interface implementation
+     */
+    protected static function execute(MigrationRecord $record): void
+    {
+        try {
+            // Get the list of posts using Media blocks.
```
mardelnet, line 37:
```suggestion
            // Get the list of posts using Social Media blocks.
```
or
```suggestion
            // Get the list of posts using Meta blocks.
```

src/Migrations/M032MigrateSocialMediaTwitterBlockToEmbedBlock.php
```diff
@@ -0,0 +1,360 @@
...
+
+            // Get all the blocks of each post.
+            foreach ($posts as $post) {
+                if (empty($post->post_content)) {
+                    continue;
+                }
+
+                $current_post_id = $post->ID; // Store the current post ID
+
+                echo 'Parsing post ', $post->ID, "\n"; // phpcs:ignore
```
mardelnet, line 63:
```suggestion
                echo 'Parsing post ', $current_post_id`, "\n"; // phpcs:ignore
```

src/Migrations/M032MigrateSocialMediaTwitterBlockToEmbedBlock.php
```diff
@@ -0,0 +1,360 @@
...
+
+                $blocks = $parser->parse($post->post_content);
+
+                if (!is_array($blocks)) {
+                    throw new \Exception("Invalid block structure for post #" . $post->ID);
```
mardelnet, line 68:
```suggestion
                    throw new \Exception("Invalid block structure for post #" . $current_post_id`);
```

src/Migrations/M032MigrateSocialMediaTwitterBlockToEmbedBlock.php
```diff
@@ -0,0 +1,360 @@
...
+                        continue;
+                    }
+
+                    // Check if the embed is a Facebook page, in which case we do nothing.
+                    if (isset($block['attrs']['embed_type']) && $block['attrs']['embed_type'] === 'facebook_page') {
+                        continue;
+                    }
+
+                    // If not, we get the social media URL.
+                    $social_media_url = self::get_social_media_url($block);
```
mardelnet, line 93: I think we don't need the `get_social_media_url` function:

 ```suggestion
                    $social_media_url = $block['attrs']['social_media_url'] ?? null
```

File: src/Migrations/M032MigrateSocialMediaTwitterBlockToEmbedBlock.php [hunk 6]
```diff
@@ -0,0 +1,360 @@
...
+
+                    // Identify the source of the media (Facebook, Instagram, Twitter, X)
+                    $source = self::identify_source($social_media_url);
+
+                    // If there is no source or social media url, the block is transformed into an empty Embed block.
+                    if (!$source || !$social_media_url) {
```
mardelnet, line 99: Why are we doing this transformation?
Can't we just keep the original Social Media block?

src/Migrations/M032MigrateSocialMediaTwitterBlockToEmbedBlock.php
```diff
@@ -0,0 +1,360 @@
...
+                    $block = self::transform_blocks($block, $social_media_url);
+                }
+
+                // Unset the reference to prevent potential issues.
+                unset($block);
+
+                // Serialize the blocks content.
+                $new_content = serialize_blocks($blocks);
+
+                $post_update = array(
```
mardelnet, line 130: Let's add an extra condition here to avoid unnecessary operations:

```php
if ($post->post_content === $new_content) {
         continue;
}
```

src/Migrations/M032MigrateSocialMediaTwitterBlockToEmbedBlock.php
```diff
@@ -0,0 +1,360 @@
...
+                    'ID' => $post->ID,
+                    'post_content' => $new_content,
+                );
+
+                // Update the post with the replaced blocks.
+                $result = wp_update_post($post_update);
+
+                if ($result === 0) {
+                    throw new \Exception("There was an error trying to update the post #" . $post->ID);
```
mardelnet, line 139:
```suggestion
                    throw new \Exception("There was an error trying to update the post #" . $current_post_id);
```

src/Migrations/M032MigrateSocialMediaTwitterBlockToEmbedBlock.php
```diff
@@ -0,0 +1,360 @@
...
+     * @return array - The transformed block.
+     */
+    private static function transform_blocks(array $block, string $social_media_url): array
+    {
+        $block_title = array_key_exists("title", $block['attrs']) ? $block['attrs']['title'] : null;
+        $block_description = array_key_exists("description", $block['attrs']) ? $block['attrs']['description'] : null;
+
+        $block = [];
+        $block['blockName'] = Utils\Constants::BLOCK_GROUP;
+        $block['attrs']['metadata']['name'] = 'Social Media Group';
```
mardelnet, line 201: I think we can call this something like "Twitter" or "X", right?
Because we're only migrating Twitter elements.

src/Migrations/M032MigrateSocialMediaTwitterBlockToEmbedBlock.php
```diff
@@ -0,0 +1,360 @@
...
+ * Migrate Social Media blocks with Twitter/X urls to Embed blocks.
+ */
+class M032MigrateSocialMediaTwitterBlockToEmbedBlock extends MigrationScript
+{
+    private const SOURCE_INSTAGRAM = 'instagram';
+    private const SOURCE_FACEBOOK = 'facebook';
+    private const SOURCE_TWITTER = 'twitter';
+    private const SOURCE_X = 'x';
+
+    private const SOURCE_PATTERNS = [
```
mardelnet, line 21: Should we double-check whether these services use alternative domains as YouTube does (for example, youtube.com, m.youtube.com, youtu.be, etc.)?

mardelnet (commented):
@mleray @comzeradd 
SonarCloud [is complaining](https://sonarcloud.io/component_measures?id=greenpeace_planet4-master-theme&pullRequest=2436&metric=new_duplicated_lines_density&view=list) about using a lot of duplicated code in the migration scripts. That situation will only get worse with the rest of the migration PRs we are opening.
Should we ignore the warning or move some functions to a separate file and reuse them? If so, should we also refactor the `M031MigrateMediaBlockToAudioVideoBlock` class that was already deployed?

comzeradd:
> SonarCloud [is complaining](https://sonarcloud.io/component_measures?id=greenpeace_planet4-master-theme&pullRequest=2436&metric=new_duplicated_lines_density&view=list) about us using a lot of duplicated code in the migration scripts. That situation will only get worse with the rest of the migration PRs we are opening.
> Should we ignore the warning or move some functions to a separate file and reuse them? If so, should we also refactor the `M031MigrateMediaBlockToAudioVideoBlock` class that was already deployed?

I'm not sure if there is much value in refactoring migrations that have already run. But for current/future ones, it does makes sense. Sonacloud provides an insight on what considers duplication (see screenshot below). Moving those out of the migration file would also decrease its complexity.

Screenshot from 2024-11-08 11-04-08
Screenshot from 2024-11-08 11-05-53

mardelnet:
> I'm not sure if there is much value in refactoring migrations that have already run. But for current/future ones, it does makes sense. Sonacloud provides an insight on what considers duplication (see screenshot below). Moving those out of the migration file would also decrease its complexity.

Thanks @comzeradd 
One more question. The migration PRs #2433 and #2435 were already approved, and have many duplicated codes, but SonarCloud is not firing any alert about them. Should we wait for the reusable functions and refactor those PRs?

src/Migrations/M032MigrateSocialMediaTwitterBlockToEmbedBlock.php
mleray: Yeah I was wondering about that myself and tbh I'm not sure 🤔 We can keep the block as is if it simplifies things, it would also remove some duplicate code

comzeradd:
> The migration PRs #2433 and #2435 were already approved, and have many duplicated codes, but SonarCloud is not firing any alert about them. Should we wait for the reusable functions and refactor those PRs?

Since #2433 seems to be the first migration to go, maybe refactor this one instead and then rebase this one and #2435 to utilize the new functions.

src/Migrations/M033MigrateSocialMediaTwitterBlockToEmbedBlock.php
```diff
@@ -0,0 +1,286 @@
...
+    private const SOURCE_FACEBOOK = 'facebook';
+    private const SOURCE_TWITTER = 'twitter';
+    private const SOURCE_X = 'x';
+
+    private const SOURCE_PATTERNS = [
+        self::SOURCE_INSTAGRAM => '/^(?:(?:http|https):\/\/)?(?:www.)?(?:instagram.com|instagr.am|instagr.com)\/.+$/i',
+        self::SOURCE_FACEBOOK => '/^(?:(?:http|https)?:\/\/)?(?:www\.)?(mbasic.facebook|m\.facebook|facebook|fb)\.(com|me)\/.+$/i',
+        self::SOURCE_TWITTER => '/^(https?\:\/\/)?(www\.)?(twitter\.com)\/.+$/i',
+        self::SOURCE_X => '/^(https?\:\/\/)?(www\.)?x\.com\/.+$/i',
+    ];
```
mardelnet, line 26: What about moving these constants to the `Utils/Constants.php` file just in case we need them in the future for another migration script?

src/Migrations/M033MigrateSocialMediaTwitterBlockToEmbedBlock.php
```diff
@@ -0,0 +1,286 @@
...
+
+    /**
+     * Perform the actual migration.
+     *
+     * @param MigrationRecord $record Information on the execution, can be used to add logs.
+     * phpcs:disable SlevomatCodingStandard.Functions.UnusedParameter -- interface implementation
+     */
+    protected static function execute(MigrationRecord $record): void
```
mardelnet, line 34: This function was completely refactored to make it reusable. You can see the new function [here](https://github.com/greenpeace/planet4-master-theme/blob/0855240dd8f336678a6ead6e97c82d7ed6b3153d/src/Migrations/Utils/Functions.php#L21) and an example of its implementation [here](https://github.com/greenpeace/planet4-master-theme/blob/0855240dd8f336678a6ead6e97c82d7ed6b3153d/src/Migrations/M032MigrateSplit2ColumnBlock.php#L21).

src/Migrations/M033MigrateSocialMediaTwitterBlockToEmbedBlock.php
```diff
@@ -0,0 +1,286 @@
...
+     * @param string $social_media_url - The Social Media URL.
+     * @return array - The transformed block.
+     */
+    private static function transform_blocks(array $block, string $social_media_url): array
+    {
+        $block_title = array_key_exists("title", $block['attrs']) ? $block['attrs']['title'] : null;
+        $block_description = array_key_exists("description", $block['attrs']) ? $block['attrs']['description'] : null;
+
+        $block = [];
+        $block['blockName'] = Utils\Constants::BLOCK_GROUP;
```
mardelnet, line 191: Can you please add a function to the `Utils/Functions.php` file to make the creation of Group blocks reusable?

src/Migrations/M033MigrateSocialMediaTwitterBlockToEmbedBlock.php
```diff
@@ -0,0 +1,286 @@
...
+    {
+        $type = 'rich';
+        $provider = self::SOURCE_TWITTER;
+
+        // IMPORTANT: DO NOT MODIFY THIS FORMAT!
+        $html_content = '<figure class="wp-block-embed is-type-' . $type . ' is-provider-' . $provider . ' wp-block-embed-' . $provider . ' wp-embed-aspect-16-9 wp-has-aspect-ratio"><div class="wp-block-embed__wrapper">
+        ' . $social_media_url . '
+        </div></figure>';
+
+        $block['blockName'] = Utils\Constants::BLOCK_EMBED;
```
mardelnet, line 257: Can you please add a function to the Utils/Functions.php file to make the creation of Embed blocks reusable?

src/Migrations/M033MigrateSocialMediaTwitterBlockToEmbedBlock.php
mleray: Yep I saw that, but in this case there's a bunch of extra `if` cases

mardelnet: That's why the `Utils\Functions::execute_block_migration` receives the callable function `$block_check_callback` as a parameter. You should create an extra function with the `if` conditions and pass it as a parameter, something like this:

```php 
public static function execute(MigrationRecord $record): void
    {
        $check_is_valid_block = function ($block) {
            return self::check_is_valid_block($block);
        };

        $transform_block = function ($block) {
            return self::transform_block($block);
        };

        Utils\Functions::execute_block_migration(
            Utils\Constants::BLOCK_SOCIAL_MEDIA,
            $check_is_valid_block,
            $transform_block,
        );
    }
``` 
```php
private static function check_is_valid_block(array $block): bool
    {
        // Check if the block is valid.
                    if (!is_array($block)) {
                        return false;
                    }

                    // Check if the block has a 'blockName' key.
                    if (!isset($block['blockName'])) {
                        return false;
                    }

                    // Check if the block is a Social Media block.
                    if ($block['blockName'] !== Utils\Constants::BLOCK_SOCIAL_MEDIA) {
                        return false;
                    }

                    // Check if the embed is a Facebook page, in which case we do nothing.
                    if (isset($block['attrs']['embed_type']) && $block['attrs']['embed_type'] === 'facebook_page') {
                        return false;
                    }

                    // If not, we get the social media URL.
                    $social_media_url = $block['attrs']['social_media_url'] ?? null;

                    // If there's no social media url, we do nothing.
                    if (!$social_media_url) {
                        return false;
                    }

                    // Identify the source of the media (Facebook, Instagram, Twitter, X)
                    $source = self::identify_source($social_media_url);

                    // If there's no source, or if it's Facebook or Instagram, we do nothing.
                    // We only want to migrate Twitter/X embeds.
                    if (!$source || $source === self::SOURCE_FACEBOOK || $source === self::SOURCE_INSTAGRAM) {
                        return false;
                    }
    }
``

Let me know if it's not clear enough.

**Review by mardelnet — COMMENTED** [review]
We should not forget to ask for this content to be updated once this PR is merged: https://planet4.greenpeace.org/content/blocks/socials/

**Comment by mleray** [comment]
> We should not forget to ask for this content to be updated once this PR is merged: https://planet4.greenpeace.org/content/blocks/socials/

@mardelnet this is why we have the `FLAG` and `documentation-needed` labels on the ticket 🙂

File: src/Migrations/M033MigrateSocialMediaTwitterBlockToEmbedBlock.php (hunk 12 above)
Comment (mleray) [reply src/Migrations/M033MigrateSocialMediaTwitterBlockToEmbedBlock.php:21]: It was clear, thank you! Updated 👍

File: src/Migrations/Utils/Functions.php [hunk 15]
```diff
@@ -382,4 +382,86 @@ public static function create_media_text_block(
...
+
+    /**
+     * Transform a block into a group of blocks.
+     *
+     * @param array $inner_blocks - The inner blocks for the Group block.
+     * @param string $name - The Group block name.
+     *
+     * @return array - The transformed block.
+     */
+    public static function transform_block_to_group(array $inner_blocks, string $name): array
```
mardelnet, line 429: I think the logic of this function can be simplified a tiny bit:

```php
public static function transform_block_to_group(array $inner_blocks, string $name): array
    {
        $html = array (
            0 => '
        <div class="wp-block-group">',
            1 => null,
            2 => '
        ',
            3 => null,
            4 => '
        ',
            5 => null,
            6 => '
        ',
            7 => null,
            8 => '</div>
        ',
        );

        $block = [];
        $block['blockName'] = Constants::BLOCK_GROUP;
        $block['attrs']['metadata']['name'] = $name;
        $block['innerBlocks'] = $inner_blocks;
        $block['innerHTML'] = $html[0] . $html[8];
        $block['innerContent'] = $html;

        return $block;
    }
```

mardelnet (commented):
@mleray 
Your code looks good to me and it works as expected.
However, I see some issues with the Twitter block and I am not sure if they're related to this PR or not. For instance, after running the migration script, the embedded content is rendered correctly in the page editor, but not in the front (I only see the embed URL). Do you experience the same?

mleray:
> However, I see some issues with the Twitter block and I am not sure if they're related to this PR or not. For instance, after running the migration script, the embedded content is rendered correctly in the page editor, but not in the front (I only see the embed URL). Do you experience the same?

@mardelnet no, I don't have that issue 🤔 It used to happen when we used `x.com` urls instead of `twitter.com` but I haven't seen it since

src/Migrations/Utils/Functions.php
mleray: I personally think that this is less clear than the previous implementation, no?

mardelnet: Let's keep the current implementation then.

mardelnet:
> @mardelnet no, I don't have that issue 🤔 It used to happen when we used `x.com` urls instead of `twitter.com` but I haven't seen it since

This is what happens in my local environment:
[Screencast from 2024-11-12 14-45-45.webm](https://github.com/user-attachments/assets/2a3fa7a4-7f15-403d-b74c-dcf74f2ba8e1)

mleray:
@comzeradd sorry to bother you but could you please test this on local? It works for me as expected but @mardelnet is not seeing the embedded tweets, just the links

comzeradd:
> @comzeradd sorry to bother you but could you please test this on local? It works for me as expected but @mardelnet is not seeing the embedded tweets, just the links

I have the same issue like @mardelnet. After the migration, the embeds both look good in the editor but in the frontend I just see the url.

Some notes:
- on the testing code you have in the description, the first example is not a status (tweet) but the profile page (not sure if that was intended).
- after the migration the `x.com` url has been replaced by `twitter.com`

mleray:
> I have the same issue like @mardelnet. After the migration, the embeds both look good in the editor but in the frontend I just see the url.

Hm 🤔 Okay, I'll test some more then

> Some notes:
> 
> * on the testing code you have in the description, the first example is not a status (tweet) but the profile page (not sure if that was intended).
> * after the migration the `x.com` url has been replaced by `twitter.com`

@comzeradd both of these are intended, the first one just to test this use case and the second one is because we found out that `x.com` links are not supported by the Embed block at the moment so we need to update them.

mleray:
Thank you to @comzeradd who figured out that there was an issue with the embed cache! @mardelnet if you encounter issues on local with just the link rendering in the frontend, you might need to run `npx wp-env run cli wp embed cache clear <post-id>`

src/Migrations/Utils/Constants.php
```diff
@@ -34,7 +35,6 @@ class Constants
         self::POST_TYPES_PAGE,
         self::POST_TYPES_POST,
         self::POST_TYPES_ACTION,
-        self::POST_TYPES_CAMPAIGN,
```
mardelnet, line 37: Should we remove this?

src/Migrations/Utils/Functions.php
```diff
@@ -382,4 +381,90 @@ public static function create_media_text_block(
...
+            3 => null,
+            4 => '
+        ',
+            5 => null,
+            6 => '
+        ',
+            7 => null,
+            8 => '</div>
+        ',
+        );
```
mardelnet, line 460: I really hate this (I know it is not your fault, actually I was the one who started implementing it this way). I just wonder if there is a cleaner way to build the blocks HTML.

mardelnet (approved):

PR #2438, opened 2024-11-13
Fix tag page when redirection is not set

comzeradd:
This is a revert of fc35da662402012b2d67ae0266ac5aac1321a49b

It seems that `get_post` returns a specific post even the argument passed as id is an empty string.

### Testing
- While on `main` branch:
  - Go to a tag and make sure it doesn't have a "Redirect Page" value (see screenshot)
  - Visit that tag page on the frontend, and instead of the tag listing page you should see a specific post
- Switch to this branch:
  - Refresh the above tag and you should see the regular tag listing page

image

### DIFF
--- a/tag.php
+++ b/tag.php
@@ -21,10 +21,10 @@ if (!is_tag()) {
 
 $tag = get_queried_object();
 $redirect_id = get_term_meta($tag->term_id, 'redirect_page', true);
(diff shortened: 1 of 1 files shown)


GP-Dan-Tovbein (approved):
Code looks good to me 👍

PR #2441, opened 2024-11-14
PLANET-7660: Fix checkbox mark alignement

GP-Dan-Tovbein:
Ref: https://jira.greenpeace.org/browse/PLANET-7660

# Description
The layout is broken not only on WP forms (i.e. comments) but also on Gravity Forms.
What I did is to consolidate the layout between both and delegate the `before` and `after` logic to the parent `selector` instead of a sibling `selector` -currently is set to `label`-.

With the current fix the parent can draw the `checkbox UI` only if it is wrapping  an`input[type="checkbox"]`. We might do the same with `radio` buttons afterwards.

## Testing
Navigate through these pages:
- [Post page](https://www-dev.greenpeace.org/test-rhea/press/1113/duis-posuere-6/)
  - Test the checkbox form element
  - Test the checkbox through the comments block _"I agree on providing my name, email and content so that my comment can be stored and displayed in the website."_
- [Search page](https://www-dev.greenpeace.org/test-rhea/?s=)
  - Test the checkboxes assigned to categories

**The mark should be always correctly aligned.**

Remember to check not only `ltr` but also `rtl` sites. You could simply set `document.body.style.direction = "rtl"` through the web console.

Screen Shot 2024-12-09 at 15 00 00

⚠️ After we merge the current PR we'll need to make a new release that fixes the [handbook](https://planet4.greenpeace.org/?s=comments&orderby=_score&f%5Bcat%5D%5BAnalytics%5D=88) also.


planet-4:

:watch: 2024.11.14 15:11:02

assets/src/scss/base/_variables.scss
```diff
@@ -6,6 +6,9 @@ $large-width:             992px;
 $extra-large-width:       1200px;
 $extra-extra-large-width: 1600px;
 
+// Override html selectors
+$checkbox-size: 20px;
```
GP-Dan-Tovbein, line 10: Using this value as variable since it's being used in both files. It will avoid to hve conflict with sizes.

mleray (commented):
The issue isn't solved on the [Search page](https://www-dev.greenpeace.org/test-rhea/?s=oceans&orderby=_score&f%5Bcat%5D%5BEnergy%5D=2), and also not in the form that I've added to your test page. It also still looks broken if the comments checkbox label is only one line long:

Screenshot 2024-11-15 at 10 54 51

mardelnet (commented):
@GP-Dan-Tobin, in the testing instance I see well the checkbox in the comments box, but not in the form:

Duis-posuere-6-Greenpeace
Duis-posuere-6-Greenpeace(1)

GP-Dan-Tovbein:
Thanks both! I've just noticed that the issue is fixed where the text is 2 lines. We need to be sure that it might works for N lines

mardelnet:
Hey @dantovbein 
I don't know whether this PR is expected to be re-reviewed yet, but I still see the same issue.

assets/src/scss/base/_mixins.scss
```diff
@@ -317,3 +317,88 @@ $nested-comment-left-margin: 50px;
   -webkit-line-clamp: $lines;
   overflow: hidden;
 }
+
```
GP-Dan-Tovbein, line 320: I move all the related code into a `mixin`. Why? To have a consistent layout between forms (P4 default and GF).

assets/src/scss/layout/_forms.scss
```diff
@@ -1,87 +1,6 @@
...
-        width: 20px;
-        height: 20px;
-        background: var(--white);
-        border: 1px solid var(--grey-500);
-        border-radius: 2px;
-        left: 0;
-        position: absolute;
-        top: calc((100% - $sp-4x) / 2);
-
-        html[dir="rtl"] & {
```
GP-Dan-Tovbein, line 46: Not neccesary to define a rule `html dir..` if you set `margin-inline-*` like here https://github.com/greenpeace/planet4-master-theme/pull/2441/files#diff-cdb1e0b8794b1c701944bd9d8e648d5915f3ca4aac5b664e823044fbd0da91beR376

GP-Dan-Tovbein:
> Hey @dantovbein I don't know whether this PR is expected to be re-reviewed yet, but I still see the same issue.

Hey @mardelnet I forget to set as draft PR before. It's now ready! Thanks for following up it.

mardelnet (commented):
Nice job @GP-Dan-Tovbein 
Your code looks good to me.
I think there's one more case we should consider: the checkboxes added to the HTML block

About-us-Greenpeace

GP-Dan-Tovbein:
> Nice job @GP-Dan-Tovbein Your code looks good to me. I think there's one more case we should consider: the checkboxes added to the HTML block
> 
> About-us-Greenpeace

Thanks for your review! This is a good catch but to be honest, that is a weird case. I don't think it is really necessary to apply the checkbox layout through this block, that's why we offer GF.

In addition, looking at the international dev or prod instance, this type checkboxes are not being customised (see https://www-dev.greenpeace.org/international/test-checkboxes/)

mardelnet (approved):

comzeradd:
I agree with Dan. Not sure in what use case, editors would start creating checkboxes inside an html block and not being part of a form element.

PR #2442, opened 2024-11-14
Fix Related Posts URL for see all posts

Osong-Michael:
**URL for see all posts link was hard coded, now it has been updated to render dynamically.**

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -302,13 +302,25 @@ function render_related_posts_block(array $attributes): string
     // Encode the query attributes to JSON for the block template
     $query_json = wp_json_encode($attributes['query_attributes'], JSON_UNESCAPED_SLASHES);
 
+    // Dynamically render link to News & Stories page
+    $news_stories_url = '';
+    $news_stories_page = (int) get_option('page_for_posts');
+
+    if ($news_stories_page) {
+        $news_stories_url = get_permalink($news_stories_page);
+    }
+
+    $see_all_link_group = !empty($news_stories_url) ?
+        '<!-- wp:navigation-link {"label":"' . __('See all posts', 'planet4-blocks') . '","url":"' . $news_stories_url . '","className":"see-all-link"} /-->'
+    : '';
+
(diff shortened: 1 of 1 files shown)


functions.php
```diff
@@ -302,13 +302,29 @@ function render_related_posts_block(array $attributes): string
     // Encode the query attributes to JSON for the block template
     $query_json = wp_json_encode($attributes['query_attributes'], JSON_UNESCAPED_SLASHES);
 
+    // Dynamically render link to News & Stories page
+    $news_stories_url = '';
+    $news_stories_page = get_posts([
```
mleray, line 307: This is not the correct way to get the News & Stories page, since the title can be changed. You can get the id the same way we do in other places in the code such as [here](https://github.com/greenpeace/planet4-master-theme/blob/main/src/ListingPage.php#L101), like this: 
```
(int) get_option('page_for_posts')
```

functions.php
```diff
@@ -302,13 +302,29 @@ function render_related_posts_block(array $attributes): string
...
+    }
+
     // Define the HTML output for the block
     $output = '<!-- wp:query ' . $query_json . ' -->
             <div class="wp-block-query posts-list p4-query-loop is-custom-layout-list"><!-- wp:group {"layout":{"type":"flex","justifyContent":"space-between"}} -->
             <div class="wp-block-group related-posts-block"><!-- wp:heading {"lock":{"move":true}} -->
             <h2 class="wp-block-heading">' . __('Related Posts', 'planet4-blocks') . '</h2>
             <!-- /wp:heading -->
-            <!-- wp:navigation-link {"label":"' . __('See all posts', 'planet4-blocks') . '","url":"http://www.planet4.test/news-stories/","className":"see-all-link"} /-->
+            ' . (!empty($news_stories_url) ?
```
mleray, line 324: Ideally we would create this chunk of code as a reusable variable, to avoid duplication

mleray (approved):
Looks good to me 👍

PR #2444, opened 2024-11-15
PLANET-7662: Fix action pages dropdown

GP-Dan-Tovbein:
Ref: https://jira.greenpeace.org/browse/PLANET-7662

The `Select Take Action Page` is now rendered based on IA.

[Demo page](https://www-dev.greenpeace.org/test-tavros/act/test-tab-demo-page/) 

## Testing
- Go to Planet 4 > Navigation and see if new IA feature is enabled
- Create a new page and include a Take Action Boxout block
- If the new IA feature is disabled the list should render only pages that its parents are the same as the one defined trough `/wp-admin/admin.php?page=planet4_settings_navigation`.
- If the new IA feature is enabled the list should include not  only the previous ones but also `p4_action` pages

<!-- Ref: Please add a url to the ticket this change is addressing.

### DIFF
--- a/assets/src/blocks/TakeActionBoxout/TakeActionBoxoutEditor.js
+++ b/assets/src/blocks/TakeActionBoxout/TakeActionBoxoutEditor.js
@@ -37,7 +37,7 @@ export const TakeActionBoxoutEditor = ({
   } = attributes;
 
   const {options: p4_options} = window.p4_vars;
-  const isNewIA = p4_options.new_ia === 'on';
+  const isNewIA = p4_options.new_ia;
 
   const {
     loading,
--- a/assets/src/blocks/TakeActionBoxout/TakeActionBoxoutEditor.js
+++ b/assets/src/blocks/TakeActionBoxout/TakeActionBoxoutEditor.js
@@ -36,6 +36,9 @@ export const TakeActionBoxoutEditor = ({
     stickyOnMobile,
   } = attributes;
 
+  const urlParams = new URLSearchParams(window.location.search);
+  const postId = urlParams.get('post');
+
   const {
     loading,
     actPageList,
@@ -63,7 +66,7 @@ export const TakeActionBoxoutEditor = ({
         (select('core').getEntityRecords('postType', 'p4_action', args) || []) :
         []
       )
-    ).sort((a, b) => {
+    ).filter(a => `${a.id}` !== postId).sort((a, b) => {
       if (a.title.raw === b.title.raw) {
         return 0;
       }
--- a/assets/src/blocks/TakeActionBoxout/TakeActionBoxoutEditor.js
+++ b/assets/src/blocks/TakeActionBoxout/TakeActionBoxoutEditor.js
@@ -51,21 +51,18 @@ export const TakeActionBoxoutEditor = ({
       per_page: -1,
       sort_order: 'asc',
       sort_column: 'post_title',
-      parent: window.p4_vars.take_action_page,
-      post_status: 'publish',
-    };
-
-    const actionsArgs = {
-      per_page: -1,
-      sort_order: 'asc',
-      sort_column: 'post_title',
       post_status: 'publish',
     };
 
     // eslint-disable-next-line no-shadow
     const actPageList = [].concat(
-      select('core').getEntityRecords('postType', 'page', args) || [],
-      select('core').getEntityRecords('postType', 'p4_action', actionsArgs) || []
+      select('core').getEntityRecords('postType', 'page', {
+        ...args, parent: window.p4_vars.options.take_action_page,
+      }) || [],
+      ...(window.p4_vars.options.new_ia === 'on' ?
+        (select('core').getEntityRecords('postType', 'p4_action', args) || []) :
+        []
+      )
     ).sort((a, b) => {
       if (a.title.raw === b.title.raw) {
         return 0;
--- a/assets/src/blocks/TakeActionBoxout/TakeActionBoxoutEditor.js
+++ b/assets/src/blocks/TakeActionBoxout/TakeActionBoxoutEditor.js
@@ -36,8 +36,8 @@ export const TakeActionBoxoutEditor = ({
     stickyOnMobile,
   } = attributes;
 
-  const urlParams = new URLSearchParams(window.location.search);
-  const postId = urlParams.get('post');
+  const {options: p4_options} = window.p4_vars;
+  const isNewIA = p4_options.new_ia === 'on';
 
   const {
     loading,
@@ -50,6 +50,7 @@ export const TakeActionBoxoutEditor = ({
     imageUrl,
     imageAlt,
   } = useSelect(select => {
+    const postId = select('core/editor').getCurrentPostId();
     const args = {
       per_page: -1,
       sort_order: 'asc',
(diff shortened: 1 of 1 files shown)


assets/src/blocks/TakeActionBoxout/TakeActionBoxoutEditor.js
```diff
@@ -47,15 +47,8 @@ export const TakeActionBoxoutEditor = ({
     imageUrl,
     imageAlt,
   } = useSelect(select => {
-    const args = {
-      per_page: -1,
-      sort_order: 'asc',
-      sort_column: 'post_title',
-      parent: window.p4_vars.take_action_page,
```
GP-Dan-Tovbein, line 54: Wrong place to get the variable. It is set `window.p4_vars.options.take_action_page`.
That's why `parent` was always `undefined` and no pages were retrieved.

assets/src/blocks/TakeActionBoxout/TakeActionBoxoutEditor.js
```diff
@@ -64,8 +57,13 @@ export const TakeActionBoxoutEditor = ({
 
     // eslint-disable-next-line no-shadow
     const actPageList = [].concat(
-      select('core').getEntityRecords('postType', 'page', args) || [],
-      select('core').getEntityRecords('postType', 'p4_action', actionsArgs) || []
+      select('core').getEntityRecords('postType', 'page', {
+        ...args, parent: window.p4_vars.options.take_action_page,
+      }) || [],
+      ...(window.p4_vars.options.new_ia === 'on'
```
GP-Dan-Tovbein, line 63: Missed to add the filter when the new IA is enabled.

planet-4:

:watch: 2024.11.15 13:43:57

assets/src/blocks/TakeActionBoxout/TakeActionBoxoutEditor.js
```diff
@@ -64,14 +56,22 @@ export const TakeActionBoxoutEditor = ({
...
+        (select('core').getEntityRecords('postType', 'p4_action', args) || []) :
+        []
+      )
     ).sort((a, b) => {
       if (a.title.raw === b.title.raw) {
         return 0;
       }
       return a.title.raw > b.title.raw ? 1 : -1;
     });
+    /* eslint-disable no-console */
```
mleray, line 72: I'm guessing we want to remove this console log?

mleray (commented):
After testing on the demo instance, I have a few comments:

- The current page shows in the dropdown, I'm wondering if we should remove it?
Screenshot 2024-11-18 at 10 27 15

- If I'm not mistaken, the "Take Action page" only exists when the new IA is enabled. For the old IA we should be using the "Act page" setting as parent instead.

GP-Dan-Tovbein:
> After testing on the demo instance, I have a few comments:
> 
>     * The current page shows in the dropdown, I'm wondering if we should remove it?
> 
> 
> Screenshot 2024-11-18 at 10 27 15
> 
>     * If I'm not mistaken, the "Take Action page" only exists when the new IA is enabled. For the old IA we should be using the "Act page" setting as parent instead.

@mleray about your first question, I totally agree to hide the current page, if fact, I mentioned that to Nikos. It has been never implemented.

About the second question, in theory that logic is solved here https://github.com/greenpeace/planet4-master-theme/pull/2444/files#diff-824c4d396b2b9bca361849bbd824b2e4afed4b8a5ed22db9bc8d6a6f59cd0d7fR62-R64. Have you tested it?

mleray:
> About the second question, in theory that logic is solved here https://github.com/greenpeace/planet4-master-theme/pull/2444/files#diff-824c4d396b2b9bca361849bbd824b2e4afed4b8a5ed22db9bc8d6a6f59cd0d7fR62-R64. Have you tested it?

I have tested it yes, and the Action pages do show as expected with the new IA, my question was more about the other part because we always use the `take_action_page` setting as parent but with the old IA it should be `act_page` instead I think.

assets/src/blocks/TakeActionBoxout/TakeActionBoxoutEditor.js
```diff
@@ -64,9 +59,14 @@ export const TakeActionBoxoutEditor = ({
...
-      select('core').getEntityRecords('postType', 'p4_action', actionsArgs) || []
-    ).sort((a, b) => {
+      select('core').getEntityRecords('postType', 'page', {
+        ...args, parent: window.p4_vars.options.take_action_page,
+      }) || [],
+      ...(window.p4_vars.options.new_ia === 'on' ?
+        (select('core').getEntityRecords('postType', 'p4_action', args) || []) :
+        []
+      )
+    ).filter(a => `${a.id}` !== postId).sort((a, b) => {
```
GP-Dan-Tovbein, line 69: Needs to filter the edited post id

assets/src/blocks/TakeActionBoxout/TakeActionBoxoutEditor.js
```diff
@@ -36,6 +36,9 @@ export const TakeActionBoxoutEditor = ({
     stickyOnMobile,
   } = attributes;
 
+  const urlParams = new URLSearchParams(window.location.search);
+  const postId = urlParams.get('post');
```
mleray, line 40: Maybe [this](https://github.com/greenpeace/planet4-master-theme/blob/main/assets/src/blocks/Articles/ArticlesEditor.js#L148) is a better way to get the post id?

assets/src/blocks/TakeActionBoxout/TakeActionBoxoutEditor.js
```diff
@@ -64,9 +59,14 @@ export const TakeActionBoxoutEditor = ({
 
     // eslint-disable-next-line no-shadow
     const actPageList = [].concat(
-      select('core').getEntityRecords('postType', 'page', args) || [],
-      select('core').getEntityRecords('postType', 'p4_action', actionsArgs) || []
-    ).sort((a, b) => {
+      select('core').getEntityRecords('postType', 'page', {
+        ...args, parent: window.p4_vars.options.take_action_page,
```
mleray, line 63: This still won't work with the old IA, because then we need to use `act_page` instead

assets/src/blocks/TakeActionBoxout/TakeActionBoxoutEditor.js
mleray: I feel like it would be better to compare the ids as numbers, no? Using `parseInt`?

GP-Dan-Tovbein: Remember that `postId` is already a string. If you prefer to compare betweens you might need to parse both values, `a.id` and `postId` as `integer`s. Do you agree?

GP-Dan-Tovbein: And that's why I've implemented as template literals

GP-Dan-Tovbein: I think is less effort to implement template strings, well, it's what is done 😆

assets/src/blocks/TakeActionBoxout/TakeActionBoxoutEditor.js
GP-Dan-Tovbein: I've pushed a new change!

assets/src/blocks/TakeActionBoxout/TakeActionBoxoutEditor.js
```diff
@@ -36,6 +36,9 @@ export const TakeActionBoxoutEditor = ({
     stickyOnMobile,
   } = attributes;
 
+  const {options: p4_options} = window.p4_vars;
+  const isNewIA = p4_options.new_ia === 'on' && true;
```
mleray, line 40: Why `&& true`?

GP-Dan-Tovbein: It is another approach to define `boolean`. Basically, if the condition is `true`, then the value is `thruty`, then is `falsy`.

If the new IA is enabled
Screen Shot 2024-11-26 at 11 07 27

If the new IA is disabled
Screen Shot 2024-11-26 at 11 07 56

mleray: But if you just do `p4_options.new_ia === 'on'` doesn't it return the same?

mleray (approved):
Tested on the demo page, works as expected 👏

PR #2450, opened 2024-11-22
PLANET-7672 Update TAB custom image width in the editor

mleray:
### Description

See [PLANET-7672](https://jira.greenpeace.org/browse/PLANET-7672)

This editor style was overridden by the frontend styles for the block. 

### Testing

Before these changes this is what a custom boxout looks like in the editor: 

GITS00250719

After these changes it should look as expected.

### DIFF
--- a/assets/src/scss/blocks/TakeActionBoxout/TakeActionBoxoutEditorStyle.scss
+++ b/assets/src/scss/blocks/TakeActionBoxout/TakeActionBoxoutEditorStyle.scss
@@ -9,25 +9,25 @@
     min-width: auto !important;
   }
 
+  // I had to put these styles on the container to get the image controls to work, and remove some from the image itself.
+  .boxout-image-container {
+    position: relative;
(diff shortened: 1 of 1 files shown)


assets/src/scss/blocks/TakeActionBoxout/TakeActionBoxoutEditorStyle.scss
```diff
@@ -27,7 +27,7 @@
   justify-content: center;
 
   img {
-    width: 100%;
+    width: 100% !important;
```
mardelnet, line 30: I think that you can move the `boxout-image-container` class inside of the `boxout` class and avoid the `!important` statement.

mleray: You are absolutely right! Updated 🙂

mardelnet (approved):

PR #2452, opened 2024-11-25
Create mixin to clamp text

mleray:
### Description

This removes some code duplication


assets/src/scss/pages/_listing-page.scss
```diff
@@ -100,27 +100,18 @@
...
   margin-bottom: $sp-1;
   line-height: var(--line-height-m--font-family-secondary);
-  line-clamp: 4;
-  -webkit-line-clamp: 4;
+  @include clamp-text(4);
 }
 
-.query-list-item-headline {
-  font-size: $font-size-lg !important;
+.query-list-item-headline.wp-block-post-title {
```
mleray, line 110: Without this extra class the selector isn't specific enough and gets overridden

mleray: Hence why there used to be `!important`

PR #2459, opened 2024-11-28
Simplify CF zone code to only remove www. from hostname

comzeradd:
We made a change to accommodate for domains with more parts (like `.org.au`), but that broke `ummah4earth.org` because they don't have `www` in their hostname.

Since we practically only want to remove `www`, this simplifies that part of the code to avoid complicated conditional logic for all use cases.

### Testing 

It's hard to test directly with Cloudflare, but I moved the code to a different file and use php cli to test out the logic with the different use cases.


planet-4:

:watch: 2024.11.28 09:56:23

sagarsdeshmukh (approved):
👍

PR #2462, opened 2024-11-29
PLANET-7482 Use proper function for displaying author bio

comzeradd:
We are currently using [the_author_meta](https://developer.wordpress.org/reference/functions/the_author_meta/).
But the function doesn't return a string, which is what [wpautop](https://developer.wordpress.org/reference/functions/wpautop/) expects. This also triggers the following warning locally:

```
Deprecated: trim(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/html/wp-includes/formatting.php on line 449
```

Instead, we should use [get_the_author_meta](https://developer.wordpress.org/reference/functions/get_the_author_meta/) as it's also mentioned in the documentation of `the_author_meta`:

>NOTE: Use `get_the_author_meta()` if you need to return (and do something with) the field, rather than just display it.

This also fixes bug [PLANET-7482](https://jira.greenpeace.org/browse/PLANET-7482).
So it includes some css changes to make the links also follow the correct design.

On the Authors block (Post view), when the bio is bigger than our character limit, the `wpautop` is not used
as it's hard to keep the formatting while breaking up the description text.

### Testing
- Use these two bio examples (short and long):
  ```
  This is test short bio.

  With a few lines and a <b>bold</b> markup.

  <a href="https://twitter.com/Greenpeace">Follow on Twitter</a>.
  ```

  ```
  This is longer bio, to test collapse/expand functionality.

  It still has a few lines and a <b>bold</b> markup, but it should appear as one paragraph on the Author block.

  Follow on <a href="https://twitter.com/Greenpeace">Twitter</a>.
  ```

- Check both the author block on Post (that requires the user is assigned a post) and the Author page

- Elements to check:
  -- Author page and Author block: links should follow the default design
  -- Author page and Author block short version: New lines should be converted to paragraphs

### DIFF
--- a/assets/src/scss/pages/_author.scss
+++ b/assets/src/scss/pages/_author.scss
@@ -49,6 +49,10 @@
   display: flex;
   flex-direction: column;
   justify-content: center;
+
+  a {
+    @include shared-link-styles;
+  }
 }
 
 .author-image {
--- a/assets/src/scss/pages/post/_author-block.scss
+++ b/assets/src/scss/pages/post/_author-block.scss
@@ -63,18 +63,19 @@
   }
 }
 
-.author-block-info {
-  font-size: $font-size-sm;
-  line-height: 1.5;
-  margin: 0;
+.author-bio {
+  p {
+    font-size: $font-size-sm;
+    line-height: 1.5;
+    margin: 0;
 
-  .collapsing {
-    display: none;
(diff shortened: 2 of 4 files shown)


planet-4:

:watch: 2024.11.29 13:26:21

PR #2465, opened 2024-12-04
PLANET-7671 Add post type filter to News & Stories page

mleray:
### Description

See [PLANET-7671](https://jira.greenpeace.org/browse/PLANET-7671)

This is in addition to the previously added category filter

**Note:** for the mobile version we don't use the modal in the end.

### Testing

You can check out the new filter and "Apply" button on the [News & Stories page](https://www-dev.greenpeace.org/test-leda/news-stories/) from the leda test instance or on local. Make sure to test several filter combinations and also all screen sizes.

### DIFF
--- a/templates/listing-page-filters.twig
+++ b/templates/listing-page-filters.twig
@@ -14,4 +14,17 @@
             </option>
         {% endfor %}
     </select>
+    <select class="form-select listing-page-select" id="post-type">
+        <option value="">{{__( 'All posts', 'planet4-master-theme' ) }}</option>
+        {% for key,post_type in post_types %}
+            {% set isSelected = current_post_type == post_type.slug %}
+            <option
+                value="{{post_type.slug}}"
+                {{isSelected ? 'selected' : ''}}
+            >
+                {{ post_type.name|raw }}
+            </option>
+        {% endfor %}
+    </select>
+    <button id="apply-filters" class="btn btn-primary">{{__('Apply', 'planet4-master-theme')}}</button>
 </div>
--- a/assets/src/js/listing_pages.js
+++ b/assets/src/js/listing_pages.js
@@ -25,21 +25,22 @@ export const setupListingPages = () => {
     return;
   }
 
-  // Functions and constants.
-  const CATEGORY_FILTER = 'category';
+  const AVAILABLE_FILTERS = ['category', 'post-type'];
 
-  const updateFilter = event => {
-    const {target: {id, value}} = event;
+  const updateFilters = () => {
     const newUrl = new URL(window.location.href.split('/page/')[0]);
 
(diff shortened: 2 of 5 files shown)


planet-4:

:watch: 2024.12.04 15:05:59

templates/listing-page-filters.twig
```diff
@@ -14,4 +14,17 @@
             </option>
         {% endfor %}
     </select>
+    <select class="form-select listing-page-select" id="post-type">
+        <option value="">{{__( 'All posts', 'planet4-master-theme' ) }}</option>
+        {% for key,post_type in post_types %}
+            {% set isSelected = current_post_type == post_type.slug %}
```
GP-Dan-Tovbein, line 20: What do you think to change to 
```suggestion
            {% set isSelected = current_post_type === post_type.slug %}
```

templates/listing-page-filters.twig
```diff
@@ -14,4 +14,17 @@
...
+            <option
+                value="{{post_type.slug}}"
+                {{isSelected ? 'selected' : ''}}
```
GP-Dan-Tovbein, line 23: I think in this case could be better if we keep the attribute with the proper boolean.
```suggestion
    selected={{isSelected}}
```
What do you think?

GP-Dan-Tovbein (commented):
Code looks good to me! I've left small comments which are not required to change.

I grabbed a video to show you the only "issue" I've found. 

https://github.com/user-attachments/assets/20da06e6-5fdf-4c77-84b9-31d189fdca6c

Do you think we can add also the layout as a param?

mleray:
> Do you think we can add also the layout as a param?

@comzeradd what do you think about this? Should we try to find a way to also preserve the layout (grid or list) when filtering?

comzeradd:
> @comzeradd what do you think about this? Should we try to find a way to also preserve the layout (grid or list) when filtering?

Yes, but if it requires refactoring some of the layout picker code let's do that in a follow up ticket.

mleray:
@GP-Dan-Tovbein @comzeradd I've created a [ticket](https://jira.greenpeace.org/browse/PLANET-7706) for keeping the layout after reloading

GP-Dan-Tovbein (approved):
Code looks good to me 🥳

templates/listing-page-filters.twig
mleray: We usually use `==` in twig templates, I see only in the google tag manager template we use `===` in one place 🤔

templates/listing-page-filters.twig
mleray: This doesn't work, it results in this:

Screenshot 2024-12-17 at 15 20 56

PR #2472, opened 2024-12-10
PLANET-7656 Prepare for WordPress 6.7 upgrade

mleray:
### Description

See [PLANET-7656](https://jira.greenpeace.org/browse/PLANET-7656)

**Related PRs:**
- https://github.com/greenpeace/planet4-base/pull/320
- https://github.com/greenpeace/planet4-develop/pull/47

**Changes made for the upgrade:**
- The core Posts List variation of the Query Loop block [has been removed](https://github.com/WordPress/gutenberg/pull/63404) so we don't need to unregister it anymore
- Some e2e tests were failing
- The Posts List and Actions List block no longer had the controls to set how many items to display, the offset and max number of pages... These [have been moved to the sidebar](https://github.com/WordPress/gutenberg/pull/58207) and we need to add them to the `allowedControls`.

**Things I noticed but don't need changes or I'm not sure they are problematic:**
- WordPress 6.7 adds `sizes=”auto”` for lazy-loaded images (see [here](https://make.wordpress.org/core/2024/10/18/auto-sizes-for-lazy-loaded-images-in-wordpress-6-7/)), I haven't seen it happen for our blocks though, maybe because we add our own `sizes` attribute?
- There is a [new setting](https://make.wordpress.org/core/2024/10/20/miscellaneous-block-editor-changes-in-wordpress-6-7/#new-preference-to-disable-the-choose-a-pattern-modal-when-adding-pages) in the editor preferences to hide the starter modal with our patterns, maybe some NROs don't use them and would be interested in this?
- We should be able to close [this bug ticket](https://jira.greenpeace.org/browse/PLANET-7040) after the upgrade (see comments there)

### Testing

You can update your local to the latest WordPress version (6.7.1) by running `npx wp-env run cli wp core update`. The [janus test instance](https://www-dev.greenpeace.org/test-janus/) used for this PR has already been updated.

### DIFF
--- a/assets/src/blocks/ActionsList/index.js
+++ b/assets/src/blocks/ActionsList/index.js
@@ -20,7 +20,7 @@ export const registerActionsListBlock = () => {
     description: __('Integrate images and text cards to automatically display tags, take action pages, or Posts in a three or four column layout displayed in a grid or carousel.', 'planet4-blocks-backend'),
     icon: 'list-view',
     scope: ['inserter'],
-    allowedControls: ['taxQuery'],
+    allowedControls: ['taxQuery', 'pages', 'offset'],
     category: 'planet4-blocks-beta',
     isActive: ({namespace, query}) => namespace === ACTIONS_LIST_BLOCK_NAME && query.postType === queryPostType,
     attributes: {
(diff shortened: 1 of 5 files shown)


planet-4:

:watch: 2024.12.10 19:45:04

comzeradd:
From the findings list, one thing we should probably fix is the "how many items to display" on our QLB variations. As we offer that option on Covers and Articles, editors may look for that option in these two new blocks too. If you think it requires a bigger amount of work we can fix that in a follow up ticket.

mleray:
> From the findings list, one thing we should probably fix is the "how many items to display" on our QLB variations. As we offer that option on Covers and Articles, editors may look for that option in these two new blocks too. If you think it requires a bigger amount of work we can fix that in a follow up ticket.

@comzeradd I've updated the PR and it should work now! However I think that we should set a `min` and `max`, because currently it's from 1 to 100, what do you think?

comzeradd (approved):
✅

PR #2475, opened 2024-12-12
PLANET-7697 Add Posts/Actions List indicators only when needed

mleray:
### Description

See [PLANET-7697](https://jira.greenpeace.org/browse/PLANET-7697)

This is for the carousel style

### Testing

Here is an [example page](https://www-dev.greenpeace.org/test-deimos/carousels/) with the different possibilities!

### DIFF
--- a/assets/src/js/query_loop_carousel.js
+++ b/assets/src/js/query_loop_carousel.js
@@ -35,7 +35,7 @@ const removeArrows = layout => ARROW_DIRECTIONS.forEach(direction => {
  */
 export const setupQueryLoopCarousel = () => {
   for (const layout of document.querySelectorAll('[class*="is-custom-layout-"]')) {
-    const hasValidLayout = layout && LAYOUTS.keys().find(l => layout.className.includes(l));
+    const hasValidLayout = layout && Object.keys(LAYOUTS).find(l => layout.className.includes(l));
     if (!hasValidLayout) {
       return;
     }
--- a/assets/src/js/query_loop_carousel.js
+++ b/assets/src/js/query_loop_carousel.js
@@ -1,5 +1,12 @@
 import {v4 as uuid} from 'uuid';
 
+const ARROW_DIRECTIONS = ['prev', 'next'];
+
+const removeArrows = layout => ARROW_DIRECTIONS.forEach(direction => {
+  const controlBtn = layout.querySelector(`.wp-block-buttons .carousel-control-${direction}`);
+  controlBtn?.parentNode.removeChild(controlBtn);
+});
+
 export const setupQueryLoopCarousel = () => {
   for (const layout of document.querySelectorAll('[class*="is-custom-layout-"]')) {
     const list = layout.querySelector('.wp-block-post-template');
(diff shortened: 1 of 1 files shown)


planet-4:

:watch: 2024.12.12 15:08:22

assets/src/js/query_loop_carousel.js
```diff
@@ -1,8 +1,16 @@
 import {v4 as uuid} from 'uuid';
 
+const ARROW_DIRECTIONS = ['prev', 'next'];
+
+const removeArrows = layout => ARROW_DIRECTIONS.forEach(direction => {
+  const controlBtn = layout.querySelector(`.wp-block-buttons .carousel-control-${direction}`);
+  controlBtn?.parentNode.removeChild(controlBtn);
```
mardelnet, line 7: Should we check whether the `parentNode` also exists?

```suggestion
if (controlBtn && controlBtn.parentNode) {
  controlBtn.parentNode.removeChild(controlBtn);
}
```

assets/src/js/query_loop_carousel.js
```diff
@@ -1,8 +1,16 @@
 import {v4 as uuid} from 'uuid';
 
+const ARROW_DIRECTIONS = ['prev', 'next'];
+
+const removeArrows = layout => ARROW_DIRECTIONS.forEach(direction => {
+  const controlBtn = layout.querySelector(`.wp-block-buttons .carousel-control-${direction}`);
```
mardelnet, line 6: Should we add an early return if `layout` is falsy?

assets/src/js/query_loop_carousel.js
```diff
@@ -21,70 +29,76 @@ export const setupQueryLoopCarousel = () => {
...
+      if (posts.length > itemsPerSlide) {
+        indicators = document.createElement('ol');
+        indicators.classList.add('carousel-indicators');
+        carousel.append(indicators);
+
+        // Update controls
+        ARROW_DIRECTIONS.forEach(direction => {
+          const controlBtn = layout.querySelector(`.wp-block-buttons .carousel-control-${direction}`);
+          controlBtn.dataset.bsTarget = `#${uniqueId}`;
+          controlBtn.dataset.bsSlide = direction;
```
mardelnet, line 47: If `controlBtn` is not found, accessing `controlBtn.dataset` or `controlBtn.querySelector` will throw an error.

assets/src/js/query_loop_carousel.js
```diff
@@ -1,8 +1,16 @@
...
+
+const removeArrows = layout => ARROW_DIRECTIONS.forEach(direction => {
+  const controlBtn = layout.querySelector(`.wp-block-buttons .carousel-control-${direction}`);
+  controlBtn?.parentNode.removeChild(controlBtn);
+});
+
 export const setupQueryLoopCarousel = () => {
   for (const layout of document.querySelectorAll('[class*="is-custom-layout-"]')) {
     const list = layout.querySelector('.wp-block-post-template');
+    let indicators = null;
```
mardelnet, line 13: Should we add an early return if `layout` is falsy?

mardelnet (commented):
Nice job, @mleray 
I have a couple of comments:

1. Can we add some JS Docs to this file?
2. I see there are a bunch of repeated values (`wp-block-buttons`, `carousel`, etc...). Should we create constants for them?
3. What about simplifying the code by replacing the nested conditions with guard classes?

assets/src/js/query_loop_carousel.js
mleray: It should be okay, I've added a check for `layout` in the main function and we only use this one there

mardelnet (commented):
LGTM @mleray 
Just out of curiosity before approving this, can't we add the arrows only when the conditions are met instead of always adding them and then removing them by calling the `removeArrows` function?

mleray:
> LGTM @mleray Just out of curiosity before approving this, can't we add the arrows only when the conditions are met instead of always adding them and then removing them by calling the `removeArrows` function?

@mardelnet we could, my guess is that these elements are a bit more complicated with all the classes than the indicators, and that's why it was easier to define them directly in the block's inner blocks

Screenshot 2024-12-16 at 14 06 11

mardelnet (approved):

PR #2487, opened 2024-12-19
PLANET-7676 Added snippet for VWO Anti Flickering

Osong-Michael:
### Summary

---

**Ref: https://jira.greenpeace.org/browse/PLANET-7676**

### Testing

1. When the _VWO A/B testing_ option is enabled in _Planet4_ > _Analytics_ the VWO script to prevent flickering during testing is added to the site. 
2. Add the account ID provided by _VWO_ as this is essential to add the _VWO_ script to the head tag for A/B testing.
3. When the option is enabled and the Account ID provided, the _VWO_ Smart code is added to the head tag.


planet-4:

:watch: 2024.12.19 11:29:36

sagarsdeshmukh (dismissed):
PR looks good to me 👍

src/Settings.php
```diff
@@ -506,12 +506,21 @@ public function __construct()
...
                     ],
+                    [
+                        'name' => __('VWO Account ID', 'planet4-master-theme-backend'),
+                        'desc' => __(
+                            'ID provided by VWO. Please make sure the VWO Testing option above is enabled.',
+                            'planet4-master-theme-backend'
+                        ),
+                        'id' => 'vwo_account_id',
+                        'type' => 'text',
+                    ],
```
comzeradd, line 523: Maybe we can have just this one setting. If it's empty, we do nothing. If there is an Account ID set here then we do both CSP and anti-flickering.

In that case we can use the `desc` text from the previous setting here to explain what will happen if they add the Accound ID. _"Adds a Content Security Policy exception for VWO and enables its anti-flicker mechanism."_

templates/head/vwo_test_script.twig
```diff
@@ -0,0 +1,12 @@
+<!-- Start VWO Async SmartCode -->
+<link rel="preconnect" href="https://dev.visualwebsiteoptimizer.com" />
+<script type='text/javascript' id='vwoCode'>
```
comzeradd, line 3: Let's move the JS code out of the twig tempalte to be consistent with the @mardelnet's recent work at #2502

comzeradd (changes requested):

assets/src/js/vwo_smart_code.js
```diff
@@ -0,0 +1,23 @@
+/* global vwoSmartScriptCodeData */
+
+if (vwoSmartScriptCodeData?.vwo_account_id) {
+  const vwoScript = `
+    <!-- Start VWO Async SmartCode -->
+    <link rel="preconnect" href="https://dev.visualwebsiteoptimizer.com" />
```
comzeradd, line 6: I know this is coming from VWO docs but maybe better to use `dns-prefetch` instead. It's a "cheaper" operation and it's consistent with what [we do already](https://github.com/greenpeace/planet4-master-theme/blob/4266c0e68b3dbcb15e027a46ef618581ce19a296/templates/html-header.twig#L40) with other services

comzeradd (approved):

PR #2492, opened 2025-01-06
PLANET-7657: Add support for Take Action pages

GP-Dan-Tovbein:
### Summary

The ActionsList block might render the same content as the Covers block.

[Demo page as a reference](https://www-dev.greenpeace.org/test-pandora/support-to-ta-pages-demo-page/)

Also play around with [this page](https://www-dev.greenpeace.org/test-pandora/take-action/page-with-act-parent/) and change its parent to `Act` or `Take Action`. The block should follows the expected behavior.

---

Ref: https://jira.greenpeace.org/browse/PLANET-7657

### Testing

1. Create a new page
2. Add two blocks: ActionsList and Cover which will be needed to compare content

Then check:
1. it's disabled, we should only fetch children pages of the page that is assigned as "Act Page" (same as the current Covers block).
2. If it's enabled, we should fetch Actions (we already do that) but also the children pages of the page that is assigned as "Take Action Page"

### Updates
Unfortunately the query loop from editor doesn't accept to pass an array of post types.
- if `IA` is `false` and post type is equal to `"page"`, then endpoint is pointed to `/wp-json/wp/v2/taxonomies`
- if `IA` is `true` and post type is equal to `"p4_action"`, then endpoint is pointed to `/wp-json/wp/v2/p4_action`
- if `IA` is `true|false` and post type is equal to an array of post types (like `["p4_action", "page"]`), then the enpoint is pointed to `/wp-json/wp/v2/types/p4_action,page`

#### Native behavior:
`/wp/v2/types`: returns all public post types
`/wp/v2/types/post`:returns info about the post post type
`/wp/v2/types/post,page`: will return a 404

### DIFF
--- a/assets/src/block-editor/setupTaxonomyBreadcrumbBlock.js
+++ b/assets/src/block-editor/setupTaxonomyBreadcrumbBlock.js
@@ -42,13 +42,15 @@ function editFunction({attributes, context}) {
       const taxonomyField = taxonomy === 'category' ? 'categories' : taxonomy;
 
       let postTypeField = post_type;
+
       if(postTypeField === 'p4_multipost') {
         postTypeField = await wp.apiFetch({path: `/wp/v2/p4_multipost/${postId}`});
-      } else {
-        for(const type of [['post', 'posts'], ['page', 'pages']]) {
-          if(postTypeField === type[0]) {
-            postTypeField = type[1];
-          }
+      }
+
+      // Pluralize post types
+      for(const type of [['post', 'posts'], ['page', 'pages']]) {
+        if(postTypeField === type[0]) {
+          postTypeField = type[1];
         }
       }
 
--- a/assets/src/block-editor/setupTaxonomyBreadcrumbBlock.js
+++ b/assets/src/block-editor/setupTaxonomyBreadcrumbBlock.js
@@ -36,19 +36,25 @@ function editFunction({attributes, context}) {
   const [term, setTerm] = wp.element.useState(null);
 
   wp.element.useEffect(() => {
-    if (!postId || !taxonomy || !post_type) {return;}
+    (async () => {
+      if (!postId || !taxonomy || !post_type) {return;}
 
-    const postTypeField = post_type === 'post' ? 'posts' : post_type;
-    const taxonomyField = taxonomy === 'category' ? 'categories' : taxonomy;
+      const taxonomyField = taxonomy === 'category' ? 'categories' : taxonomy;
+      let postTypeField = post_type === 'post' ? 'posts' : post_type;
 
-    wp.apiFetch({path: `/wp/v2/${postTypeField}/${postId}`}).then(post => {
-      const termIds = post[taxonomyField];
-      if (termIds && termIds.length > 0) {
-        wp.apiFetch({path: `/wp/v2/${taxonomyField}/${termIds[0]}`}).then(termData => {
-          setTerm(termData.name);
-        });
+      if(post_type === 'p4_multipost') {
+        postTypeField = await wp.apiFetch({path: `/wp/v2/p4_multipost/${postId}`});
       }
-    });
+
+      wp.apiFetch({path: `/wp/v2/${postTypeField}/${postId}`}).then(post => {
+        const termIds = post[taxonomyField];
+        if (termIds && termIds.length > 0) {
+          wp.apiFetch({path: `/wp/v2/${taxonomyField}/${termIds[0]}`}).then(termData => {
+            setTerm(termData.name);
+          });
+        }
+      });
+    })();
   }, [postId, taxonomy]);
 
   const linkAttrs = {href: '', onClick: e => e.preventDefault()};
--- a/assets/src/block-editor/setupTaxonomyBreadcrumbBlock.js
+++ b/assets/src/block-editor/setupTaxonomyBreadcrumbBlock.js
@@ -68,8 +68,8 @@ function editFunction({attributes, context}) {
   const linkAttrs = {href: '', onClick: e => e.preventDefault()};
   const contentAttrs = {className: 'taxonomy-category wp-block-post-terms'};
 
-  const link = wp.element.createElement('a', linkAttrs, term || 'Loading...');
-  const content = wp.element.createElement('div', contentAttrs, link);
+  const link = term ? wp.element.createElement('a', linkAttrs, term || 'Loading...') : null;
+  const content = link ? wp.element.createElement('div', contentAttrs, link) : null;
 
   return content;
 }
--- a/assets/src/block-editor/setupTaxonomyBreadcrumbBlock.js
+++ b/assets/src/block-editor/setupTaxonomyBreadcrumbBlock.js
@@ -40,10 +40,16 @@ function editFunction({attributes, context}) {
       if (!postId || !taxonomy || !post_type) {return;}
 
       const taxonomyField = taxonomy === 'category' ? 'categories' : taxonomy;
-      let postTypeField = post_type === 'post' ? 'posts' : post_type;
 
-      if(post_type === 'p4_multipost') {
+      let postTypeField = post_type;
+      if(postTypeField === 'p4_multipost') {
         postTypeField = await wp.apiFetch({path: `/wp/v2/p4_multipost/${postId}`});
+      } else {
+        for(const type of [['post', 'posts'], ['page', 'pages']]) {
+          if(postTypeField === type[0]) {
+            postTypeField = type[1];
+          }
+        }
       }
 
       wp.apiFetch({path: `/wp/v2/${postTypeField}/${postId}`}).then(post => {
--- a/assets/src/blocks/ActionsList/index.js
+++ b/assets/src/blocks/ActionsList/index.js
@@ -66,7 +66,7 @@ export const registerActionsListBlock = () => {
         ['core/group', {}, [
           [TAX_BREADCRUMB_BLOCK_NAME, {
             taxonomy: LISTS_BREADCRUMBS[0].value,
-            post_type: queryPostType,
+            post_type: postType,
           }],
           ['core/post-title', {isLink: true}],
           ['core/post-excerpt'],
--- a/assets/src/blocks/ActionsList/index.js
+++ b/assets/src/blocks/ActionsList/index.js
@@ -12,11 +12,7 @@ export const ACTIONS_LIST_LAYOUT_TYPES = [
 export const registerActionsListBlock = () => {
   const {registerBlockVariation} = wp.blocks;
   const {__} = wp.i18n;
-
-  const IS_NEW_IA = window.p4_vars.options.new_ia;
-  const ACT_PAGE = window.p4_vars.options.take_action_page || -1;
-
-  const queryPostType = IS_NEW_IA ? 'p4_action' : 'page';
+  const postType = 'p4_query_loop_interceptor';
 
   registerBlockVariation('core/query', {
     name: ACTIONS_LIST_BLOCK_NAME,
@@ -26,7 +22,7 @@ export const registerActionsListBlock = () => {
     scope: ['inserter'],
     allowedControls: ['taxQuery', 'pages', 'offset'],
     category: 'planet4-blocks-beta',
-    isActive: ({namespace, query}) => namespace === ACTIONS_LIST_BLOCK_NAME && query.postType === queryPostType,
+    isActive: ({namespace, query}) => namespace === ACTIONS_LIST_BLOCK_NAME && query.postType === postType,
     attributes: {
       namespace: ACTIONS_LIST_BLOCK_NAME,
       className: 'actions-list p4-query-loop is-custom-layout-grid',
@@ -34,17 +30,14 @@ export const registerActionsListBlock = () => {
         pages: 0,
         perPage: 3,
         offset: 0,
-        order: 'desc',
-        orderBy: 'date',
         author: '',
         search: '',
         exclude: [],
         sticky: '',
         inherit: false,
-        postType: queryPostType,
+        postType,
         postIn: [],
         hasPassword: false,
-        ...!IS_NEW_IA && {postParent: ACT_PAGE},
       },
       layout: {
         type: 'grid',
--- a/assets/src/blocks/ActionsList/index.js
+++ b/assets/src/blocks/ActionsList/index.js
@@ -12,7 +12,7 @@ export const ACTIONS_LIST_LAYOUT_TYPES = [
 export const registerActionsListBlock = () => {
   const {registerBlockVariation} = wp.blocks;
   const {__} = wp.i18n;
-  const postType = 'p4_query_loop_interceptor';
+  const postType = 'p4_multipost';
 
   registerBlockVariation('core/query', {
     name: ACTIONS_LIST_BLOCK_NAME,
--- a/src/Api/QueryLoopExtension.php
+++ b/src/Api/QueryLoopExtension.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace P4\MasterTheme\Api;
 
 use WP_REST_Response;
+use WP_REST_Request;
 
 /**
  * Instance QueryLoopExtension API
@@ -26,8 +27,13 @@ class QueryLoopExtension
             '/p4_multipost/(?P<id>\d+)',
             [
                 'methods' => 'GET',
-                'callback' => function () {
-                    return new WP_REST_Response([], 200);
+                'callback' => function (WP_REST_Request $request) {
+                    global $wpdb;
+                    $post_type = $wpdb->get_col($wpdb->prepare(
+                        "SELECT post_type FROM {$wpdb->posts} WHERE ID = %s",
+                        $request->get_param('id')
+                    ));
+                    return new WP_REST_Response($post_type[0] ?? "", 200);
                 },
                 'permission_callback' => '__return_true',
             ]
--- a/src/Api/QueryLoopExtension.php
+++ b/src/Api/QueryLoopExtension.php
@@ -1,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace P4\MasterTheme\Api;
+
+use WP_REST_Response;
+
+/**
+ * Instance QueryLoopExtension API
+ */
+class QueryLoopExtension
+{
+    /**
+     * Register endpoint to avoid 404 response when fetching multiples
+     * posts within the Actions List block.
+     * Please check /src/Blocks/QueryLoopExtension.php for more reference.
+     *
+     * @example GET /wp-json/wp/v2/p4_multipost/193
+     */
+
+    public static function register_endpoint(): void
+    {
+        register_rest_route(
+            'wp/v2',
+            '/p4_multipost/(?P<id>\d+)',
+            [
+                'methods' => 'GET',
+                'callback' => function () {
+                    return new WP_REST_Response([], 200);
+                },
+                'permission_callback' => '__return_true',
+            ]
+        );
+    }
+}
--- a/src/Blocks/QueryLoopExtension.php
+++ b/src/Blocks/QueryLoopExtension.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace P4\MasterTheme\Blocks;
 
+use WP_REST_Request;
+
 /**
  * Handle Posts/Actions List block manual override query with custom `postIn` filter.
  * `include` is not used because it generates some issues with getEntityRecords filters.
@@ -19,11 +21,10 @@ class QueryLoopExtension
 
     public static function registerEditorQuery(): void
     {
-
         // This is used to handle Actions List block with multiple Post Types
         register_post_type('p4_multipost', [
-            'label' => 'P4 Query Loop Interceptor',
-            'description' => 'This post type is used to filter filter queries since arrays are not supported natively.',
+            'label' => 'P4 multipost',
+            'description' => 'This post type is used to filter queries since arrays are not supported natively.',
             'public' => false,
             'show_ui' => false,
             'show_in_rest' => true,
@@ -62,12 +63,12 @@ class QueryLoopExtension
     /**
      * Filter the query to include the postIn and hasPassword parameters
      *
-     * @param array $args The query
-     * @param array $request The request
+     * @param array           $args The query
+     * @param WP_REST_Request $request The request
      *
      * @return array The new parsed query
      */
-    public static function postInFilter(array $args, array $request): array
+    public static function postInFilter(array $args, WP_REST_Request $request): array
     {
         $postIn = $request->get_param('postIn');
         if (!empty($postIn)) {
--- a/src/Blocks/QueryLoopExtension.php
+++ b/src/Blocks/QueryLoopExtension.php
@@ -19,22 +19,24 @@ class QueryLoopExtension
 
     public static function registerEditorQuery(): void
     {
-        $postInFilter = function ($args, $request) {
-            $postIn = $request->get_param('postIn');
-            if (!empty($postIn)) {
-                $args['post__in'] = array_map('intval', (array) $postIn);
-                $args['orderby'] = 'post__in';
-            }
-            if ($request->has_param('hasPassword')) {
-                $hasPassword = $request->get_param('hasPassword');
-                $args['has_password'] = $hasPassword !== false && $hasPassword !== 'false';
-            }
-            return $args;
-        };
+        // This is used to handle Actions List block with multiple Post Types
+        register_post_type('p4_query_loop_interceptor', [
+            'label' => 'P4 Query Loop Interceptor',
+            'description' => 'This post type is used to filter filter queries since arrays are not supported natively.',
+            'public' => false,
+            'show_ui' => false,
+            'show_in_rest' => true,
+            'capability_type' => 'post',
+            'supports' => [],
+            'query_var' => false,
+        ]);
 
-        add_filter('rest_post_query', $postInFilter, 10, 2);
-        add_filter('rest_page_query', $postInFilter, 10, 2);
-        add_filter('rest_p4_action_query', $postInFilter, 10, 2);
+        add_filter('rest_post_query', [self::class, 'postInFilter'], 10, 2);
+        add_filter('rest_page_query', [self::class, 'postInFilter'], 10, 2);
+        add_filter('rest_p4_action_query', [self::class, 'postInFilter'], 10, 2);
+        add_filter('rest_p4_query_loop_interceptor_query', function ($query, $request) {
+            return self::sanitizePostTypes($query, $request->get_params());
+        }, 10, 2);
     }
 
     public static function registerFrontendQuery(): void
@@ -42,18 +44,93 @@ class QueryLoopExtension
         add_filter(
             'query_loop_block_query_vars',
             function ($query, $block) {
-                $blockQuery = $block->context['query'] ?? [];
-                if (!empty($blockQuery['postIn'])) {
-                    $query['post__in'] = array_map('intval', (array) $blockQuery['postIn']);
-                    $query['orderby'] = 'post__in';
+                if ($block->context['query']['postType'] === 'p4_query_loop_interceptor') {
+                    return self::sanitizePostTypes(
+                        $query,
+                        $block->context['query'],
+                    );
                 }
-                if (isset($blockQuery['hasPassword'])) {
-                    $query['has_password'] = (bool) $blockQuery['hasPassword'];
-                }
+
                 return $query;
             },
             10,
             2
         );
+    }
+
+    /**
+     * Filter the query to include the postIn and hasPassword parameters
+     *
+     * @param array $args The query
+     * @param array $request The request
+     *
+     * @return array The new parsed query
+     */
+    public static function postInFilter(array $args, array $request): array
+    {
+        $postIn = $request->get_param('postIn');
+        if (!empty($postIn)) {
+            $args['post__in'] = array_map('intval', (array) $postIn);
+            $args['orderby'] = 'post__in';
+        }
+        if ($request->has_param('hasPassword')) {
+            $hasPassword = $request->get_param('hasPassword');
+            $args['has_password'] = $hasPassword !== false && $hasPassword !== 'false';
+        }
+        return $args;
+    }
+
+    /**
+     * Sanitize the query received by the editor and frontend
+     *
+     * @param array $query The query
+     * @param array $params The query or request params
+     *
+     * @return array The new parsed query
+     */
+    public static function sanitizePostTypes(array $query, array $params = []): array
+    {
+        $is_new_ia = !empty(planet4_get_option('new_ia'));
+        $query['post_status'] = 'publish';
+
+        if (!$is_new_ia) {
+            $query['post_type'] = ['page'];
+            $query['post_parent'] = planet4_get_option('act_page');
+
+            if (!empty($params['postIn'])) {
+                $query['post__in'] = array_map('intval', (array) $params['postIn']);
+            }
+        } else {
+            $query['post_type'] = ['page', 'p4_action'];
+
+            global $wpdb;
+            $post_ids = $wpdb->get_col($wpdb->prepare(
+                "
+                (SELECT ID FROM {$wpdb->posts} WHERE post_type = %s)
+                UNION ALL
+                (SELECT ID FROM {$wpdb->posts} WHERE post_type = %s AND post_parent = %d)",
+                'p4_action',
+                'page',
+                planet4_get_option('take_action_page')
+            ));
+
+            if (!empty($post_ids)) {
+                $query['post__in'] = $post_ids;
+            } else {
+                $query['post__in'] = [0];
+            }
+
+            if (!empty($params['hasPassword'])) {
+                $query['has_password'] = $params['hasPassword'] !== false && $params['hasPassword'] !== 'false';
+            }
+
+            $query['orderby'] = [
+                'menu_order' => 'ASC',
+                'post_date' => 'DESC',
+                'post_title' => 'ASC',
+                'post__in' => 'ASC',
+            ];
+        }
+        return $query;
     }
 }
--- a/src/Blocks/QueryLoopExtension.php
+++ b/src/Blocks/QueryLoopExtension.php
@@ -19,8 +19,9 @@ class QueryLoopExtension
 
     public static function registerEditorQuery(): void
     {
+
         // This is used to handle Actions List block with multiple Post Types
-        register_post_type('p4_query_loop_interceptor', [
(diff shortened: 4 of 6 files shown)


planet-4:

:watch: 2025.01.07 11:06:24

src/Blocks/QueryLoopExtension.php
```diff
@@ -43,13 +43,26 @@ public static function registerFrontendQuery(): void
             'query_loop_block_query_vars',
             function ($query, $block) {
                 $blockQuery = $block->context['query'] ?? [];
+
+                // Sanitize query only applied to ActionsList block.
+                if ((bool) array_key_exists('isCustomQueryLoop', $blockQuery)) {
```
GP-Dan-Tovbein, line 48: Need to check this to avoid apply this changes to the native `core/query` query.

assets/src/block-editor/QueryLoopBlockExtension/index.js
```diff
@@ -108,6 +108,20 @@ export const setupQueryLoopBlockExtension = () => {
           wp.data.select('core/block-editor').getSelectedBlock(),
         ]);
 
+        useEffect(() => {
+          const IS_NEW_IA = window.p4_vars.options.new_ia;
+          const queryPostType = IS_NEW_IA ? 'p4_action' : 'page';
+
+          if(query.postType !== queryPostType) {
```
GP-Dan-Tovbein, line 115: Need to implement this since the query passed through attributes (1) it's not the same as the one needed by the query loop (2)

For instance:
(1)
```json
{
   "post_type":"page",
   "order":"DESC",
   "orderby":"date",
   "post__not_in":[
      2067
   ],
   "tax_query":[
      
   ],
   "offset":0,
   "posts_per_page":3,
   "author__in":[]
}
```
(2)
```json
{
   "pages":0,
   "perPage":3,
   "offset":0,
   "order":"desc",
   "orderBy":"date",
   "postStatus":"publish",
   "author":"",
   "search":"",
   "exclude":[
      2067
   ],
   "sticky":"",
   "inherit":false,
   "postType":"page",
   "postIn":[
      
   ],
   "hasPassword":false,
   "parent":9,
   "post_parent":9
}
```

src/Blocks/QueryLoopExtension.php
```diff
@@ -43,6 +43,29 @@ public static function registerFrontendQuery(): void
             'query_loop_block_query_vars',
             function ($query, $block) {
                 $blockQuery = $block->context['query'] ?? [];
+
+                $is_new_ia = !empty(planet4_get_option('new_ia'));
+
+                $query['post_type'] = ['page'];
+
+                // This applies only to Actions List block
+                if ($blockQuery['isCustom']) {
```
mleray, line 52: I'm wondering if we need this extra query variable 🤔 In JS we use the [namespace attribute](https://github.com/greenpeace/planet4-master-theme/blob/main/assets/src/block-editor/QueryLoopBlockExtension/index.js#L22), any chance we can access it here too?

GP-Dan-Tovbein: Looking at the `$block` params we receive, I don't see any related to `namespace` :/

GP-Dan-Tovbein: Remember that this hook is only related to the `query` but not the `block`.

mleray: I understand but we do get the `block` parameter which according to the function docs is a `WP_Block` so its attributes should be accessible ([WP_Block docs](https://developer.wordpress.org/reference/classes/wp_block/))

GP-Dan-Tovbein: In this case the `$block` refers to `core/query-no-results` and `core/post-template` but not to the `planet4-blocks/actions-list`. But I'm still checking if there is another better approach.

mleray: Weird, feels like it should be `core/query`

GP-Dan-Tovbein: This filter is being called for the inner blocks `core/query-no-results` and `core/post-template`, but you cannot access to `namespace` from `core/query`. That's why wee need to implement a custom `isCustom` flag.

src/Blocks/QueryLoopExtension.php
```diff
@@ -43,6 +43,35 @@ public static function registerFrontendQuery(): void
...
+
+                    if (!$is_new_ia) {
+                        $query['post_parent'] = planet4_get_option('take_action_page');
+                    } else {
+                        $query['post_type'][] = 'p4_action';
+                        $query['post_parent__in'] = [planet4_get_option('take_action_page')];
+                        $query['posts_per_page'] = -1;
+                    }
+
+                    if (is_array($blockQuery['orderBy'])) {
```
GP-Dan-Tovbein, line 60: Take [this](https://github.com/greenpeace/planet4-master-theme/pull/2492/files#diff-ad9c111941a060027e21059b76d6136f4b1edb3aa4744e8f094f82ad200e3260R33-R34) or set by default https://github.com/greenpeace/planet4-master-theme/pull/2492/files#diff-5731321646bfd7fe6c8f5582c07ece1a8c4078579f40d518eafc522c4f89f421R66-R70

mardelnet (commented):
@GP-Dan-Tovbein 
On the test instance, adding an Actions List block produces a critical error:

image

GP-Dan-Tovbein:
> @GP-Dan-Tovbein On the test instance, adding an Actions List block produces a critical error:
> 
> image

I think the issue happens when you define an Array of `orderBy`

assets/src/blocks/ActionsList/index.js
```diff
@@ -30,8 +30,7 @@ export const registerActionsListBlock = () => {
         pages: 0,
         perPage: 3,
         offset: 0,
-        order: 'desc',
```
GP-Dan-Tovbein, line 33: I've removed this filters since by default `orderby` is set to `date` and `order` is set to `desc`.

Check it out the [WP's official doc about arguments section](https://developer.wordpress.org/rest-api/reference/posts/#arguments)

GP-Dan-Tovbein: Also this filters are being overriden [here](https://github.com/greenpeace/planet4-master-theme/pull/2492/files#diff-5731321646bfd7fe6c8f5582c07ece1a8c4078579f40d518eafc522c4f89f421R66-R70)

GP-Dan-Tovbein: Last but not least, unfortunately passing an Array to `order` or `orderBy` within JS is not possible for now.
[These are a couple open issues](https://github.com/WordPress/gutenberg/issues/40170) that explains the need to set an array of strings, the same as we do in [PHP](https://github.com/greenpeace/planet4-master-theme/pull/2492/files#diff-5731321646bfd7fe6c8f5582c07ece1a8c4078579f40d518eafc522c4f89f421R66-R70).

mardelnet (commented):
Hey @GP-Dan-Tovbein, I still see some issues in the test environment:

**Scenario 1:**
- New Information Architecture is enabled.
- "Take Action" is the selected "Take Action" page.
- "Climate" and "Sitemap" are the children of the "Take Action" page.
- In the editor, the block shows 3 actions but no children pages (same as Covers).
- On the front, the block shows the children pages (with no featured image) but no actions.
 
**Scenario 2:**
- New Information Architecture is disabled.
- "Act" is the selected "Act Page".
- "All Blocks Page" and "All blocks. Evergreen template" are the children of the "Act Page".
- In the editor, the block shows the 2 children pages of the "Act Page" (same as Covers).
- On the front, the block shows the children pages of the "Take Action" page (with no featured image).

GP-Dan-Tovbein:
> Hey @GP-Dan-Tovbein, I still see some issues in the test environment:
> 
> **Scenario 1:**
> 
>     * New Information Architecture is enabled.
> 
>     * "Take Action" is the selected "Take Action" page.
> 
>     * "Climate" and "Sitemap" are the children of the "Take Action" page.
> 
>     * In the editor, the block shows 3 actions but no children pages (same as Covers).
> 
>     * On the front, the block shows the children pages (with no featured image) but no actions.
> 
> 
> **Scenario 2:**
> 
>     * New Information Architecture is disabled.
> 
>     * "Act" is the selected "Act Page".
> 
>     * "All Blocks Page" and "All blocks. Evergreen template" are the children of the "Act Page".
> 
>     * In the editor, the block shows the 2 children pages of the "Act Page" (same as Covers).
> 
>     * On the front, the block shows the children pages of the "Take Action" page (with no featured image).

Thanks for testing @mardelnet !
I'll check all of them and will go back to you

GP-Dan-Tovbein:
> Hey @GP-Dan-Tovbein, I still see some issues in the test environment:
> 
> **Scenario 1:**
> 
>     * New Information Architecture is enabled.
> 
>     * "Take Action" is the selected "Take Action" page.
> 
>     * "Climate" and "Sitemap" are the children of the "Take Action" page.
> 
>     * In the editor, the block shows 3 actions but no children pages (same as Covers).
> 
>     * On the front, the block shows the children pages (with no featured image) but no actions.
> 
> 
> **Scenario 2:**
> 
>     * New Information Architecture is disabled.
> 
>     * "Act" is the selected "Act Page".
> 
>     * "All Blocks Page" and "All blocks. Evergreen template" are the children of the "Act Page".
> 
>     * In the editor, the block shows the 2 children pages of the "Act Page" (same as Covers).
> 
>     * On the front, the block shows the children pages of the "Take Action" page (with no featured image).

@mardelnet  The issue is also happing on the front, no?

src/Blocks/QueryLoopExtension.php
```diff
@@ -43,6 +43,50 @@ public static function registerFrontendQuery(): void
...
+                    if (!$is_new_ia) {
+                        $query['post_type'] = ['page'];
+                        $query['post_parent'] = planet4_get_option('act_page');
+                    } else {
+                        $query['post_type'] = ['page', 'p4_action'];
+                        $query['post_status'] = 'publish';
+                        $query['posts_per_page'] = -1;
+
+                        global $wpdb;
+                        $post_ids = $wpdb->get_col($wpdb->prepare(
```
GP-Dan-Tovbein, line 60: I'm returning back to the original idea by using an SQL. I know it's a tricky implementation but infortunately is not possible to use `tax_query` and follow the expected behaviour.

```
$query = [
            'post_type' => ['page', 'p4_action'],
            'post_status' => 'publish',
            'posts_per_page' => -1,
            'meta_query' => [
                'relation' => 'OR',
                [
                    'key' => 'post_type',
                    'value' => 'p4_action',
                    'compare' => '=',
                ],
                [
                    'relation' => 'AND',
                    [
                        'key' => 'post_type',
                        'value' => 'page',
                        'compare' => '=',
                    ],
                    [
                        'key' => 'post_parent',
                        'value' => $take_action_page,
                        'compare' => '=',
                    ],
                ],
            ],
        ];
```

What does this query? Returns all posts id that follow the condition, and then set to `post__in`.

src/Blocks/QueryLoopExtension.php
```diff
@@ -43,6 +43,50 @@ public static function registerFrontendQuery(): void
...
+                            (SELECT ID FROM {$wpdb->posts} WHERE post_type = %s AND post_parent = %d)",
+                            'p4_action',
+                            'page',
+                            planet4_get_option('take_action_page')
+                        ));
+
+                        if (!empty($post_ids)) {
+                            $query['post__in'] = $post_ids;
+                        } else {
+                            $query['post__in'] = [0];
```
GP-Dan-Tovbein, line 73: Prevent to return all posts when `post_ids` is empty

src/Blocks/QueryLoopExtension.php
GP-Dan-Tovbein: Also this kind of implementation "follows" the one [related to Covers](https://github.com/greenpeace/planet4-master-theme/blob/main/src/Blocks/Covers.php#L368).

mardelnet (commented):
Thanks for the changes, @GP-Dan-Tovbein 
I still see some problems:

**Scenario:**
- New Information Architecture is enabled.
- "Take Action" is the selected "Take Action" page.
- "Climate" and "Sitemap" are the children of the "Take Action" page.
- In the editor, the block shows only actions but no children pages.
- On the front, the block always shows all the action and children pages without paying attention to the "Posts per page" limit.

GP-Dan-Tovbein:
> @GP-Dan-Tovbein, Actions List (the first 3 columns) and Covers block (the latter 3) do not look exactly the same or have the same content. Is that expected?
> 
> image

Looking at [Figma](https://www.figma.com/design/9NtbY8n3at8uOEJTsLrETb/P4-Design-System?node-id=192-14857&t=7erFKVzFrIsBqYrI-4), it looks like a Covers issue.

GP-Dan-Tovbein:
> > @GP-Dan-Tovbein, Actions List (the first 3 columns) and Covers block (the latter 3) do not look exactly the same or have the same content. Is that expected?
> > image
> 
> Looking at [Figma](https://www.figma.com/design/9NtbY8n3at8uOEJTsLrETb/P4-Design-System?node-id=192-14857&t=7erFKVzFrIsBqYrI-4), it looks like a Covers issue.

I mean about the content, for example the missed excerpt

GP-Dan-Tovbein:
@mardelnet Let me know your thoughts. I will create a ticket to follow up the Covers's issue.

mardelnet:
> @mardelnet Let me know your thoughts. I will create a ticket to follow up the Covers's issue.

@GP-Dan-Tovbein, sure. If there are troubles with the Covers block, let's tackle them in another ticket.
Have you seen my previous comment? https://github.com/greenpeace/planet4-master-theme/pull/2492#pullrequestreview-2746141340

mardelnet (commented):
Hey @GP-Dan-Tovbein, just for confirmation: are [these changes](https://github.com/greenpeace/planet4-master-theme/pull/2492#pullrequestreview-2746141340) done? Because I still see the issues in the testing environment. Thanks!

mardelnet:
> Hey @GP-Dan-Tovbein, just for confirmation: are [these changes](https://github.com/greenpeace/planet4-master-theme/pull/2492#pullrequestreview-2746141340) done? Because I still see the issues in the testing environment. Thanks!

Nevermind! 
I'll test this PR on my local environment.
It seems the last changes were not deployed to the testing environment yet, as it is stuck.

mardelnet (commented):
Thanks for the changes, @GP-Dan-Tovbein 
I still see some problems:

**Scenario 1:**
- New Information Architecture is enabled.
- In the editor, the block shows only actions but no children pages of the "Take Action" page.
- On the front, the block always shows all the action and children pages without paying attention to the "Posts per page" limit.
- Also on the front, the block shows actions and children pages, but not great-children pages of the "Take Action" page.

**Scenario 2:**
- New Information Architecture is enabled.
- An Actions List block is added to a page.
- I disable the New Information Architecture.
- In the editor, the Actions Lists block is turned into a Query block.
- On the front, the block shows nothing.

assets/src/blocks/ActionsList/index.js
```diff
@@ -8,11 +8,7 @@ export const ACTIONS_LIST_LAYOUT_TYPES = [
 export const registerActionsListBlock = () => {
   const {registerBlockVariation} = wp.blocks;
   const {__} = wp.i18n;
-
-  const IS_NEW_IA = window.p4_vars.options.new_ia;
-  const ACT_PAGE = window.p4_vars.options.take_action_page || -1;
-
-  const queryPostType = IS_NEW_IA ? 'p4_action' : 'page';
+  const postType = 'p4_query_loop_interceptor';
```
GP-Dan-Tovbein, line 11: Since you cannot pass an array of post types from editor, I'd need to pass this custom post type and then intercept it [here](https://github.com/greenpeace/planet4-master-theme/pull/2492/files#diff-5731321646bfd7fe6c8f5582c07ece1a8c4078579f40d518eafc522c4f89f421R37-R39)

src/Blocks/QueryLoopExtension.php
```diff
@@ -19,41 +19,118 @@ public static function registerHooks(): void
...
+
+    /**
+     * Filter the query to include the postIn and hasPassword parameters
+     *
+     * @param array $args The query
+     * @param array $request The request
+     *
+     * @return array The new parsed query
+     */
+    public static function postInFilter(array $args, array $request): array
```
GP-Dan-Tovbein, line 69: Move out this function from `registerEditorQuery` scope

mardelnet (commented):
@GP-Dan-Tovbein 
Thanks for the updates!
However, the Actions List block seems not to be working in the editor, both in the local and testing environments. 
The posts are not loaded and the console shows the following error: 
`GET | http://www.planet4.test/wp-json/wp/v2/types/p4_query_loop_interceptor?context=edit&_locale=user - NOT FOUND`

GP-Dan-Tovbein:
> @GP-Dan-Tovbein Thanks for the updates! However, the Actions List block seems not to be working in the editor, both in the local and testing environments. The posts are not loaded and the console shows the following error: `GET | http://www.planet4.test/wp-json/wp/v2/types/p4_query_loop_interceptor?context=edit&_locale=user - NOT FOUND`

Thank you ❤️ ! Can you please check it again? I've modified the lenght of post type since it seemed to be exceding the limit of characters.

mardelnet (commented):
Nice job, @GP-Dan-Tovbein 
The functionality seems to be working as expected, so now I can proceed with the proper code review. In the meantime, I have some questions:

1. The children's pages are included correctly. Should we include grandchildren, too?
2. While action posts have a tag above the title, that doesn't happen with pages. Should we add something to keep the same layout?

aaa-Greenpeace

src/Loader.php
mardelnet, line 1: Is this change necessary?

GP-Dan-Tovbein: From my point of view yes, because in that case you're making sure that the hooks are being registered before adding blocks. But we can keep it at the previous line if you agree.

GP-Dan-Tovbein:
> Nice job, @GP-Dan-Tovbein The functionality seems to be working as expected, so now I can proceed with the proper code review. In the meantime, I have some questions:
> 
>     1. The children's pages are included correctly. Should we include grandchildren, too?
> 
>     2. While action posts have a tag above the title, that doesn't happen with pages. Should we add something to keep the same layout?
> 
> 
> aaa-Greenpeace

@mardelnet  Thanks for your review!

> The children's pages are included correctly. Should we include grandchildren, too?
As the ticket mentions, we need to show only pages with parent ids. So, I understand that children pages are only the pages to include.
> While action posts have a tag above the title, that doesn't happen with pages. Should we add something to keep the same layout?
What do you mean to add something? This is part of the native [core template behavior](https://github.com/greenpeace/planet4-master-theme/blob/main/assets/src/blocks/ActionsList/index.js#L71-L84). If you ask me for an opinion, I don't see any layout issue here.

What do you think?

src/Blocks/QueryLoopExtension.php
```diff
@@ -19,41 +19,119 @@ public static function registerHooks(): void
...
-                $hasPassword = $request->get_param('hasPassword');
-                $args['has_password'] = $hasPassword !== false && $hasPassword !== 'false';
-            }
-            return $args;
-        };
 
-        add_filter('rest_post_query', $postInFilter, 10, 2);
-        add_filter('rest_page_query', $postInFilter, 10, 2);
-        add_filter('rest_p4_action_query', $postInFilter, 10, 2);
+        // This is used to handle Actions List block with multiple Post Types
```
mardelnet, line 23: There's an empty line above that could be removed.

src/Blocks/QueryLoopExtension.php
```diff
@@ -19,41 +19,119 @@ public static function registerHooks(): void
...
+        register_post_type('p4_multipost', [
+            'label' => 'P4 Query Loop Interceptor',
+            'description' => 'This post type is used to filter filter queries since arrays are not supported natively.',
```
mardelnet, line 26:
```suggestion
            'description' => 'This post type is used to filter queries since arrays are not supported natively.',
```

src/Blocks/QueryLoopExtension.php
```diff
@@ -19,41 +19,119 @@ public static function registerHooks(): void
...
-            }
-            return $args;
-        };
 
-        add_filter('rest_post_query', $postInFilter, 10, 2);
-        add_filter('rest_page_query', $postInFilter, 10, 2);
-        add_filter('rest_p4_action_query', $postInFilter, 10, 2);
+        // This is used to handle Actions List block with multiple Post Types
+        register_post_type('p4_multipost', [
+            'label' => 'P4 Query Loop Interceptor',
```
mardelnet, line 25: Should we change the label as the block is no longer called "Interceptor"?

src/Blocks/QueryLoopExtension.php
```diff
@@ -19,41 +19,119 @@ public static function registerHooks(): void
...
+
+    /**
+     * Filter the query to include the postIn and hasPassword parameters
+     *
+     * @param array $args The query
+     * @param array $request The request
+     *
+     * @return array The new parsed query
+     */
+    public static function postInFilter(array $args, array $request): array
```
mardelnet, line 70: The `$request` parameter is an object, not an array.

assets/src/blocks/ActionsList/index.js
mardelnet, line 1: `queryPostType` is not defined in this file. Check like 69.

GP-Dan-Tovbein: Sorry, I don't understand what you mean 🙏

src/Blocks/QueryLoopExtension.php
GP-Dan-Tovbein: Yes, sure! In general words, It works like a interceptor since whenever it receives this `postType` the query is going to be re-implemented. What text do you suggest then?

assets/src/blocks/ActionsList/index.js
mardelnet: Check line 69 of the `assets/src/blocks/ActionsList/index.js` file. You'll see that the variable `queryPostType` does not exist. You have to replace it with `postType`.

GP-Dan-Tovbein: Oh I see! You're correct. Good catch 👏

mardelnet (commented):
Hey @GP-Dan-Tovbein, on my local environment:

Adding an Actions List block produces a bunch of errors:
image

The Posts List block is no longer working (posts are not loaded):
image

Can you please take a look at those issues?

GP-Dan-Tovbein:
@mardelnet I've just updated the code

src/Api/QueryLoopExtension.php
```diff
@@ -0,0 +1,36 @@
...
+     * Please check /src/Blocks/QueryLoopExtension.php for more reference.
+     *
+     * @example GET /wp-json/wp/v2/p4_multipost
+     */
+
+    public static function register_endpoint(): void
+    {
+        register_rest_route(
+            'wp/v2',
+            '/p4_multipost/(?P<id>\d+)',
```
GP-Dan-Tovbein, line 26: This is the real interceptor to avoid `404` posts when adding any Actions List block into a page.  It returns an empty 200 response.

mardelnet (commented):
@GP-Dan-Tovbein, when no "Act" page or "Take Action" page is chosen from the Planet 4 setting page, the block shows all the pages. I think it should show none of them.

mardelnet (commented):
@GP-Dan-Tovbein, I think you have to update the `assets/src/block-editor/setupTaxonomyBreadcrumbBlock.js` file as the taxonomy terms do not load in the editor:

Edit-Page-“bbb”-‹-Greenpeace-—-WordPress

GP-Dan-Tovbein:
> @GP-Dan-Tovbein, I think you have to update the `assets/src/block-editor/setupTaxonomyBreadcrumbBlock.js` file as the taxonomy terms do not load in the editor:
> 
> Edit-Page-“bbb”-‹-Greenpeace-—-WordPress

TBH, I'm curious about this question since I've just noticed that it happens only with items without tags or categories but it's something completely new. Don't you think it is start happening after https://github.com/greenpeace/planet4-master-theme/commit/d1fa83eab88813ff4e32ea1c850954a5158bcb12 🤔 ?

mardelnet:
Hey @GP-Dan-Tovbein 
This is just an idea, but I think there's a simple way to get the pages we want without registering an additional post type and an extra endpoint. Check it out [here](https://github.com/greenpeace/planet4-master-theme/compare/PLANET-7657_add-support-for-take-action-pages...PLANET-7657_add-support-for-take-action-pages-v2).

GP-Dan-Tovbein:
> Hey @GP-Dan-Tovbein This is just an idea, but I think there's a simple way to get the pages we want without registering an additional post type and an extra endpoint. Check it out [here](https://github.com/greenpeace/planet4-master-theme/compare/PLANET-7657_add-support-for-take-action-pages...PLANET-7657_add-support-for-take-action-pages-v2).

I took the time to review the proposed changes:

1)
> https://github.com/greenpeace/planet4-master-theme/compare/PLANET-7657_add-support-for-take-action-pages...PLANET-7657_add-support-for-take-action-pages-v2#diff-ad9c111941a060027e21059b76d6136f4b1edb3aa4744e8f094f82ad200e3260R69

In case we need to display actions we're showing page tags. Is that correct?

2)
> https://github.com/greenpeace/planet4-master-theme/compare/PLANET-7657_add-support-for-take-action-pages...PLANET-7657_add-support-for-take-action-pages-v2#diff-ad9c111941a060027e21059b76d6136f4b1edb3aa4744e8f094f82ad200e3260R69

- We already invoke `rest_page_query`. Do you think it's okay to run it again?
- Do you agree to render pages by default even if we want to render `p4_action`s?

mardelnet:
> > Hey @GP-Dan-Tovbein This is just an idea, but I think there's a simple way to get the pages we want without registering an additional post type and an extra endpoint. Check it out [here](https://github.com/greenpeace/planet4-master-theme/compare/PLANET-7657_add-support-for-take-action-pages...PLANET-7657_add-support-for-take-action-pages-v2).

> In case we need to display actions we're showing page tags. Is that correct?

Maybe we should change it to `p4_action` instead of pages. My branch is just a proof of concept and should be definitely polished.

> We already invoke `rest_page_query`. Do you think it's okay to run it again?

We could call the `sanitizePostTypes` function from `postInFilter` the function if only for pages and avoid calling the `rest_page_query` twice.

> Do you agree to render pages by default even if we want to render `p4_action`s?

Same as before: maybe we should change it to `p4_action` instead of pages.

src/Api/QueryLoopExtension.php
```diff
@@ -0,0 +1,42 @@
...
+            [
+                'methods' => 'GET',
+                'callback' => function (WP_REST_Request $request) {
+                    global $wpdb;
+                    $post_type = $wpdb->get_col($wpdb->prepare(
```
GP-Dan-Tovbein, line 32: This will returns the post_type used by breadcrumbs

assets/src/block-editor/setupTaxonomyBreadcrumbBlock.js
```diff
@@ -36,19 +36,25 @@ function editFunction({attributes, context}) {
   const [term, setTerm] = wp.element.useState(null);
 
   wp.element.useEffect(() => {
-    if (!postId || !taxonomy || !post_type) {return;}
```
mardelnet, line 39: We also have to consider the pages, otherwise, the category is never loaded for pages in the editor, and a bunch of errors appear in the browser console. I tried this solution and it works:

```js
function editFunction({attributes, context}) {
  const {taxonomy, post_type} = attributes;
  const postId = context.postId;
  const [term, setTerm] = wp.element.useState(null);

  wp.element.useEffect(() => {
    (async () => {
      if (!postId || !taxonomy || !post_type) {return;}

      const taxonomyField = taxonomy === 'category' ? 'categories' : taxonomy;
      let postTypeField = '';

      if (post_type === 'p4_multipost') {
        postTypeField = await wp.apiFetch({path: `/wp/v2/p4_multipost/${postId}`});
      }

      if (postTypeField === 'post') {
        postTypeField = 'posts';
      } else if (postTypeField === 'page') {
        postTypeField = 'pages';
      } else {
        postTypeField = post_type;
      }

      wp.apiFetch({path: `/wp/v2/${postTypeField}/${postId}`}).then(post => {
        const termIds = post[taxonomyField];
        if (termIds && termIds.length > 0) {
          wp.apiFetch({path: `/wp/v2/${taxonomyField}/${termIds[0]}`}).then(termData => {
            setTerm(termData.name);
          });
        }
      });
    })();
  }, [postId, taxonomy]);

  const linkAttrs = {href: '', onClick: e => e.preventDefault()};
  const contentAttrs = {className: 'taxonomy-category wp-block-post-terms'};

  const link = term ? wp.element.createElement('a', linkAttrs, term || 'Loading...') : null;
  const content = link ? wp.element.createElement('div', contentAttrs, link) : null;

  return content;
}
```

GP-Dan-Tovbein: FYI `p4_multiposts` also include `pages`, not only `p4_action`. Do you think it's still necessary this change then?

mardelnet: Without this change, the taxonomy of the pages is never loaded in the editor:

Screenshot-from-2025-04-30-11-07-41

assets/src/block-editor/setupTaxonomyBreadcrumbBlock.js
```diff
@@ -36,19 +36,31 @@ function editFunction({attributes, context}) {
...
-    wp.apiFetch({path: `/wp/v2/${postTypeField}/${postId}`}).then(post => {
-      const termIds = post[taxonomyField];
-      if (termIds && termIds.length > 0) {
-        wp.apiFetch({path: `/wp/v2/${taxonomyField}/${termIds[0]}`}).then(termData => {
-          setTerm(termData.name);
+      let postTypeField = post_type;
+      if(postTypeField === 'p4_multipost') {
+        postTypeField = await wp.apiFetch({path: `/wp/v2/p4_multipost/${postId}`});
+      } else {
+        [['post', 'posts'], ['page', 'pages']].forEach(type => {
```
GP-Dan-Tovbein, line 48: Based on https://github.com/greenpeace/planet4-master-theme/pull/2492#discussion_r2068202215

mardelnet (commented):
@GP-Dan-Tovbein, I went through all the possible scenarios and found one where the logic fails: when the new IA is enabled but the Take Action parent page is not selected from the Settings page, the P4 Actions are not shown.

| New IA | P4 Actions | Parent Page | Children Pages | Result | Comments
|--------|------------|-------------|----------------|--------|----------------------------
| Yes    | Yes        | Yes         | Yes            |   ✔    |
| Yes    | Yes        | Yes         | No             |   ✔    |
| Yes    | Yes        | No          | Yes            |        | No P4 Actions are shown
| Yes    | Yes        | No          | No             |        | No P4 Actions are shown
| Yes    | No         | Yes         | Yes            |   ✔    |
| Yes    | No         | Yes         | No             |   ✔    |
| Yes    | No         | No          | Yes            |   ✔    |
| Yes    | No         | No          | No             |   ✔    |
| No     | -          | Yes         | Yes            |   ✔    |
| No     | -          | Yes         | No             |   ✔    |
| No     | -          | No          | Yes            |   ✔    |
| No     | -          | No          | No             |   ✔    |

_References:_
- New IA: whether the new IA is enabled.
- P4 Actions: whether posts of type P4 Actions exist and are published.
- Parent Page: whether the main "Take Action" or "Act" page is selected.
- Children Pages: whether children's pages of the Parent Page exist and are published.

assets/src/block-editor/setupTaxonomyBreadcrumbBlock.js
mardelnet: Was this issue addressed, @GP-Dan-Tovbein?
Whilst the taxonomy for the pages is correctly shown in the frontend, in the editor I only see "Loading..." and still errors in the console (on my local environment).

GP-Dan-Tovbein: Fixed [here](https://github.com/greenpeace/planet4-master-theme/pull/2492/files#diff-1da62b883ad950e328a5e5d0658e50a0341d769c63e3e745686a475103711ebcR48-R52) . Is that okay?

GP-Dan-Tovbein: Let me review it again

GP-Dan-Tovbein:
> @GP-Dan-Tovbein, I went through all the possible scenarios and found one where the logic fails: when the new IA is enabled but the Take Action parent page is not selected from the Settings page, the P4 Actions are not shown.
> New IA 	P4 Actions 	Parent Page 	Children Pages 	Result 	Comments
> Yes 	Yes 	Yes 	Yes 	✔ 	
> Yes 	Yes 	Yes 	No 	✔ 	
> Yes 	Yes 	No 	Yes 		No P4 Actions are shown
> Yes 	Yes 	No 	No 		No P4 Actions are shown
> Yes 	No 	Yes 	Yes 	✔ 	
> Yes 	No 	Yes 	No 	✔ 	
> Yes 	No 	No 	Yes 	✔ 	
> Yes 	No 	No 	No 	✔ 	
> No 	- 	Yes 	Yes 	✔ 	
> No 	- 	Yes 	No 	✔ 	
> No 	- 	No 	Yes 	✔ 	
> No 	- 	No 	No 	✔ 	
> 
> _References:_
> 
>     * New IA: whether the new IA is enabled.
> 
>     * P4 Actions: whether posts of type P4 Actions exist and are published.
> 
>     * Parent Page: whether the main "Take Action" or "Act" page is selected.
> 
>     * Children Pages: whether children's pages of the Parent Page exist and are published.

@mardelnet do you have time to validate this table again? Thanks!

assets/src/block-editor/setupTaxonomyBreadcrumbBlock.js
mardelnet: This problem still persists.
Maybe changing this could fix it:

```js
const link = term ? wp.element.createElement('a', linkAttrs, term || 'Loading...') : null;
const content = link ? wp.element.createElement('div', contentAttrs, link) : null;
```

mardelnet:
> > @GP-Dan-Tovbein, I went through all the possible scenarios and found one where the logic fails: when the new IA is enabled but the Take Action parent page is not selected from the Settings page, the P4 Actions are not shown.

> @mardelnet do you have time to validate this table again? Thanks!

Fixed, great job!

assets/src/block-editor/setupTaxonomyBreadcrumbBlock.js
GP-Dan-Tovbein: Updated!

mardelnet (approved):
Excellent job, @GP-Dan-Tovbein

PR #2493, opened 2025-01-07
PLANET-7489: Check number of columns in the P4 columns block

mardelnet:
### Summary

This is a temporary fix to solve the issue mentioned by Kelli in the ticket [PLANET-7489](https://jira.greenpeace.org/browse/PLANET-7489).
This PR adds an additional column to the Blocks Report page only once the P4 Columns block is selected, to show the number of columns for each block. If more than 4 columns were added to the block (the maximum allowed) a warning message is shown.

Report-‹-Greenpeace-—-WordPress
---

Ref: https://jira.greenpeace.org/browse/PLANET-7489 

### Testing

1. Go to the Blocks Report page (`/wp-admin/admin.php?page=plugin_blocks_report`)
2. Filter only P4 Columns block.
3. Check the Columns table's column.
4. Check if there are blocks using more than 4 columns. If so, a warning should be shown. If not, create a new page, add the following block code and check the report again:

``` 
<!-- wp:planet4-blocks/columns {"columns_title":"Test Column","columns":[{"title":"Column A","description":"This is a test description","cta_text":"A test button!"},{"title":"Column B","cta_text":"A test button!","description":"This is a test description"},{"title":"Column C","cta_text":"A test button!","description":"This is a test description"},{"title":"Column D","cta_text":"A test button!","description":"This is a test description"},{"title":"Column E","description":"This is a test description","attachment":0,"cta_link":"","cta_text":"A test button!","link_new_tab":false}],"className":"is-style-no_image"} /-->
```

### DIFF
--- a/src/BlockReportSearch/Block/BlockUsageTable.php
+++ b/src/BlockReportSearch/Block/BlockUsageTable.php
@@ -473,9 +473,9 @@ class BlockUsageTable extends WP_List_Table
      */
     public function column_block_attrs($item): string
     {
-        $content = $item['block_attrs'] ?? "";
+        $content = $item['block_attrs'] ?? null;
         if (empty($content)) {
-            return $content;
+            return '';
         }
 
 		//phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_print_r , Squiz.PHP.DiscouragedFunctions.Discouraged
--- a/src/BlockReportSearch/Block/BlockUsageTable.php
+++ b/src/BlockReportSearch/Block/BlockUsageTable.php
@@ -473,9 +473,9 @@ class BlockUsageTable extends WP_List_Table
      */
     public function column_block_attrs($item): string
     {
-        $content = $item['block_attrs'] ?? null;
+        $content = $item['block_attrs'] ?? "";
         if (empty($content)) {
-            return '';
+            return $content;
         }
 
 		//phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_print_r , Squiz.PHP.DiscouragedFunctions.Discouraged
--- a/src/BlockReportSearch/Block/BlockUsageTable.php
+++ b/src/BlockReportSearch/Block/BlockUsageTable.php
@@ -272,6 +272,12 @@ class BlockUsageTable extends WP_List_Table
             'post_status' => 'Status',
         ];
 
+        // This is a temporary fix to identify problems with the columns migrated here:
+        // https://jira.greenpeace.org/browse/PLANET-7489
+        if (isset($_GET['name']) && $_GET['name'] === 'planet4-blocks/columns') {
+            $default_columns['block_cols'] = 'Columns';
+        }
+
         $this->columns = array_merge(
             [ $this->group_by => $default_columns[ $this->group_by ] ],
             $default_columns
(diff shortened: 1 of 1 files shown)


planet-4:

:watch: 2025.01.07 10:43:09

src/BlockReportSearch/Block/BlockUsageTable.php
```diff
@@ -487,6 +493,33 @@ public function column_block_attrs($item): string
...
+     */
+    public function column_block_cols($item): string
+    {
+        $content = $item['block_attrs'] ?? null;
+
+        if (empty($content)) {
+            return '';
+        }
+
+        $cols = $content['columns'] ? count($content['columns']) : '0';
```
GP-Dan-Tovbein, line 514: What do you think about this suggestion? I think you need to cast only within the `if` clause.
```suggestion
               $cols = $content['columns'] ? $content['columns'] : '0';

        if (count($cols) > 4) {
            return "<span style='color: red;'>$cols - Fix needed!</span>";
        }

        return $cols;
```

mardelnet: Unless I'm wrong, with your suggestion we will be returning `$cols` as an array (equal to `$content['columns']`) but we need to return the number of elements of the array (`count($content['columns'])`)

src/BlockReportSearch/Block/BlockUsageTable.php
```diff
@@ -487,6 +493,33 @@ public function column_block_attrs($item): string
...
+     * We need this function to identify P4 columns blocks using more than 4 columns.
+     *
+     * @param array $item Item.
+     * @phpcs:disable SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingNativeTypeHint
+     */
+    public function column_block_cols($item): string
+    {
+        $content = $item['block_attrs'] ?? null;
+
+        if (empty($content)) {
```
mleray, line 510: If I'm not mistaken, `empty()` actually returns false with `null` 🤔 So maybe we should set `content` to an empty string instead?

mleray: Just checked it [there](https://stackoverflow.com/questions/5615747/what-is-the-difference-between-null-and-empty)

mleray (commented):
FYI I've created [this page](https://www-dev.greenpeace.org/test-mars/more-columns/) on the mars instance, and the functionality works fine 👍
But in the page's frontend the block doesn't display, is that expected?

mardelnet:
> FYI I've created [this page](https://www-dev.greenpeace.org/test-mars/more-columns/) on the mars instance, and the functionality works fine 👍 But in the page's frontend the block doesn't display, is that expected?

@mleray, I think there was a problem with the block code I had added to the PR description. I've updated the description and the page you created, too. [Would you mind rechecking it](https://www-dev.greenpeace.org/test-mars/more-columns/)?

src/BlockReportSearch/Block/BlockUsageTable.php
mardelnet: Nice catch! Fixed!

mleray: You didn't fix the right one 😉

mardelnet: I reverted the change because there was no mistake: if a variable is `null` it is also considered "empty". 
Check it out: https://onlinephp.io?s=s7EvyCjg5VLJK83JiS9LLFKwVQAxrXm5UpMz8hVScwtKKjXgsppAcXs7Xi4A&v=8.2.20

mardelnet: You can also see this in the link you shared before:
image

mleray: Ah, it seems you're right, maybe the thread I found was old then 🤔 okay, nevermind!

mleray: Ah I misunderstood it 🤦‍♀️

sagarsdeshmukh (approved):
👍

PR #2494, opened 2025-01-08
PLANET-6530 Add new Secondary Navigation Block

Osong-Michael:
### Summary

---

Ref: 

### Testing

1. 
2. 
3.

### DIFF
--- a/assets/src/blocks/SecondaryNavigation/SecondaryNavigationBlock.js
+++ b/assets/src/blocks/SecondaryNavigation/SecondaryNavigationBlock.js
@@ -1,0 +1,39 @@
+import {SecondaryNavigationEditor} from './SecondaryNavigationEditor';
+import {example} from './example';
+
+const {__} = wp.i18n;
+
+const BLOCK_NAME = 'planet4-blocks/secondary-navigation';
+
+export const registerSecondaryNavigationBlock = () => {
+  const {registerBlockType} = wp.blocks;
+
+  registerBlockType(BLOCK_NAME, {
+    title: 'Secondary Navigation Menu',
+    description: __('Inserts a secondary navigation menu to the page that leads to different sections of the same page.', 'planet4-blocks-backend'),
+    icon: 'welcome-widgets-menus',
+    category: 'planet4-blocks',
+    attributes: {
+      levels: {
+        type: 'array',
+        default: [{heading: 2, link: true}],
+      },
+      exampleMenuItems: { // Used for the block's preview, which can't extract items from anything.
+        type: 'array',
+      },
+    },
+    isExample: {
+      type: 'boolean',
+      default: false,
+    },
+    supports: {
+      multiple: false, // Use the block just once per post.
+      html: false,
+    },
+    edit: SecondaryNavigationEditor,
+    save() {
+      return null;
+    },
+    example,
+  });
+};
--- a/assets/src/blocks/SecondaryNavigation/SecondaryNavigationEditor.js
+++ b/assets/src/blocks/SecondaryNavigation/SecondaryNavigationEditor.js
@@ -1,0 +1,69 @@
+import {getHeadingsFromBlocks} from './generateHeadingsForBlock';
+
+const {useSelect} = wp.data;
+const {InspectorControls} = wp.blockEditor;
+const {PanelBody} = wp.components;
+const {__} = wp.i18n;
+
+const renderEdit = () => {
+  return (
+    <InspectorControls>
+      <PanelBody title={__('Learn more about this block', 'planet4-blocks-backend')} initialOpen={false}>
+        <p className="components-base-control__help">
+          <a target="_blank" href="https://planet4.greenpeace.org/content/blocks/table-of-contents/" rel="noreferrer">
+            P4 Handbook P4 Secondary Navigation Menu
+          </a>
+          {' '} &#128203;
+        </p>
+      </PanelBody>
+    </InspectorControls>
+  );
+};
+
+const renderView = attributes => {
+  const {
+    levels,
+    isExample,
+    exampleMenuItems,
+  } = attributes;
+
+  const blocks = useSelect(select => select('core/block-editor').getBlocks(), []);
+
+  const flatHeadings = getHeadingsFromBlocks(blocks, levels);
+
+  const menuItems = isExample ? exampleMenuItems : flatHeadings;
+
+  return (
+    <section className="block secondary-navigation-block">
+      {menuItems.length > 0 ?
+        <div className="secondary-navigation-menu">
+          <ul className="secondary-navigation-item">
+            {menuItems.map(({anchor, text, shouldLink}) => (
+              <li key={anchor} className={`list-style-none ${shouldLink ? 'list-link' : 'list-heading'}`}>
+                {shouldLink ?
+                  <a
+                    className="icon-link secondary-navigation-link"
+                    href={`#${anchor}`}
+                  >
+                    {text}
+                  </a> :
+                  <span className="secondary-navigation-heading">{text}</span>
+                }
+              </li>
+            ))}
+          </ul>
+        </div> :
+        <div className="EmptyMessage">
+          {__('There are not any pre-established headings that this block can display in the form of a secondary navigation menu. Please add headings to your page.', 'planet4-blocks-backend')}
+        </div>
+      }
+    </section>
+  );
+};
+
+export const SecondaryNavigationEditor = ({attributes, isSelected}) => (
+  <>
+    {isSelected && renderEdit()}
+    {renderView(attributes)}
+  </>
+);
--- a/assets/src/blocks/SecondaryNavigation/SecondaryNavigationFrontend.js
+++ b/assets/src/blocks/SecondaryNavigation/SecondaryNavigationFrontend.js
@@ -1,0 +1,27 @@
+import {getHeadingsFromDom} from '../TableOfContents/getHeadingsFromDom';
+
+export const SecondaryNavigationFrontend = ({levels}) => {
+  const headings = getHeadingsFromDom(levels);
+
+  return (
+    <section className="block secondary-navigation-block">
+      <div className="secondary-navigation-menu">
+        <ul className="secondary-navigation-item">
+          {headings.map(({anchor, text, shouldLink}) => (
+            <li key={anchor} className={`list-style-none ${shouldLink ? 'list-link' : 'list-heading'}`}>
+              {shouldLink ?
+                <a
+                  className="icon-link secondary-navigation-link"
+                  href={`#${anchor}`}
+                >
+                  {text}
+                </a> :
+                <span className="secondary-navigation-heading">{text}</span>
+              }
+            </li>
+          ))}
+        </ul>
+      </div>
+    </section>
+  );
+};
--- a/assets/src/blocks/TableOfContents/getHeadingsFromDom.js
+++ b/assets/src/blocks/TableOfContents/getHeadingsFromDom.js
@@ -9,7 +9,9 @@ export const getHeadingsFromDom = selectedLevels => {
   }
 
   // Get all heading tags that we need to query
-  const headingsSelector = selectedLevels.map(level => `:not(.table-of-contents-block) > h${level.heading}`);
+  const headingsSelector = selectedLevels.map(
+    level => `:not(.table-of-contents-block):not(.secondary-navigation-block) > h${level.heading}`
+  );
 
   const usedAnchors = [];
 
--- a/src/Blocks/SecondaryNavigation.php
+++ b/src/Blocks/SecondaryNavigation.php
@@ -1,0 +1,73 @@
+<?php
+
+/**
+ * SecondaryNavigation block class.
+ *
+ * @package P4\MasterTheme
+ * @since 0.1
+ */
+
+ namespace P4\MasterTheme\Blocks;
+
+/**
+ * Class SecondaryNavigation
+ *
+ * @package P4\MasterTheme\Blocks
+ */
+class SecondaryNavigation extends BaseBlock
+{
+    /**
+     * Block name.
+     *
+     * @const string BLOCK_NAME.
+     */
+    public const BLOCK_NAME = 'secondary-navigation';
+
+    /**
+     * SecondaryNavigation constructor.
+     */
+    public function __construct()
+    {
+
+        $this->register_secondary_navigation_block();
+    }
+
+    /**
+     * Register SecondaryNavigation block.
+     */
+    public function register_secondary_navigation_block(): void
+    {
+        register_block_type(
+            self::get_full_block_name(),
+            [
+                'render_callback' => [ self::class, 'render_frontend' ],
+                'attributes' => [
+                    'levels' => [
+                        'type' => 'array',
+                        'default' => [
+                            [
+                                'heading' => 2,
+                                'link' => false,
+                            ],
+                        ],
+                    ],
+                ],
+            ]
+        );
+
+        add_action('enqueue_block_editor_assets', [ self::class, 'enqueue_editor_assets' ]);
+        add_action('wp_enqueue_scripts', [ self::class, 'enqueue_frontend_assets' ]);
+    }
+
+    /**
+     * Required by the `Base_Block` class.
+     *
+     * @param array $fields Unused, required by the abstract function.
+     * @phpcs:disable SlevomatCodingStandard.Functions.UnusedParameter.UnusedParameter
+     */
+    public function prepare_data(array $fields): array
+    {
+        return [];
+    }
+    // @phpcs:enable SlevomatCodingStandard.Functions.UnusedParameter.UnusedParameter
+}
--- a/assets/src/blocks/SecondaryNavigation/example.js
+++ b/assets/src/blocks/SecondaryNavigation/example.js
@@ -1,0 +1,27 @@
+export const example = {
+  viewportWidth: 992,
+  attributes: {
+    isExample: true,
+    exampleMenuItems: [
+      {
+        text: 'Title 1',
+        anchor: 'Title 1',
+        level: 2,
+        shouldLink: true,
+      },
+      {
+        text: 'Title 2',
+        anchor: 'Title 2',
+        level: 2,
+        shouldLink: true,
+      },
+      {
+        text: 'Title 3',
+        anchor: 'Title 3',
+        level: 2,
+        shouldLink: true,
+      },
+    ],
+  },
+};
+  
--- a/assets/src/blocks/SecondaryNavigation/generateHeadingsForBlock.js
+++ b/assets/src/blocks/SecondaryNavigation/generateHeadingsForBlock.js
@@ -1,0 +1,30 @@
+import {generateAnchor} from '../TableOfContents/generateAnchor';
+import {unescape} from '../../functions/unescape';
+
+const stripTags = str => str.replace(/(<([^>]+)>)/ig, ''); //NOSONAR
+  
+export const getHeadingsFromBlocks = (blocks, selectedLevels) => {
+  const headings = [];
+  blocks.forEach(block => {
+    if (block.name === 'core/heading') {
+      const blockLevel = block.attributes.level;
+  
+      const levelConfig = selectedLevels.find(selected => selected.heading === blockLevel);
+  
+      if (!levelConfig) {
+        return;
+      }
+  
+      const anchor = block.attributes.anchor || generateAnchor(block.attributes.content, headings.map(h => h.anchor));
+  
+      headings.push({
+        level: blockLevel,
+        content: unescape(stripTags(block.attributes.content)),
+        anchor,
+        shouldLink: levelConfig.link,
+      });
+    }
+  });
+  
+  return headings;
+};
--- a/assets/src/editorIndex.js
+++ b/assets/src/editorIndex.js
@@ -12,6 +12,7 @@ import {registerColumnsBlock} from './blocks/Columns/ColumnsBlock';
 import {registerBlockStyles} from './block-styles';
 import {registerBlockVariations} from './block-variations';
 import {registerActionButtonTextBlock} from './blocks/ActionCustomButtonText';
+import {registerSecondaryNavigationBlock} from './blocks/SecondaryNavigation/SecondaryNavigationBlock';
 
 wp.domReady(() => {
   // Blocks
@@ -21,6 +22,7 @@ wp.domReady(() => {
   registerHappyPointBlock();
   registerSocialMediaBlock();
   registerTimelineBlock();
+  registerSecondaryNavigationBlock();
 
   // Block Templates
   registerBlockTemplates();
--- a/assets/src/frontendIndex.js
+++ b/assets/src/frontendIndex.js
@@ -8,12 +8,14 @@ import {TableOfContentsFrontend} from './blocks/TableOfContents/TableOfContentsF
 import {HappyPointFrontend} from './blocks/HappyPoint/HappyPointFrontend';
 import {ColumnsFrontend} from './blocks/Columns/ColumnsFrontend';
 import {setupLightboxForImages} from './blocks/components/Lightbox/setupLightboxForImages';
+import {SecondaryNavigationFrontend} from './blocks/SecondaryNavigation/SecondaryNavigationFrontend';
 
 // Render React components
 const COMPONENTS = {
   'planet4-blocks/submenu': TableOfContentsFrontend,
   'planet4-blocks/happypoint': HappyPointFrontend,
   'planet4-blocks/columns': ColumnsFrontend,
+  'planet4-blocks/secondary-navigation': SecondaryNavigationFrontend,
 };
 
 document.addEventListener('DOMContentLoaded', () => {
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -203,6 +203,7 @@ final class Loader
         new Blocks\TableOfContents();//NOSONAR
         new Blocks\TakeActionBoxout();//NOSONAR
         new Blocks\Timeline();//NOSONAR
+        new Blocks\SecondaryNavigation();//NOSONAR
 
         register_block_pattern_category(
             'page-headers',


planet-4:

:watch: 2025.01.10 21:36:14

assets/src/blocks/SecondaryNavigation/SecondaryNavigationBlock.js
```diff
@@ -0,0 +1,39 @@
...
+const BLOCK_NAME = 'planet4-blocks/secondary-navigation';
+
+export const registerSecondaryNavigationBlock = () => {
+  const {registerBlockType} = wp.blocks;
+
+  registerBlockType(BLOCK_NAME, {
+    title: 'Secondary Navigation Menu',
+    description: __('Inserts a secondary navigation menu to the page that leads to different sections of the same page.', 'planet4-blocks-backend'),
+    icon: 'menu-alt3',
+    category: 'planet4-blocks-beta',
```
mardelnet, line 15: It looks like the block can be used even when the Beta Block feature is disabled.

assets/src/blocks/SecondaryNavigation/SecondaryNavigationEditor.js
```diff
@@ -0,0 +1,66 @@
...
+                  className="secondary-navigation-link"
+                  href={`#${anchor}`}
+                >
+                  {content}
+                </a>
+              </li>
+            ))}
+          </ul>
+        </div> :
+        <div className="EmptyMessage">
```
mardelnet, line 53: There's a misalignment with the empty message:

image

assets/src/blocks/TableOfContents/getHeadingsFromDom.js
```diff
@@ -9,7 +9,9 @@ export const getHeadingsFromDom = selectedLevels => {
   }
 
   // Get all heading tags that we need to query
-  const headingsSelector = selectedLevels.map(level => `:not(.table-of-contents-block) > h${level.heading}`);
+  const headingsSelector = selectedLevels.map(
```
mardelnet, line 12: Are we only using `H2` headings for a reason? Is that part of the requirements?

mardelnet (commented):
Well done, @Osong-Michael.
I have some initial comments:

assets/src/blocks/TableOfContents/getHeadingsFromDom.js
Osong-Michael: Yes, `h2` elements it what we are using for the first iteration.

assets/src/js/app.js
```diff
@@ -32,4 +33,5 @@ document.addEventListener('DOMContentLoaded', () => {
   removeRelatedPostsSection();
   setupClickabelActionsListCards();
   setupCountrySelector();
+  makeSecondaryNavigationStickyonScroll();
```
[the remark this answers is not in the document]
mardelnet: Can't we move it to the `SecondaryNavigationFrontend.js` file?

assets/src/js/sticky_on_scroll_sn.js
```diff
@@ -0,0 +1,20 @@
+// Adds class to Secondary Navigation block to make it sticky on scroll
+export const makeSecondaryNavigationStickyonScroll = () => {
+  setTimeout(() => {
```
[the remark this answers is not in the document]
Osong-Michael: This [question](https://github.com/greenpeace/planet4-master-theme/pull/2494#discussion_r1957823249) and this one are related, so I will answer both here.

Generally the scripts in `app.js` run only after the `DOMContentLoaded` event listener, but during my testing most operations on blocks fail if you immediately run them after the event above, so adding a delay after the event is a safe guard to ensure everything has mounted and is visible.

assets/src/js/app.js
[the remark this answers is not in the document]
Osong-Michael: 🤔 we could, I don't see why not

assets/src/js/sticky_on_scroll_sn.js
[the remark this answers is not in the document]
mardelnet: You can remove the `sticky_on_scroll_sn.js` file and move the code to the `SecondaryNavigationFrontend.js` file as a separate function. Then, inside the `SecondaryNavigationFrontend` component, you call the function using a callback in `useEffect`. That way, the script is only added when the block is used, and you can get rid of the `setTimeOut`:

```js 
const makeSecondaryNavigationStickyonScroll = () => {
  const stickyElement = document.querySelector('.secondary-navigation-block');
  const container = document.querySelector('.page-content');
  const offset = -500;

  window.addEventListener('scroll', () => {
    const containerRect = container.getBoundingClientRect();
    const stickyRect = stickyElement.getBoundingClientRect();

    if (containerRect.top <= offset && containerRect.bottom > stickyRect.height + offset) {
      stickyElement.classList.add('stuck');
    } else {
      stickyElement.classList.remove('stuck');
    }
  });
};
``` 
```js
useEffect(() => {
    makeSecondaryNavigationStickyonScroll();
  }, []);
```

assets/src/blocks/SecondaryNavigation/SecondaryNavigationEditor.js
```diff
@@ -0,0 +1,72 @@
+import {getHeadingsFromBlocks} from '../TableOfContents/getHeadingsFromBlocks';
+import {initializeJustifyContentAdjustment} from './adjustNavWidth';
+
+const {useSelect} = wp.data;
+const {useEffect} = wp.element;
+const {InspectorControls} = wp.blockEditor;
+const {PanelBody} = wp.components;
+const {__} = wp.i18n;
+
+const renderEdit = () => {
```
mleray, line 10: Very tiny suggestion, we can remove the `return` here and directly use the parentheses:
```
const renderEdit = () => (...);
```

assets/src/blocks/SecondaryNavigation/SecondaryNavigationEditor.js
```diff
@@ -0,0 +1,72 @@
...
+            P4 Handbook - P4 Secondary Navigation Menu
+          </a>
+          {' '} &#128203;
+        </p>
+      </PanelBody>
+    </InspectorControls>
+  );
+};
+
+const renderView = attributes => {
```
mleray, line 25: We could declare the various attributes directly:
```
const renderView = ({levels, isExample, exampleMenuItems}) => {...}
```

assets/src/blocks/SecondaryNavigation/adjustNavWidth.js
```diff
@@ -0,0 +1,19 @@
+function adjustJustifyContent() {
+  const container = document.querySelector('.secondary-navigation-item');
+  if (!container) {return;}
+
+  const children = Array.from(container.children);
+  const containerWidth = container.clientWidth;
+  let totalChildrenWidth = (children.length - 1) * 48;
+
+  children.forEach(child => {
```
mleray, line 9: This can be one line:
```
children.forEach(child => totalChildrenWidth += child.offsetWidth);
```

assets/src/scss/blocks/SecondaryNavigation/SecondaryNavigationStyle.scss
```diff
@@ -0,0 +1,83 @@
...
+  left: 0;
+}
+
+.secondary-navigation-item {
+  list-style: none;
+  display: flex;
+  overflow-x: auto;
+  flex-wrap: nowrap;
+  scroll-padding-inline-start: 24px;
+  gap: 48px;
```
mleray, line 39: Can we try using variables wherever possible?

assets/src/blocks/SecondaryNavigation/SecondaryNavigationFrontend.js
```diff
@@ -0,0 +1,213 @@
+import {useState, useEffect, useRef} from '@wordpress/element';
+import {getHeadingsFromDom} from '../TableOfContents/getHeadingsFromDom';
+import {initializeJustifyContentAdjustment} from './adjustNavWidth';
+
+const makeSecondaryNavigationStickyonScroll = () => {
```
mleray, line 5: To respect the case, I think that `on` should also start with a capital letter

mleray: Also, can we add some JS docs to this function?

assets/src/blocks/SecondaryNavigation/SecondaryNavigationFrontend.js
```diff
@@ -0,0 +1,213 @@
...
+  const [currentHeaderLink, setCurrentHeaderLink] = useState('');
+  const [showLeftArrow, setShowLeftArrow] = useState(false);
+  const [showRightArrow, setShowRightArrow] = useState(false);
+  const isClicking = useRef(false);
+  const isManualScroll = useRef(false);
+  const hasLoaded = useRef(false);
+  const navListRef = useRef(null);
+
+  const isMobile = window.innerWidth <= 991;
+
```
mleray, line 37: Unnecessary extra empty line here

assets/src/blocks/SecondaryNavigation/SecondaryNavigationFrontend.js
```diff
@@ -0,0 +1,213 @@
...
+    };
+
+    const observer = new IntersectionObserver(observerCallback, observerOptions);
+
+    headings.forEach(({anchor}) => {
+      const link = document.getElementById(anchor);
+      if (link) {observer.observe(link);}
+    });
+
+    setTimeout(() => {
```
mleray, line 85: I think that this can be shortened to 1 line: `setTimeout(() => hasLoaded.current = true, 500);`

assets/src/blocks/SecondaryNavigation/SecondaryNavigationFrontend.js
```diff
@@ -0,0 +1,213 @@
...
+      headings.forEach(({anchor}) => {
+        const link = document.getElementById(anchor);
+        if (link) {observer.unobserve(link);}
+      });
+    };
+  }, [headings]);
+
+  useEffect(() => {
+    // For smaller screens to update the new Navigation with the active element.
+    if (isMobile) {
```
mleray, line 99: Can we use early return if `isMobile` is false?

assets/src/blocks/SecondaryNavigation/SecondaryNavigationFrontend.js
```diff
@@ -0,0 +1,213 @@
...
+  const handleDropdown = () => {
+    setIsDropdownOpen(prev => !prev);
+  };
+
+  const scrollLeft = () => {
+    isManualScroll.current = true;
+    navListRef.current?.scrollBy({left: -150, behavior: 'smooth'});
+
+    // Reset after short delay
+    setTimeout(() => {
```
mleray, line 154: I think that this can also be simplified into 1 line

assets/src/blocks/SecondaryNavigation/SecondaryNavigationFrontend.js
```diff
@@ -0,0 +1,213 @@
...
+      isManualScroll.current = false;
+    }, 500);
+  };
+
+  const scrollRight = () => {
+    isManualScroll.current = true;
+    navListRef.current.scrollBy({left: 150, behavior: 'smooth'});
+
+    setTimeout(() => {
```
mleray, line 163: Same here, we can probably simplify it

assets/src/blocks/SecondaryNavigation/SecondaryNavigationFrontend.js
```diff
@@ -0,0 +1,213 @@
...
+          {showLeftArrow && (
+            <button className="nav-arrow left" onClick={scrollLeft}>
+              &#8592;
+            </button>
+          )}
+          <div className="secondary-nav-scroll-wrapper">
+            <ul className="secondary-navigation-item" ref={navListRef}>
+              {headings.map(({anchor, content}) => (
+                <li
+                  key={anchor}
```
mleray, line 190: This doesn't need to be on a separate line

mleray:
The position is off if the mobile tabs are disabled:

Screenshot 2025-07-30 at 14 52 38

Osong-Michael:
> admin

That is because you are logged in, seems to affect the space in the page layout hence affecting the position of the secondary navigation, happens even on larger screens.

mleray (commented):
On mobile you have to scroll quite far into the bar before it sticks to the top, whereas on desktop it's pretty much immediate

Screenshot 2025-07-31 at 13 16 28

mleray:
In all screen sizes, the height of the secondary menu is bigger than the search bar, so it looks off when both are present:

Screenshot 2025-07-31 at 13 18 53

mleray:
Is it expected that the items are not vertically centered?

Screenshot 2025-07-31 at 13 20 04

assets/src/blocks/SecondaryNavigation/SecondaryNavigationBlock.js
```diff
@@ -0,0 +1,39 @@
...
+    attributes: {
+      levels: {
+        type: 'array',
+        default: [{heading: 2, link: true}],
```
mleray, line 19: Is there an option to change this?

Osong-Michael: I will assume at some point we may want to use this for different heading levels, maybe give editors the option to use it for h2 only or h3 only

assets/src/blocks/SecondaryNavigation/SecondaryNavigationEditor.js
```diff
@@ -0,0 +1,64 @@
...
+import {getHeadingsFromBlocks} from '../TableOfContents/getHeadingsFromBlocks';
+import {initializeJustifyContentAdjustment} from './adjustNavWidth';
+
+const {useSelect} = wp.data;
+const {InspectorControls} = wp.blockEditor;
+const {PanelBody} = wp.components;
+const {__} = wp.i18n;
+
+const renderEdit = () =>  (
+  // Update with correct link
```
mleray, line 11: Can you please add a `TODO` here?

mleray:
In the editor, the empty block message has a dark green band on top, it looks weird and shouldn't be there:

Screenshot 2025-08-07 at 16 33 50

assets/src/blocks/SecondaryNavigation/SecondaryNavigationEditor.js
```diff
@@ -0,0 +1,64 @@
+import {useEffect} from '@wordpress/element';
+import {getHeadingsFromBlocks} from '../TableOfContents/getHeadingsFromBlocks';
+import {initializeJustifyContentAdjustment} from './adjustNavWidth';
+
+const {useSelect} = wp.data;
+const {InspectorControls} = wp.blockEditor;
+const {PanelBody} = wp.components;
+const {__} = wp.i18n;
+
+const renderEdit = () =>  (
```
mleray, line 10: Maybe we should just comment this out for now, until the documentation is updated? It's weird to redirect to another block's docs

mleray:
The tabbing behaviour is broken when you get to the "edge" links in the secondary navigation, meaning the ones that are hidden and would need the arrows to be shown:

https://github.com/user-attachments/assets/9f628533-1956-42f2-8009-d1c42bf920db

assets/src/blocks/SecondaryNavigation/SecondaryNavigationEditor.js
```diff
@@ -0,0 +1,65 @@
...
+
+  const flatHeadings = getHeadingsFromBlocks(blocks, levels);
+
+  const menuItems = isExample ? exampleMenuItems : flatHeadings;
+
+  return (
+    <section className="block secondary-navigation-block">
+      {menuItems.length > 0 ?
+        <div className="secondary-navigation-menu">
+          <ul className="secondary-navigation-item">
```
mleray, line 38: This is really nit-picky but shouldn't the classname here be `secondary-navigation-items` with an 's' since it's the container?

assets/src/blocks/SecondaryNavigation/SecondaryNavigationFrontend.js
```diff
@@ -0,0 +1,330 @@
...
+    document.addEventListener('click', handleClickOutside);
+
+    return () => document.removeEventListener('click', handleClickOutside);
+  }, [isDropdownOpen, isMobile]);
+
+  useEffect(() => {
+    if (!isMobile) {return;}
+
+    const footer = document.querySelector('footer');
+    if (!footer || !dropdownRef.current) {return;}
```
mleray, line 225: Why isn't `!dropdownRef.current` together with `!isMobile` right before?

assets/src/blocks/SecondaryNavigation/SecondaryNavigationFrontend.js
```diff
@@ -0,0 +1,330 @@
+import {useState, useEffect, useRef} from '@wordpress/element';
+import {getHeadingsFromDom} from '../TableOfContents/getHeadingsFromDom';
```
mleray, line 2: It's confusing to use a function from the `TableOfContents` folder here, can we move this function somewhere else instead? Maybe in `src/functions` folder?

assets/src/scss/blocks/SecondaryNavigation/SecondaryNavigationStyle.scss
```diff
@@ -0,0 +1,322 @@
+.secondary-navigation-block {
+  margin-left: calc(-50vw - -50%);
```
mleray, line 2: Can't we use `+ 50%` instead of `- -50%`? 😅

assets/src/scss/blocks/SecondaryNavigation/SecondaryNavigationStyle.scss
```diff
@@ -0,0 +1,322 @@
...
+    position: absolute;
+    left: 0;
+    height: 0;
+    padding: 0;
+    overflow: hidden;
+    transition: height 0.3s linear;
+  }
+
+  .EmptyMessage {
+    text-align: center;
```
mleray, line 23: Is this something that we'd want to add for all `EmptyMessage` components?

assets/src/scss/blocks/SecondaryNavigation/SecondaryNavigationStyle.scss
```diff
@@ -0,0 +1,322 @@
...
+.secondary-navigation-menu {
+  position: relative;
+
+  .nav-arrow {
+    position: absolute;
+    top: 15px;
+    background: none;
+    border: none;
+    color: var(--white);
+    font-size: 20px;
```
mleray, line 36: Maybe we can use variables or tokens for font sizes?

assets/src/scss/blocks/SecondaryNavigation/SecondaryNavigationStyle.scss
```diff
@@ -0,0 +1,322 @@
...
+    display: none;
+  }
+
+  @include medium-and-less {
+    flex-direction: column;
+    height: auto;
+    overflow: hidden;
+    margin: auto;
+    width: 100vw;
+    margin-left: calc(-50vw - -50%);
```
mleray, line 223: Same comment here about `+`

assets/src/scss/blocks/SecondaryNavigation/SecondaryNavigationStyle.scss
```diff
@@ -0,0 +1,322 @@
...
+}
+
+.current-active-class,
+.dropdown-btn {
+  margin-top: 15px;
+}
+
+.sn-wrapper {
+  width: 100vw;
+  margin-left: calc(-50vw - -50%);
```
mleray, line 296: Same comment here about `+`

src/Blocks/SecondaryNavigation.php
```diff
@@ -0,0 +1,73 @@
...
+     * @const string BLOCK_NAME.
+     */
+    public const BLOCK_NAME = 'secondary-navigation';
+
+    /**
+     * SecondaryNavigation constructor.
+     */
+    public function __construct()
+    {
+
```
mleray, line 31: Unnecessary empty line

assets/src/scss/blocks/SecondaryNavigation/SecondaryNavigationStyle.scss
Osong-Michael: I can't say for sure, seems to work well in the Guestbook block but then again that is used in a p tag

assets/src/blocks/SecondaryNavigation/adjustNavWidth.js
```diff
@@ -0,0 +1,17 @@
+const adjustJustifyContent = () => {
+  const container = document.querySelector('.secondary-navigation-item');
```
mleray, line 2: Shouldn't this be `.secondary-navigation-items` now?

assets/src/js/header/setupMobileTabsMenuScroll.js
```diff
@@ -76,6 +76,8 @@ export default () => {
       lastScrollDir = dir;
       lastScrollRef = lastScrollTop;
       menu.classList.add(HIDDEN_TABS_CLASS);
+      document.body.classList.remove('stuck-open');
```
mardelnet, line 79: Can we use a constant value for `stuck-open`?

mardelnet (commented):
@Osong-Michael, I have some comments for you. You can check them [here](https://www-dev.greenpeace.org/test-phobos/pedro-test-page).

1. The navigation only gets fixed when a Page Header pattern exists. Is that part of the requirements?
2. When two or more titles are visible on screen at the same time, both are underlined in the navigation, and the page URL changes constantly from one anchor to another.
3. In the editor, not all the links are shown in the block (check it [here](https://www-dev.greenpeace.org/test-phobos/wp-admin/post.php?post=4720&action=edit))

assets/src/js/header/setupMobileTabsMenuScroll.js
Osong-Michael: This is just a class we add to the body so we can use it's presence or absence to style certain components differently. It is the same thing I do with the presence or absence of the `admin-bar` class, that is automatically added by WP. This solution makes it cleaner to just use `SCSS`, in my opinion.

Osong-Michael:
> @Osong-Michael, I have some comments for you. You can check them [here](https://www-dev.greenpeace.org/test-phobos/pedro-test-page).
> 
> 1. The navigation only gets fixed when a Page Header pattern exists. Is that part of the requirements?
> 2. When two or more titles are visible on screen at the same time, both are underlined in the navigation, and the page URL changes constantly from one anchor to another.
> 3. In the editor, not all the links are shown in the block (check it [here](https://www-dev.greenpeace.org/test-phobos/wp-admin/post.php?post=4720&action=edit))

So here are some points to clarify @mardelnet, we did have some meetings with the design team on this and this was the resolution for the MVP for this new block.

1. For now the navigation only works best when a Page Header pattern is present, this is also reflected in the [designs](https://www.figma.com/design/9NtbY8n3at8uOEJTsLrETb/P4-Design-System?node-id=8995-14821&p=f&t=ZG5qYsjHyGQxRcgc-0) and will be included in the Documentation so editors can be aware of this.

2. This is as a result of you not having enough content on your page which defeats the purpose of the block. It is meant for when pages have a lot of content and users want to navigate quickly without having to scroll a lot. The reason you are experiencing that issue is because we use an `IntersectionObserver` that automatically updates the navigation block with the current header title as that sections becomes visible, in your example they are all visible.

3. Currently, this block only works with h2 headers,  so any other level won't be picked up.

assets/src/blocks/SecondaryNavigation/SecondaryNavigationFrontend.js
mardelnet, line 1: Can you please document all the functions in this file to make it easier to read and maintain?

mardelnet (commented):
I think the mobile navigation should not use rounded borders on focus:

Screenshot-from-2025-08-27-11-11-12

Osong-Michael:
> I think the mobile navigation should not use rounded borders on focus:
> 
> Screenshot-from-2025-08-27-11-11-12

Already had a discussion with Houssam about the focus for this, and he said its ok

assets/src/blocks/SecondaryNavigation/SecondaryNavigationBlock.js
```diff
@@ -0,0 +1,39 @@
...
+      },
+      exampleMenuItems: { // Used for the block's preview, which can't extract items from anything.
+        type: 'array',
+      },
+    },
+    isExample: {
+      type: 'boolean',
+      default: false,
+    },
```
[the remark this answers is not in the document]
Osong-Michael: In the case below, when I hover over the Posts List block, I get the preview like this.
Screenshot 2025-08-27 at 12 22 20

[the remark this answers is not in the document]
mardelnet: Cool, thanks!
I only see the default message, so the data from the `example.js` file is not being used.
Also, I commented the properties `attributes.exampleMenuItems` and `isExample` and nothing changed. Are they necessary? 

Screenshot from 2025-08-27 13-39-05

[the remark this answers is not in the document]
Osong-Michael: To be fair I have not really worked much on the preview, just had it there for now while i build the actual block

comzeradd:
So regarding the two issue discussed in the sprint review:

- Page Header presence: if the Design team is fine having a different behavior based on whether the Page Header block is there or not, then all is good for this first iteration.

- Having two headers visible on screen: Regardless of the guidelines we provide, the current behavior seems "broken". Ideally we should find a way to predict those cases somehow. For instance, if there is more than one header visible we could underline the first one and ignore the rest.

assets/src/blocks/SecondaryNavigation/SecondaryNavigationBlock.js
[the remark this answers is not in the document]
Osong-Michael: I removed this completely, according to WP, the data you pass to the block is the same format you need to pass to the example file in order for the preview to work. In our case we use that data just to get back the actual data we display, so the preview always defaults to the error message, so now there is no preview

Osong-Michael:
> The `z-index` of the block has to be increased:
> 
> aaa-Greenpeace(3)

What block is this with the image?

mardelnet (changes requested):
The mobile version does not update the highlighted link when this one changes:

[AwesomeScreenshot-9_3_2025,10_55_14AM.webm](https://github.com/user-attachments/assets/e6a68d7a-6922-4059-a0e0-28d63623122c)

mardelnet (commented):
Should we remove this space to keep the nav items aligned with the rest of the page?

aaa-Greenpeace(4)

Osong-Michael:
> Should we remove this space to keep the nav items aligned with the rest of the page?
> 
> aaa-Greenpeace(4)

It is not aligned with the content of the rest of the page on either side, its meant to be slight inside because we have the fading effect at the edges, you can see that in the menu with the last header element

Osong-Michael:
> The mobile version does not update the highlighted link when this one changes:
> 
>  AwesomeScreenshot-9_3_2025.10_55_14AM.webm

Hey @mardelnet, does this happen only on that browser? Because I don't get the same behaviour on chrome and safari

mardelnet:
> > The mobile version does not update the highlighted link when this one changes:
> > AwesomeScreenshot-9_3_2025.10_55_14AM.webm
> 
> Hey @mardelnet, does this happen only on that browser? Because I don't get the same behaviour on chrome and safari

I don't know, I've only tested it in Firefox

mardelnet:
> > > > The `z-index` of the block has to be increased:
> > 
> > 
> > > What block is this with the image?
> > 
> > 
> > Is the Media and Text block, but the same happens with the page footer and probably with other blocks, as well.
> 
> There is an override at the end of the scss file for the secondary navigation for this implementation, we override the z-index for certain blocks when the Secondary Nav is present reason I asked so I can include them to the override

Sorry, I don't follow your response.
If you test the new block in your local instance within the All Blocks page, you'll see that a number of blocks overlap it.

assets/src/blocks/SecondaryNavigation/SecondaryNavigationFrontend.js
```diff
@@ -0,0 +1,367 @@
...
+  const stickyElement = document.querySelector('.sn-wrapper');
+  const container = document.querySelector('.page-content');
+
+  if (!pageHeader || !stickyElement || !container) {return;}
+
+  // Create spacer to add space between the header element on the page and the navbar so the
+  // header is visble when the page scrolls into view
+  const spacer = document.createElement('div');
+  spacer.style.display = 'none';
+  stickyElement.parentNode.insertBefore(spacer, stickyElement.nextSibling);
```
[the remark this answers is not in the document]
Osong-Michael: So the moment the `.stuck` class is applied the secondary navigation block becomes `position:fixed`, this takes it out of the normal flow of the document so the content below it, which is the `h2` elements jump suddenly to the next available position which was previously occupied by the nav, this creates a weird animation, to prevent this, we use this spacer which is the same height as the nav, to avoid the sudden jump and make the transition smooth on scroll

Osong-Michael:
> > > > > The `z-index` of the block has to be increased:
> > > 
> > > 
> > > > What block is this with the image?
> > > 
> > > 
> > > Is the Media and Text block, but the same happens with the page footer and probably with other blocks, as well.
> > 
> > 
> > There is an override at the end of the scss file for the secondary navigation for this implementation, we override the z-index for certain blocks when the Secondary Nav is present reason I asked so I can include them to the override
> 
> Sorry, I don't follow your response. If you test the new block in your local instance within the All Blocks page, you'll see that a number of blocks overlap it.

The way our SCSS is currently setup, there is a lot of overlaps, an easy way to mitigate this was to override the `z-index` for certain `blocks/elements `only when the Secondary navigation block is present on that same page, it is done [here](https://github.com/greenpeace/planet4-master-theme/blob/secondary-nav-block/assets/src/scss/blocks/SecondaryNavigation/SecondaryNavigationStyle.scss#L313).

assets/src/blocks/SecondaryNavigation/SecondaryNavigationFrontend.js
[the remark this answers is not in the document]
Osong-Michael: To test is out, inside the `onScroll` function, comment out the code that gives the spacer its height, then reload the page. You will notice when scrolling, as soon as the _navbar_ becomes sticky, there is like a jump animation from the next element to kinda account for the space left behind by the block leaving the document flow.

assets/src/blocks/SecondaryNavigation/SecondaryNavigationFrontend.js
```diff
@@ -0,0 +1,367 @@
...
+const {__} = wp.i18n;
+const pageHeader = document.querySelector('.is-pattern-p4-page-header');
+
+/**
+ * Makes the secondary navigation sticky when the page header scrolls past
+ * a defined offset and there's enough space in the content container.
+ *
+ * Adds/removes the 'stuck' class on `.sn-wrapper` based on scroll position.
+ */
+const makeSecondaryNavigationStickyOnScroll = () => {
```
mardelnet, line 13: I think this function can be significantly simplified if we use `position: sticky` instead of switching between the positions absolute, fixed, and none. Would you mind checking this snippet? It could require some adjustments, but it reduces the code and the complexity, and removes the scroll listener.

```js
const makeSecondaryNavigationStickyOnScroll = () => {
  const block = document.querySelector('div[data-render="planet4-blocks/secondary-navigation"]');

  if (!block) {return;}

  document.body.classList.add('secondary-nav-class');
  document.body.classList.add('overflow-visible');
  block.style.position = 'sticky';

  // Get the offset value for different screens and conditions
  const getDynamicOffset = () => {
    const mobile = window.innerWidth <= 991;
    const hasAdminBar = !!document.getElementById('wpadminbar');
    const hasStuckOpen = document.body.classList.contains('stuck-open');

    if (mobile) {
      if (hasStuckOpen) {return 110;}
      if (hasAdminBar) {return 86;}
      return 60;
    }
    if (hasAdminBar) {return 52;}
    return 20;
  };

  const onScroll = () => {
    const offset = getDynamicOffset();
    block.style.top = `${offset}px`;
  };

  onScroll();

  window.addEventListener('resize', onScroll);
};
```

assets/src/blocks/SecondaryNavigation/SecondaryNavigationFrontend.js
```diff
@@ -0,0 +1,367 @@
...
+    // Reset after short delay
+    setTimeout(() => isManualScroll.current = false, 500);
+  };
+
+  const scrollRight = () => {
+    isManualScroll.current = true;
+    navListRef.current.scrollBy({left: isRTL ? -150 : 150, behavior: 'smooth'});
+
+    setTimeout(() => isManualScroll.current = false, 500);
+  };
```
mardelnet, line 306: Can we combine these functions and differentiate them by passing a parameter?
Also, what is the `setTimeout` used for?

assets/src/blocks/SecondaryNavigation/SecondaryNavigationFrontend.js
```diff
@@ -0,0 +1,367 @@
...
+    };
+
+    document.addEventListener('click', handleClickOutside);
+
+    return () => document.removeEventListener('click', handleClickOutside);
+  }, [isDropdownOpen, isMobile]);
+
+  // This is is to ensure if the user scrolls further down on mobile to where the
+  // Footer intercepts the dropdown, it closes the dropdown automatically
+  useEffect(() => {
```
mardelnet, line 247: What about using an `IntersectionObserver` instead of a manual scroll handler that fires on every scroll event and has to do the `getBoundingClientRect()` calculations every time?

assets/src/blocks/SecondaryNavigation/SecondaryNavigationFrontend.js
Osong-Michael: it is used to reset the `isManualScroll` variable to `false` after some time, when `isManaulScroll` is true, the observer is inactive.

assets/src/blocks/SecondaryNavigation/SecondaryNavigationFrontend.js
```diff
@@ -0,0 +1,343 @@
...
+            <p className="current-active-class">{currentHeaderLink}</p>
+            <p className={`dropdown-btn ${isDropdownOpen ? 'active' : ''}`}/>
+          </div>
+          <div
+            className={`block secondary-navigation-block ${isDropdownOpen ? 'show' : ''}`}
+            ref={dropdownRef}
+          >
+            <div className="secondary-navigation-menu container">
+              {showLeftArrow && (
+                <button className="nav-arrow left" onClick={() => scrollNav('left')}/>
```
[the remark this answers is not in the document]
Osong-Michael: We don't use them during tabbing, they are there only visually. Tabbing goes through every element in the header already.

assets/src/blocks/SecondaryNavigation/SecondaryNavigationFrontend.js
```diff
@@ -0,0 +1,343 @@
...
+                        tabIndex={isMobile && !isDropdownOpen ? -1 : 0}
+                      >
+                        {content}
+                      </a>
+                    </li>
+                  ))}
+                </ul>
+              </div>
+              {showRightArrow && (
+                <button className="nav-arrow right" onClick={() => scrollNav('right')}/>
```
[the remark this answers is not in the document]
Osong-Michael: Same as above

assets/src/blocks/SecondaryNavigation/SecondaryNavigationFrontend.js
```diff
@@ -0,0 +1,343 @@
...
+  };
+
+  const onScroll = () => {
+    const offset = getDynamicOffset();
+    block.style.top = `${offset}px`;
+  };
+
+  onScroll();
+
+  window.addEventListener('scroll', onScroll);
```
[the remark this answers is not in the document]
mardelnet: Thanks! In what case is the scroll listener needed?
I have removed it and see everything working alright, but I'm probably missing something.

mardelnet (commented):
The mobile version needs to be scrollable, otherwise some elements cannot be reached when there are too many:

All-Blocks-Page-Greenpeace

PR #2501, opened 2025-01-16
Fix <ol> selector issue

GP-Dan-Tovbein:
### Summary

The carousel indicator numbers are visible and might be hidde and it is happening only in Chrome.

Screen Shot 2025-01-16 at 09 49 18

---

Ref: 

### Testing

1. Go to the assigned [demo page](https://www-dev.greenpeace.org/test-phoebe/fix-covers-block-layout-issue/)
3. Check if the carousel indicator numbers are hidden
4. Ideally confirm on Chrome but could be great if there is another browser to test. Besides Chrome, I've also tested on FF and Safari.


sagarsdeshmukh (approved):
Nice work @GP-Dan-Tovbein , This fix is working for me 👍

PR #2502, opened 2025-01-16
PLANET-7625: Move JS scripts away from Twig templates

mardelnet:
### Summary
This PR removes JS scripts from the Twig templates and enqueue them correctly using WordPress methods and Webpack.

**Related PRs:**
- https://github.com/greenpeace/planet4-master-theme/pull/2503
- https://github.com/greenpeace/planet4-master-theme/pull/2509

---

Ref: https://jira.greenpeace.org/browse/PLANET-7625

### Testing

1. Visit a page using a comments form (i.e., [this one](https://www-dev.greenpeace.org/test-nix/press/1111/duis-posuere-5/)) and check if the behavior is consistent with the logic of `assets/src/js/toggle_comment_submit.js`. Check the script is not loading on the pages not using a comments form.
2. Visit a page using share buttons (i.e., [this one](https://www-dev.greenpeace.org/test-nix/press/1111/duis-posuere-5/)) and check if the behavior is consistent with the logic of `assets/src/js/share_buttons.js`. Check the script is not loading on the pages that are not using share buttons.
3. Visit a page using share buttons once a Gravity Form is submitted and check if the behavior is consistent with the logic of `assets/src/js/share_buttons.js`.
4. Visit any page (i.e., [this one](https://www-dev.greenpeace.org/test-nix/)) and check if the behavior is consistent with the logic of `assets/src/js/hubspot_cookie.js`

PR #2503, opened 2025-01-17
PLANET-7625: Move Google Tag Manager scripts away from Twig templates

mardelnet:
### Summary
This PR removes Google Tag Manager scripts from the Twig templates and enqueue them correctly using WordPress methods and Webpack. 

**Related PRs:**
- https://github.com/greenpeace/planet4-master-theme/pull/2502
- https://github.com/greenpeace/planet4-master-theme/pull/2509

---

Ref: https://jira.greenpeace.org/browse/PLANET-7625

### Testing
Check that Google Tag Manager keeps working as expected.

---

### Notes

1. I'm not quite happy with this implementation since it depends on firing the custom action `pass_gtm_data` every time the function `Timber::render` is called. I've tried to hook that last function to fire the custom action without the need to call it every time that `Timber::render` is also called, but I couldn't. Apparently, there is no such a hook. Please, let me know your thoughts and new ideas.


planet-4:

:watch: 2025.01.17 14:18:31

src/Search/SearchPage.php
```diff
@@ -328,6 +328,8 @@ public function render(): void
             add_filter('wp_robots', 'wp_robots_no_robots');
         }
 
+        do_action('enqueue_google_tag_manager_script', $this->context);
+
```
Osong-Michael, line 332: Hey @mardelnet is it possible to define a wrapper function that we can call and pass parameters to and it calls both the `do_action` and `Timber::render` functions?

mardelnet: Yes, but what would be the actual benefit?
I mean, we will end up with 1 line of code (the wrapper function) instead of 2 (the `do_action` and `Timber::render` functions). Maybe I'm missing something else.

assets/src/js/google_tag_manager.js
```diff
@@ -0,0 +1,100 @@
...
+    gtag('set', 'url_passthrough', true);
+    gtag('set', 'ads_data_redaction', capabilities.ad_storage === 'denied');
+    dataLayer.push({event: 'defaultConsent', ...capabilities});
+  }
+
+  dataLayer.push({
+    'pageType' : googleTagManagerData.page_category,
+    'signedIn' : googleTagManagerData.p4_signedin_status,
+    'visitorType' : googleTagManagerData.p4_visitor_type,
+    'userID' : '',
```
Osong-Michael, line 56: Is this value meant to be empty?

src/Search/SearchPage.php
Osong-Michael: In case we ever have to change something, say a script name, we can do that in one place and not have to find every single place it was called at.

But its just a suggestion

assets/src/js/google_tag_manager.js
mardelnet: It seems so. Check line 56 of file `templates/blocks/google_tag_manager.twig`

Osong-Michael (approved):
👍 tested on local with africa and it works as expected

mardelnet:
@comzeradd since this is a sensitive change, how can we check that GTM keeps working as expected before merging this?

comzeradd:
> @comzeradd since this is a sensitive change, how can we check that GTM keeps working as expected before merging this?

I think that Molly should be able to verify the setup from a dev site. So pick one of them to deploy this branch before merging.

mardelnet:
> > @comzeradd since this is a sensitive change, how can we check that GTM keeps working as expected before merging this?
> 
> I think that Molly should be able to verify the setup from a dev site. So pick one of them to deploy this branch before merging.

Thanks, @comzeradd 
Just to be sure, if I want to use the international site, would replacing this in the `development.json` file be sufficient?
```json
{
    "require": {
      "greenpeace/planet4-master-theme": "PLANET-7625_move-gtm-scripts-from-twig-templates"
    }
}
```

comzeradd:
> Just to be sure, if I want to use the international site, would replacing this in the `development.json` file be sufficient?

Composer needs a `dev-` prefix to understand this is a branch, so it should be `dev-PLANET-7625_move-gtm-scripts-from-twig-templates`. But I think @Osong-Michael is already using international for testing, so maybe pick a different site.

PR #2506, opened 2025-01-20
Register footer menus as theme locations

stduerre:
### Summary

Register footer menus so they are available as positions in WordPress admin. 

Currently, the menus in the footer (Footer Primary, Footer Secondary, Social) are only defined by naming convention. That doesn't work for multilingual setups because menu names are unique and menus that work with the master theme can not be created in more than one language at a time. Registering the menu as theme locations  enabled the use of WordPress admin to select what menus to show in the footer, independent of menu names.

If no menu is set for the location, the old naming convention is still in use to ensure backwards compatibility.

### Testing

1.  Make sure existing menus are still shown in the footer after installing this update (backwards compatibility)
2. Create or edit menus and assign them to the appropriate theme locations
3. Make sure the menus display correctly

### DIFF
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -740,7 +740,7 @@ class MasterSite extends TimberSite
             $footer_social_menu = new TimberMenu('footer-social-menu');
             $context['footer_social_menu'] = $footer_social_menu->get_items();
         } else {
-            $context['footer_primary_menu'] = wp_get_nav_menu_items('Footer Social');
+            $context['footer_social_menu'] = wp_get_nav_menu_items('Footer Social');
         }
 
         if (has_nav_menu('footer-primary-menu')) {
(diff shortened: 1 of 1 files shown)


Osong-Michael (approved):
👍 I have tested on a few sites locally and have not had any issues.

src/MasterSite.php
```diff
@@ -732,9 +735,32 @@ function ($item) {
...
-        $context['footer_social_menu'] = wp_get_nav_menu_items('Footer Social');
-        $context['footer_primary_menu'] = wp_get_nav_menu_items('Footer Primary');
-        $context['footer_secondary_menu'] = wp_get_nav_menu_items('Footer Secondary');
+
+        if (has_nav_menu('footer-social-menu')) {
+            $footer_social_menu = new TimberMenu('footer-social-menu');
+            $context['footer_social_menu'] = $footer_social_menu->get_items();
+        }
+        else {
+            $context['footer_primary_menu'] = wp_get_nav_menu_items('Footer Social');
```
comzeradd, line 744: That should be `'footer_social_menu'`, right?

comzeradd:
@stduerre can you do some minor syntax fixes to make `phpcs` happy? The `else` statement should be on the same line as the previous closing bracket.

```php
if (condition) {
  do someting
} else {
  do something else
}
```

src/MasterSite.php
stduerre: Yes! Thanks for spotting it.

comzeradd:
Thanks! Merged at 8c4f0dce36d06b2f1881d00f16709280e068f185

PR #2514, opened 2025-01-29
Fix icons not showing up in footer social menu

stduerre:
### Summary

Fixes the empty social media menu after assigning the menu to the theme position.

The recent change that allows the footer menus to be assigned to theme positions creates an issue with the social media menu: Once the menu is assigned to the menu positions, the icons for the social channels are not displayed anymore in the frontend (the menu appears empty). 
It doesn't happen when the menu isn't assigned and just loaded by naming convention.

### Testing

1. Go to Design > Menus in WP Admin
2. Select the Footer Social menu from the dropdown and click the "Select" button next to it$
grafik
3. Scroll down to the Menu Settings section (below the menu items) and set "Position in Theme" to "Footer Social Menu" (if that's set already, leave it as is)
grafik
4. Save the menu
5. Reload the frontend and check the social media menu in the footer

**Expected Result:** You can see the social media menu on the right hand side of the footer. There should be a text "Follow us" and a list of the social media icons as defined in the menu.

Example (but will show different icons depending on the menu):
grafik

**Without this fix:** The menu only shows a text "Follow us", but no icons.

grafik

### DIFF
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -740,7 +740,7 @@ class MasterSite extends TimberSite
             $footer_social_menu = new TimberMenu('footer-social-menu');
             $context['footer_social_menu'] = $footer_social_menu->get_items();
         } else {
-            $context['footer_primary_menu'] = wp_get_nav_menu_items('Footer Social');
+            $context['footer_social_menu'] = wp_get_nav_menu_items('Footer Social');
         }
 
         if (has_nav_menu('footer-primary-menu')) {
@@ -749,8 +749,7 @@ class MasterSite extends TimberSite
         } else {
             $context['footer_primary_menu'] = wp_get_nav_menu_items('Footer Primary');
         }
-
-
+        
         if (has_nav_menu('footer-secondary-menu')) {
             $footer_secondary_menu = new TimberMenu('footer-secondary-menu');
             $context['footer_secondary_menu'] = $footer_secondary_menu->get_items();
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -739,16 +739,14 @@ class MasterSite extends TimberSite
         if (has_nav_menu('footer-social-menu')) {
             $footer_social_menu = new TimberMenu('footer-social-menu');
             $context['footer_social_menu'] = $footer_social_menu->get_items();
-        }
-        else {
+        } else {
             $context['footer_primary_menu'] = wp_get_nav_menu_items('Footer Social');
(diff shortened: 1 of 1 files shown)


comzeradd:
It doesn't work for me locally. Looking at the inspector I can see that instead of having the css class name as a filter for the svg icon there a `#menu-item-50` instead.

image

GP-Dan-Tovbein:
@stduerre is this PR ready for review or still in draft?

stduerre:
@comzeradd that's what it looks like before this PR and what it's trying to fix. You either don't have this PR applied, the additional class is in the menu item settings in WP Admin or it doesn't work.

Can you double check 
- if your  `src/MasterSite.php` file contains the changes [in this commit](https://github.com/greenpeace/planet4-master-theme/pull/2514/commits/2808633a50356e225b677ff87c569f710b44dee6)? 
- If your menu item setting contains only the class `facebook`?
Bildschirmfoto vom 2025-02-03 11-05-19


I ask because this PR uses `wp_get_nav_menu_items()` instead of `TimberMenu::get_items()`. 
 `TimberMenu::get_items()`  returns a  Timber\MenuItem object that has multiple classes (including `#menu-item-50`), which causes the original issue. But `wp_get_nav_menu_items()` returns the WP_Post object that only has the classes entered in the menu in Wordpress admin, which should be only `facebook`. So unless you also have these additional classes set in the menu items, I can't explain how they still appear.

I ran another test to try again:

1. Installed a completely new dev environment (current version without this PR, plain version, not an NRO install)
**--> Social icons are displayed correctly**
3. Went to Appearance > Menus > Manage Locations and assigned the Footer menu to the Footer position
**--> Icons are gone and the code looks like in your screenshot**

4. Checked out this PR (`gh pr checkout 2514`)
**--> Icons show up and the html looks like in this screenshot.** 
Bildschirmfoto vom 2025-02-03 09-47-11

This is a classic "it works on my machine" situation, but I can't comprehend how you're getting a different result than me except if the patch wasn't applied properly. Or if that additional class somehow ended up in the menu item settings, but it's not there by default.

@GP-Dan-Tovbein it should be ready, pending @comzeradd's issue that it doesn't work for him.

comzeradd:
Thanks for detailed info. I was either on a wrong branch or something wrong with my local twig cache. It works perfectly fine for me too.

Merged at b1eec196a9fc1347487778e9d2889e0c89e951ba

PR #2515, opened 2025-01-29
PLANET-7681 Make existing social share buttons configurable

mleray:
### Description

See [PLANET-7681](https://jira.greenpeace.org/browse/PLANET-7681)

**Requirements**

- [x] A new setting under Planet 4 > Social
  - Label: "Choose social sharing options"
  - This should be a multiple selection option field (checkboxes), showing the 4 sharing options we already have (WhatsApp, Facebook, X, Email).
- [x] We should create a migration to have those options enabled by default
- [x] Settings should also apply on the Gravity Forms Share button settings
  - Remove the current customization menu
  - Native share should still be displayed in mobile
  - We can keep the text fields for custom share url/text
- [x] Adjust the social sharing template to dynamically include the selected options

### Testing

- To test the migration on local, you can run `npx wp-env run cli wp p4-run-activator` and the new settings should all be checked by default. You can run `npm run db:reset` beforehand if necessary.
- To test the new settings, either on local or on the test instance in `Planet 4 > Social` you can try (un)checking the various social media sharing options. Then you can make sure that they work as expected in any post, or Gravity Forms confirmation that has share buttons. 
- Please also make sure that the social share platforms checkboxes have been removed from the Gravity Forms confirmation settings.

### DIFF
--- a/src/Migrations/M039EnableNewSocialSharePlatforms.php
+++ b/src/Migrations/M039EnableNewSocialSharePlatforms.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace P4\MasterTheme\Migrations;
+
+use P4\MasterTheme\MigrationRecord;
+use P4\MasterTheme\MigrationScript;
+use P4\MasterTheme\Settings;
+
+/**
+ * Enable all social share options via the new P4 setting.
+ */
+class M039EnableNewSocialSharePlatforms extends MigrationScript
+{
+    /**
+     * Perform the actual migration.
+     *
+     * @param MigrationRecord $record Information on the execution, can be used to add logs.
+     * phpcs:disable SlevomatCodingStandard.Functions.UnusedParameter -- interface implementation
+     */
+    protected static function execute(MigrationRecord $record): void
+    {
+        $options = get_option('planet4_options');
+        $options['social_share_options'] = array_keys(Settings::SOCIAL_SHARE_OPTIONS);
+        update_option('planet4_options', $options);
+    }
+    // phpcs:enable SlevomatCodingStandard.Functions.UnusedParameter
+}
--- a/src/Post.php
+++ b/src/Post.php
@@ -335,18 +335,16 @@ class Post extends TimberPost
 
     /**
      * Get values for share buttons content.
-     *
-     * @return string[]
      */
     public function social_share_platforms(): array
     {
         $social_share_options = planet4_get_option('social_share_options', []);
 
         return [
-            'facebook' => $social_share_options['facebook'] ?? false,
-            'twitter' => $social_share_options['twitter'] ?? false,
-            'whatsapp' => $social_share_options['whatsapp'] ?? false,
-            'email' => $social_share_options['email'] ?? false,
+            'facebook' => in_array('facebook', $social_share_options),
+            'twitter' => in_array('twitter', $social_share_options),
+            'whatsapp' => in_array('whatsapp', $social_share_options),
+            'email' => in_array('email', $social_share_options),
             // We might add a setting for this one in the future, but for now we always disable it in Posts.
             'native' => false,
         ];
--- a/src/Post.php
+++ b/src/Post.php
@@ -334,6 +334,25 @@ class Post extends TimberPost
     }
 
     /**
+     * Get values for share buttons content.
+     *
+     * @return string[]
+     */
+    public function social_share_platforms(): array
+    {
+        $social_share_options = planet4_get_option('social_share_options', []);
+
+        return [
+            'facebook' => $social_share_options['facebook'] ?? false,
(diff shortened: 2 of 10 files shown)


planet-4:

:watch: 2025.01.29 15:01:56

src/Settings.php
```diff
@@ -682,6 +698,31 @@ public function p4_render_category_dropdown(CMB2_Field $field_args, $value): voi
...
+
+            printf(
+                '<label><input type="checkbox" id="%1$s" name="%2$s" value="%3$s" %4$s />%5$s</label>',
+                $field_args->id(),
+                $field_args->id() . '[' . $key . ']',
+                $key,
+                $selected ? 'checked' : '',
+                $label,
+            );
+            echo '<br/><br/>';
```
mardelnet, line 721: Can we replace the `br` tags with a CSS class or at least an inline style?

src/Settings.php
```diff
@@ -682,6 +698,31 @@ public function p4_render_category_dropdown(CMB2_Field $field_args, $value): voi
     }
     // phpcs:enable SlevomatCodingStandard.Functions.UnusedParameter
 
+    /**
+     * Render category dropdown.
```
mardelnet, line 702: This description is not accurate.

src/Migrations/M039EnableNewSocialSharePlatforms.php
```diff
@@ -0,0 +1,27 @@
...
+ */
+class M039EnableNewSocialSharePlatforms extends MigrationScript
+{
+    /**
+     * Perform the actual migration.
+     *
+     * @param MigrationRecord $record Information on the execution, can be used to add logs.
+     * phpcs:disable SlevomatCodingStandard.Functions.UnusedParameter -- interface implementation
+     */
+    protected static function execute(MigrationRecord $record): void
```
mardelnet, line 20: After resetting the database and running the migration script, all the social checkboxes are marked but they're not visible in the front. I have to manually press the Save button on the Planet 4 > Social page to make it work. Do you see the same?

src/Settings.php
```diff
@@ -397,6 +406,12 @@ public function __construct()
                             'type' => 'text',
                         ],
                     ],
+                    [
+                        'name' => __('Choose social sharing options', 'planet4-master-theme-backend'),
+                        'id' => 'social_share_options',
+                        'type' => 'social_share_checkboxes',
```
mardelnet, line 412: Can you explain how can `social_share_checkboxes` be a `type`?

src/Settings.php
```diff
@@ -682,6 +698,32 @@ public function p4_render_category_dropdown(CMB2_Field $field_args, $value): voi
     }
     // phpcs:enable SlevomatCodingStandard.Functions.UnusedParameter
 
+    /**
+     * Render social share checkboxes.
+     *
+     * @param CMB2_Field $field_args Field arguments.
+     * @param array $value Value.
+     */
+    public function p4_render_social_share_checkboxes(CMB2_Field $field_args, array $value): void
```
mardelnet, line 707: This is super picky and optional: can we align the title with the boxes?

image

src/Post.php
```diff
@@ -333,6 +333,25 @@ public function share_meta(): array
         ];
     }
 
+    /**
+     * Get values for share buttons content.
+     *
+     * @return string[]
```
mardelnet, line 339:
```suggestion
     * @return array<string, string|bool>
```

src/Settings.php
mleray: This is similar to what we do for page dropdowns in other settings, we register this new type [here](https://github.com/greenpeace/planet4-master-theme/pull/2515/files/42e70bca98f8a455702b8aff3b68f12afa227c27#diff-eb1710f64250974d6d1550d421a31054e6079592ae3a8428cd9530bc086bdd94R595) with the render function

src/Post.php
mleray: Can you please explain this? We use `string[]` in a lot of places in the code

mardelnet: The function returns an associative array where the values are booleans, not strings.

mleray: Right.. I've replaced it with just `array`, is that good enough?

mleray: Well, actually, that can be removed because it's already in the function declaration

src/Migrations/M039EnableNewSocialSharePlatforms.php
mardelnet: Out of curiosity: how did you fix it, @mleray?

mleray: I'm now using `in_array` to check if the setting is checked! Like I do in the settings themselves actually, not sure why I didn't do that in the first place

mardelnet (approved):
Well done, @mleray

PR #2523, opened 2025-02-06
 PLANET-7645: Implement image replacement feature

mardelnet:
### Summary

This PR implements a new feature to replace images in Google Cloud Storage through the WP Stateless plugin. Also, the PR creates the image thumbnails using PHP native functions since, for unknown reasons, the ones created by WordPress are not uploaded to Google Cloud Storage. Even if that happened, the new thumbnails might differ in size from the originals, so the file names and their URLs would differ, making the replacement feature unusable.

---

Ref: https://jira.greenpeace.org/browse/PLANET-7645

### Testing

**Testing note:** for local testing, go to [the Gutenberg instance](https://www-dev.greenpeace.org/gutenberg/wp-admin/) and copy the WP Stateless configuration from Media > Stateless settings.

1. Go to the Media Library and upload a new image.
2. Go to Google Cloud Storage ([Gutenberg link](https://console.cloud.google.com/storage/browser/_details/planet4-gutenberg-stateless-develop/), [Janus link](https://console.cloud.google.com/storage/browser/planet4-test-janus-stateless-develop)) and check the new image and all the thumbnails were uploaded. Also, check the images' metadata.
3. In the Media Library, select the uploaded image and click on the "Replace Media" button.
4. Pick a new image to replace it.
5. Wait for the process to end and check the admin notice for success or failure.
6. If the image was replaced, go to Google Cloud Storage and check it. All the thumbnails should be replaced too, keeping the same size and file name.
7. Check the metadata of the replaced file. It should be the same as step 2.
8. Repeat steps 1 to 6 using another image file type (`jpg`, `png`, or `gif`).
9. Repeat steps 1 to 6 using a non-image file type (document, video, audio) to be sure that the previous implementation still works correctly.

---

### Notes

1. The Cloudflare purger was temporarily removed as it was failing. A new PR will be opened to fix it.

### DIFF
--- a/src/MediaReplacer.php
+++ b/src/MediaReplacer.php
@@ -9,7 +9,8 @@ use WP_Post;
  *
  * This class is used to replace media elements in WordPress.
  */
-class MediaReplacer {
+class MediaReplacer
+{
     private array $replacement_status;
     private array $user_messages;
 
--- a/src/MediaReplacer.php
+++ b/src/MediaReplacer.php
@@ -224,7 +224,6 @@ class MediaReplacer
             // If the file was not uploaded, abort
             if (!$movefile) {
                 throw new \Exception($this->user_messages['media']);
-                return;
             }
 
             // Get the old file path
@@ -241,7 +240,6 @@ class MediaReplacer
             // If the file was not renamed, abort
             if (!$file_renamed) {
                 throw new \Exception($this->user_messages['media']);
-                return;
             }
 
             // Update the attachment metadata with new information
@@ -260,7 +258,6 @@ class MediaReplacer
             // If the post was not updated, abort
             if (is_wp_error($post_updated) || $post_updated === 0) {
                 throw new \Exception($this->user_messages['media']);
-                return;
             }
 
             // Update file metadata
@@ -286,7 +283,6 @@ class MediaReplacer
             // If it's not loaded, abort.
             if (!extension_loaded('gd')) {
                 throw new \Exception($this->user_messages['gd']);
-                return;
             }
 
             $new_image_path = $file['tmp_name'];
@@ -298,7 +294,6 @@ class MediaReplacer
             // Validate image type against allowed MIME types
             if (!isset(self::IMAGE_MIME_TYPES[$new_image_type])) {
                 throw new \Exception($this->user_messages['image']);
-                return;
             }
 
             // Load the image dynamically
@@ -306,14 +301,12 @@ class MediaReplacer
             $image = call_user_func($image_data['create'], $new_image_path);
             if (!$image) {
                 throw new \Exception($this->user_messages['image']);
-                return;
             }
 
             // Get the image metadata.
             $old_image_meta = get_post_meta($id, 'sm_cloud')[0];
             if (!$old_image_meta) {
                 throw new \Exception($this->user_messages['image']);
-                return;
             }
 
             $old_image_dirname = pathinfo($old_image_meta['name'], PATHINFO_DIRNAME);
@@ -322,17 +315,16 @@ class MediaReplacer
 
             if (!function_exists('ud_get_stateless_media')) {
                 throw new \Exception($this->user_messages['image']);
-                return;
             }
 
             // Create metadata for uploading the main image.
             $metadata = [
                 'width' => $new_image_width,
                 'height' => $new_image_height,
-                'file-hash' => md5($old_image_filename),
                 'size' => '__full',
                 'object-id' => $id,
-                'source-id' => md5($file . ud_get_stateless_media()->get('sm.bucket')),
+                'source-id' => md5($file . ud_get_stateless_media()->get('sm.bucket')), //NOSONAR
+                'file-hash' => md5($old_image_filename), //NOSONAR
             ];
 
             // Replace the main image.
@@ -354,7 +346,7 @@ class MediaReplacer
                 $metadata = [
                     'width' => $new_image_width,
                     'height' => $new_image_height,
-                    'file-hash' => md5($image_name . '-' . $old_image_width . 'x' . $old_image_height),
+                    'file-hash' => md5($image_name . '-' . $old_image_width . 'x' . $old_image_height),  //NOSONAR
                     'size' => $size,
                     'child-of' => $id,
                 ];
@@ -472,7 +464,6 @@ class MediaReplacer
 
             if (!function_exists('ud_get_stateless_media')) {
                 throw new \Exception($this->user_messages['image']);
-                return;
             }
 
             // Upload the image (main or variant)
@@ -483,7 +474,6 @@ class MediaReplacer
                 return;
             }
             throw new \Exception($this->user_messages['image']);
-            return;
         } catch (\Exception $e) {
             $this->error_handler($e->getMessage());
             return;
--- a/src/MediaReplacer.php
+++ b/src/MediaReplacer.php
@@ -113,6 +113,8 @@ class MediaReplacer
 
     /**
      * Adds a button to replace media files in the attachments editor.
+     *
+     * @param WP_Post $post The attachment post object.
      */
     public function render_replace_media_metabox(WP_Post $post): void
     {
@@ -121,7 +123,11 @@ class MediaReplacer
     }
 
     /**
-     * Adds a button to replace media files to the Attachment details modal.
+     * Adds a button to replace media files in the Attachment details modal.
+     *
+     * @param array $form_fields Existing attachment form fields.
+     * @param WP_Post $post The attachment post object.
+     * @return array Modified attachment form fields.
      */
     public function add_replace_media_button(array $form_fields, WP_Post $post): array
     {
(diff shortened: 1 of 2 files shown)


planet-4:

:watch: 2025.02.06 11:43:37

src/MediaReplacer.php
```diff
@@ -249,84 +247,298 @@ private function replace_media_file(int $old_file_id, string $new_file_path): bo
...
+            'file',
+            $this->user_messages['success'],
+            $this->user_messages['error'],
+        );
     }
 
     /**
      * Renders admin notices.
-     *
-     * @param string $transient_key The Transient key.
```
GP-Dan-Tovbein, line 316: Why did we remove this?

src/MediaReplacer.php
```diff
@@ -249,84 +247,298 @@ private function replace_media_file(int $old_file_id, string $new_file_path): bo
...
-            set_transient('cloudflare_purge_error', $e->getMessage(), 5);
+            $this->error_handler($e->getMessage());
             return;
         }
     }
 
     /**
-     * Displays admin notices for media replacement messages.
+     * Handles errors in replacements.
+     * Sets error messages.
```
GP-Dan-Tovbein, line 476: I think it would be nice if we add `@param` and `@return` in most of cases and then we have a clear PHPDoc.

src/MediaReplacer.php
```diff
@@ -249,84 +247,298 @@ private function replace_media_file(int $old_file_id, string $new_file_path): bo
...
+
+                // Free memory
+                imagedestroy($thumb);
+            }
+            // Free memory
+            imagedestroy($image);
+
+            // Send JSON success data.
+            wp_send_json_success();
+            return;
```
GP-Dan-Tovbein, line 372: Maybe I'm missing something else, but do we really need to `return` void here?

src/MediaReplacer.php
mardelnet: PHPDoc is not mandatory in our codebase; we only have to define the parameters type and function returns as part of the functions themselves.

src/MediaReplacer.php
mardelnet: Alright, I'll re-add them later on.

mardelnet: Done!

GP-Dan-Tovbein:
@mardelnet I've just tested this feature and seems to working great the replacement on GCS but see how it works on Media Library

https://drive.google.com/file/d/1viuNxGW2sNQ6S4xDarnGikOmCJycu_dy/view?usp=sharing

src/MediaReplacer.php
```diff
@@ -11,25 +11,70 @@
...
+
     private const IMAGE_MIME_TYPES = [
-        'image/jpeg',
-        'image/png',
-        'image/gif',
-        'image/bmp',
-        'image/webp',
-        'image/tiff',
-        'image/svg+xml',
+        IMAGETYPE_JPEG => [
```
GP-Dan-Tovbein, line 18: This is nothing relevant but I think `IMAGE_TYPE_***` is more cleaner. What do you think?

mardelnet:
> @mardelnet I've just tested this feature and seems to working great the replacement on GCS but see how it works on Media Library

Thanks @GP-Dan-Tovbein! Just to be clear: you see the image replaced in GCS but not in the Media Library? If so, that's probably a cache issue. Try to copy the image URL in the Media Library and open it in a different tab adding some extra parameters to the URL and see if that works. That happens because I've to complete the Cloudflare purging as I mentioned in the Notes of the PR description.

GP-Dan-Tovbein:
> > @mardelnet I've just tested this feature and seems to working great the replacement on GCS but see how it works on Media Library
> 
> Thanks @GP-Dan-Tovbein! Just to be clear: you see the image replaced in GCS but not in the Media Library? If so, that's probably a cache issue. Try to copy the image URL in the Media Library and open it in a different tab adding some extra parameters to the URL and see if that works. That happens because I've to complete the Cloudflare purging as I mentioned in the Notes of the PR description.

See from 1:40 to 2:10. The thumbnail it's different than the one is shown through the modal. Please confirm that that is because of Cloudflare 🙏

mardelnet:
> See from 1:40 to 2:10. The thumbnail it's different than the one is shown through the modal. Please confirm that that is because of Cloudflare 🙏

I confirm it was a cache issue, @GP-Dan-Tovbein. Now I can see that the thumbnail and the main image are the same: [AwesomeScreenshot-2_17_2025,1_56_23PM.webm](https://github.com/user-attachments/assets/7c13d6c5-0a80-4c63-8975-afe4c447f2d6)

GP-Dan-Tovbein:
I've tested deeply the metadata it's pushed and I also see a difference between the same file that were replaced. Can you confirm it please?

Testing instructions:
1. Upload a new image and do the same process you request
2. From GCC, go to that picture and download any
3. Upload to https://www.metadata2go.com/ and retrieve its metadata
4. Replace the uploaded image in Media Library with the same it's upladed to GCC
5. Repeat item 3

The author is misising for example.

Here is a comparisson between the attached picture.

underwater-view-768x329

 
[2025_02_c3144146-underwater-view-768x329-1__2.json-metadata.pdf](https://github.com/user-attachments/files/18826225/2025_02_c3144146-underwater-view-768x329-1__2.json-metadata.pdf)
[2025_02_c3144146-underwater-view-768x329-1__1.json-metadata.pdf](https://github.com/user-attachments/files/18826227/2025_02_c3144146-underwater-view-768x329-1__1.json-metadata.pdf)

mardelnet:
> > > I've tested deeply the metadata it's pushed and I also see a difference between the same file that were replaced. Can you confirm it please?
> > > What do you think?
> > 
> > 
> > I think that keeping the same metadata in Google Cloud Storage, and not the whole file metadata, would be enough. However, I'd like to know @comzeradd opinion on this matter.

> But I noticed that the author is missing which is super important that value

Are you sure that the image used for the replacement had a value for the author, @GP-Dan-Tovbein?

GP-Dan-Tovbein:
> > > > I've tested deeply the metadata it's pushed and I also see a difference between the same file that were replaced. Can you confirm it please?
> > > > What do you think?
> > > 
> > > 
> > > I think that keeping the same metadata in Google Cloud Storage, and not the whole file metadata, would be enough. However, I'd like to know @comzeradd opinion on this matter.
> 
> > But I noticed that the author is missing which is super important that value
> 
> Are you sure that the image used for the replacement had a value for the author, @GP-Dan-Tovbein?

**The one that I attached, yes**

Screen Shot 2025-02-17 at 14 29 51

underwater-view-768x329

Using [this tool](https://www.metadata2go.com)

mardelnet:
> > > But I noticed that the author is missing which is super important that value
> > 
> > 
> > Are you sure that the image used for the replacement had a value for the author, @GP-Dan-Tovbein?
> 
> **The one that I attached, yes, Using [this tool](https://www.metadata2go.com)**

I'm sorry @GP-Dan-Tovbein, but I still don't get it. We are only replacing the metadata in Google Storage. Why should we also replace the file metadata? If the second image has a different author or no author at all, we should not replace that information with data from the first image. Otherwise, we could end up with wrong information, such as pictures from an incorrect author.

comzeradd:
I'm also a bit confused. If the metadata of the new image are those of the newly uploaded image then that's fine.

@mardelnet is the Janus instance updated? I'm trying to test the image replacement and though I'm getting a "File was succesfully replaced!" message the image remains the same.

mardelnet:
> I'm also a bit confused. If the metadata of the new image are those of the newly uploaded image then that's fine.
> 
> @mardelnet is the Janus instance updated? I'm trying to test the image replacement and though I'm getting a "File was succesfully replaced!" message the image remains the same.

@comzeradd, yes, it's working, but:
1. Images in WordPress take time to be replaced because of the cache (I temporarily disabled the Cloudflare purger because it wasn't working as expected) but you can compare them in the Google bucket.
2. I just realized that `jpg` images are not replacing `jpeg` images, and vice-versa. I'll fix it.

src/MediaReplacer.php
mardelnet: In this case, we cannot change it. `IMAGETYPE_***` is a [constant in PHP](https://www.php.net/manual/en/image.constants.php#constant.imagetype-jpeg) that represents the image type. It is used by PHP's image handling functions to identify the type of image file when working with image resources

GP-Dan-Tovbein (approved):
Excelent job @mardelnet 👏👏👏 

Even if I approve the PR, please check the missing comment if it requires a change before you merge .

mardelnet:
@GP-Dan-Tovbein @comzeradd 
The problem with `jpeg` images was fixed. You can test it by uploading a `jpeg` file and then replacing it with a `jpg` one.
I still cannot purge Cloudflare after the replacement, making the replaced image still be shown in the Media Library even when the replacement succeded. @comzeradd, if you don't mind I'd rather fix that on a different PR.

src/MediaReplacer.php
mardelnet: Removed!

PR #2529, opened 2025-02-12
Replace special characters in the content of the posts

mardelnet:
### Summary

This PR replaces the special characters `u003c`, `u003e`,  and `u0022` in the content of the posts that were affected by previous migrations, for example [here](https://www.greenpeace.org/sweden/klimat/#vad-du-kan-gora).

---

### Testing

1. Run this PR in your local environment.
2. Add the following text (it contains wrong markups) to a new post or page (or edit an existing one): `This should be replaced: u003cau003eu0022testu0022u003c/au003e, but this should NOT: \u003ca\u003e\u0022test\u0022\u003c/a\u003e` 
4. Run the migrator command: `npx wp-env run cli wp p4-run-activator`
5. Check whether the migration worked as expected and the special characters were fixed.
6. To reset the environment database back to the latest default content and repeat the tests, run: `npm run db:reset`

### DIFF
--- a/src/Migrations/M040ReplaceSpecialCharactersInPostsContent.php
+++ b/src/Migrations/M040ReplaceSpecialCharactersInPostsContent.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace P4\MasterTheme\Migrations;
+
+use P4\MasterTheme\MigrationRecord;
+use P4\MasterTheme\MigrationScript;
+
+/**
+ * Replace special characters ("u003c", "u003e", "u0022") in the posts content.
+ */
+class M040ReplaceSpecialCharactersInPostsContent extends MigrationScript
+{
+    /**
+     * Perform the actual migration.
+     *
+     * @param MigrationRecord $record Information on the execution, can be used to add logs.
+     * phpcs:disable SlevomatCodingStandard.Functions.UnusedParameter -- interface implementation
+     */
+    public static function execute(MigrationRecord $record): void
+    {
+        echo 'Replacing special characters in posts content...', "\n"; // phpcs:ignore
+
+        $args = [
+            'post_type' => get_post_types(['public' => true], 'names'),
+            'post_status' => 'any',
+            'posts_per_page' => -1,
+        ];
+
+        $results = get_posts($args) ?? [];
+
+        foreach ((array) $results as $post) {
+            if (empty($post->post_content)) {
+                continue;
+            }
+
+            $current_post_id = $post->ID;
+
+            echo 'Parsing post ', $current_post_id, "\n"; // phpcs:ignore
+
+            $post_content = $post->post_content;
+
+            // Only replace standalone "u003c", "u003e", "u0022" (not inside quotes or escape sequences)
+            $post_content = preg_replace('/(?<!\\\\)u003c/', '<', $post_content);
+            $post_content = preg_replace('/(?<!\\\\)u003e/', '>', $post_content);
+            $post_content = preg_replace('/(?<!\\\\)u0022/', '"', $post_content);
+
+            $post_args = [
+                'ID' => $current_post_id,
+                'post_content' => $post_content,
+            ];
+
+            wp_update_post(wp_slash($post_args));
+        }
+    }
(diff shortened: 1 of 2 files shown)


planet-4:

:watch: 2025.02.12 09:02:40

src/Migrations/M040ReplaceSpecialCharactersInPostsContent.php
```diff
@@ -0,0 +1,56 @@
...
+            $post_content = preg_replace('/(?<!\\\\)u003c/', '<', $post_content);
+            $post_content = preg_replace('/(?<!\\\\)u003e/', '>', $post_content);
+            $post_content = preg_replace('/(?<!\\\\)u0022/', '"', $post_content);
+
+            $post_args = [
+                'ID' => $current_post_id,
+                'post_content' => $post_content,
+            ];
+
+            wp_update_post(wp_slash($post_args));
```
sagarsdeshmukh, line 52: The wp_slash() function works with array values as well, but can't we use it like this instead: `'post_content' => wp_slash($post_content)` 🤔

src/Migrations/M040ReplaceSpecialCharactersInPostsContent.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        echo 'Replacing special characters in posts content...', "\n"; // phpcs:ignore
+
+        $args = [
+            'post_type' => get_post_types(['public' => true], 'names'),
+            'post_status' => 'any',
+            'posts_per_page' => -1,
+        ];
+
+        $results = get_posts($args) ?? [];
```
sagarsdeshmukh, line 29: I am concerned with this operation as we are loading huge data almost the whole table in memory, which will cause performance issues. can we try WP_Query with pagination instead of get_posts()? and process records in batches

mardelnet: What about the post type `attachment`?

mardelnet: Done, @sagarsdeshmukh

src/Migrations/M040ReplaceSpecialCharactersInPostsContent.php
```diff
@@ -0,0 +1,69 @@
...
+
+                $updated_content = preg_replace('/(?<!\\\\)u003c/', '<', $content);
+                $updated_content = preg_replace('/(?<!\\\\)u003e/', '>', $updated_content);
+                $updated_content = preg_replace('/(?<!\\\\)u0022/', '"', $updated_content);
+
+                if ($updated_content === $content) {
+                    continue;
+                }
+
+                wp_update_post([
```
sagarsdeshmukh, line 58: Can we print the update status, just in case if we want to check which posts are updated in migration?

mardelnet:
@comzeradd 
@sagarsdeshmukh found a case where the migration script fails: https://www-dev.greenpeace.org/international/story/49721/story-italy-single-use-plastic-bottle-toxic-fossil-fuels/
After running the script, the gallery block disappears. That's because the `data-attributes` have to be wrapped within `&quot;` instead of `"`. However, if the script is changed to replace the `u0022` with `&quot;` some other elements break. 
Any thoughts on how to fix this?

sagarsdeshmukh:
While testing the migration script with the international site DB dump on local, I found that only 12 records were updated out of the 12.5k dataset.

We can log the updated post IDs in the migration script, which would be helpful for tracking updated records during migration.

Additionally, I suspect the gallery block issue occurs with an older version of the gallery block. For example, [this post](https://www-dev.greenpeace.org/international/story/49721/story-italy-single-use-plastic-bottle-toxic-fossil-fuels/) was last updated three years ago.

sagarsdeshmukh:
Great work on this PR, @mardelnet! 🎉  The solution looks solid. I tested the migration locally, and it works well. Other than the gallery block issue, everything looks good from my side. 👍

sagarsdeshmukh (approved):
Well done @mardelnet 👍

PR #2532, opened 2025-02-13
PLANET-7706 Keep News & Stories page layout when reloading/filtering

mleray:
### Description

See [PLANET-7706](https://jira.greenpeace.org/browse/PLANET-7706)

### Testing

You can check this new behaviour (with pagination) on the titan instance:
- in a normal listing page ([example](https://www-dev.greenpeace.org/test-titan/press/))
- in the [News & Stories page](https://www-dev.greenpeace.org/test-titan/news-stories/) (for this one make sure the filtering also still works as expected)

### DIFF
--- a/assets/src/js/listing_pages.js
+++ b/assets/src/js/listing_pages.js
@@ -1,25 +1,34 @@
 const {__} = wp.i18n;
 
 export const setupListingPages = () => {
+  const listingPageContent = document.getElementById('listing-page-content');
+  // If the current page is not a listing page, we do nothing.
+  if (!listingPageContent) {
+    return;
+  }
+
   // Setup behaviour for list/grid toggle.
   const listViewToggle = document.querySelector('.list-view-toggle');
   const gridViewToggle = document.querySelector('.grid-view-toggle');
 
-  const listingPageContent = document.getElementById('listing-page-content');
-
-  if (!listingPageContent || !listViewToggle || !gridViewToggle) {
+  if (!listViewToggle && !gridViewToggle) {
     return;
   }
 
-  const switchViews = () => {
-    listingPageContent.classList.toggle('wp-block-query--list');
-    listingPageContent.classList.toggle('wp-block-query--grid');
-    gridViewToggle.classList.toggle('d-none');
-    listViewToggle.classList.toggle('d-none');
+  listingPageContent.classList.toggle('wp-block-query--list', gridViewToggle);
+  listingPageContent.classList.toggle('wp-block-query--grid', listViewToggle);
+
+  const switchViews = layout => {
+    const newUrl = new URL(window.location.href);
+    newUrl.searchParams.set('layout', layout);
+    window.location.href = newUrl.href;
   };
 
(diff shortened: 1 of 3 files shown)


planet-4:

:watch: 2025.02.13 10:58:08

templates/listing-page.twig
```diff
@@ -6,12 +6,15 @@
         {% else %}
             <h2>{{ listing_page_title ?? __( 'All articles', 'planet4-master-theme' ) }}</h2>
         {% endif %}
-        <span class="grid-view-toggle" title="{{__( 'Grid layout', 'planet4-master-theme' ) }}">
-            {{ 'grid-view'|svgicon }}
-        </span>
-        <span class="list-view-toggle d-none" title="{{__( 'List layout', 'planet4-master-theme' ) }}">
-            {{ 'list-view'|svgicon }}
-        </span>
+        {% if layout == 'list' %}
```
mardelnet, line 9: Should we use a strict comparison `===`?

src/ListingPage.php
```diff
@@ -47,6 +47,9 @@ private function update_context(): void
         $this->context['page_category'] = is_home() ? 'News' : 'Listing Page';
         $this->context['og_type'] = isset($this->context['author']) ? 'profile' : 'website';
 
+        // Set layout (grid or list)
+        $this->set_layout();
```
mardelnet, line 51: Does it make sense to create a single-line function that is not reused, either?
Maybe we can just go with:
```suggestion
        $this->context['layout'] = $_GET['layout'] ?? 'list';
```

assets/src/js/listing_pages.js
```diff
@@ -1,25 +1,34 @@
...
+  }
+
   // Setup behaviour for list/grid toggle.
   const listViewToggle = document.querySelector('.list-view-toggle');
   const gridViewToggle = document.querySelector('.grid-view-toggle');
 
-  const listingPageContent = document.getElementById('listing-page-content');
-
-  if (!listingPageContent || !listViewToggle || !gridViewToggle) {
+  if (!listViewToggle && !gridViewToggle) {
```
mardelnet, line 14: I think this should be an `OR` instead of `AND`
We need both elements on lines 18 and 19.

mardelnet (commented):
Great job, @mleray

assets/src/js/listing_pages.js
mleray: It's an `AND` because in the Twig template we only show one or the other based on the layout. On lines 18 and 19, it still works fine when one is not there because we need a boolean value.

mleray: That's also why I had to use an `if ... else` statement on lines 27 to 31

src/ListingPage.php
mleray: Updated!

templates/listing-page.twig
mardelnet: Thanks, I didn't know that!

mardelnet (approved):

PR #2533, opened 2025-02-14
PLANET-7701: Fix main nav menu accessibility issues

mardelnet:
### Summary
This PR enhances the experience of the main navigation menu by:
- Displaying the Navigation Menu drop-downs with keyboard navigation ([see reference](https://www.apple.com/)).
- Adding a focus ring to the reset button of the search field.

---

Ref: https://jira.greenpeace.org/browse/PLANET-7701

### Testing
1. Navigate the main nav menu by tabbing its elements. It should mimic the behavior of [this menu](https://www.apple.com/). 
2. Check the reset button of the search fields has a visible focus state.

PR #2543, opened 2025-02-24
PLANET-7703 Fix country selector close button focus ring

mleray:
### Description

See [PLANET-7703](https://jira.greenpeace.org/browse/PLANET-7703)

This commit also adds a visually hidden span for better accessibility.

### Testing

Either on local or on the [titan instance](https://www-dev.greenpeace.org/test-titan/), make sure that the close button now has a visible focus state when navigating with tabs. Note that on local you will need to clear the Timber cache (`npx wp-env run cli wp timber clear_cache`) to test this.

### DIFF
--- a/templates/footer_country_selector.twig
+++ b/templates/footer_country_selector.twig
@@ -20,9 +20,7 @@
         data-bs-target="#country-selector"
         aria-expanded="false"
         aria-label="{{ __( 'Close country selector', 'planet4-master-theme' ) }}"
-    >
-        &#10005;
-    </button>
+    ></button>
     <ul class="countries" aria-label="{{ __('Worldwide site list', 'planet4-master-theme') }}">
         <li class="international"><a href="{{ international.url }}">{{ international.name }}</a></li>
         {% for initial, countries in countries_by_initials[1:] %}
--- a/templates/footer_country_selector.twig
+++ b/templates/footer_country_selector.twig
@@ -19,7 +19,9 @@
         data-bs-toggle="open"
         data-bs-target="#country-selector"
         aria-expanded="false"
-        aria-label="{{ __( 'Close country selector', 'planet4-master-theme' ) }}">
+        aria-label="{{ __( 'Close country selector', 'planet4-master-theme' ) }}"
+    >
+        &#10005;
     </button>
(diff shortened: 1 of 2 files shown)


planet-4:

:watch: 2025.02.24 10:55:02

templates/footer_country_selector.twig
```diff
@@ -19,7 +19,12 @@
         data-bs-toggle="open"
         data-bs-target="#country-selector"
         aria-expanded="false"
-        aria-label="{{ __( 'Close country selector', 'planet4-master-theme' ) }}">
+        aria-label="{{ __( 'Close country selector', 'planet4-master-theme' ) }}"
+    >
+        &#10005;
```
mardelnet, line 24: What about using this image instead?
https://github.com/greenpeace/planet4-master-theme/blob/main/images/cross.svg

templates/footer_country_selector.twig
```diff
@@ -19,7 +19,12 @@
...
+        <span class="visually-hidden">
```
mardelnet, line 25: Can you please explain how this improves the accessibility?

mleray: I added it as a label for screen readers, but I now realise that the button already has a `aria-label` so actually it's probably not needed 🤔

templates/footer_country_selector.twig
mleray: This is what we used as a background mask before, but then the focus state wouldn't show. Do you think it would be better to use a svg rather than this code?

mardelnet: We are replacing the mask with an SVG image [here](https://github.com/greenpeace/planet4-master-theme/pull/2533/files), too. Let's decide whether the code or the image is better and use the same approach in both PRs for consistency.

mleray: I'm not sure whether or not the focus will show for an empty button with just a background, but I'll try it out! It would be best to be consistent indeed

mardelnet: Maybe you'll need to set the height and width for the empty button.

mleray: It was set already and not working with the mask 🤔

mleray: Updated!

mardelnet (approved):
Well done, @mleray

PR #2544, opened 2025-02-26
PLANET-7775 Add same favicon but with better resolution

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-7775

### Summary

Same icon with transparency, just much better resolution

### Testing

1. It's already shown in the test instance


mleray:
It also seems to be a slightly lighter color, is that expected?

Screenshot 2025-02-26 at 10 38 03

comzeradd:
> It also seems to be a slightly lighter color, is that expected?

Yeah, I think it's fine because this actually the same as the full logo we use

mleray (approved):
👍

PR #2540, opened 2025-02-19
PLANET-7766: Purge Cloudflare on file replacement

mardelnet:
### Summary

This PR:
- Purges the Cloudflare cache after file replacements.
- Reduces the calls to the `ud_get_stateless_media` function.

---

Ref: https://jira.greenpeace.org/browse/PLANET-7766

### Testing

1. Go to the [Media Library](https://www-dev.greenpeace.org/test-nix/wp-admin/upload.php).
2. Upload a new file, preferably an image.
3. Replace it by clicking the Replace Media button.
4. Check that the file was replaced and the user can see and interact with the new file (for instance, if an image was replaced, the new one has to be visible in the back and the front).

PR #2551, opened 2025-03-03
Refresh images after replacing them using the Media Replacer

mardelnet:
### Summary

This PR introduces a mechanism to refresh the images after replacing them using the Media Replacer feature.

---

Ref: https://jira.greenpeace.org/browse/PLANET-7766

### Notes
@comzeradd: I am not quite happy with this solution but it works.
Purging Cloudflare using the methods of the `CloudflarePurger` class showed not to be effective to reflect the changes of the replaced image. Instead of that, this PR adds a query parameter to the image URLs using JavaScript to force the browser to show the updated version of the image. If you're okay with this implementation, I'll ask the reviewers to check the code in depth.

### Testing

1. Replace an image using the Media Replacer feature.
2. Check the image was replaced in Google Cloud Storage.
3. Check that the Media Library shows the newly replaced image.
4. Check that the links of the Stateless metabox in the image editor open the newly replaced image.

### DIFF
--- a/admin/js/media_replacer_cache_cleaner.js
+++ b/admin/js/media_replacer_cache_cleaner.js
@@ -1,0 +1,36 @@
+/* global cacheCleaner */
+
+/**
+ * JavaScript implementation for forcing images to update after replacing,
+ * by introducing an extra parameter to the URL.
+ */
+document.addEventListener('DOMContentLoaded', () => {
+  const cleaner = '?cache_cleaner=' + Date.now();
+
+  // Check if we are on the media library page
+  if (window.location.href.indexOf('upload.php') > -1) {
+    const selectors = '.attachments-wrapper img, .edit-attachment-frame img, #the-list img';
+    const images = document.querySelectorAll(selectors);
+
+    images.forEach(img => {
+      if (cacheCleaner.some(substring => img.getAttribute('src').includes(substring))) {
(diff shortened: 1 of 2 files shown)


planet-4:

:watch: 2025.03.03 15:38:28

comzeradd:
> Instead of that, this PR adds a query parameter to the image URLs using JavaScript to force the browser to show the updated version of the image.

I'm not sure if that achieves what we want 🤔  Mostly because this is just for the Media Library and therefor just for the admins. It doesn't affect the image in the frontend, if it has already been added in a published Post. If we don't purge the Cloudflare cache at all, then the end users would still see the older image.

As long as the API request we do through `CloudflarePurger` is successful, we can adjust the dashboard message to let the admins know that it may take a couple of minutes to work and they may have to clear their browser cache. I think it makes more sense for the implementation to work better for end users even if it doesn't work perfectly for admins, than the other way around.

mardelnet:
> > Instead of that, this PR adds a query parameter to the image URLs using JavaScript to force the browser to show the updated version of the image.

> As long as the API request we do through `CloudflarePurger` is successful, we can adjust the dashboard message to let the admins know that it may take a couple of minutes to work and they may have to clear their browser cache. I think it makes more sense for the implementation to work better for end users even if it doesn't work perfectly for admins, than the other way around.

You're absolutely right, @comzeradd 
I'm closing this PR and will continue working on #2540

PR #2554, opened 2025-03-04
PLANET-7749 Add Bluesky share button

sagarsdeshmukh:
Ref. https://jira.greenpeace.org/browse/PLANET-7749

### Requirements

- Add Bluesky as an option to Planet 4 > Social > Social sharing options (not enabled by default)
- Add Bluesky as an option on Forms sharing settings
- Use Bluesky's compose url to generate the sharing action
- We already have the Bluesky icon in our svg sprite so we can use the twig tag we have, same as we do for [Whatsapp](https://github.com/greenpeace/planet4-master-theme/blob/v1.308.0/templates/blocks/share_buttons.twig#L46). We can use #1185fe as the svg [fill color](https://github.com/greenpeace/planet4-master-theme/blob/c31f6667ff235855b2ac33759f6a35fe76659833/assets/src/scss/base/_colors.scss#L2).

### Testing Instructions:

1. Login to the [test instance](https://www-dev.greenpeace.org/test-titan/wp-admin/) or test it on your local environment.
2. Navigate to **Planet 4 > Social > Social Sharing Options** and enable the newly added **Bluesky** social share option.
3. Create a new post and check the frontend—you should see the Bluesky share option available in the social sharing list.
4. Create a new Gravity Form (GF) or edit an existing one:

      - Go to **Settings > Confirmation page**.
      
      - Under Confirmation Type 'text', verify that the **Bluesky** share button option is available.
      
      - If enabled, it should appear in the form confirmation's share buttons.

5. Verify the Bluesky compose URL and check if any changes are needed.

### DIFF
--- a/src/GravityFormsExtensions.php
+++ b/src/GravityFormsExtensions.php
@@ -80,6 +80,11 @@ class GravityFormsExtensions
             'name' => 'native',
             'default_value' => 1,
         ],
+        [
+            'label' => 'Bluesky',
+            'name' => 'bluesky',
+            'default_value' => 0,
+        ],
     ];
 
     public array $current_entry = [];
@@ -509,6 +514,7 @@ class GravityFormsExtensions
             'whatsapp' => $current_confirmation['whatsapp'] ?? true,
             'native' => $current_confirmation['native'] ?? true,
             'email' => $current_confirmation['email'] ?? true,
+            'bluesky' => $current_confirmation['bluesky'] ?? true,
         ];
 
         $apply_social_sharing_options = planet4_get_option('apply_social_sharing_options', 'posts_and_forms');
@@ -520,6 +526,7 @@ class GravityFormsExtensions
                 'twitter' => in_array('twitter', $social_share_options),
                 'whatsapp' => in_array('whatsapp', $social_share_options),
                 'email' => in_array('email', $social_share_options),
+                'bluesky' => in_array('bluesky', $social_share_options),
                 // We might add a setting for this one in the future, but for now we always enable it in GF.
                 'native' => true,
             ];
--- a/templates/blocks/share_buttons.twig
+++ b/templates/blocks/share_buttons.twig
@@ -39,11 +39,12 @@
     {% endif %}
     <!-- Bluesky -->
     {% if share_platforms.bluesky %}
-        <a href="https://bsky.app/intent/compose?text={{ social.title|url_encode }}{% if share_text or social.description %} - {{ (share_text ?? social.description)|striptags|url_encode }}{% endif %}, via @{{ social_accounts.bluesky ?? 'greenpeace' }} {{ (socialLink ~ '&utm_source=bluesky')|url_encode }}"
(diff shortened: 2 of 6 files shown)


planet-4:

:watch: 2025.03.04 10:59:40

src/GravityFormsExtensions.php
```diff
@@ -80,6 +80,11 @@ class GravityFormsExtensions
             'name' => 'native',
             'default_value' => 1,
         ],
+        [
+            'label' => 'Bluesky',
+            'name' => 'bluesky',
+            'default_value' => 0,
```
sagarsdeshmukh, line 86: As like at `Planet 4 > Social > Social sharing options`, I kept it not enabled by default

templates/blocks/share_buttons.twig
```diff
@@ -37,6 +37,15 @@
             <span class="visually-hidden">{{ __( 'Share via', 'planet4-master-theme' ) }} Email</span>
         </a>
     {% endif %}
+    <!-- Bluesky -->
+    {% if share_platforms.bluesky %}
+        <a href="https://bsky.app/intent/compose?text={{ social.title|url_encode }}{% if share_text or social.description %} - {{ (share_text ?? social.description)|striptags|url_encode }}{% endif %}, via @{{ social_accounts.bluesky ?? 'greenpeace' }} {{ (socialLink ~ '&utm_source=bluesky')|url_encode }}"
```
sagarsdeshmukh, line 42: Here (`via @{{ social_accounts.bluesky ?? 'greenpeace' }}`) I added 'greenpeace' as a fallback case.

comzeradd: I don't think there is a `greenpeace` account in Bluesky at the moment. Maybe it's better to not have a fallback at all if a Bluesky account is not defined, same as we do for X.

templates/blocks/share_buttons.twig
```diff
@@ -37,6 +37,15 @@
...
+           onclick="window.dataLayerPush('Bluesky');"
+           target="_blank" class="share-btn bluesky">
+            {{ 'bluesky'|svgicon }}
+            <span class="visually-hidden">{{ __( 'Share on', 'planet4-master-theme' ) }} Bluesky</span>
```
mleray, line 46: I'm wondering if this might lead to translation issues in some languages? 🤔 Maybe in some cases "Bluesky" would not be at the end of the sentence for example... it could be better to use a variable? I know it's the same for other platforms though so it can be handled in a follow-up

templates/blocks/share_buttons.twig
sagarsdeshmukh: Fixed.

templates/blocks/share_buttons.twig
```diff
@@ -37,6 +37,16 @@
             <span class="visually-hidden">{{ __( 'Share via', 'planet4-master-theme' ) }} Email</span>
         </a>
     {% endif %}
+    <!-- Bluesky -->
+    {% if share_platforms.bluesky %}
+        <a href="https://bsky.app/intent/compose?text={{ social.title|url_encode }}{% if share_text or social.description %} - {{ (share_text ?? social.description)|striptags|url_encode }}{% endif %}, via @{{ social_accounts.bluesky }} {{ (socialLink ~ '&utm_source=bluesky')|url_encode }}"
+           onclick="window.dataLayerPush('Bluesky');"
```
comzeradd, line 43: If there is no bluesky account set up in footer social accounts, we get something like this: `via @ <url>`. Which doesn't look nice. I think we have the same issue for X.

comzeradd (approved):

PR #2555, opened 2025-03-04
PLANET-7702: Add Actions List Block CTA text for Screen Reader

mardelnet:
### Summary
This PR introduces an `aria-label` attribute to the Actions CTA to be read by screen readers. 
The text of the attribute can be edited from the sidebar of the Actions post type editor.
Also, some texts have been adjusted.

---

Ref: https://jira.greenpeace.org/browse/PLANET-7702

### Testing
1. Go to the editor of an Action post.
2. Check the new field ("Add descriptive text for screen readers...") exists in the sidebar.
3. Add text to the field and save the changes.
4. Visit a page containing an Actions List block with the Action edited in the previous steps.
5. Check that the CTA of that action in the Actions List block contains an `aria-label` attribute with the value added in step 3 (i.e., the Oceans action in [this page](https://www-dev.greenpeace.org/test-neptune/actions-list-test-page/)).
6. Check that a screen reader reads the value of the `aria-label` correctly.

### DIFF
--- a/src/ActionPage.php
+++ b/src/ActionPage.php
@@ -215,7 +215,7 @@ class ActionPage
             'action_button_accessibility_text',
             array_merge(
                 $args,
-                [ 'default' => __('Get Involved', 'planet4-master-theme') ]
+                [ 'default' => __('Take action', 'planet4-master-theme') ]
             )
         );
 
--- a/src/ActionPage.php
+++ b/src/ActionPage.php
@@ -215,7 +215,7 @@ class ActionPage
             'action_button_accessibility_text',
             array_merge(
                 $args,
-                [ 'default' => __('Take action', 'planet4-master-theme') ]
+                [ 'default' => __('Get Involved', 'planet4-master-theme') ]
             )
         );
 
--- a/src/ActionPage.php
+++ b/src/ActionPage.php
@@ -210,6 +210,15 @@ class ActionPage
             )
         );
 
+        register_post_meta(
+            self::POST_TYPE,
+            'action_button_accessibility_text',
+            array_merge(
+                $args,
+                [ 'default' => __('Take action', 'planet4-master-theme') ]
+            )
+        );
+
         foreach (self::META_FIELDS as $field) {
             register_post_meta(self::POST_TYPE, $field, $args);
         }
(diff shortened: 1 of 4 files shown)


planet-4:

:watch: 2025.03.04 14:11:10

src/ActionPage.php
```diff
@@ -210,6 +210,15 @@ public function register_post_meta(): void
             )
         );
 
+        register_post_meta(
+            self::POST_TYPE,
+            'action_button_accessibility_text',
+            array_merge(
+                $args,
+                [ 'default' => __('Take action', 'planet4-master-theme') ]
```
sagarsdeshmukh, line 218: Can we use the same text domain for the 'Take action' text here? If we try to unify text domains later, we might face challenges in deciding which translation version to keep—one of them might be translated while the other is not.

sagarsdeshmukh (approved):
I tested the PR on local and the screen reader functionality for CTA button works as expected 👍

src/ActionPage.php
mardelnet: Which textdomain in particular, @sagarsdeshmukh?
I found that for the same sentence ("Take action") there are a bunch of different textdomains: `planet4-master-theme`, `planet4-blocks`, `planet4-blocks-backend`...

sagarsdeshmukh: We have 3 instances in the current PR where the 'Take action' translation is used—
line #218 in `src/ActionPage.php` file
`[ 'default' => __('Take action', 'planet4-master-theme') ]`

line #84 & #94 in `src/Blocks/ActionButtonText.php` file
`$button_text = __('Take action', 'planet4-blocks');`
`$button_acc_text = __('Take action', 'planet4-blocks');`

So my suggestion was to use a 'planet4-master-theme' texdomain in all 3 places. it's already used at [here](https://github.com/greenpeace/planet4-master-theme/blob/main/src/ActionPage.php#L218) and [here](https://github.com/greenpeace/planet4-master-theme/blob/main/src/Search/SearchPage.php#L140) in master theme.
Nevermind, we will fix it when we work on [PLANET-7684](https://jira.greenpeace.org/browse/PLANET-7684)

PR #2556, opened 2025-03-05
PLANET-7762 Fixed Posts list layout and made links clickable

Osong-Michael:
### Summary

---

Ref: https://jira.greenpeace.org/browse/PLANET-7762

### Testing
- Install the India NRO locally
- Inspect the Publications page, the Posts List block should have posts with no clickable titles like this [example](https://www-dev.greenpeace.org/india/en/latest-publications/).
- Switch to this branch locally and run the migration using `npx wp-env run cli wp p4-run-activator`
- After the migration runs the Posts List block should be updated and the links are clickable.
- Also, the post titles should now be `h4` elements instead of `h2` and the styling should still match


mardelnet:
@Osong-Michael, I think you can get the block title and description by using recursive functions:

```php
private static function getTitle(array $blocks) {
        foreach ($blocks as $block) {
            if (isset($block['innerHTML']) && str_contains($block['innerHTML'], 'wp-block-heading')) {
                return $block; // Return the found value
            }
            if (isset($block['innerBlocks']) && is_array($block['innerBlocks'])) {
                $result = self::getTitle($block['innerBlocks']);
                if ($result !== false) {
                    return $result; // Return early if a match is found in recursion
                }
            }
        }
        return false;
    }

    private static function getDescription(array $blocks) {
        foreach ($blocks as $block) {
            if (isset($block['attrs']['placeholder']) && str_contains($block['attrs']['placeholder'], 'Enter description')) {
                return $block; // Return the found value
            }
            if (isset($block['innerBlocks']) && is_array($block['innerBlocks'])) {
                $result = self::getDescription($block['innerBlocks']);
                if ($result !== false) {
                    return $result; // Return early if a match is found in recursion
                }
            }
        }
        return false;
    }
```

You can call those functions from the `get_posts_list_block_attrs` function, like this:

```php
$title = self::getTitle($existing_block['innerBlocks']);
$description = self::getDescription($existing_block['innerBlocks']);
```

And change the return of the `get_posts_list_block_attrs` function, like this:

```php
'title' => $title['innerHTML'],
'description' => $description['attrs']['placeholder'],
```

Please, test it thoroughly and try to improve the code to avoid so much duplication.

src/Migrations/M042FixPostsListMigration.php
```diff
@@ -0,0 +1,445 @@
...
+                ) {
+                    return $block;
+                }
+            }
+
+            if (empty($block['innerBlocks'])) {
+                continue;
+            }
+
+            $found = find_block($block['innerBlocks'], $targetBlockName, $attributeKey, $attributeValue);
```
mardelnet, line 417:
```suggestion
            $found = self::find_block($block['innerBlocks'], $targetBlockName, $attributeKey, $attributeValue);
```

src/Migrations/M042FixPostsListMigration.php
```diff
@@ -0,0 +1,445 @@
...
+        $query['inherit'] = false;
+        $query['hasPassword'] = false;
+        $query['postIn'] = $post_in;
+
+        if (!empty($per_page)) {
+            $query['perPage'] = $per_page;
+        }
+
+        if (empty($per_page)) {
+            if ($layout === 'default') {
```
mardelnet, line 173: Undefined variable `$layout`

mardelnet (commented):
Some initial comments:

src/Migrations/M042FixPostsListMigration.php
```diff
@@ -0,0 +1,445 @@
...
+        $inner_blocks = [
+            self::get_head_group_block($existing_block_attrs['title']),
+            self::get_paragraph_block($existing_block_attrs['description']),
+            self::get_query_no_results_block(),
+            self::get_post_template(),
+            self::get_buttons_block(),
+            self::get_nav_links_block(),
+        ];
+
+        return Utils\Functions::create_block_query(
```
mardelnet, line 91: I wonder if it is really necessary to re-create the whole block.
Can't we just modify the `$block` array to change the layout and add the link?

Osong-Michael: I am open to suggestions.
But since there are only a few Posts List blocks I think we can go with this solution.

mardelnet: I think this is the only code we need:

```php
private static function transform_block(array $block): array {
        self::add_attribute_to_post_title($block['innerBlocks']);
        return $block;
}

private static function add_attribute_to_post_title(&$blocks) {
        foreach ($blocks as &$block) {
            if (isset($block['blockName']) && $block['blockName'] === 'core/post-title') {
                if (!isset($block['attrs'])) {
                    $block['attrs'] = [];
                }
                $block['attrs']['isLink'] = true;
                $block['attrs']['level'] = 4;
            }
            if (isset($block['innerBlocks']) && is_array($block['innerBlocks'])) {
                self::add_attribute_to_post_title($block['innerBlocks']);
            }
        }
}
```

By doing this, we remove hundreds of lines of code, avoiding a ton of duplicated code (which is making fail the Sonar Cloud Analysis) and making the code review way easier.

mardelnet: @Osong-Michael, to make things clearer I opened this branch with the previously mentioned changes: 
https://github.com/greenpeace/planet4-master-theme/compare/main...fix/posts-list-layout-pedro-test 
You can check it locally.
That branch only fixes the link issue, not the layout one, but that can be easily applied.

Osong-Michael: I may be missing something, but re-arranging the order of the `post-terms` attribute array, does not seem to update the layout issue. 🤔

mardelnet: Try something like this:
```php
if (isset($block['blockName']) && $block['blockName'] === 'core/post-terms') {
                if (isset($block['attrs']['term']) && $block['attrs']['term'] === 'post_tag') {
                    $block['attrs']['term'] = 'category';
                    $block['attrs']['separator'] = ' | ';
                } else if (isset($block['attrs']['term']) && $block['attrs']['term'] === 'category') {
                    $block['attrs']['term'] = 'post_tag';
                    $block['attrs']['separator'] = ' ';
                }
            }
```

Osong-Michael: > ```
> if (isset($block['attrs']['term']) && $block['attrs']['term'] === 'post_tag') {
>                     $block['attrs']['term'] = 'category';
>                     $block['attrs']['separator'] = ' | ';
>                 } else if (isset($block['attrs']['term']) && $block['attrs']['term'] === 'category') {
>                     $block['attrs']['term'] = 'post_tag';
>                     $block['attrs']['separator'] = ' ';
>                 }
> ```

Same outcome, layout does not change.

Osong-Michael: Yes, I changed the whole code as you suggested. Then reinstalled the India NRO locally, edited one of the posts to have tags so the layout shows. Ran the migration with the new update, refreshed the page, everything else works but tags still come before the category

planet-4:

:watch: 2025.03.10 18:18:48

mardelnet (commented):
@Osong-Michael 
1. We have to find a way of switching terms with categories and vice-versa only in the blocks that need it. Currently, the script switches all the Posts list blocks.
2. Should we add some kind of text transformation when the cursor hovers the title to make the link more noticeable?
3. The dot separator has to be placed before the category name. Also, some spacing is needed below the title:
All-blocks-Evergreen-template-Greenpeace

Osong-Michael:
> pluto

Hey @mardelnet for the space is your server for the theme repo running? because I added the space already. 

For the terms, I think we have only the Posts list block which were formerly content Covers block which means they have the issue we are trying to solve.

For the title links, we are just following the designs unless you have something you want to suggest to the design team. 

The Dot separator is where it is suppose to be you can build a new Posts list block in the editor and see the layout.

assets/src/scss/blocks/PostsList/PostsListEditorStyle.scss
```diff
@@ -22,6 +22,10 @@
     }
   }
 
+  .wp-block-post-title {
+    margin-bottom: 24px;
```
mleray, line 26: Can we use a variable here?

assets/src/scss/blocks/PostsList/PostsListStyle.scss
```diff
@@ -92,6 +92,7 @@
 
   .wp-block-post-title {
     font-size: 18px;
+    margin-bottom: 24px;
```
mleray, line 95: Can we use a variable here?

mardelnet:
> Hey @mardelnet for the space is your server for the theme repo running? because I added the space already.

Okay!
 
> For the terms, I think we have only the Posts list block which were formerly content Covers block which means they have the issue we are trying to solve.

I just tested the migration and the script affects all the Posts lists blocks, which means that if the NROs have added new blocks or already edited the existing ones, the position of the terms will be switched.

> For the title links, we are just following the designs unless you have something you want to suggest to the design team.

Let's keep them as they are.
 
> The Dot separator is where it is suppose to be you can build a new Posts list block in the editor and see the layout.

Okay!

mardelnet (commented):
In this screenshot you can see a comparison between an already existing Posts list block (the top one) and a new one (down below). After running the migration script, I can see some spacing is missing below the header and the dots are different. Also, as the terms position is switched, the new block has the terms in a wrong position.

Screenshot-2025-03-11-at-10-31-52-People-Greenpeace

assets/src/js/add_style_posts_lists.js
```diff
@@ -0,0 +1,19 @@
+export const addMarginToPTag = () => {
```
mardelnet, line 1: I don't understand why this function is needed.
Why aren't we passing the margin using CSS or the migration script?
What elements are affected by this?

assets/src/scss/blocks/PostsList/PostsListEditorStyle.scss
```diff
@@ -22,6 +22,10 @@
     }
   }
 
+  .wp-block-post-title {
```
mardelnet, line 25: The posts' titles should be underlined on mouse hover.

Osong-Michael: The title is underlined by default on hover, part of the styles we have for a tags

https://github.com/user-attachments/assets/ee9e86f6-2f65-42bc-9867-40723a1c50b2

Osong-Michael: Is this on all browsers or just a particular one, because I still get the hover on my end even after the migration

src/Migrations/M042FixPostsListMigration.php
```diff
@@ -0,0 +1,137 @@
...
+                }
+
+                $core_post_terms_block[0]['attrs']['term'] = 'category';
+                $core_post_terms_block[0]['attrs']['separator'] = ' | ';
+                $core_post_terms_block[1]['attrs']['term'] = 'post_tag';
+                $core_post_terms_block[1]['attrs']['separator'] = ' ';
+            }
+
+            if (isset($block['blockName']) && $block['blockName'] === 'core/paragraph') {
+                if (isset($block['attrs']['placeholder']) && $block['attrs']['placeholder'] === 'Enter description') {
```
mardelnet, line 115: I think this condition should be either removed or changed.
If website editors have changed the placeholder description for something different than `Enter description` the margins will be not added.

mardelnet: The tricky part will be if the editors have added another paragraph... 🤔

Osong-Michael: 🤔, i don't think so, we set that value by default, editors can change the content of the paragraph block but the placeholder value will remain the same.

mardelnet: I see! I think you're right!

assets/src/scss/blocks/PostsList/PostsListEditorStyle.scss
mardelnet: Check Firefox

Osong-Michael: fixed

mardelnet (commented):
Nice job, @Osong-Michael 
On my local installation with the default P4 dev database, the "See all stories" button of the already existing blocks is pointing to an incorrect URL. Would you check that?
Also, the e2e tests were fixed. Please, update this PR against the main branch.

Osong-Michael:
> Nice job, @Osong-Michael On my local installation with the default P4 dev database, the "See all stories" button of the already existing blocks is pointing to an incorrect URL. Would you check that? Also, the e2e tests were fixed. Please, update this PR against the main branch.

🤔  Seems to be an issue only on local from the content covers migration, but does not affect dev-instances or live sites

Osong-Michael:
> Nice job, @Osong-Michael On my local installation with the default P4 dev database, the "See all stories" button of the already existing blocks is pointing to an incorrect URL. Would you check that? Also, the e2e tests were fixed. Please, update this PR against the main branch.

Hey @mardelnet I rebased my PR, but the audio-video block spec still fails on this PR.

mardelnet (commented):
Your code looks good to me, @Osong-Michael 
However, before approving this and considering that this is a sensitive PR, I'd like to ask @mleray for an extra review.

mleray:
Hey @Osong-Michael, I think the design is still a bit off. If I compare the Related Posts block ([example](https://www-dev.greenpeace.org/test-pluto/press/1113/duis-posuere-6/)) and Posts List block ([example](https://www-dev.greenpeace.org/test-pluto/just-a-posts-list-block/)) to the listing pages ([example](https://www-dev.greenpeace.org/test-pluto/tag/renewables/)), in certain screen sizes they look different:

Screenshot 2025-03-24 at 10 18 43

Screenshot 2025-03-24 at 10 18 50

mleray:
Also @Osong-Michael, on local the migration didn't work for me with an existing Posts List block (I didn't switch to the NRO, just the general local env) After running it I still see a `h2` heading instead of a `h4`.

Osong-Michael:
> Also @Osong-Michael, on local the migration didn't work for me with an existing Posts List block (I didn't switch to the NRO, just the general local env) After running it I still see a `h2` heading instead of a `h4`.

Can I see the element you are talking about, curious to see why it didn't work. 

Please send like an image of the inspected element

mleray:
> Please send like an image of the inspected element

@Osong-Michael 

Screenshot 2025-03-25 at 10 09 38

Osong-Michael:
> @Osong-Michael I've been able to reproduce my local issue on [the venus test instance](https://www-dev.greenpeace.org/test-venus/just-a-posts-list-block/). I've added the Posts List block and then [switched the instance to your branch](https://github.com/greenpeace/planet4-test-venus/blob/main/composer-local.json) (so the migration ran) and yet the block still uses `h2`.

Hey @mleray I just tested on local too and I think I understand why, the migration scripts seems to work too well 😅. It ignores Posts List blocks which are ok (the post terms are in order), from my testing it fixes just the blocks with issues skips the others without. Which is kinda strange in my opinion. If you check this [page](https://www-dev.greenpeace.org/test-venus/get-informed/people/) on the venus instance you can see it updated the Post list block here while skipping the one you created.

mleray:
@Osong-Michael maybe we should ask for UAT on this 🤔 It seems to me that there are still some slight design differences between Posts List and listing pages (I'm aware that this is a bit nit-picky, sorry!)

Screenshot 2025-03-25 at 11 15 19

Screenshot 2025-03-25 at 11 15 36

mleray:
> @Osong-Michael maybe we should ask for UAT on this 🤔 It seems to me that there are still some slight design differences between Posts List and listing pages (I'm aware that this is a bit nit-picky, sorry!)

@Osong-Michael this could also be done in a follow-up ticket if the fix is urgent

mleray:
> Hey @mleray I noticed too, but may be easier to fix in a follow up ticket as design wise they all use the same style but for some reason the blocks especially the post terms section seems to look slightly different. Might be easier to fix after the migration runs.

@Osong-Michael that's fine by me, can you please create the follow-up ticket? However we still need the migration to work for all Posts List blocks so that they all use `h4` from now on instead of `h2`.

Osong-Michael:
> > Hey @mleray I noticed too, but may be easier to fix in a follow up ticket as design wise they all use the same style but for some reason the blocks especially the post terms section seems to look slightly different. Might be easier to fix after the migration runs.
> 
> @Osong-Michael that's fine by me, can you please create the follow-up ticket? However we still need the migration to work for all Posts List blocks so that they all use `h4` from now on instead of `h2`.

Hey @mleray and @mardelnet, I split the code into two separate functions so the title is updated in all Posts List blocks. Should work as expected now.

src/Migrations/M042FixPostsListMigration.php
```diff
@@ -0,0 +1,158 @@
...
+
+    /**
+     * Update post titles.
+     *
+     * @param array $blocks - The block.
+     */
+    private static function update_posts_list_block_title(array &$blocks): void
+    {
+        foreach ($blocks as &$block) {
+            if (isset($block['blockName']) && $block['blockName'] === 'core/post-title') {
```
mardelnet, line 81: We could use a constant from the `Utils\Constants` class here.

src/Migrations/M042FixPostsListMigration.php
```diff
@@ -0,0 +1,158 @@
...
+    private static function update_posts_list_block_terms(array &$blocks): void
+    {
+        foreach ($blocks as &$block) {
+            if (isset($block['blockName']) && $block['blockName'] === 'core/post-title') {
+                if (!isset($block['attrs'])) {
+                    $block['attrs'] = [];
+                }
+                $block['attrs']['isLink'] = true;
+                $block['attrs']['level'] = 4;
+            }
```
mardelnet, line 107: Isn't this doing the same as the `update_posts_list_block_title` function?
If so, why are we using 2 different functions making the iteration through all the blocks twice?

src/Migrations/M042FixPostsListMigration.php
```diff
@@ -0,0 +1,167 @@
...
+        if (!is_array($block)) {
+            return false;
+        }
+
+        // Check if the block has a 'blockName' key.
+        if (!isset($block['blockName'])) {
+            return false;
+        }
+
+        // Check if the block is a Posts List block.
```
mleray, line 56: Technically this checks if the block is Query Loop, not Posts List

src/Migrations/M042FixPostsListMigration.php
```diff
@@ -0,0 +1,167 @@
...
+            }
+
+            self::update_posts_list_block_title($block['innerBlocks']);
+        }
+    }
+
+     /**
+     * Update Posts List block post terms and paragraph.
+     *
+     * @param array $blocks - The block.
```
mleray, line 96: This is an array of blocks, not just "the block"

src/Migrations/M042FixPostsListMigration.php
```diff
@@ -0,0 +1,171 @@
...
+    /**
+     * Update the post title, post terms and paragraph of the
+     * Posts List block
+     *
+     * @param array $block - The current posts list block.
+     * @return array - array of blocks.
+     */
+    private static function transform_block(array $block): array
+    {
+        if ($block['attrs']['namespace'] === Utils\Constants::BLOCK_POSTS_LIST) {
```
mleray, line 69: I would add this in the `check_is_valid_block` function instead since we don't want to migrate otherwise

mardelnet (commented):
@Osong-Michael whilst migrated blocks (green) are using `h4` for the titles, the new blocks (orange) are still using `h2`. Also, migrated blocks need extra spacing around the title.

Screenshot-from-2025-04-01-10-50-17

src/Migrations/M042FixPostsListMigration.php
```diff
@@ -0,0 +1,171 @@
...
+                    if (isset($block['innerHTML'])) {
+                        $block['innerHTML'] = self::addStyleIfMissing($block['innerHTML'], $style);
+                    }
+
+                    if (isset($block['innerContent'][0])) {
+                        $block['innerContent'][0] = self::addStyleIfMissing($block['innerContent'][0], $style);
+                    }
+                }
+            }
+
```
mardelnet, line 146: I think that `$block['innerHTML']` and `$block['innerContent'][0]` are the same, so this logic can be simplified and the `addStyleIfMissing` function removed:

```php
            if (isset($block['blockName']) && $block['blockName'] === Utils\Constants::BLOCK_PARAGRAPH) {
                if (isset($block['attrs']['placeholder']) && $block['attrs']['placeholder'] === 'Enter description') {
                    $style = 'margin-top: "24px" margin-bottom: "36px"';

                    if (!preg_match('/<p[^>]*\sstyle=["\']([^"\']*)["\']/', $style)) {
                        $paragraph = preg_replace('/<p([^>]*)>/', '<p\1 style="' . $style . '">', $block['innerHTML'], 1);

                        if (isset($block['innerHTML'])) {
                            $block['innerHTML'] = $paragraph;
                        }
                        if (isset($block['innerContent'][0])) {
                            $block['innerContent'][0] = $paragraph;
                        }
                    }
                }
            }
```

Osong-Michael: Quick question @mardelnet , what is the if condition checking?

mardelnet: Unless I'm wrong, it is the same condition you added to the `addStyleIfMissing` function.
It could be wrong, as I haven't tested it thoroughly.

Osong-Michael: My bad, I didn't read it correctly

Osong-Michael: What I wanted to say was, the regular expression checks the string passed which is the p tag in my case to see if there is a style attribute on it, in your case you are passing the style variable instead which does not have the paragraph tag.

mardelnet: Okay, my bad, forget the if condition and the regular expression.
The important thing is that there's no need to call the `addStyleIfMissing` function twice since the values of `$block['innerHTML']` and `$block['innerContent'][0] are the same.

mardelnet: Thanks for this change, @Osong-Michael 
As the regular expression is looking for the existence of the string "style=", the condition will be false if the editors have added custom styles to the paragraph. However, I think this is a very edge case, so I'm okay to leave it as it is now.

Osong-Michael:
> @Osong-Michael whilst migrated blocks (green) are using `h4` for the titles, the new blocks (orange) are still using `h2`. Also, migrated blocks need extra spacing around the title, and the title loses the underline on hover.
> 
> Screenshot-from-2025-04-01-10-50-17

I already fixed this h2 but let me double check, but for the spacing, ensure the server is running locally so the styles get loaded

mardelnet (approved):
Well done, @Osong-Michael!

PR #2559, opened 2025-03-05
Fix tag redirect functionality

comzeradd:
### Summary

Follow up from 580fa49e23ee51f9239a025dc5965073dc42f0ed

Tried a different way of making the redirect work, while keeping Sentry happy.

---

### Testing

1. Currently if you pick a page as a redirect while editing a Tag it doens't work in the frontend
2. In this branch the redirect works just fine

image

### DIFF
--- a/tag.php
+++ b/tag.php
@@ -20,7 +20,7 @@ if (!is_tag()) {
 }
 
 $tag = get_queried_object();
-$redirect_id = $tag !== null && get_term_meta($tag->term_id, 'redirect_page', true);
+$redirect_id = isset($tag, $tag->term_id) ? get_term_meta($tag->term_id, 'redirect_page', true) : false;
 
 if ($redirect_id) {
     global $wp_query;


mardelnet (approved):
LGTM!

tag.php
```diff
@@ -20,7 +20,7 @@
 }
 
 $tag = get_queried_object();
-$redirect_id = $tag !== null && get_term_meta($tag->term_id, 'redirect_page', true);
+$redirect_id = isset($tag) ? get_term_meta($tag->term_id, 'redirect_page', true) : false;
```
mardelnet, line 23: Do you think this can be safer for Sentry?

```suggestion
isset($tag, $tag->term_id) ? get_term_meta($tag->term_id, 'redirect_page', true) : false;
```

comzeradd: Yes, good idea. Amended 👍🏻

PR #2561, opened 2025-03-06
PLANET-7782 & PLANET-7738 Improve comments form

mleray:
### Description

See [PLANET-7782](https://jira.greenpeace.org/browse/PLANET-7782). The new style is applied to links within comments, and also to the Reply button.

See [PLANET-7738](https://jira.greenpeace.org/browse/PLANET-7738). 
- In RTL sites, the "cancel reply" link is missing spacing
- When the logged in user has a long name, in some screen sizes the "Logged in as..." message overlaps with the "cancel reply" link

### Testing

You can check the changes on [this post](https://www-dev.greenpeace.org/test-jupiter/press/1109/duis-posuere-4) for example.

### DIFF
--- a/assets/src/scss/layout/_forms.scss
+++ b/assets/src/scss/layout/_forms.scss
@@ -155,6 +155,11 @@ textarea.form-control {
       opacity: 0;
       transition: all .3s ease;
       margin-bottom: 0;
+
+      html[dir="rtl"] & {
+        right: 0;
+        left: auto;
+      }
     }
 
     .form-control:not(:placeholder-shown) {
--- a/assets/src/scss/pages/post/_comments-block.scss
+++ b/assets/src/scss/pages/post/_comments-block.scss
@@ -27,6 +27,10 @@
   line-height: 1.5;
   margin-bottom: 8px;
   overflow-wrap: break-word;
+
+  a {
+    @include shared-link-styles;
+  }
 }
 
 .single-comment-meta {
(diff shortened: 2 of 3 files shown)


planet-4:

:watch: 2025.03.06 13:00:09

mardelnet (commented):
Your code looks good to me, @mleray

Should we add some extra spacing to the labels in RTL sites? 
Duis-posuere-4-Greenpeace

I have noticed that WordPress is throwing this error when a user attempts to respond to an unapproved message. I wonder if we should handle this in a different way (in another PR!)
Screenshot from 2025-03-11 10-08-19

mleray:
> I have noticed that WordPress is throwing this error when a user attempts to respond to an unapproved message. I wonder if we should handle this in a different way (in another PR!) 

@mardelnet if this is WordPress default behaviour, we should probably leave it as is? Also, might be only an issue when logged in 🤔

mardelnet:
> > I have noticed that WordPress is throwing this error when a user attempts to respond to an unapproved message. I wonder if we should handle this in a different way (in another PR!)
> 
> @mardelnet if this is WordPress default behaviour, we should probably leave it as is? Also, might be an issue only when logged in 🤔

This happens for not logged in users. And despite being the default behaviour, it seems to me a bit rough showing the users a full white error screen. But this is a very edge case, we don't have to pay too much attention to it right now.

mardelnet (approved):
Well done!

PR #2565, opened 2025-03-12
PLANET-7603 Add an option to change TAB heading size 

mleray:
### Description

See [PLANET-7603](https://jira.greenpeace.org/browse/PLANET-7603)

**Requirements**

- Heading Sizes: The title should be able to be adjusted with the following options:
  - **Desktop XXL, XL, & L**
    - XL: 28px/1.75rem
    - L: 22px/1.1375rem
    - M: 20px/1.25rem - default
    - S: 18px/1.125rem
  
  - **Mobile M&S**
    - XL: 22px/1.1375rem
    - L: 20px/1.25rem
    - M: 18px/1.125rem - default
    - S: 16px/1rem

- Default Setting: The default title size will be set to Heading M.
- Use design tokens for font sizes
- Ideally use rem points for font sizes

[FontSizePicker documentation
](https://developer.wordpress.org/block-editor/reference-guides/components/font-size-picker/)

### Testing

You can test all the new variations on [this post](https://www-dev.greenpeace.org/test-sinope/press/1113/duis-posuere-6/) for example (different font sizes + custom vs selected Action card)

### DIFF
--- a/assets/src/blocks/TakeActionBoxout/HeadingFontSizes.js
+++ b/assets/src/blocks/TakeActionBoxout/HeadingFontSizes.js
@@ -1,0 +1,32 @@
+const {__} = wp.i18n;
+
+export const FONT_SIZES = [
+  {
+    name: __('Small'),
+    slug: 'small',
+    size: '1.125rem',
+  },
+  {
+    name: __('Medium'),
+    slug: 'medium',
+    size: '1.25rem',
+  },
+  {
+    name: __('Large'),
+    slug: 'large',
+    size: '1.375rem',
+  },
+  {
+    name: __('Extra Large'),
+    slug: 'x-large',
+    size: '1.75rem',
+  },
+];
+
+export const getHeadingFontSizeClassName = fontSize => {
+  if (!fontSize || !FONT_SIZES.find(font => font.size === fontSize)) {
+    return 'medium';
+  }
+
+  return FONT_SIZES.find(font => font.size === fontSize).slug;
+};
--- a/assets/src/blocks/TakeActionBoxout/HeadingFontSizes.js
+++ b/assets/src/blocks/TakeActionBoxout/HeadingFontSizes.js
@@ -4,29 +4,21 @@ export const FONT_SIZES = [
   {
     name: __('Small'),
     slug: 'small',
-    size: '1.125rem',
+    size: 'small',
   },
   {
     name: __('Medium'),
     slug: 'medium',
-    size: '1.25rem',
+    size: 'medium',
   },
   {
     name: __('Large'),
     slug: 'large',
-    size: '1.375rem',
+    size: 'large',
   },
   {
     name: __('Extra Large'),
     slug: 'x-large',
-    size: '1.75rem',
+    size: 'x-large',
   },
 ];
-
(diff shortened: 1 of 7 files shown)


mardelnet (commented):
Well done, @mleray 
I think there's a chance to avoid hardcoding the font size:

1. In the array `FONT_SIZES` use the slug value for the size property (ie `size: 'small'` or `size: 'x-large'`) and get rid of the `slug`
2. Remove the `getHeadingFontSizeClassName` function as it's no longer needed.
3. Replace `${getHeadingFontSizeClassName(headingFontSize)}` with `headingFontSize` 
4. Remove the hardcoded values in `HEADING_FONT_SIZES` and `headingFontSize

mleray:
> Well done, @mleray I think there's a chance to avoid hardcoding the font size:

@mardelnet for the FontSizePicker component we need to have this structure `{slug, name, size}`

mardelnet:
> > Well done, @mleray I think there's a chance to avoid hardcoding the font size:
> 
> @mardelnet for the FontSizePicker component we need to have this structure `{slug, name, size}`

You can use the same value for `slug` and `size` 
If we use a hardcoded value for the size, IMO, the whole point of using tokens stops making sense.
If the value of the token is changed in the future, we'll have to update the hardcoded sizes, too, as they're not dynamic.

mleray:
> You can use the same value for `slug` and `size`

@mardelnet if I do that then the component no longer correctly updates the selected size. We need to provide actual font sizes there. For the tokens, I'm not sure I understand what you mean, we use them in the CSS so if they are updated they will be applied to this component too?

mardelnet:
> > You can use the same value for `slug` and `size`
> 
> @mardelnet if I do that then the component no longer correctly updates the selected size. We need to provide actual font sizes there. For the tokens, I'm not sure I understand what you mean, we use them in the CSS so if they are updated they will be applied to this too?

Maybe it's easier this way: https://github.com/greenpeace/planet4-master-theme/commit/b6ee4348fa0e703ddd6b6f5406f74738759d2b5c

mleray:
> Maybe it's easier this way: [b6ee434](https://github.com/greenpeace/planet4-master-theme/commit/b6ee4348fa0e703ddd6b6f5406f74738759d2b5c)

The [documentation](https://developer.wordpress.org/block-editor/reference-guides/components/font-size-picker/#fontsizes-fontsize) specifically says you should have slug, name and size

mardelnet:
> > Maybe it's easier this way: [b6ee434](https://github.com/greenpeace/planet4-master-theme/commit/b6ee4348fa0e703ddd6b6f5406f74738759d2b5c)
> 
> @mardelnet the [documentation](https://developer.wordpress.org/block-editor/reference-guides/components/font-size-picker/#fontsizes-fontsize) specifically says you should have slug, name and size

We can use the same value for the slug and the size.
While hardcoding the font size in JS and PHP files won't stop the tokens in CSS from working as expected, it won't be consistent if the future, for example, the medium size changes from 1.25rem to 1.5rem or any other value, forcing us to change the values in the JS and PHP files as well.
Maybe I am too nick-picky with this 😆 If so, just ignore this change and I'll happily approve your PR (which works great, btw).

mardelnet (approved):

PR #2566, opened 2025-03-12
Fix lists' font size in the editor

mardelnet:
### Summary

This PR fixes the lists in the editor (not in the frontend) that have a smaller font size than normal paragraphs.

---

Ref: https://greenpeace.slack.com/archives/C014UMRC4AJ/p1741786765906429

### Testing
1. In the editor, add a list block and a paragraph block.
2. Add content to both blocks.
3. The font size of the blocks has to be the same.

### DIFF
--- a/assets/src/scss/editorOverrides.scss
+++ b/assets/src/scss/editorOverrides.scss
@@ -20,10 +20,6 @@
 }
 
 .edit-post-visual-editor {
-  li {
-    font-size: unset;
-  }
-
   @include table;
 }
 


assets/src/scss/editorOverrides.scss
```diff
@@ -20,10 +20,6 @@
 }
 
 .edit-post-visual-editor {
-  li {
```
mardelnet, line 23: I don't know why this property was introduced and whether removing it will cause an unexpected problem.
I could only identify the PR that introduced this change: https://github.com/greenpeace/planet4-master-theme/pull/1337/files

mleray: I don't know why either 🤷‍♀️ It looks like it was just moved from the plugin without questioning it

mardelnet: So... do we keep it or not? cc @GP-Dan-Tovbein

planet-4:

:watch: 2025.03.13 09:08:46

mleray (approved):
Doesn't seem to cause any issues 👍

PR #2572, opened 2025-03-19
PLANET-7718 Improve search page accessibility

mleray:
### Description

See [PLANET-7718](https://jira.greenpeace.org/browse/PLANET-7718)

**Requirements**

- [x] Add a Skip Link to the results section
- [x] Update the placeholder text of the search input to make it more descriptive (e.g., "Search by name, keyword, or topic").
- [x] Add an `aria-label` for filter checkboxes to ensure they clearly communicate the number of results (e.g., "People, 3 results").
- [x] Add an `aria-label` to active filter button removal (e.g. "Remove Energy filter")
- [x] Use proper HTML tags on the page
  - Use `<section>` or `<aside>` HTML tags for Filters
  - ~Use `<form>` for the Search Input.~ (already implemented)

### Testing

You can test all of these changes on the oberon instance [Search page](https://www-dev.greenpeace.org/test-oberon/?s=&orderby=_score).

### DIFF
--- a/templates/search.twig
+++ b/templates/search.twig
@@ -56,11 +56,7 @@
                 </div>
             </div>
             <div class="search-filter-results row clearfix">
-                <a
-                    class="skip-link"
-                    href="#search-results"
-                    aria-hidden="true"
-                >
+                <a class="skip-link" href="#search-results">
                     {{ __( 'Skip to search results', 'planet4-master-theme' ) }}
                 </a>
                 <aside class="col-lg-4">
--- a/templates/search.twig
+++ b/templates/search.twig
@@ -5,7 +5,7 @@
     {% set collapsed = posts or (not has_aggregation) ? '' : 'collapsed' %}
     {% set show      = posts or (not has_aggregation)? 'show' : '' %}
     {% set expanded  = posts ? true : false %}
-    {% set search_label = __( 'Search', 'planet4-master-theme' ) %}
+    {% set search_label = __( 'Search by name, keyword, or topic', 'planet4-master-theme' ) %}
 
     <div class="search-block">
         <div class="container">
@@ -56,7 +56,14 @@
                 </div>
             </div>
             <div class="search-filter-results row clearfix">
-                <div class="col-lg-4">
+                <a
+                    class="skip-link"
+                    href="#search-results"
+                    aria-hidden="true"
+                >
+                    {{ __( 'Skip to search results', 'planet4-master-theme' ) }}
+                </a>
+                <aside class="col-lg-4">
                     <div class="filter-sidebar">
                         <div class="filter-button d-lg-none">
                             <button class="btn btn-filter btn-secondary {{ disabled }}" data-bs-toggle="modal" data-bs-target="#filtermodal">
@@ -104,7 +111,11 @@
                                     <div class="activefilter-list">
                                         {% for filter_type in filters %}
                                             {% for filter in filter_type %}
-                                                <button class="activefilter-tag" data-id="{{ filter.id }}">
+                                                <button
+                                                    aria-label="{{__('Remove %s filter', 'planet4-master-theme')|format(filter.name)}}"
+                                                    class="activefilter-tag"
+                                                    data-id="{{ filter.id }}"
+                                                >
                                                     {{ filter.name|e('wp_kses_post')|raw }} <i class="icon-cross">{{ 'times'|svgicon }}</i>
                                                 </button>
                                             {% endfor %}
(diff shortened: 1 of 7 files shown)


templates/search/action_types.twig
```diff
@@ -20,6 +20,7 @@
                             data-ga-category="Search Page"
                             data-ga-action="Content Filter"
                             data-ga-label="{{ action_type.name|e('wp_kses_post')|raw }}"
+                            aria-label="{{__('%s, %d results', 'planet4-master-theme')|format(action_type.name, action_type.results)}}"
```
mardelnet, line 23: I know this change fits the requirements, but, is it clear enough?
I don't see any previous attribute that lets the screen reader explain what these filters are.
Maybe we should try some other messaging such as: `Filter results by the action type "Petitions". 5 results were found.`

templates/search.twig
```diff
@@ -56,7 +56,14 @@
                 </div>
             </div>
             <div class="search-filter-results row clearfix">
-                <div class="col-lg-4">
+                <a
```
mardelnet, line 59: Don't we need an `aria-label` attribute for screen readers here, too?

mleray: No, because in this case the label is clear enough!

templates/search/action_types.twig
mleray: That's a good point 🤔 I'll discuss it with Houssam!

mleray: Houssam approved, code updated!

templates/search/action_types.twig
```diff
@@ -20,6 +20,7 @@
                             data-ga-category="Search Page"
                             data-ga-action="Content Filter"
                             data-ga-label="{{ action_type.name|e('wp_kses_post')|raw }}"
+                            aria-label="{{__('Filter results by action type %s, %d results were found', 'planet4-master-theme')|format(action_type.name, action_type.results)}}"
```
mardelnet, line 23: This is super picky (and optional): we should also consider the text for one single result and use "was" instead of "were".

templates/search.twig
mardelnet: Hmm... is not the `aria-hidden="true"` attribute preventing the screen reader from reading it?

templates/search.twig
```diff
@@ -5,7 +5,7 @@
     {% set collapsed = posts or (not has_aggregation) ? '' : 'collapsed' %}
     {% set show      = posts or (not has_aggregation)? 'show' : '' %}
     {% set expanded  = posts ? true : false %}
-    {% set search_label = __( 'Search', 'planet4-master-theme' ) %}
+    {% set search_label = __( 'Search by name, keyword, or topic', 'planet4-master-theme' ) %}
 
     <div class="search-block">
```
mardelnet, line 10: I wonder whether this element should use the `main` tag as it wraps `aside` and `section`

templates/search.twig
mleray: Hm, that's a good question actually 🤔 I kinda took inspiration from [our existing skip links](https://github.com/greenpeace/planet4-master-theme/blob/4cc988b70fc6b9fbcb3b7dee9a955c25e460ce88/templates/base.twig#L21) and they do use `aria-hidden`

mardelnet: I haven't tested it with an actual screen reader, but I think it should be read just like the rest of the buttons. Maybe this is something to discuss with the design team and expand to the rest of the skip buttons.

mleray: Updated!

templates/search/action_types.twig
mleray: Updated!

templates/search.twig
mleray: I feel like this might be something to consider for our other pages too, I think we don't use `main` anywhere but maybe we should 🤔

mardelnet: I agree. Let's discuss it with the design team and apply it globally in another PR/ticket.

templates/search/action_types.twig
```diff
@@ -10,6 +10,11 @@
         <ul class="list-unstyled">
             {% for id, action_type in action_types %}
                 {% if (action_type.results > 0) or (not has_aggregation) %}
+                {% if content_type.results == 1 %}
```
mardelnet, line 13: I think this condition is not working:

Screenshot-from-2025-03-27-10-41-25

mleray: I'm an idiot, it's checking the content type instead of post type 🤦‍♀️

mleray: Updated!

mardelnet (approved):
Good job!

PR #2587, opened 2025-04-07
Test switching to CircleCI images for testing

comzeradd:


planet-4:

:watch: 2025.04.07 07:57:59

.circleci/config.yml
```diff
@@ -89,26 +97,22 @@ jobs:
 
   frontend-tests:
     docker:
-      - image: greenpeaceinternational/p4-unit-tests:node20.18.0
+      - image: cimg/node:20.18.3
```
GP-Dan-Tovbein, line 100: Do you think if there isn't a problem if we're currently using `20.18.0`?

admin/css/archive-picker.css
GP-Dan-Tovbein, line 1: Is this stylesheet related to the PR?

admin/css/dashboard.css
GP-Dan-Tovbein, line 1: Same here

.circleci/config.yml
comzeradd: I just picked the latest `20.18`. Where do you mean we use `20.18.0`? On the develop environment we just have `20` for nvm, so it should pick the latest minor release available

admin/css/archive-picker.css
comzeradd: Not directly. But without those changes `stylelint` fails. Not sure why it wasn't failing before since those look obvious violations of our stylelint configuration.

.circleci/config.yml
GP-Dan-Tovbein: For example [here](https://github.com/greenpeace/planet4-circleci/blob/main/config.default#L18) but maybe I'm wrong and that is not a problem!

comzeradd: Ah good catch. Minor releases shouldn't contain breaking changes, but just to be safe I switched all references to `20.18.0`

GP-Dan-Tovbein (approved):
Code looks good to me 🧉

PR #2589, opened 2025-04-08
PLANET-7778 Added Actions and Posts List to blocks report

Osong-Michael:
### Summary

---

**Ref: https://jira.greenpeace.org/browse/PLANET-7778**

### Testing

Now when checking the Blocks Report, under the planet4-blocks category, we should have the new Actions and Posts List blocks


planet-4:

:watch: 2025.04.09 11:46:54

src/BlockReportSearch/Block/BlockUsage.php
```diff
@@ -73,6 +73,23 @@ private function get_filtered_blocks(array $posts_ids, Parameters $params): arra
         return $this->blocks;
     }
 
+    /**
+     * Function to filter Query Loop block
+     *
+     * @param array $blocks     array of posts/pages.
```
mleray, line 79: Is this param description accurate? Isn't it an array of blocks?

mleray:
@Osong-Michael the functionality doesn't seem to work for me 🤔 I've added a Posts List to a page, but it doesn't appear in the blocks report

mleray (commented):
It seems to be working well 🎉 Would it be possible to show the block namespace instead of `core/query`?

Screenshot 2025-04-29 at 15 19 16

src/BlockReportSearch/Block/BlockUsage.php
```diff
@@ -123,13 +133,19 @@ private function filter_blocks(Parameters $params): void
...
 
         if (! empty($filters['block_type'])) {
-            $filtered = array_filter(
-                $filtered,
-                function ($i) use ($filters) {
-                    return $i['block_type'] === $filters['block_type']
-                        || $i['local_name'] === $filters['local_name'];
-                }
-            );
+            if (self::POSTS_LIST_NAME === $filters['block_type']) {
```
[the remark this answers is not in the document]
mleray: We could also use `in_array()` instead of the `||` condition

src/BlockReportSearch/Block/BlockUsageApi.php
```diff
@@ -51,6 +51,21 @@ public function get_count(): array
             $this->fetch_items();
         }
 
+        $posts_list_block = $this->usage->filter_query_blocks($this->items, $this->usage::POSTS_LIST_NAME);
+        $actions_list_block = $this->usage->filter_query_blocks($this->items, $this->usage::ACTIONS_LIST_NAME);
+
+        // Merge updated Posts List array with items array for API
+        if (!empty($posts_list_block)) {
```
mleray, line 58: This code is the same as the one below for Actions List, maybe it could be shortened by using a `for` loop?

src/BlockReportSearch/Block/Sql/SqlQuery.php
```diff
@@ -52,12 +52,10 @@ function ($r) {
     private function get_sql_query(Block\Query\Parameters ...$params_list): string
     {
         // Prepare query parameters.
-        $like = [];
```
mleray, line 55: Why did you remove this?

mleray:
@Osong-Michael I'm wondering if we should add the Actions List block only if beta blocks are enabled 🤔

src/BlockReportSearch/Block/Sql/SqlQuery.php
Osong-Michael: So the _Like Class_ is responsible for checking if the `post_content` contains a block that matches a particular type, maybe `core/query` or `planet4-blocks/something`, issue is we don't get back results for the _Actions/Posts List_ blocks since they don't appear in the `post_content`. Since that was the first layer of filtering for `post_ids`, I just removed that line so we always return all the post IDs, then we can later use our own custom filter which we already had, to filter and find the `post_types` with the block being filtered for.

Osong-Michael:
> @Osong-Michael I'm wondering if we should add the Actions List block only if beta blocks are enabled 🤔

🤔 I don't know, maybe @comzeradd can guide us on this, but it is not going to be an issue I think, when the block is not enabled we just return nothing for it in the blocks report anyways

comzeradd:
> it is not going to be an issue I think, when the block is not enabled we just return nothing for it in the blocks report anyways

Yes, I think it's fine to add it already. We can start working on its migration soon enough anyway.

src/BlockReportSearch/Block/BlockUsage.php
```diff
@@ -73,6 +65,23 @@ private function get_filtered_blocks(array $posts_ids, Parameters $params): arra
...
+     */
+    public function filter_query_blocks(array $blocks, string $block_name): array
+    {
+        $found_blocks = array_filter($blocks, function ($post) use ($block_name) {
+            return 'core/query' === $post['block_type']
+                && isset($post['block_attrs']['namespace'])
+                && $block_name === $post['block_attrs']['namespace'];
+        });
+
+        return $found_blocks;
```
mleray, line 82: We can skip creating this variable, and just return the `array_filter` directly

src/BlockReportSearch/Block/BlockUsage.php
```diff
@@ -49,17 +52,6 @@ public function get_blocks(Parameters $params): array
         return $this->get_filtered_blocks($this->posts_ids, $params);
     }
 
-    /**
-     * @param Parameters $params Query parameters.
-     * @return int[]
-     */
-    public function get_posts(Parameters $params): array
```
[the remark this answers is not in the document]
Osong-Michael: The function is never used, I tested it and also had @sagarsdeshmukh test it but we don't seem to be using it anywhere.

mleray (approved):
I've added a small comment but it's not blocking, so I'll let it up to you whether you address it or not! Looks good 👏

PR #2607, opened 2025-04-22
PLANET-7665: Fix visual errors in the Search Results page

mardelnet:
### Summary

This PR implements the following updates to the Search Results page:
- Keep the same spacing in the search form.
- Avoid overlapping the title with the navigation on tablet devices.

---

Ref: https://jira.greenpeace.org/browse/PLANET-7665

### Testing

1. Check [the search results page in the test instance](https://www-dev.greenpeace.org/test-neptune/?s=oceans&orderby=_score).
2. Check that the 2 mentioned changes are applied correctly.
3. Check that the changes look fine on any window size.

### DIFF
--- a/assets/src/scss/pages/search/_search-bar.scss
+++ b/assets/src/scss/pages/search/_search-bar.scss
@@ -58,9 +58,6 @@
 
   .search-btn {
     font-size: $font-size-sm;
-    display: flex;
-    align-content: center;
-    align-items: center;
 
     svg {
       margin-right: 4px;
@@ -69,7 +66,7 @@
     }
 
     @include medium-and-up {
-      margin-left: $sp-3;
+      margin-inline-start: $sp-3;
     }
 
     i {
--- a/templates/search.twig
+++ b/templates/search.twig
@@ -40,7 +40,7 @@
                             </div>
                             <button
                                 type="submit"
-                                class="btn btn-primary search-btn btn-block  mt-2 mt-md-0"
+                                class="btn btn-primary search-btn btn-block d-flex align-items-center align-content-center mt-2 mt-md-0"
                                 data-ga-category="Search Page"
                                 data-ga-action="Search Button"
                                 data-ga-label="n/a" >


templates/search.twig
```diff
@@ -25,32 +25,28 @@
...
+                                    aria-label="{{ __( 'Clear search', 'planet4-master-theme' ) }}"
+                                    type="button"
+                                    onclick="document.getElementById('search-page-input').value=null;"
+                                >
+                                    <span class="visually-hidden">{{ __( 'Clear search', 'planet4-master-theme' ) }}</span>
+                                </button>
                             </div>
+                            <button
+                                type="submit"
+                                class="btn btn-primary search-btn btn-block  mt-2 mt-md-0"
```
mleray, line 43: Do we still need these `mt-2 mt-md-0` classes? It seems that removing them does nothing

assets/src/scss/pages/search/_search-bar.scss
```diff
@@ -62,25 +58,18 @@
...
-      }
-    }
-
-    @include large-and-up {
-      margin-top: 2px;
-
-      svg {
-        margin-right: 4px;
-      }
+      margin-left: $sp-3;
```
mleray, line 72: We should use `margin-inline-start` instead so that it works properly in RTL sites as well

assets/src/scss/pages/search/_search-bar.scss
```diff
@@ -62,25 +58,18 @@
 
   .search-btn {
     font-size: $font-size-sm;
+    display: flex;
```
mleray, line 61: Instead of adding this line to our CSS, we could add the `d-md-flex` class to the button

mleray: I believe there are classes to set the alignment as well!

mardelnet: Fixed!

assets/src/scss/pages/search/_search-bar.scss
mardelnet: Fixed!

mleray (approved):
Looks good to me 👏

PR #2610, opened 2025-04-25
PLANET-7683 Re-design core Quote block

mleray:
### Summary

See [PLANET-7683](https://jira.greenpeace.org/browse/PLANET-7683)

**Requirements:**
- [x] The Quote should occupy the same width as the body text
- [x] The Quote font should remain italic
- [x] Update font sizes, `$font-size-lg: 1.375rem` on small screens, `$font-size-xl: 1.75rem` on large screens
- [x] The Quote text should have the correspondent line spacing to it's font size. -> _I've reduced the line height from 1.4 to 1.2._
- [x] Is it possible to add a big "quotation" symbol at the beginning, just like in the reference? Also, a closing  " " " would have to be added automatically at the end of the quote, if possible.
- [x] I think that the author of the quote can currently be placed at any side, we can keep that.
- [x] Add on the backend sidebar the following recommendation: "Try to keep your Quote short and concise, so that it stands out effectively"
- [x] Fix Quote block editor styles
- [x] Remove core block styles

### Testing

You can check out the block on this [post](https://www-dev.greenpeace.org/test-deimos/press/1113/duis-posuere-6/).

### DIFF
--- a/assets/src/block-editor/BlockFilters/index.js
+++ b/assets/src/block-editor/BlockFilters/index.js
@@ -77,7 +77,8 @@ const addQuoteBlockExtraDescription = () => addFilter('blocks.registerBlockType'
     return settings;
   }
 
-  settings.description += '\n' + __('Try to keep your Quote short and concise, so that it stands out effectively', 'planet4-blocks-backend');
+  const {description} = settings;
+  settings.description = `${description}. ${__('Try to keep your Quote short and concise, so that it stands out effectively.', 'planet4-blocks-backend')}`;
 
   return settings;
 });


assets/src/scss/blocks.scss
```diff
@@ -22,14 +22,15 @@
...
 // Native Blocks
-@import "blocks/core-overrides/Image";
 @import "blocks/core-overrides/BlockSpacing";
-@import "blocks/core-overrides/Separator";
-@import "blocks/core-overrides/Table";
-@import "blocks/core-overrides/MediaText";
 @import "blocks/core-overrides/Columns";
-@import "blocks/core-overrides/Heading";
 @import "blocks/core-overrides/Embed";
+@import "blocks/core-overrides/Heading";
```
mleray, line 28: Just reorganised these in alphabetical order

assets/src/scss/style.scss
```diff
@@ -28,7 +28,6 @@ Text Domain: planet4-master-theme
 @import "base/shame";
 
 // Components
-@import "components/blockquote";
```
mleray, line 31: Since the Quote is a core WordPress block, I thought it made more sense to move this to the `core-overrides` folder

planet-4:

:watch: 2025.04.25 10:09:06

assets/src/scss/blocks/core-overrides/BlockSpacing.scss
```diff
@@ -36,7 +36,6 @@
   }
 }
 
-.wp-block-quote,
```
mleray, line 39: With this condition, the Quote would stick to the left in the editor in larger screens

assets/src/scss/blocks/core-overrides/Quote.scss
```diff
@@ -0,0 +1,37 @@
+.wp-block-quote {
+  font-style: italic;
+  quotes: auto;
+
+  > p,
+  > p.block-editor-rich-text__editable { // This is for the editor.
+    font-size: $font-size-lg;
+    line-height: 1.2;
+
+    &:before {
```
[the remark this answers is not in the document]
mleray: We don't have a token for `5rem`, closest we have is [4.25rem](https://github.com/greenpeace/planet4-master-theme/blob/b99e5aa80ff8ff47c05a317d932550167c42bacb/assets/src/scss/base/_tokens.scss#L26)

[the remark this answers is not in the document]
mleray: I've updated the other font sizes to use tokens!

assets/src/scss/blocks/core-overrides/BlockSpacing.scss
mardelnet, line 39: Is this change necessary?
I've reverted it and see no difference in the block.

assets/src/block-editor/BlockFilters/index.js
```diff
@@ -68,3 +70,14 @@ const addGravityFormsBlockFilter = () => addFilter('blocks.registerBlockType', '
 
   return settings;
 });
+
+// Add more details to Quote block description.
+const addQuoteBlockExtraDescription = () => addFilter('blocks.registerBlockType', 'planet4-blocks/filters/quote', (settings, name) => {
```
mardelnet, line 75: Can we add a line break or some sort of separation?
Otherwise, this sentence gets merged with Julio Cortázar's 🇦🇷  quote.

image

mleray:
> This is out of the PR scope, but I just realized that any block can be placed within the Quotes block. Shouldn't we filter it to only accept a Paragraph block as child?

@mardelnet this was actually a bit of a conversation in the ticket comments, about the headings in particular. Maybe indeed we could look into removing other options, not sure it'll be possible though!

mardelnet (approved):
Well done, @mleray

mardelnet:
> > This is out of the PR scope, but I just realized that any block can be placed within the Quotes block. Shouldn't we filter it to only accept a Paragraph block as child?
> 
> @mardelnet this was actually a bit of a conversation in the ticket comments, about the headings in particular. Maybe indeed we could look into removing other options, not sure it'll be possible though!

Should we create a ticket for this, @mleray?

mleray:
> Should we create a ticket for this, @mleray?

@mardelnet [done](https://jira.greenpeace.org/browse/PLANET-7830)!

GP-Dan-Tovbein:
> > Code looks good to me! However, before to approve I'd prefer to apply tokens into styles.
> 
> @GP-Dan-Tovbein do we want to remove [these variables](https://github.com/greenpeace/planet4-master-theme/blob/main/assets/src/scss/base/_fonts.scss#L17-L27) and use tokens everywhere instead? It's a bit strange to have both 🤔 This would have to be a separate PR though.

I completely agree with that idea! We should use always tokens instead. Before removing we'll need to make a research of how hard is this change, not only within our code, but also any NRO which is currently using these CSS variables !

Thanks Maud ❤️

PR #2611, opened 2025-04-28
PLANET-7720: Use semantic HTML to convey content meaning

mardelnet:
### Summary

This PR introduces the usage of the following HTML tags:
- `main` for the central content of the homepage, posts, and most internal pages.
- `section` for P4 blocks.
- `nav` for navigation menus (both in desktop and mobile) and the navigation breadcrumbs.
- `articles` for some instances of the Posts List block.
- `button` for the Donate buttons.
- The `aria-disabled` attribute is now used for the Comments button.

---

Ref: https://jira.greenpeace.org/browse/PLANET-7720

### Testing

1. Check that the [homepage](https://www-dev.greenpeace.org/test-saturn/) fulfills the ticket requirements.
2. Check that [posts](https://www-dev.greenpeace.org/test-saturn/press/1113/duis-posuere-6/) fulfill the ticket requirements.
3. Check that the [P4 blocks](https://www-dev.greenpeace.org/test-saturn/all-p4-blocks-page/) fulfill the ticket requirements (pay attention to the notes added below).

---

### Notes

1. Only the P4 blocks have been wrapped by the `section` tag. Replacing the tags in the native blocks could be a bit tricky as it would require filtering the block HTML before rendering. Do you think it is worth it to replace it anyway? cc @comzeradd 
2. I could not replace the wrapping tag in the following P4 blocks: Carousel Header, Spreadsheet, and Counter. All of them are wrapped by a `div` with the class `wp-block-planet4-blocks-[BLOCK-NAME]`. Any idea on how to fix that?

### DIFF
--- a/assets/src/blocks/Counter/CounterBlock.js
+++ b/assets/src/blocks/Counter/CounterBlock.js
@@ -54,12 +54,12 @@ export const registerCounterBlock = () => {
     edit: CounterEditor,
     save: props => {
       const markup = renderToString(
-        <div
+        <section
           data-hydrate={BLOCK_NAME}
           data-attributes={JSON.stringify(props.attributes)}
         >
           <CounterFrontend {...props.attributes} />
-        </div>
+        </section>
       );
       return <RawHTML>{markup}</RawHTML>;
     },
--- a/assets/src/blocks/Counter/CounterBlock.js
+++ b/assets/src/blocks/Counter/CounterBlock.js
@@ -54,12 +54,12 @@ export const registerCounterBlock = () => {
     edit: CounterEditor,
     save: props => {
       const markup = renderToString(
-        <section
+        <div
           data-hydrate={BLOCK_NAME}
           data-attributes={JSON.stringify(props.attributes)}
         >
           <CounterFrontend {...props.attributes} />
-        </section>
+        </div>
       );
       return <RawHTML>{markup}</RawHTML>;
     },
--- a/templates/burger-menu.twig
+++ b/templates/burger-menu.twig
@@ -1,6 +1,6 @@
 <div id="nav-main" class="burger-menu d-lg-none">
     <div class="burger-menu-header">
-        <a class="site-logo" href="{{ data_nav_bar.home_url }}">
+        <a class="site-logo" href="{{ data_nav_bar.home_url }}" aria-label="Greenpeace Homepage">
             {% include "blocks/site_logo.twig" %}
         </a>
 
--- a/templates/burger-menu.twig
+++ b/templates/burger-menu.twig
@@ -36,7 +36,7 @@
     </div>
 
     <div class="burger-menu-footer">
-        <a
+        <button
             class="btn btn-donate"
             href="{{ donate_menu_items[0].link }}"
             target="{{ item.target }}"
@@ -44,7 +44,7 @@
             data-ga-action="Donate"
             data-ga-label="{{ page_category }}">
             {{ donate_menu_items[0].title }}
-        </a>
+        </button>
     </div>
 
     <button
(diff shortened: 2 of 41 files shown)


planet-4:

:watch: 2025.04.28 08:28:55

mleray (commented):
I'm not 100% sure, but you mention that some blocks are still rendered within a `div` and I think it might be due to the [render_frontend](https://github.com/greenpeace/planet4-master-theme/blob/main/src/Blocks/BaseBlock.php#L75) and [hydrate_frontend](https://github.com/greenpeace/planet4-master-theme/blob/main/src/Blocks/BaseBlock.php#L234) PHP functions 🤔

templates/burger-menu.twig
```diff
@@ -1,6 +1,6 @@
 <div id="nav-main" class="burger-menu d-lg-none">
     <div class="burger-menu-header">
-        <a class="site-logo" href="{{ data_nav_bar.home_url }}">
+        <a class="site-logo" href="{{ data_nav_bar.home_url }}" aria-label="Greenpeace Homepage">
```
mleray, line 3: Do we need to translate this label?

templates/burger-menu.twig
```diff
@@ -36,15 +36,15 @@
     </div>
 
     <div class="burger-menu-footer">
-        <a
+        <button
             class="btn btn-donate"
             href="{{ donate_menu_items[0].link }}"
```
mleray, line 41: I don't think that buttons support the `href` attribute?

templates/desktop-menu.twig
```diff
@@ -34,14 +34,14 @@
             {% if donate_menu_items is not empty %}
                 {% set donate_menu = donate_menu_items|first %}
                 <li class="nav-donate {{ donate_menu.children is not empty ? 'has-children' : '' }}">
-                    <a
+                    <button
```
mleray, line 37: Same comment as earlier, a button with `href`?

templates/burger-menu.twig
mleray: Yep, just tested this change, nothing happens when you click on the button

templates/desktop-menu.twig
mardelnet: Nice catch, fixed!

templates/burger-menu.twig
mardelnet: Nice catch, fixed!

templates/burger-menu.twig
mardelnet: Fixed, thanks!

assets/src/blocks/Counter/CounterBlock.js
```diff
@@ -54,12 +54,12 @@ export const registerCounterBlock = () => {
     edit: CounterEditor,
     save: props => {
       const markup = renderToString(
-        <div
+        <section
```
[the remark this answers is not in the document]
mardelnet: I've removed this as it doesn't fix the problem mentioned in Note 2 of the PR description.

comzeradd:
> 1. Only the P4 blocks have been wrapped by the `section` tag. Replacing the tags in the native blocks could be a bit tricky as it would require filtering the block HTML before rendering. Do you think it is worth it to replace it anyway?

No, I don't think we should touch the gutenberg blocks. But now I'm thinking that in this case we shouldn't change our blocks either, since that would make the markup inconsistent within the same post if some blocks are `<div>` and some are `<section>`.

Reading [about section](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/section), it looks like it's not really necessary to use it for blocks. Especially since many of our blocks don't necessarily come with a heading.

mardelnet:
> > 1. Only the P4 blocks have been wrapped by the `section` tag. Replacing the tags in the native blocks could be a bit tricky as it would require filtering the block HTML before rendering. Do you think it is worth it to replace it anyway?
> 
> No, I don't think we should touch the gutenberg blocks. But now I'm thinking that in this case we shouldn't change our blocks either, since that would make the markup inconsistent within the same post if some blocks are `<div>` and some are `<section>`.

Thanks, @comzeradd 
I've reverted those changes.

mleray (approved):
Looks good to me 👍

PR #2646, opened 2025-06-12
PLANET-7832: Fix issues in the Posts List and Actions List blocks

mardelnet:
### Summary

This PR fixes the following issues in the Posts List and the Actions List blocks:
- In the carousel layout of the Posts List block, column widths become distorted when a post excerpt contains a very long unbroken word ([example](https://www-dev.greenpeace.org/gutenberg/test-page-6/)).
- Private posts are incorrectly included in the query results.
- Password-protected posts are also being included in the query results.

---

Ref: https://jira.greenpeace.org/browse/PLANET-7832

### Testing

1. Run this branch locally, or use the test instance.
2. Add a very long, unbroken word to a post excerpt.
3. Add that post to a carousel layout Post List block.
4. Check that the width of the Post List block columns is equal.
5. Check that the Posts List block only includes published posts and excludes password-protected or private posts (in both, editor and the frontend).
6. Enable the new IA. Check that the Actions List block only includes published actions and child pages of the Take Action page, and excludes password-protected or private actions/pages (in both, editor and the frontend).
7. Disable the new IA. Check that the Actions List block only includes published child pages of the Act page, and excludes password-protected or private pages (in both, editor and the frontend).

### DIFF
--- a/assets/src/scss/blocks/PostsList/PostsListStyle.scss
+++ b/assets/src/scss/blocks/PostsList/PostsListStyle.scss
@@ -167,6 +167,7 @@
       margin-bottom: $sp-1;
       font-size: var(--font-size-m--font-family-secondary);
       @include clamp-text(3);
+      word-break: break-word;
 
       @include large-and-up {
         line-height: 1.5;
--- a/src/Blocks/QueryLoopExtension.php
+++ b/src/Blocks/QueryLoopExtension.php
@@ -24,7 +24,6 @@ class QueryLoopExtension
     {
         $postInFilter = function ($args, $request) {
             $postIn = $request->get_param('postIn');
-            $hasPassword = $request->get_param('hasPassword');
             $block_name = $request->get_param('block_name');
 
             if ($block_name === self::ACTIONS_LIST_BLOCK) {
@@ -39,7 +38,8 @@ class QueryLoopExtension
                 $args['orderby'] = 'post__in';
             }
             // Exclude posts with the custom property `has_password` set to false
-            if ($hasPassword) {
+            if ($request->has_param('hasPassword')) {
+                $hasPassword = $request->get_param('hasPassword');
                 $args['has_password'] = $hasPassword !== false && $hasPassword !== 'false';
             }
             return $args;
--- a/src/Blocks/QueryLoopExtension.php
+++ b/src/Blocks/QueryLoopExtension.php
@@ -24,6 +24,7 @@ class QueryLoopExtension
     {
         $postInFilter = function ($args, $request) {
             $postIn = $request->get_param('postIn');
+            $hasPassword = $request->get_param('hasPassword');
             $block_name = $request->get_param('block_name');
 
             if ($block_name === self::ACTIONS_LIST_BLOCK) {
@@ -36,6 +37,10 @@ class QueryLoopExtension
             if (!empty($postIn)) {
                 $args['post__in'] = array_map('intval', (array) $postIn);
                 $args['orderby'] = 'post__in';
+            }
+            // Exclude posts with the custom property `has_password` set to false
+            if ($hasPassword) {
+                $args['has_password'] = $hasPassword !== false && $hasPassword !== 'false';
             }
             return $args;
         };
@@ -63,6 +68,10 @@ class QueryLoopExtension
                     $query['post__in'] = array_map('intval', (array) $blockQuery['postIn']);
                     $query['orderby'] = 'post__in';
                     $query['ignore_sticky_posts'] = true;
+                }
+                // Exclude posts with the custom property `has_password` set to false
+                if (isset($blockQuery['hasPassword'])) {
+                    $query['has_password'] = (bool) $blockQuery['hasPassword'];
                 }
                 return $query;
             },
--- a/src/Blocks/QueryLoopExtension.php
+++ b/src/Blocks/QueryLoopExtension.php
@@ -62,6 +62,7 @@ class QueryLoopExtension
                 if (isset($blockQuery['hasPassword'])) {
                     $query['has_password'] = (bool) $blockQuery['hasPassword'];
                 }
+                $query['post_status'] = 'publish'; // Ensure only published posts are queried
                 return $query;
             },
             10,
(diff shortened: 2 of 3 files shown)


planet-4:

:watch: 2025.06.12 12:48:36

src/Blocks/QueryLoopExtension.php
```diff
@@ -80,7 +81,8 @@ function ($query, $block) {
     private static function buildActionListQuery(array $query, array $params = []): array
```
mleray, line 81: This function seems to be for the Actions List block, not Posts List

mardelnet: You're right. Both blocks have issues. I have to update the JIRA ticket and the PR title.

assets/src/scss/blocks/PostsList/PostsListStyle.scss
```diff
@@ -167,6 +167,7 @@
       margin-bottom: $sp-1;
       font-size: var(--font-size-m--font-family-secondary);
       @include clamp-text(3);
+      word-break: break-all;
```
mleray, line 170: I would not recommend using this, because it will also break normal length words randomly (see `like` in the below example)

Screenshot 2025-06-13 at 08 55 20

src/Blocks/QueryLoopExtension.php
```diff
@@ -62,6 +62,7 @@ function ($query, $block) {
                 if (isset($blockQuery['hasPassword'])) {
                     $query['has_password'] = (bool) $blockQuery['hasPassword'];
                 }
+                $query['post_status'] = 'publish'; // Ensure only published posts are queried
```
mleray, line 65: This will impact all Query Loop blocks I think, not only the Actions List and Posts List, which is not what we want

assets/src/scss/blocks/PostsList/PostsListStyle.scss
mleray: Maybe try [word-wrap: break-word](https://www.w3schools.com/cssref/css3_pr_word-wrap.php) instead?

src/Blocks/QueryLoopExtension.php
mardelnet: Isn't that expected, @mleray? I think the Query Loop block only includes published posts.

mleray: Then we shouldn't need to add this parameter?

mardelnet: If I remove it, the front adds private and password-protected posts, but not the editor. I don't know why.

mleray: It's ok to add it, but I don't think we should add it to all Query Loop blocks, only Posts List and Actions List blocks

src/Blocks/QueryLoopExtension.php
```diff
@@ -121,10 +123,6 @@ private static function buildActionListQuery(array $query, array $params = []):
                 $query['post__in'] = array_map('intval', (array) $params['postIn']);
             }
 
-            if (!empty($params['hasPassword'])) {
-                $query['has_password'] = $params['hasPassword'] !== false && $params['hasPassword'] !== 'false';
```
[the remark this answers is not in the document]
mleray: This is something we added ourselves, specifically to know when to remove password-protected posts, so if we handle that differently now it should be redundant?

assets/src/scss/blocks/PostsList/PostsListStyle.scss
mardelnet: Fixed, thanks!

src/Blocks/QueryLoopExtension.php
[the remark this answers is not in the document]
mardelnet: Fixed!

src/Blocks/QueryLoopExtension.php
mardelnet: Done!

mleray:
@mardelnet I now see password-protected posts in the Posts List block again:

Screenshot 2025-06-13 at 11 13 23

mardelnet:
> @mardelnet it works for new blocks actually, so I figured out that this is because existing blocks will not have the new `blockName` attribute... so we would need a migration to apply that change! Do think it's worth it? Or we revert to using `hasPassword` for Posts List blocks?

I've reverted the removal of the `hasPassword` property, @mleray.

src/Blocks/QueryLoopExtension.php
```diff
@@ -23,17 +24,22 @@ public static function registerEditorQuery(): void
...
+                $args['has_password'] = false; // Exclude password-protected posts
+            }
             if (!empty($postIn)) {
                 $args['post__in'] = array_map('intval', (array) $postIn);
                 $args['orderby'] = 'post__in';
             }
-            if ($request->has_param('hasPassword')) {
-                $hasPassword = $request->get_param('hasPassword');
+            // Exclude posts with the custom property `has_password` set to false
+            if ($hasPassword) {
```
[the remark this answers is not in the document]
mardelnet: Good catch, fixed!

src/Blocks/QueryLoopExtension.php
```diff
@@ -23,17 +24,22 @@ public static function registerEditorQuery(): void
...
         $postInFilter = function ($args, $request) {
             $postIn = $request->get_param('postIn');
+            $hasPassword = $request->get_param('hasPassword');
             $block_name = $request->get_param('block_name');
 
             if ($block_name === self::ACTIONS_LIST_BLOCK) {
                 return self::buildActionListQuery($args, $request->get_params());
             }
+            if ($block_name === self::POSTS_LIST_BLOCK) {
+                $args['post_status'] = 'publish'; // Ensure only published posts are queried
```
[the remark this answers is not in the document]
mardelnet: I'm running out of ideas for this.
Option 1: We remove the condition so that all Query Loop blocks only show published posts.
Option 2: We use a migration script to set the `blockName` property to all the Posts List blocks.
Any other ideas?

mardelnet:
Closing this PR in favour of #2649

PR #2651, opened 2025-06-17
PLANET-7832: Exclude current post from query

mardelnet:
### Summary

This PR excludes the current post from the posts query in the Actions List and Posts List blocks.

---

Ref: https://jira.greenpeace.org/browse/PLANET-7832

### Testing

1. Run this branch locally, or use the test instance.
2. Check that the Actions List blocks added to an Action post exclude the current post, both in the editor and the front-end.
3. Check that the Actions List blocks added to a children page of the Act/Take Action page exclude the current page, both in the editor and the front-end.
4. Check that the Posts List blocks added to a post exclude the current post, both in the editor and the front-end.
5. Check that the Related Posts exclude the current post.


mleray (approved):
Works great 👏

PR #2653, opened 2025-06-17
PLANET-7840 Enabled Greenpeace Media on Cover block

Osong-Michael:
### Summary

---

**Ref: https://jira.greenpeace.org/browse/PLANET-7840**

### Testing

GP Media is now enabled for the Cover block. For testing, make sure to enable all blocks in _Planet4 -> Features_, then on the editor, you can add the Cover block and use the GP Media tab to add and update the images.

### DIFF
--- a/admin/css/archive-picker.css
+++ b/admin/css/archive-picker.css
@@ -296,7 +296,7 @@
   justify-content: space-between;
   width: 100%;
   min-height: 20px;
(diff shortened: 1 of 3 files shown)


Osong-Michael:
> admin

Oh, then that is good to know

mleray (approved):
Tested on the instance, works great 👏

PR #2661, opened 2025-06-23
PLANET-7805 Fix mobile navigation tabbing behaviour

mleray:
### Summary

Ref: [PLANET-7805](https://jira.greenpeace.org/browse/PLANET-7805)

- [x] Disable tabs keyboard navigation when mobile tabs are hidden
- [x] Disable menu items keyboard navigation when the mobile menu is closed

### Testing

Try using the navigation bar, mobile tabs, and burger menu on mobile using only the keyboard. It should now behave as expected, meaning that hidden elements shouldn't be accessible.

### DIFF
--- a/assets/src/js/header.js
+++ b/assets/src/js/header.js
@@ -59,14 +59,16 @@ const toggleNavElement = element => {
 
     // Update tab index for keyboard navigation depending on burger menu being open or not.
     const burgerMenu = document.querySelector('.burger-menu');
-    const tabbingItems = [
-      burgerMenu.querySelector('.site-logo'),
-      burgerMenu.querySelector('.btn-donate'),
-      burgerMenu.querySelector('.nav-menu-close'),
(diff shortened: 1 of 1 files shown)


mleray:
Thank you for your review, @mardelnet! I've updated the code according to your suggestions, and I've created a [ticket](https://jira.greenpeace.org/browse/PLANET-7849) for the refactor (feel free to edit it for more information if needed)

mardelnet:
> > @mleray, I noticed something weird happening when a user tabs further than the second navigation menu and then goes back. The focus turns red and marks an empty space:
> 
> @mardelnet which browser are you using? On Chrome I cannot reproduce 🤔

@mleray, it happens with Firefox only when the width of the desktop version goes under 990px, but not when using the Responsive Design Mode (an actual mobile device). It looks like a very edge case, maybe we can ignore it. What do you think?

mardelnet (approved):

PR #2662, opened 2025-06-23
Replace a fallback solution for loading a language files

sagarsdeshmukh:
From wp6.7 the translation files path location change to 'wp-content/languages/themes/' for now we have keep language files in theme language folder and use load_textdomain() to load translations

### Summary

---

[Ref: ](https://wordpress.org/support/topic/translations-are-no-longer-loading-from-the-theme-folder/#post-18132409)

### Testing

TBD


comzeradd (approved):
Tested locally and it does fix the issue 👍🏻

PR #2666, opened 2025-06-25
PLANET-7851 Add a migration script which restore a previous migration…

sagarsdeshmukh:
… script changes

Ref. https://jira.greenpeace.org/browse/PLANET-7851

### Summary

---

Ref: 

### Testing

1. 
2. 
3.


src/Migrations/M052RollbackToPreviousRevision.php
```diff
@@ -0,0 +1,95 @@
...
+        $num_revision_rollback = 0;
+
+        foreach ($revisions as $revision) {
+            // Check if previous revision existed and has post_author 0.
+            if ($prev_revision && 0 == $prev_revision->post_author) { // phpcs:ignore
+                $rollback_to_revision = $revision->ID;
+                break;
+            }
+            $prev_revision = $revision;
+            $num_revision_rollback++;
```
sagarsdeshmukh, line 82: Using this count, we know how many revisions are rolled back.

mardelnet:
I am hesitant with this implementation, @greenpeace/planet-4-developers 
While at first glance the PR's code seems safe and well-made, there are cases besides our previous migration where WordPress creates posts or revisions with `post_author = 0` (such as plugins or the usage of the `wp_update_post` function), and we could be restoring them unwillingly.
Wouldn't it be easier to try some kind of NROs-oriented solution? Is there a way to target our migrations to fix the specific character problems in specific websites?

mardelnet:
@sagarsdeshmukh 
If extra revisions were added after the targeted one, those will be deleted, right?
Maybe editors have already changed the content and fixed the character issues, and we'll be reverting those changes.
Is there a way to identify those cases and check them before reverting to a previous state?

mardelnet:
> > If extra revisions were added after the targeted one, those will be deleted, right? Maybe editors have already changed the content and fixed the character issues, and we'll be reverting those changes. Is there a way to identify those cases and check them before reverting to a previous state?
> 
> But in those cases the revisions would have an post_author, right?

Yes, but if I understood the code correctly, when the targeted revision is found, this one replaces the current post content, overwriting the changes added after the targeted date. Am I wrong, @sagarsdeshmukh?

sagarsdeshmukh:
> @sagarsdeshmukh If extra revisions were added after the targeted one, those will be deleted, right? Maybe editors have already changed the content and fixed the character issues, and we'll be reverting those changes. Is there a way to identify those cases and check them before reverting to a previous state?


Maybe my wording in some places caused confusion(like the word rollback) — just to clarify, we are not rolling back the post revision/s. Instead, we are restoring a specific revision using the [wp_restore_post_revision()](https://developer.wordpress.org/reference/functions/wp_restore_post_revision/) wp function.

In cases where the editor has made changes to the post after the targeted revision, we can track them in migration log output. For example, if the log message is:
"Post revision restored successfully: ID 37096 revision ID: 37131 count: 1"
Here if the count is > 1, it indicates that the post was edited after the targeted revision.

So as we are restoring a revision. Hence the all changes made by editor after targeted revision are also stays in revision history.

sagarsdeshmukh:
> > > > > If extra revisions were added after the targeted one, those will be deleted, right? Maybe editors have already changed the content and fixed the character issues, and we'll be reverting those changes. Is there a way to identify those cases and check them before reverting to a previous state?
> > > > 
> > > > 
> > > > But in those cases the revisions would have an post_author, right?
> > > 
> > > 
> > > Yes, but if I understood the code correctly, when the targeted revision is found, this one replaces the current post content, overwriting the changes added after the targeted date. Am I wrong, @sagarsdeshmukh?
> > 
> > 
> > You are right @mardelnet . We are restoring a revision, so the restored version(the version before the targeted revision) will be the current post content.
> 
> Can we then skip the restoration in cases where the targeted revision is not the last one?
> 
> Just thinking out loud, but if an editor has already made changes on a post they may have already fix any issue. Even if they didn't, it seems like an acceptable trade off to fix those posts manually than altering live content.

Sounds good to me 👍 We can also log the skipped posts for tracking.

PR #2673, opened 2025-07-04
PLANET-7823: Enable Wordpress builtin Details block

mardelnet:
### Summary

This PR:
- Enables the native Details block.
- Extends the native Details block with extra styles in order to mimic the Accordion block visuals.

image

---

### Notes

- While the `AccordionStyle.scss` file could look very different (and thus hard to review), I haven't applied any style changes. I just rearranged the styles using `mixins` in order to apply them to both the Accordion and the Details block without code duplication.
- As the Details block has no wrapper for the inner content, I added a custom class named `p4-details-group` that can be added to a Group block and thus mimic the layout of the Accordion block.

---

Ref: https://jira.greenpeace.org/browse/PLANET-7823

### Testing

1. In your local or test instance, create a new page or edit an existing one.
2. Add a Details block.
3. Check if the block works as expected.
4. Change the styles of the Details block from the editor, using the "Block Styles" controls.
5. Compare the custom styles of the Details block with the Accordion block; they should look the same.

### DIFF
--- a/assets/src/scss/blocks/Accordion/AccordionStyle.scss
+++ b/assets/src/scss/blocks/Accordion/AccordionStyle.scss
@@ -200,10 +200,10 @@
   }
 
   &.p4-details--light summary {
-      @include headline-light;
+    @include headline-light;
   }
 
   &.p4-details--outline summary {
-      @include headline-outline;
+    @include headline-outline;
   }
 }


sagarsdeshmukh (approved):
Great work, @mardelnet  👏
I’ve tested the PR on both local and the test instance — everything looks good to me. 👍
Apologies for the delay in reviewing it!

PR #2675, opened 2025-07-07
Skip the M051 migration from execution

sagarsdeshmukh:
### Summary

- Remove M051ReplaceSpecialCharactersInPostsContent migration
- Add migration status page
---

### Testing

- Navigate to the newly added [Migration Status](https://www-dev.greenpeace.org/test-phoebe/wp-admin/admin.php?page=planet4_migration_status) page in the admin panel.
- Review the displayed table and its fields.
- Please share any feedback, suggestions, or requests for changes to the table structure or field content.


planet-4:

:watch: 2025.07.07 06:38:18

comzeradd:
I guess we can also just delete all the code related to `M051`, right?

src/MigrationStatus.php
```diff
@@ -0,0 +1,132 @@
...
+            [ $this, 'admin_page_display' ]
+        );
+    }
+
+    /**
+     * Admin page markup. Mostly handled by CMB2.
+     */
+    public function admin_page_display(): void
+    {
+        echo '<h1>P4 Migration Status</h1>' . "\n";
```
comzeradd, line 63: Let's use `<h2>` here, same as other pages in that section. I think "Development" is the only other section that has `h1`. We should probably change that one too.

src/MigrationStatus.php
```diff
@@ -0,0 +1,132 @@
...
+            echo '<tr>
+                    <td>' . $key . '</td>
+                    <td>' . str_replace('P4\MasterTheme\Migrations\\', '', $gp_migration['id']) . '</td>
+                    <td>
+                        <div class="datetime">' . $gp_migration['start_time']->date . '</div>
+                        <div class="object-property">' . $gp_migration['start_time']->timezone . '</div>
+                    </td>
+                    <td>
+                        <div class="datetime">' . $gp_migration['end_time']->date . '</div>
+                        <div class="object-property">' . $gp_migration['end_time']->timezone . '</div>
```
mardelnet, line 101: Can we place this in a single `div` so there's no line break?

src/MigrationStatus.php
```diff
@@ -0,0 +1,132 @@
...
+            $migration_status = $gp_migration['success'] ? 'Success' : 'Failed';
+            $migration_status_class = $gp_migration['success'] ? 'success' : 'error';
+            $migration_logs = implode('\n', $gp_migration['logs']);
+
+            echo '<tr>
+                    <td>' . $key . '</td>
+                    <td>' . str_replace('P4\MasterTheme\Migrations\\', '', $gp_migration['id']) . '</td>
+                    <td>
+                        <div class="datetime">' . $gp_migration['start_time']->date . '</div>
+                        <div class="object-property">' . $gp_migration['start_time']->timezone . '</div>
```
mardelnet, line 97: Can we place this in a single `div` so there's no line break?

src/MigrationStatus.php
```diff
@@ -0,0 +1,132 @@
...
+
+    /**
+     * Admin page markup. Mostly handled by CMB2.
+     */
+    public function admin_page_display(): void
+    {
+        echo '<h2>P4 Migration Status</h2>' . "\n";
+        $gp_migrations = get_option('planet4_migrations');
+
+        if (!$gp_migrations) {
```
mardelnet, line 66: Perhaps, we could check for `gp_migrations` in the `__construct` or the `hooks` function, and do not add the page at all if it's `false`.

src/Loader.php
```diff
@@ -89,6 +89,7 @@ private function load_services(array $services): void
             PostReportController::class,
             Cookies::class,
             DevReport::class,
+            MigrationStatus::class,
```
mardelnet, line 92: Should we move this to the `if (is_admin())` condition on line 105?

src/MigrationStatus.php
```diff
@@ -0,0 +1,132 @@
+<?php
+
+namespace P4\MasterTheme;
```
mardelnet, line 3: This is very nit-picky, and we don't have to do it now, but I think it would be nice to move all the classes that create admin pages to the same folder (something like `src/customAdminPages/[fileName].php`) to gather them all in a single place. 
cc @comzeradd

admin/css/options.css
```diff
@@ -38,3 +38,40 @@
...
+    vertical-align: top;
+  }
+
+  th {
+    background-color: #f2f2f2;
+    width: 150px;
+  }
+
+  .array-item {
+    margin-left: 20px;
```
mardelnet, line 64: I think the margin is not necessary.
Maybe we could add a max-height and `overflow:scroll` in case some logs go too long.

src/MigrationStatus.php
comzeradd: I think just "Data migrations" is fine

PR #2672, opened 2025-07-03
PLANET-7857 Adjust See All link based on taxonomy filters

mleray:
### Summary

**Ref:** [PLANET-7857](https://jira.greenpeace.org/browse/PLANET-7857)

**Related PR:** https://github.com/greenpeace/planet4-master-theme/pull/2677 (migration script for existing blocks)

### Testing

On local or on the test instance, add a Posts List block to a page and update the taxonomy filters in the sidebar. After doing so, check in the editor that the corresponding "see all" link has been updated, save the page and make sure that the custom link also shows in the frontend.

PR #2677, opened 2025-07-07
Add migration to customise See All Stories link

mleray:
### Summary

Ref: [PLANET-7857](https://jira.greenpeace.org/browse/PLANET-7857) 

Related PR: https://github.com/greenpeace/planet4-master-theme/pull/2672

### Testing

On local,
1. Run `npx wp-env run cli wp p4-run-activator`. 
2. Make sure that the "see all" link has been correctly updated, for example on [this page](http://www.planet4.test/get-informed/people/).

If needed, you can first run `npm run db:reset` to make sure that you are using the default database.


planet-4:

:watch: 2025.07.07 13:29:21

src/Migrations/M053CustomisePostsListSeeAllLink.php
```diff
@@ -0,0 +1,125 @@
...
+        $post_type_id = isset($filters['p4-page-type']) ? (int) $filters['p4-page-type'][0] : null;
+
+        // Add tag filter
+        if ($tag_id) {
+            $tag = get_term_by('id', $tag_id, 'post_tag');
+            $new_see_all_url .= 'tag=' . $tag->slug . '&';
+        }
+        // Add category filter
+        if ($category_id) {
+            $category = get_category_by_id($category_id);
```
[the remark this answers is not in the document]
mleray: 🤦‍♀️ maybe we should just use `get_term_by` everywhere

src/Migrations/M053CustomisePostsListSeeAllLink.php
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * Update "see all" link to add the taxonomy filters.
+     *
+     * @param array $block - A block data array.
+     * @return array - The transformed block.
+     */
+    private static function transform_block(array $block): array
+    {
+        // Get the News & Stories page url.
+        $news_page_id = (int) get_option('page_for_posts');
```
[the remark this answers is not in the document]
mleray: I've added a function to get the id, let me know what you think! I think a class constant doesn't work in this case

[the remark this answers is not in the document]
mardelnet: By using only a function, we still call `get_option('page_for_posts')` multiple times.
What about something like this, a class variable and a function?

```php 
private static ?int $news_page_id = null;
```

```php 
private static function news_page_id(): int
{
    if (self::$news_page_id === null) {
        self::$news_page_id = (int) get_option('page_for_posts');
    }
    return self::$news_page_id;
}
```

src/Migrations/M053CustomisePostsListSeeAllLink.php
```diff
@@ -0,0 +1,136 @@
...
+            $new_see_all_url .= 'category=' . $category->slug . '&';
+        }
+        // Add post type filter
+        if ($post_type_id) {
+            $post_type = get_term_by('id', $post_type_id, 'p4-page-type');
+            $new_see_all_url .= 'post-type=' . $post_type->slug;
+        }
+
+        // Update the "see all" link in inner blocks.
+        foreach ($block['innerBlocks'] as &$innerBlock) {
```
mardelnet, line 112: I think we'll need a recursive function in order to reach blocks within other blocks. Otherwise, in cases such as the Carousel layout, not all the links are updated:
Edit-Page-“About-us”-‹-Greenpeace-—-WordPress

src/Migrations/M053CustomisePostsListSeeAllLink.php
```diff
@@ -0,0 +1,136 @@
...
+     * Check whether a block is a Posts List block with taxonomy filters.
+     *
+     * @param array $block - A block data array.
+     */
+    private static function check_is_valid_block(array $block): bool
+    {
+        // Check if the News & Stories page is defined in the settings.
+        if (!self::news_page_id()) {
+            return false;
+        }
```
Osong-Michael, line 49: 🤔 this is just for reflection, but don't we already check if the News and Stories page exist before showing this block on the frontend? Meaning if the class `see-all-link` is present we can assume the _News and Stories_ page exists too and just update the URL from the `core/navigation-link` block attributes without needing to check this right?

mleray: Yes, we do! But this allows us to abort the migration much quicker, rather than look through all inner blocks if there is a navigation link or not

src/Migrations/M053CustomisePostsListSeeAllLink.php
mleray: Updated!

src/Migrations/M053CustomisePostsListSeeAllLink.php
```diff
@@ -0,0 +1,152 @@
...
+     * @param string $new_link - The new url value.
+     */
+    private static function update_see_all_links(array &$blocks, string $new_link): void
+    {
+        foreach ($blocks as &$block) {
+            if (
+                $block['blockName'] === Utils\Constants::BLOCK_NAV_LINK &&
+                $block['attrs']['className'] === 'see-all-link'
+            ) {
+                $block['attrs']['url'] = $new_link;
```
Osong-Michael, line 142: So I have a question here, because the last time this was brought up, I was DoD and suggested editors could edit the _See All_ link manually and a URL to a custom page, assuming some editors have done that, do we need to check first? Ccing @comzeradd for clarity too.

comzeradd: Yes, but unfortunately we can't fit all use cases in the migration. So the ones who have edited the link to something completely different, we will send them a heads up to let me know that their changes will be overridden. See [this discussion](https://greenpeace-gpi.slack.com/archives/G015K63081W/p1751623439142169)

PR #2689, opened 2025-07-21
PLANET-7849 Refactor header.js file

mleray:
### Summary

Ref: [PLANET-7849](https://jira.greenpeace.org/browse/PLANET-7849)

The code is still the same, it was just moved around, some functions were created. Only notable change is that I removed mentions of the `navbar-search-toggle` class because I couldn't find it used anywhere else in our code 🤷‍♀️ 

### Testing

The navigation bar and menu should still work as expected 🤞


planet-4:

:watch: 2025.07.21 09:16:23

assets/src/js/header/setupMobileTabs.js
```diff
@@ -0,0 +1,97 @@
...
+  try {
+    const opts = Object.defineProperty({}, 'passive', {
+      get: () => {
+        supportsPassive = true;
+        return supportsPassive;
+      },
+    });
+    window.addEventListener('testPassive', null, opts);
+    window.removeEventListener('testPassive', null, opts);
+  } catch (e) {} // eslint-disable-line no-empty
```
mardelnet, line 29: Can we log something here, pass it to Sentry or similar?
If so, we can remove the disabling.

assets/src/js/header.js
mardelnet, line 1: What do you think about moving the remaining logic to another file so we keep this one extra clean (with only imports and function calling)?

assets/src/js/header/setupMobileTabs.js
mleray: I've added a console, log I saw that this is what we do in other similar cases! Maybe we can implement a better way in a follow-up and update all instances?

mleray: I've created a [ticket](https://jira.greenpeace.org/browse/PLANET-7881) for it, feel free to add info!

assets/src/js/header.js
```diff
@@ -1,239 +1,23 @@
...
-    updateGaAction(countryDropdownToggle, 'Country Selector');
-  }
-
-  if (countrySelectorToggle) {
-    updateGaAction(countrySelectorToggle, 'Country Selector');
-  }
-
-  if (navbarSearchToggle) {
-    updateGaAction(navbarSearchToggle, 'Search');
-  }
```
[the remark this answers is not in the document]
mleray: No, because it was about the `navbar-search-toggle` class that I removed!

assets/src/js/header/setupGaAction.js
```diff
@@ -0,0 +1,34 @@
...
+ * Update data-ga-action attribute used in GTM tracking for various elements.
+ */
+export const toggleGaActionAttribute = () => {
+  const countryDropdownToggle = document.querySelector('.country-dropdown-toggle');
+  const countrySelectorToggle = document.querySelector('.country-selector-toggle');
+  const navMenuToggle = document.querySelector('.nav-menu-toggle');
+
+  if (!countryDropdownToggle && !countrySelectorToggle && !navMenuToggle) {
+    return;
+  }
```
mardelnet, line 21: I think this condition is not necessary as we're checking the elements again down below.

mleray: My idea was to exit the function early in case none of these are present

PR #2690, opened 2025-07-21
Move metabox search functionalities out of MasterSite.php

mleray:
### Summary

This is part of our efforts to make this file shorter 🩳 

### Testing

This box should still be present in the sidebar for Posts and Pages, and work as expected:

Screenshot 2025-07-21 at 11 53 14

### DIFF
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -123,8 +123,6 @@ class MasterSite extends TimberSite
         add_action('wp_enqueue_scripts', [PublicAssets::class, 'enqueue_js']);
         add_filter('safe_style_css', [$this, 'set_custom_allowed_css_properties']);
         add_filter('wp_kses_allowed_html', [$this, 'set_custom_allowed_attributes_filter'], 10, 2);
-        add_action('add_meta_boxes', [$this, 'add_meta_box_search'], 10, 2);
-        add_action('save_post', [$this, 'save_meta_box_search'], 10, 2);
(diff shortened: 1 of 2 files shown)


src/Loader.php
```diff
@@ -99,6 +99,7 @@ private function load_services(array $services): void
             BlockSettings::class,
             EnqueueController::class,
             YouTubeHandler::class,
+            MetaBoxSearchHandler::class,
```
mardelnet, line 102: Should we move this within the `is_admin()` condition on line 104?

src/MetaBoxSearchHandler.php
```diff
@@ -0,0 +1,108 @@
...
+     *
+     * @param string  $post_type Post type.
+     * @param WP_Post|WP_Comment $post      Post object.
+     * phpcs:disable SlevomatCodingStandard.Functions.UnusedParameter -- add_meta_boxes callback
+     */
+    public function add_meta_box_search(string $post_type, $post): void
+    {
+        add_meta_box(
+            'meta-box-search',
+            'Search',
```
mardelnet, line 33: I know this is not part of the PR scope, but should we add a description to the metabox?
The title "Search" looks a bit vague to me, and possibly hard to understand for web editors.

src/Loader.php
mleray: Good catch! I've actually reused our already existing `MetaboxRegister` class, let me know what you think

mardelnet (approved):
Well done, @mleray

PR #2699, opened 2025-07-22
Move close nav on document click function to separate file

mleray:
### Summary

This allows us to simplify `header.js` and hopefully makes the code easier to maintain. 

Ref: [PLANET-7849](https://jira.greenpeace.org/browse/PLANET-7849)

### Testing

When the mobile menu is open, clicking outside of it should close it down.

[PLANET-7849]: https://greenpeace-planet4.atlassian.net/browse/PLANET-7849?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


assets/src/js/header/setupDocumentClick.js
mardelnet, line 1: What about renaming the recently introduced file `setupCloseNavMenuButton.js` in a more generic manner and moving this logic there as it is also related to "closing" elements? 
I know we want to have more small and modular files, but if we end up with a million files with 2 functions each it won't be great either!

mleray: Updated!

PR #2705, opened 2025-07-25
PLANET-7608: Fix Quick Links layout issues on small screens

GP-Dan-Tovbein:
Ref: https://greenpeace-planet4.atlassian.net/browse/PLANET-7608

### Summary
The currently layout displays the items fine, however, some times the users don't understand if there are more quick links availble to scroll within small screens. The idea is to display at least a tiny part of the last item visible on viewport to let them know that there are more items available to see.

#### Extra implementations
Convert block to wide and remove the grey spaces in both sides.

Demo page: https://www-dev.greenpeace.org/test-pandora/quick-links-demo/

https://github.com/user-attachments/assets/c99d1f22-3887-4c8d-b7f4-5a43cc1cdf6a

---

Ref: https://jira.greenpeace.org/browse/PLANET-7608

### Testing
Use small sizes `< 576px` (mobile recommended).

### DIFF
--- a/assets/src/block-templates/quick-links/template.js
+++ b/assets/src/block-templates/quick-links/template.js
@@ -26,6 +26,7 @@ const template = ({
   backgroundColor = 'beige-100',
 }) => ([
   ['core/group', {
+    className: 'block',
     align: 'full',
     backgroundColor,
   }, [
--- a/assets/src/block-templates/quick-links/template.js
+++ b/assets/src/block-templates/quick-links/template.js
@@ -26,7 +26,6 @@ const template = ({
   backgroundColor = 'beige-100',
 }) => ([
   ['core/group', {
-    className: 'block',
     align: 'full',
     backgroundColor,
   }, [
@@ -35,7 +34,7 @@ const template = ({
       ['core/heading', {level: 4, placeholder: __('Enter title', 'planet4-blocks-backend'), content: title}],
       ['core/columns', {
         isStackedOnMobile: false,
-        className: 'is-style-mobile-carousel',
+        className: 'is-style-mobile-carousel quick-links',
       },
       [...Array(5).keys()].map(() => category)],
     ]],
--- a/assets/src/scss/blocks/core-overrides/Columns.scss
+++ b/assets/src/scss/blocks/core-overrides/Columns.scss
@@ -35,11 +35,11 @@
       padding-left: $sp-1x;
       padding-right: $sp-1x;
 
-      @media (max-width: 340px) {
+      @media (max-width: 350px) {
         @include apply-item-width(0, 2);
       }
 
-      @media (min-width: 420px) {
+      @media (min-width: 480px) {
         @include apply-item-width(0, 4);
       }
     }
--- a/assets/src/scss/blocks/core-overrides/Columns.scss
+++ b/assets/src/scss/blocks/core-overrides/Columns.scss
@@ -1,3 +1,15 @@
+@mixin apply-item-width($gap, $items-count) {
+  $gap: 0;
+  $total-gap: calc($gap * $items-count);
+  $apply-item-width: calc(100vw / calc($items-count + .5));
+
+  gap: $gap !important;
+
+  .wp-block-column {
+    flex-basis: $apply-item-width;
+  }
+}
+
 .wp-block-columns.is-not-stacked-on-mobile.is-style-mobile-carousel {
   @include large-and-less {
     flex-wrap: nowrap;
@@ -18,30 +30,17 @@
   &.quick-links {
     @include mobile-only {
       @include wide-block;
-
-      $item-width: calc(100vw / 4);
+      @include apply-item-width(0, 3);
 
       padding-left: $sp-1x;
       padding-right: $sp-1x;
 
-      .wp-block-column {
-        flex-basis: $item-width;
-      }
-
-      @media (max-width: 380px) {
-        $item-width: calc(100vw / 3);
-
-        .wp-block-column {
-          flex-basis: $item-width;
-        }
+      @media (max-width: 340px) {
+        @include apply-item-width(0, 2);
       }
 
       @media (min-width: 420px) {
-        $item-width: calc(100vw / 5);
-
-        .wp-block-column {
-          flex-basis: $item-width;
-        }
+        @include apply-item-width(0, 4);
       }
     }
   }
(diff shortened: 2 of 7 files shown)


assets/src/block-templates/quick-links/template.js
```diff
@@ -35,7 +34,7 @@ const template = ({
       ['core/heading', {level: 4, placeholder: __('Enter title', 'planet4-blocks-backend'), content: title}],
       ['core/columns', {
         isStackedOnMobile: false,
-        className: 'is-style-mobile-carousel',
+        className: 'is-style-mobile-carousel quick-links',
```
mardelnet, line 37: @GP-Dan-Tovbein, do you think a migration script will be necessary to add this new class to the existing Quick Links?

assets/src/scss/blocks/core-overrides/Columns.scss
```diff
@@ -14,6 +14,38 @@
     }
   }
 
+  // This styles will be applied to quick links block only
```
mardelnet, line 17:
```suggestion
  //These styles will be applied to the quick links block only
```

assets/src/block-templates/quick-links/template.js
GP-Dan-Tovbein: Yes, we will need it because the class is not updated to the rest of Quick Links. 

FYI, this ticket started as an investigation. If you and the design team agree with the code/ui we could proceed to run the migration.

assets/src/scss/blocks/core-overrides/Columns.scss
```diff
@@ -1,3 +1,15 @@
+@mixin item-width($gap, $items-count) {
+  $gap: 0;
+  $total-gap: calc($gap * $items-count);
+  $item-width: calc(100vw / calc($items-count + .5));
+
+  gap: $gap !important;
```
GP-Dan-Tovbein, line 6: Force to overrides it

assets/src/scss/blocks/core-overrides/Columns.scss
```diff
@@ -1,3 +1,15 @@
+@mixin apply-item-width($gap, $items-count) {
```
GP-Dan-Tovbein, line 1: Crated a mixin since the logic-style it's used within different media screens

assets/src/block-templates/quick-links/template.js
mardelnet, line 37: Dan, it looks like the class `is-style-mobile-carousel` is only used by the Quick Links template. Perhaps we could add a comment to the SCSS to note that the class is used for Quick Links, despite its name, remove the `quick-links` class you introduced, and avoid the migration altogether.

mardelnet (commented):
The code looks good to me, Dan.
Let's wait for the design team's decision, and in the meantime, check this comment:

assets/src/block-templates/quick-links/template.js
GP-Dan-Tovbein: That class is added when you set as Mobile Carousel to `core/columns`, so, it's not only referred to `Quick Links`, it could be any other `core/columns`.

Screen Shot 2025-07-31 at 18 57 47

src/Migrations/M054UpdateQuickLinksClassName.php
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * Update class name
+     *
+     * @param array $block - A block data array.
+     * @return array - The transformed block.
+     */
+    private static function transform_block(array &$block): array // Using the parameter as a reference to update it
+    {
+        $iterate = function (array &$innerBlocks, string $blockName, array $path = []) use (&$iterate): ?array {
```
GP-Dan-Tovbein, line 63: Use `&$innerBlocks` as a reference and update it directly from loop, instead of copying a value of it

assets/src/scss/blocks/core-overrides/Columns.scss
```diff
@@ -1,3 +1,15 @@
...
+  $gap: 0;
+  $total-gap: calc($gap * $items-count);
+  $apply-item-width: calc(100vw / calc($items-count + .5));
```
GP-Dan-Tovbein, line 4: This will display a half part of the first hidden element

GP-Dan-Tovbein:
@mleray FYI, UAT is already passed

src/Migrations/M056UpdateQuickLinksClassName.php
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * Update class name
+     *
+     * @param array $block - A block data array.
+     * @return array - The transformed block.
+     */
+    private static function transform_block(array &$block): array // Using the parameter as a reference to update it
+    {
+        $iterate = function (array &$innerBlocks, string $blockName, array $path = []) use (&$iterate): ?array {
```
mleray, line 63: Instead of using this `$iterate`, can't we just create a separate function like we do [here](https://github.com/greenpeace/planet4-master-theme/blob/main/src/Migrations/M050AddTagsBackInPostsListBlock.php#L72) for example?

src/Migrations/M056UpdateQuickLinksClassName.php
```diff
@@ -44,7 +42,7 @@ public static function execute(MigrationRecord $record): void
     private static function check_is_valid_block(array $block): bool
     {
        // Check if the block is an array, has a 'blockName' key.
-        if (!is_array($block) || !isset($block['blockName'])) {
+        if (!is_array($block) || !isset($block['blockName']) && !empty($block['innerBlocks'])) {
```
mleray, line 45: Shouldn't it be `||` instead of `&&`?

GP-Dan-Tovbein:
> I added more items in the demo page and now it's unclear on tablets that there are more items available:
> Screenshot 2025-08-20 at 14 11 06

I see the point but I also think it's hard to control what the editors do by adding or remivung items. 
I mean, we need to make sure that what we provide to editors need to work as expected. So, if they want to use only 2 items or 3, we couldn't confirm that the behaviour will be the same, so it's not p4 compatible.

What do you think?

mleray (approved):
Looks good to me, following our Slack conversation with the design team it seems we shouldn't worry about editors adding more items 👍

GP-Dan-Tovbein:
> Looks good to me, following our Slack conversation with the design team it seems we shouldn't worry about editors adding more items 👍

Here is the [thread on Slack](https://greenpeace-gpi.slack.com/archives/G015K63081W/p1755705066517099)

PR #2708, opened 2025-07-31
PLANET-7865 Enable Load More functionality in Actions List

mleray:
### Summary

Ref: [PLANET-7865](https://greenpeace-planet4.atlassian.net/browse/PLANET-7865)

**Requirements**

- [x] Load More Button every n cards, in the Grid Version of the Action List block: after the 6th card for  XXL, XL, L screen sizes, after 4 in M sizes, and after 3 in S.
- [x] Hide the button when there are no more items to show.
- [x] Write migration to add this new button to existing Actions List blocks.

### Testing

You can test the functionality on [this page](https://www-dev.greenpeace.org/test-janus/actions-list-test/) for example.

For the migration, on local you can use the [Take Action page](http://www.planet4.test/take-action/) which already has an Actions List. At first it shouldn't have the load more button, but when running `npx wp-env run cli wp p4-run-activator` you should see it (maybe only when inspecting, it's hidden if there are less than 6 items in the block).

To reset your local content you can use `npm run db:reset`

[PLANET-7865]: https://greenpeace-planet4.atlassian.net/browse/PLANET-7865?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/src/blocks/ActionsList/index.js
+++ b/assets/src/blocks/ActionsList/index.js
@@ -4,7 +4,7 @@ import {TAX_BREADCRUMB_BLOCK_NAME} from '../../block-editor/setupTaxonomyBreadcr
 export const ACTIONS_LIST_BLOCK_NAME = 'planet4-blocks/actions-list';
 
 export const ACTIONS_LIST_LAYOUT_TYPES = [
-  {label: 'Grid', value: 'grid', columnCount: 3, perPage: 100},
+  {label: 'Grid', value: 'grid', columnCount: 3, perPage: 24},
   {label: 'Carousel', value: 'flex', columnCount: 6},
 ];
 
@@ -21,7 +21,7 @@ export const ACTIONS_LIST_BLOCK_ATTRIBUTES = {
   className: 'actions-list p4-query-loop is-custom-layout-grid',
   query: {
     pages: 0,
-    perPage: 100,
+    perPage: 24,
     offset: 0,
     order: 'desc',
     orderBy: 'date',
--- a/assets/src/blocks/ActionsList/index.js
+++ b/assets/src/blocks/ActionsList/index.js
@@ -4,7 +4,7 @@ import {TAX_BREADCRUMB_BLOCK_NAME} from '../../block-editor/setupTaxonomyBreadcr
 export const ACTIONS_LIST_BLOCK_NAME = 'planet4-blocks/actions-list';
 
 export const ACTIONS_LIST_LAYOUT_TYPES = [
-  {label: 'Grid', value: 'grid', columnCount: 3, perPage: 24},
+  {label: 'Grid', value: 'grid', columnCount: 3, maxPosts: 24},
   {label: 'Carousel', value: 'flex', columnCount: 6},
 ];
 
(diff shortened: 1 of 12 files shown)


planet-4:

:watch: 2025.07.31 12:22:35

assets/src/js/actions_list_clickable_cards.js
```diff
@@ -5,7 +5,7 @@
  * @function setupClickabelActionsListCards
  */
 
-export const setupClickabelActionsListCards = () => {
+export const setupClickableActionsListCards = () => {
```
mleray, line 8: Just a small typo fix

assets/src/blocks/ActionsList/index.js
```diff
@@ -21,7 +21,7 @@ export const ACTIONS_LIST_BLOCK_ATTRIBUTES = {
   className: 'actions-list p4-query-loop is-custom-layout-grid',
   query: {
     pages: 0,
-    perPage: 3,
+    perPage: 100,
```
mleray, line 24: @comzeradd in terms of performance (and also maybe design) should make this limit lower?

Osong-Michael, line 24: Hey @mleray, a bit confused about this. So for example, if I set the value of `perPage` to 3 or 6 in the sidebar settings, I don't get the Load More button, but when its higher like 10 or above then I get it, I was thinking editors could still determine the number of actions to show per page then the load more button takes that number into account and adds n number of actions to the page when clicked.

[Edit Page “Test new Actions” ‹ Greenpeace Middle East and North Africa — WordPress.webm](https://github.com/user-attachments/assets/3b78d09a-f8b0-4bd8-9a47-045a9b418942)

Osong-Michael: So what if an editor still wants to have just the 3 actions per page on initial load, but load more actions, 3 at a time, when the load more is clicked, they won't get the load more button, is that what we want?

src/Migrations/M056ActionsListBlockLoadMore.php
```diff
@@ -0,0 +1,125 @@
...
+            10 => '
+        ',
+            11 => null,
+            12 => '</div>
+        ',
+        );
+
+        // Change "perPage" query attribute to 100 if layout is grid.
+        if (str_contains($block['attrs']['className'], 'is-custom-layout-grid')) {
+            $block['attrs']['query']['perPage'] = 100;
```
Osong-Michael, line 120: Shouldn't we do this only if `perPage` >= 10?  Since we offer the load more button only after n number of posts

mleray: I just don't know how we would spot these specific cases?

mleray: According to [this Slack thread](https://greenpeace-gpi.slack.com/archives/C0160MX64AG/p1756114433775519), yes we should update `perPage` in the migration. But it will now be 24 instead of 100.

PR #2726, opened 2025-08-21
Adjust ES sync cron to work for multilingual sites

comzeradd:
### Summary

If multilingual, iterate through active languages to run a full index sync for each one of them

### DIFF
--- a/src/CronJob.php
+++ b/src/CronJob.php
@@ -16,6 +16,25 @@ class CronJob
     }
 
     /**
+     * Helper function to perform the ElasticPress full index.
+     */
+    private function sync_helper(): void
+    {
+        try {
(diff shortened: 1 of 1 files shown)


src/CronJob.php
```diff
@@ -33,17 +33,55 @@ public function p4_full_es_sync(): void
...
+                    return $lang_index;
+                });
+
+                try {
+                    \ElasticPress\IndexHelper::factory()->full_index([
+                        'put_mapping' => true,
+                        'method' => 'dashboard',
+                        'network_wide' => false,
+                        'show_errors' => false,
+                        'trigger' => 'manual_' . $lang_code,
```
sagarsdeshmukh, line 64: if we add `_{lang_code}` here it will display as below on `Sync Settings` page.
image

PR #2728, opened 2025-08-22
PLANET-7842: Reorder completed actions

GP-Dan-Tovbein:
### Summary
Re order actions only when they are set as completed.

---

Ref: https://greenpeace-planet4.atlassian.net/browse/PLANET-7842

### Testing

1. Create a new resistance hub page
2. Add at least 3 actions into it (manual override)
3. Visit at least the last one

Expected behaviour:
The last one might be displayed first and the rest 2 following the same order


mleray (commented):
Looks good to me, but let's wait for UAT 👀

assets/src/js/resistance-hub-campaign.js
```diff
@@ -1,5 +1,22 @@
 const {__} = wp.i18n;
 
+// Re-order completed posts in a actions list block into resistance hub.
+const reorderCompletedPosts = () => {
+  const actionsList = document.querySelector('.actions-list .wp-block-post-template');
```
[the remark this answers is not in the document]
GP-Dan-Tovbein: Yep! That sounds good!

assets/src/js/resistance-hub-campaign.js
```diff
@@ -1,5 +1,21 @@
 const {__} = wp.i18n;
 
+// Re-order completed posts in a actions list block into resistance hub.
+const reorderCompletedPosts = () => {
+  const posts = document.querySelector('.actions-list .wp-block-post-template > li');
```
mleray, line 5: I think that you need `querySelectorAll`

assets/src/js/resistance-hub-campaign.js
```diff
@@ -1,5 +1,21 @@
 const {__} = wp.i18n;
 
+// Re-order completed posts in a actions list block into resistance hub.
+const reorderCompletedPosts = () => {
+  const posts = document.querySelectorAll('.actions-list .wp-block-post-template > li');
```
mleray, line 5: Actually I think you need to iterate it, like so `[...document.querySelectorAll('.actions-list .wp-block-post-template > li');]` because right now there is an error

assets/src/js/resistance-hub-campaign.js
```diff
@@ -1,5 +1,21 @@
...
+const reorderCompletedPosts = () => {
+  const posts = [...document.querySelectorAll('.actions-list .wp-block-post-template > li')];
+  if (posts.length === 0) {
+    return;
+  }
+
+  const completed = posts.filter(post => post.classList.contains('completed'));
+  const notCompleted = posts.filter(post => !post.classList.contains('completed'));
+  if (completed.length > 0) {
+    posts.innerHTML = '';
```
mardelnet, line 13: I don't know if in this case it should be considered or not, but resetting `innerHTML` also removes the associated event listeners. If we need to keep them (maybe there are not any now, but it could change in the future), we could use something like:
``` js
const container = document.querySelector('.actions-list .wp-block-post-template');

while (container.firstChild) {
      container.removeChild(container.firstChild);
}
```

assets/src/js/resistance-hub-campaign.js
```diff
@@ -1,5 +1,14 @@
 const {__} = wp.i18n;
 
+// Reorder completed posts in the Actions List block on the Resistance Hub template.
+const reorderCompletedPosts = () => {
+  [...document.querySelectorAll('.actions-list .wp-block-post-template > li:not(.completed)')]
+    .reverse()
+    .forEach(node => {
+      node.parentNode.insertBefore(node, node.parentNode.firstChild);
```
GP-Dan-Tovbein, line 8: Using `insertBefore` instead of `innerHTML` to update the original parent element

mardelnet (approved):
Looks good to me, I'd only suggest checking whether `node.parentNode` exists.

GP-Dan-Tovbein:
> Looks good to me, I'd only suggest checking whether `node.parentNode` exists.

If your suggestions is something like this:
```js
document.querySelectorAll('.actions-list .wp-block-post-template > li:not(.completed)')]
    .reverse()
    .forEach(node => {
      if(node.parentNode) {
        node.parentNode.insertBefore(node, node.parentNode.firstChild);
      }
    });
```

I don't think if it's neccesary since you are already iterating the `parentNode`:
```js
document.querySelectorAll('.actions-list .wp-block-post-template > li:not(.completed)')]
    .reverse()
    .forEach(node => {
      node.parentNode.insertBefore(node, node.parentNode.firstChild);
    });
```
where `node.parentNode` refers to `document.querySelector('.actions-list .wp-block-post-template`)

PR #2730, opened 2025-08-28
PLANET-7904 Fix Sentry TypeError in GF form filter

sagarsdeshmukh:
Ref. https://greenpeace-planet4.atlassian.net/browse/PLANET-7904

### Summary

The `gform_pre_render` filter throughs a TypeError due to strict return type of array, now we make it array|bool (use php Union type)

> TypeError
> P4\MasterTheme\GravityFormsExtensions::p4_client_side_gravityforms_prefill(): Return value must be of type array, bool returned

### DIFF
--- a/src/GravityFormsExtensions.php
+++ b/src/GravityFormsExtensions.php
@@ -251,7 +251,7 @@ class GravityFormsExtensions
      * @return array|bool The form setting.
      *
      */
-    public function enqueue_share_buttons(array $form): array|bool
+    public function enqueue_share_buttons($form): array|bool
     {
         if (!is_array($form['confirmations'])) {
             return $form;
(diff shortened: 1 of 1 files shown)


src/GravityFormsExtensions.php
mardelnet, line 1: Can we make the PHP Docs more consistent?

The `enqueue_share_buttons` function adds the type to the docs and also to the parameter itself, but the `p4_client_side_gravityforms_prefill` only to the docs.
The `enqueue_share_buttons` function uses the "or" connector for multiple types, while the `p4_client_side_gravityforms_prefill` uses the type `mixed` instead.

Maybe we can get rid of the type definition in the docs to keep it simpler.

planet-4:

:watch: 2025.09.03 09:47:11

src/GravityFormsExtensions.php
sagarsdeshmukh: If we remove the type definition from the docs, PHPCS will complain with `SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingAnyTypeHint`. 
I’ll make it consistent in both places.

mardelnet (approved):

PR #2731, opened 2025-08-28
PLANET-7894 Prevent unauthenticated users from accessing users endpoint

mleray:
### Summary

Ref: [PLANET-7894 ](https://greenpeace-planet4.atlassian.net/browse/PLANET-7894)

### Testing

Make sure that everything still works as expected, especially the parts where we are displaying author information (eg. author page, author block, listing pages, Posts List block, etc).

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -554,3 +554,14 @@ add_filter(
     2
 );
 // phpcs:enable Generic.Files.LineLength.MaxExceeded
+
+// Prevent unauthenticated users from accessing the /wp-json/wp/v2/users endpoint.
+add_filter('rest_endpoints', function ($endpoints) {
+    if (isset($endpoints['/wp/v2/users'])) {
+        unset($endpoints['/wp/v2/users']);
+    }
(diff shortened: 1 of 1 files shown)


sagarsdeshmukh:
While searching the `/wp/v2/users` endpoint, I found that gravity-form & ACF plugins are using this endpoint in
`gravityforms/includes/settings/config/class-gf-settings-config-admin.php` file on line no.45,  
`advanced-custom-fields-pro/includes/fields/class-acf-field-user.php` file at line no. 637 &
`advanced-custom-fields-pro/includes/fields/class-acf-field.php` at line no 267
Can we add a current_user_can() permission check and allow admin users to access this endpoint?

mleray:
@sagarsdeshmukh updated!

mleray:
> Can we add a current_user_can() permission check and allow admin users to access this endpoint?

@sagarsdeshmukh this check doesn't work in this situation, I'm looking into other solutions

mleray:
We have decided to no longer implement this in the end, since we now have a strong password policy in place it doesn't represent a major security risk.

PR #2734, opened 2025-09-01
PLANET-7649: Fix lightbox not using largest image

mardelnet:
### Summary

This PR fixes the images of the Image block not using their largest size when expanded by the lightbox, but the one selected by the editor.

---

Ref: https://greenpeace-planet4.atlassian.net/browse/PLANET-7649

### Testing

1. In a post or page, add the following blocks: Image (with caption), Image (without caption), Media Text, and Paragraph.
2. Add images to all the blocks mentioned in step 1. In the case of the Paragraph block, add an inline image.
3. Change the size of the image and use one different to of "full" or "original".
4. In the front, click on every image. The lightbox should be fired, and the image should expand to full size regardless of the size chosen on step 3.

### DIFF
--- a/assets/src/blocks/components/Lightbox/Lightbox.js
+++ b/assets/src/blocks/components/Lightbox/Lightbox.js
@@ -37,6 +37,19 @@ export const Lightbox = ({index = 0, isOpen, items, onClose = () => {}}) => {
       photoSwipeOptions
     );
 
+    const getData = (i, galleryItem) => {
+      if (!galleryItem.w || !galleryItem.h) {
+        const img = new Image();
+        img.onload = function () {
+          galleryItem.w = this.width;
+          galleryItem.h = this.height;
+          photoSwipe.updateSize(true);
+        };
+        img.src = galleryItem.src;
+      }
+    };
+
+    photoSwipe.listen('gettingData', getData);
     photoSwipe.listen('destroy', onClose);
     photoSwipe.listen('close', onClose);
     photoSwipe.listen('afterChange', () => setCurrentIndex(photoSwipe.getCurrentIndex()));
@@ -44,7 +57,9 @@ export const Lightbox = ({index = 0, isOpen, items, onClose = () => {}}) => {
     photoSwipe.init();
 
     return () => {
-      photoSwipe.destroy();
+      if (photoSwipe && photoSwipe._listeners) {
+        photoSwipe.destroy();
+      }
     };
   }, [isOpen, items, photoSwipeOptions, onClose]);
 
--- a/assets/src/blocks/components/Lightbox/Lightbox.js
+++ b/assets/src/blocks/components/Lightbox/Lightbox.js
@@ -1,30 +1,27 @@
 import PhotoSwipe from '../../../../../node_modules/photoswipe/dist/photoswipe.js';
 import PhotoSwipeUI_Default from '../../../../../node_modules/photoswipe/dist/photoswipe-ui-default.js';
 
-const {useEffect, useRef, useState} = wp.element;
+const {useEffect, useRef, useState, useMemo} = wp.element;
 
 // `items` should be an array of object with this shape:
 // [{ src, w, h, title }, ...]
 // See: https://photoswipe.com/documentation/getting-started.html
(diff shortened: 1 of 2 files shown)


assets/src/blocks/components/Lightbox/Lightbox.js
```diff
@@ -26,21 +26,19 @@ export const Lightbox = ({index, isOpen, items, onClose = () => {}}) => {
...
-    photoSwipe.listen('gettingData', (index, galleryItem) => {
-      if (galleryItem.w < 1 || galleryItem.h < 1) {
-        const imageSizeHandler = new Image();
-        imageSizeHandler.onload = function() {
-          galleryItem.w = this.width;
-          galleryItem.h = this.height;
-          photoSwipe.updateSize(true);
-        };
-        imageSizeHandler.src = galleryItem.src;
+    const preloadItems = items.map(item => {
```
Osong-Michael, line 29: We could use a derived array instead
```suggestion
    const preloadItems = items.map(item => ({
       src: item.originalSrc?.url || item.src
       w: item.originalSrc?.width || item.w,
       h: item.originalSrc?.height || item.h,
       title: item.title,
       }));
```

Osong-Michael (commented):
It works for me, I added a few notes and suggestions for improvement.

assets/src/blocks/components/Lightbox/setupLightboxForImages.js
```diff
@@ -1,17 +1,74 @@
...
+ * @return {Promise<{url: string, width: number|null, height: number|null}|null>}
+ *   An object containing the image URL and dimensions, or null if not found.
+ */
+const getImageMeta = async imgEl => {
+  const imgId = getImageIdFromClass(imgEl);
+  if (!imgId) { return null; }
+
+  try {
+    const response = await fetch(`/wp-json/wp/v2/media/${imgId}`);
+    if (!response.ok) { throw new Error('Failed to fetch media info'); }
```
GP-Dan-Tovbein, line 33: Thinking loud, do we really need to `throw` en `Error`? What if we log that as an error instead?

assets/src/blocks/components/Lightbox/Lightbox.js
mardelnet: Done, thanks!

Osong-Michael (commented):
Hey @mardelnet just curious, these changes seem to affect the Gallery block. On the test instance if I click on the image, i don't get a preview, but it works on other dev instances, can you look into that?

Screenshot 2025-09-22 at 09 18 01

assets/src/blocks/components/Lightbox/setupLightboxForImages.js
```diff
@@ -1,19 +1,79 @@
...
+        data.source_url; // fallback to original
+
+    const width = data.media_details.width || null;
+    const height = data.media_details.height || null;
+
+    return {url, width, height};
+  }
+
+  // fallback if no media_details
+  return {url: data.source_url, width: null, height: null};
```
GP-Dan-Tovbein, line 53: Is it expected to return height as null?

GP-Dan-Tovbein (dismissed):
Code looks goos to me! Just left some comment that are not blocking the merge. Feel free to update it if you consider necessary.

assets/src/blocks/components/Lightbox/Lightbox.js
```diff
@@ -1,71 +1,55 @@
...
-export const Lightbox = ({index, isOpen, items, onClose = () => {}}) => {
-  let photoSwipeElement = useRef(null);
+export const Lightbox = ({index = 0, isOpen, items, onClose = () => {}}) => {
+  const photoSwipeElement = useRef(null);
+  const [currentIndex, setCurrentIndex] = useState(index);
 
-  const photoSwipeOptions = {
-    index: index || 0,
+  const photoSwipeOptions = useMemo(() => ({
+    index,
```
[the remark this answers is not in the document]
Osong-Michael: I think i suggested this, so we don't have to create the options every time there is a re-render

mardelnet:
Closing this PR in favor of https://github.com/greenpeace/planet4-master-theme/pull/2759, which uses a much simpler implementation.

PR #2736, opened 2025-09-04
PLANET-7907 Strengthen Account Password Policy

Osong-Michael:
### Summary

---

Ref: https://greenpeace-planet4.atlassian.net/browse/PLANET-7907

### Testing

1.  Creating a new user should trigger an error if password does not validate conditions
2. Also triggers errors on user update and password reset.
3. On user creation/update the `Confirm Weak password` checkbox is no longer visible, passwords must meet certain standards to be acceptable.

### DIFF
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -161,6 +161,8 @@ class MasterSite extends TimberSite
                 return $roles;
             }
         );
+        add_action('user_profile_update_errors', [$this, 'validate_password_policy'], 10, 3);
+        add_action('validate_password_reset', [$this, 'validate_password_reset'], 10, 2);
 
         /**
          * Apply wpautop to non-block content.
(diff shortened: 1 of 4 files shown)


planet-4:

:watch: 2025.09.04 09:33:49

src/MasterSite.php
```diff
@@ -1427,4 +1429,90 @@ public function save_global_project_id(int $post_id): void
...
+
+    // @phpcs:disable SlevomatCodingStandard.Functions.UnusedParameter.UnusedParameter
+    /**
+     * Validates password on user creatiion/update.
+     *
+     * @param WP_Error   $errors Error object.
+     * @param bool       $update True/False if user is being updated.
+     * @param \stdClass    $user   User object.
+     */
+    public function validate_password_policy(WP_Error $errors, bool $update, \stdClass $user): void
```
mardelnet, line 1441: Can't we combine this function with the `validate_password_reset` function as their logic is the same?
I know they require different parameters, but I think that we could workaround it using something like this:
`WP_Error $errors, $arg2 = null, $arg3 = null`

src/MasterSite.php
```diff
@@ -1427,4 +1429,90 @@ public function save_global_project_id(int $post_id): void
...
+        $errors->add('pass', $check);
+    }
+
+    /**
+     * Validates password on password reset.
+     *
+     * @param WP_Error    $errors Error object if any.
+     * @param \stdClass     $user   User object.
+     */
+    public function validate_password_reset(WP_Error $errors, \stdClass $user): void
```
mardelnet, line 1461: Is it fine to use `\stdClass` as the type for `$user`?
According to the [docs](https://developer.wordpress.org/reference/hooks/validate_password_reset/), it should be `WP_User|WP_Error`

Osong-Michael:
> There are cases where the password is considered weak by WordPress even when the requirements are fulfilled, for example: `blablablablablaB1`
> 
> image

Yea, that is why there is the checkbox below to allow for weak passwords. I am just following the ticket requirements.

Osong-Michael:
> @Osong-Michael can you do a quick research on WP documentation to see what WP considers weak or medium?

So from what I can see WP themselves don't have any logic in the backend to prevent "weak" passwords. They have[ best practises](https://wordpress.org/documentation/article/password-best-practices/) which they suggest, but the validation for the password is done only on the [frontend](https://github.com/WordPress/WordPress/blob/5.3.2/wp-admin/js/user-profile.js#L31) I think using [zxcvbn](https://github.com/dropbox/zxcvbn) from dropbox. Reason the checkbox for allow weak passwords is present.

comzeradd:
> So from what I can see WP themselves don't have any logic in the backend to prevent "weak" passwords. They have[ best practises](https://wordpress.org/documentation/article/password-best-practices/) which they suggest, but the validation for the password is done only on the [frontend](https://github.com/WordPress/WordPress/blob/5.3.2/wp-admin/js/user-profile.js#L31) I think using [zxcvbn](https://github.com/dropbox/zxcvbn) from dropbox. Reason the checkbox for allow weak passwords is present.

Ok, that makes sense. I think it's fine to keep the visual hint, as the dropbox library just has some rating based on the actual password (regardless of length). But ideally we should remove the checkbox confirmation.

Osong-Michael:
> > So from what I can see WP themselves don't have any logic in the backend to prevent "weak" passwords. They have[ best practises](https://wordpress.org/documentation/article/password-best-practices/) which they suggest, but the validation for the password is done only on the [frontend](https://github.com/WordPress/WordPress/blob/5.3.2/wp-admin/js/user-profile.js#L31) I think using [zxcvbn](https://github.com/dropbox/zxcvbn) from dropbox. Reason the checkbox for allow weak passwords is present.
> 
> Ok, that makes sense. I think it's fine to keep the visual hint, as the dropbox library just has some rating based on the actual password (regardless of length). But ideally we should remove the checkbox confirmation to avoid confusion.

So just for clarity, hide the checkbox, but allow the visual hint and use this current implementation to enforce our polices.

mardelnet (commented):
1. In the user admin page, should we render somewhere the new password rules, so editors can be aware of them?
2. When trying to update an existing user password, sometimes the "Update" button remains disabled even when the password requirements are fulfilled:

image

src/ControlPanel.php
```diff
@@ -301,4 +302,37 @@ public function load_wpml_assets(): void
...
     }
+
+    /**
+     * Hide Checkbox for weak passwords
+     * Checkbox is checked by default
+     */
+    public function hide_weak_password_checkbox(): void
+    {
+        $screen = get_current_screen();
+        if (!in_array($screen->id, ['user', 'user-edit'])) {
```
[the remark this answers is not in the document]
Osong-Michael: I don't think it is necessary, it works too for that, you can test it out.

Osong-Michael:
> 1. In the user admin page, should we render somewhere the new password rules, so editors can be aware of them?
>
We could add the best practise [suggestions](https://wordpress.org/documentation/article/password-best-practices/) from WP in our documentation so they are aware.
> 2. When trying to update an existing user password, sometimes the "Update" button remains disabled even when the password requirements are fulfilled:
> 
WP is still running validation on password length and patterns, since we still want the visual warnings, but I just created a bypass for the password length only so we get the buttons enabled.
> image

mardelnet:
> > 2. When trying to update an existing user password, sometimes the "Update" button remains disabled even when the password requirements are fulfilled:

> WP is still running validation on password length and patterns, since we still want the visual warnings, but I just created a bypass for the password length only so we get the buttons enabled.

Sorry, I don't understand your answer.
On my local, if I try to reset an existing password with a new one that fulfills our requirements (for example, `qwertyuioP12`), the "Update User" button does not get enabled.

mardelnet (dismissed):
Well done!

PR #2618, opened 2025-05-07
PLANET-7712 Carousel Header: enhance keyboard navigation accessibility

GP-Dan-Tovbein:
### Summary

- Re-implement Carousel functionality and refactor code
- Move most of the logic into the slides hook
- ~Import Hammer as an npm dep instead of cdn~ removed in favor of https://github.com/greenpeace/planet4-master-theme/pull/2738
- Add stop pause buttons
- Follow up accessibility requirements
- Implement new layout
- Make strucuture more semantic
- Implement memoize pattern to avoid re-renders

Demo page https://www-dev.greenpeace.org/test-janus/carousel-demo-page/

---

Ref: https://jira.greenpeace.org/browse/PLANET-7712

### Testing

1. Create a new page
2. Insert a Carousel Header
3. Test a11y aspects, ideally using a built-in screen reader

PR #2738, opened 2025-09-08
PLANET-7899: Add integrity Sub-Resource to 3rd party libraries

GP-Dan-Tovbein:
- It is applied to CDN libraries as a base64-encoded SHA-512 hash
- Jquery, Hammer and rellax are the spotted libraries

### Summary

---

Ref: https://greenpeace-planet4.atlassian.net/browse/PLANET-7899

### Testing

1. 
2. 
3.

### DIFF
--- a/src/Blocks/CarouselHeader.php
+++ b/src/Blocks/CarouselHeader.php
@@ -123,6 +123,11 @@ class CarouselHeader extends BaseBlock
                 '2.0.8',
                 true
             );
+            wp_script_add_data(
+                'hammer',
+                'integrity',
+                'sha512-UXumZrZNiOwnTcZSHLOfcTs0aos2MzBWHXOHOuB0J/R44QB0dwY5JgfbvljXcklVf65Gc4El6RjZ+lnwd2az2g=='
+            );
         }
         parent::enqueue_frontend_assets();
     }
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -69,6 +69,11 @@ final class Loader
                     '1.12.1',
                     true
                 );
+                wp_script_add_data(
+                    'rellax',
+                    'integrity',
+                    'sha512-f5HTYZYTDZelxS7LEQYv8ppMHTZ6JJWglzeQmr0CVTS70vJgaJiIO15ALqI7bhsracojbXkezUIL+35UXwwGrQ=='
+                );
             }
         );
     }
--- a/src/Blocks/Gallery.php
+++ b/src/Blocks/Gallery.php
@@ -133,6 +133,11 @@ class Gallery extends BaseBlock
                 '2.0.8',
                 true
             );
+            wp_script_add_data(
+                'hammer',
+                'integrity',
+                'sha512-UXumZrZNiOwnTcZSHLOfcTs0aos2MzBWHXOHOuB0J/R44QB0dwY5JgfbvljXcklVf65Gc4El6RjZ+lnwd2az2g=='
+            );
         }
         parent::enqueue_frontend_assets();
     }
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -1031,6 +1031,11 @@ class MasterSite extends TimberSite
             '3.3.1',
             true
         );
+        wp_script_add_data(
+            'jquery-3',
+            'integrity',
+            'sha512-+NqPlbbtM1QqiK8ZAo4Yrj2c4lNQoGv8P79DPtKzj++l5jnN39rHA/xsqn8zE9l0uSoxaCdrOgFs6yjyfbBxSg=='
+        );
     }
 
     /**


src/Blocks/CarouselHeader.php
```diff
@@ -123,6 +123,11 @@ public static function enqueue_frontend_assets(): void
                 '2.0.8',
                 true
             );
+            wp_script_add_data(
+                'hammer',
```
GP-Dan-Tovbein, line 127: Since Hammer is being used twice (Carousel Header and Gallery), the idea is to install it through `NPM` like we do in [this PR](https://github.com/greenpeace/planet4-master-theme/pull/2618/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R56).

planet-4:

:watch: 2025.09.08 15:27:44

mardelnet (commented):
Nice job, @GP-Dan-Tovbein 
I have a couple of questions:

1. We are trying to reduce the size of some big PHP files and switch them to more specific and easy-to-maintain ones. We already began with the `MasterSite` class and the `functions.php` file. Do you think we should create a separate class/file to enqueue external scripts?
2. Where does the integrity value come from before being encoded?

src/Loader.php
```diff
@@ -59,17 +59,11 @@ private function __construct(array $services)
...
-            static function (): void {
-                wp_enqueue_script(
-                    'rellax',
-                    'https://cdnjs.cloudflare.com/ajax/libs/rellax/1.12.1/rellax.min.js',
-                    [],
-                    '1.12.1',
-                    true
-                );
-            }
+        self::load_external_script_with_integrity(
```
mardelnet, line 62: It looks like this library is only used by the Page Header template.
Can we only load it when the template is used?

mardelnet:
> > 2. Where does the integrity value come from before being encoded?
> > ```
> 
> See the **Testing** section at [**Description**](https://github.com/greenpeace/planet4-master-theme/pull/2738#issue-3394565627). I had shared the [Helper](https://www.srihash.org/).

Thanks, @GP-Dan-Tovbein 
Maybe we can include the link to [this website](load_external_script_with_integrity) as part of the documentation of the `load_external_script_with_integrity` for future reference.

src/Loader.php
mardelnet: I think it's the best in terms of performance and green code, so it should be done.
If you consider that a change out of this PR scope, let's create a ticket to tackle it later.

GP-Dan-Tovbein:
> btw, I think that the best source of truth for the SRI hashes are from the CDN we use directly. For instance, in the [rellax library](https://cdnjs.com/libraries/rellax), the last icon is providing the SRI hash.
> image

Thanks @comzeradd! I'll double check the Integrity value. I took it from [Mozilla](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity#tools_for_generating_sri_hashes)

GP-Dan-Tovbein:
> > btw, I think that the best source of truth for the SRI hashes are from the CDN we use directly. For instance, in the [rellax library](https://cdnjs.com/libraries/rellax), the last icon is providing the SRI hash.
> > image
> 
> Thanks @comzeradd! I'll double check the Integrity value. I took it from [Mozilla](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity#tools_for_generating_sri_hashes)

Just validated and all are the same.

mardelnet (commented):
The code looks good to me, Dan.
Let me know what you decided to do with [this](https://github.com/greenpeace/planet4-master-theme/pull/2738/files#r2332810610), so I can proceed with the approval.

src/Loader.php
```diff
@@ -59,17 +59,11 @@ private function __construct(array $services)
         add_action('init', [self::class, 'add_blocks'], 20);
 
         // Load parallax library for Media & Text block.
```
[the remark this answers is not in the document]
GP-Dan-Tovbein: Yup! Because this is the block that use this dependency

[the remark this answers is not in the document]
GP-Dan-Tovbein: Because of https://github.com/greenpeace/planet4-master-theme/pull/2738#discussion_r2332797310

src/Loader.php
comzeradd: If we bring Hammer into `package.json` then it looks that Rellax is now the only external library that we fetch from a CDN? (at least for frontend)

If so, isn't it simpler to move both of them already to simplify our code?

GP-Dan-Tovbein: > What do you mean by “on demand”? We can still lazy-load only them when a component/block need them

As "on deman" I mean something like this https://github.com/greenpeace/planet4-master-theme/blob/main/src/Blocks/CarouselHeader.php#L118-L126

GP-Dan-Tovbein:
@mardelnet @comzeradd  FYI, I've updated the code installing `npm`instead of `cdn` libraries.

PR #2745, opened 2025-09-12
PLANET-7712: Fix the play/pause button in the new version of the Carousel Header

mardelnet:
### Summary

This PR is based on [this one](https://github.com/greenpeace/planet4-master-theme/pull/2618/files), and it fixes the errors with the play/pause button mentioned [here](https://github.com/greenpeace/planet4-master-theme/pull/2618#issuecomment-3284090791).

---

Ref: https://jira.greenpeace.org/browse/PLANET-7712

### Testing

1. Add a Carousel Header block.
2. Check that the pause/play button works as expected when the autoplay is enabled.
3. Check that the pause/play button is not shown when the autoplay is disabled.


GP-Dan-Tovbein (commented):
I'm testing the dev instance and when you're in the editor and click the pause button it doesn't change its state

PR #2748, opened 2025-09-19
Fix search e2e test

comzeradd:
### Summary

Search test is failing even now that Elasticsearch is back on test instances (see screenshot from the report below). Looking at the test code, it looks like we are testing for the tag name twice, but the first time we are looking at the wrong place.

Not sure why it used to work before. Maybe something changed because if [this commit](https://github.com/greenpeace/planet4-master-theme/commit/56f4f2834a185fc80f428fef089e7db3ed315975#diff-3a54bde3f5a7b056c7bb6cfd1a84b704e2412d26ae66ea39e0de84e7f6783737)? 

Screenshot 2025-09-19 at 12 05 30

### DIFF
--- a/tests/e2e/search.spec.js
+++ b/tests/e2e/search.spec.js
@@ -6,7 +6,6 @@ test.useAdminLoggedIn();
 test('check search works', async ({page, requestUtils}) => {
   const testId = `testsearch-${Math.floor(Math.random() * 10000)}`; //NOSONAR
   const tagName = `Tag ${testId}`;
-  const tagPageTitle = `#Tag ${testId}`;
   const postTitle = `Test Post ${testId}`;
 
   const tag = await requestUtils.rest({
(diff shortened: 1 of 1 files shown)


tests/e2e/search.spec.js
```diff
@@ -44,9 +44,8 @@ test('check search works', async ({page, requestUtils}) => {
     const searchTags = await page.locator('.search-result-item-tag').allInnerTexts();
 
     expect(searchResult).toContain(testId);
-    expect(searchPage).toContain(tagPageTitle);
     expect(searchPage).toContain(postTitle);
-    expect(searchTags).toContain(`#${tagName}`);
+    expect(searchTags).toContain(tagPageTitle);
```
mleray, line 48: Maybe we don't need this variable? It's only called once, can't we directly use `#${tagName}`?

PR #2752, opened 2025-09-24
Remove transparent navigation in screens smaller than large

mleray:
### Summary

Otherwise the search input also has a transparent background and it's not user friendly

Ref: [Slack thread](https://greenpeace-gpi.slack.com/archives/G015K63081W/p1758621515810969)

### Testing

The transparent navigation should no longer be applied when the screen width is lower than 1200px.


Osong-Michael (approved):
👍

PR #2758, opened 2025-09-29
PLANET-7896 Login screen user enumeration prevented

Osong-Michael:
### Summary

When a user tries to login and fails, instead of getting an error message with the user name, we get a generic text about either username or password could be invalid

---

Ref: https://greenpeace-planet4.atlassian.net/browse/PLANET-7896

### Testing

 When logging in on test instance or local, even if the user name is correct and password wrong, we no longer get the error that lets you know the user name is in the DB, now we get a generic text.


planet-4:

:watch: 2025.09.29 09:39:32

src/MasterSite.php
```diff
@@ -355,6 +355,18 @@ function ($classes) {
                 return $classes;
             }
         );
+
+        // Override error messgae for login screen
+        // @phpcs:disable SlevomatCodingStandard.Functions.UnusedParameter.UnusedParameter,
+        add_filter(
```
mardelnet, line 361: Should we move this to the `LoginHandler` class?

Osong-Michael: Thanks for this, I remember creating a ticket to move all Auth code to a a separate class, in that case will update this then

src/LoginHandler.php
```diff
@@ -56,6 +56,18 @@ public function __construct()
         add_filter('login_headerurl', [$this, 'add_login_logo_url']);
         add_filter('login_headertext', [$this, 'add_login_logo_url_title']);
         add_action('login_enqueue_scripts', [$this, 'add_login_stylesheet']);
+
+        // Override error messgae for login screen
+        // @phpcs:disable SlevomatCodingStandard.Functions.UnusedParameter.UnusedParameter,
+        add_filter(
+            'login_errors',
+            function ($error): string {
+                return __('Invalid username or password', 'planet4-master-theme');
```
mardelnet, line 65: We're always displaying the same message no matter what error is actually happening.
We should not display it for cases such as disabled cookies or problems with the reset link.

mardelnet (changes requested):

src/LoginHandler.php
```diff
@@ -183,4 +184,21 @@ public function add_login_stylesheet(): void
...
+
+    /**
+     * Returns new generic error message for login page.
+     *
+     * @param string $error The original login error message generated by WordPress.
+     * @return string Modified error message.
+     */
+    public function custom_login_error(string $error): string
+    {
+        $action = isset($_REQUEST['action']) ? $_REQUEST['action'] : 'login';
```
mardelnet, line 196: Why does `$action` become `login` when `$_REQUEST['action']` is not set?

src/LoginHandler.php
```diff
@@ -183,4 +184,21 @@ public function add_login_stylesheet(): void
...
+
+        if ($action === 'login') {
```
mardelnet, line 198: Can you briefly explain why this ensures that only login-related errors are considered?

Osong-Michael: As explained [here](https://github.com/greenpeace/planet4-master-theme/pull/2758#discussion_r2387548752).

PR #2761, opened 2025-09-30
PLANET-7908 Fix weak authentication lockout threshold

mleray:
### Summary

Ref: [PLANET-7908](https://greenpeace-planet4.atlassian.net/browse/PLANET-7908)

### Testing

1. Try to login with a username (e.g. `admin`) and the wrong password.
2. You should see our default login error "Invalid username or password".
3. Try again 4 more times, always with a wrong password.
4. After the 5th time, you should see the new error message.
5. For 15 minutes after that, you should no longer be able to login with this username, even with the right password.
6. After the 15 minutes, you should again be able to login with this username.

[PLANET-7908]: https://greenpeace-planet4.atlassian.net/browse/PLANET-7908?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/LoginHandler.php
```diff
@@ -53,10 +54,85 @@ public function __construct()
...
+
+    /**
+     * Tracks how many attempts failed at login for a specific username.
+     *
+     * @param string $username The username of the user.
+     */
+    // phpcs:ignore SlevomatCodingStandard.Functions.UnusedParameter
+    public function custom_track_failed_login(string $username): void
+    {
+        $ip = $_SERVER['REMOTE_ADDR'];
```
mardelnet, line 87: @mleray:
I think, but I am not 100% sure, that if users visit the website via a web proxy, `$_SERVER['REMOTE_ADDR']` would be the proxy’s IP instead of the visitor’s real IP.
So, if one user fails and gets blocked, all users behind the same proxy will be blocked, too.
Perhaps, we should use `HTTP_X_FORWARDED_FOR` or `HTTP_CLIENT_IP` to get the original visitor’s IP.
cc: @comzeradd

comzeradd: There is this edge case indeed, but if we want to block per IP. Most proxies or VPNs don’t relay the actual IP user at all, for privacy reasons.

But we are talking about non-GP users, falsely entering credentials multiple times while being behind a proxy/VPN shared by other non-GP users. So really an edge case 😁 We could start with this implementation and if we get reports that is blocking the wrong users we can adjust it to rely on other identifiers (eg. we could use a combination of IP and username).

mleray: If it's an edge case I think it's ok?

PR #2784, opened 2025-10-17
Experimentation Day: reduce and optimize database queries

mardelnet:
### Summary

As part of the Experimentation Day, this PR attempts to reduce and optimize some of the most time-consuming database queries on our websites, according to the information provided by the Query Monitor plugin.

---

### Results

By reducing the calls to the functions `get_post_meta`, `wp_get_nav_menu_items`, `wp_get_object_terms`, and `get_terms`, the number of query calls made by the master theme was reduced from 59 to 48 on the homepage and from 55 to 44 on a single post (about 20% in both cases).

The most time-consuming queries, by far, are the ones related to functions used to get menu items in the `MasterSite` class (for example: `$menu = Timber::get_menu('navigation-bar-menu');`). I tried to reduce the number of calls for these functions, without success.

Homepage Before:
home-before

Homepage After:
home-after

Single Post Before:
post-before

Single Post After:
post-after

### DIFF
--- a/src/Post.php
+++ b/src/Post.php
@@ -44,9 +44,12 @@ class Post extends \Timber\Post
      */
     public array $data_layer;
 
+    public array $post_meta = [];
+
     public static function build(\WP_Post $wp_post): self
     {
         $post = parent::build($wp_post);
+        $post->post_meta = get_post_meta($wp_post->ID);
         $post->set_page_types();
         $post->set_author();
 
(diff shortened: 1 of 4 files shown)


src/Post.php
```diff
@@ -44,9 +44,12 @@ class Post extends \Timber\Post
      */
     public array $data_layer;
 
+    public array $post_meta = [];
+
     public static function build(\WP_Post $wp_post): self
     {
         $post = parent::build($wp_post);
+        $post->post_meta = get_post_meta($wp_post->ID);
```
GP-Dan-Tovbein, line 52: FYI, `get_post_meta` is already cached by Wordpress. Thus, I don't think this is a very important change.

We can refactor the Class? Yes, but following the query performance from DB it does not affect too much.

Here is a very helpful post
https://stackoverflow.com/questions/61661896/does-get-post-meta-makes-a-separate-query-to-the-database

PR #2851, opened 2026-01-14
Planet 8081 fix translations loaded too early

sagarsdeshmukh:
### Summary

- When switching the language to pt_BR, the following notice is triggered:

`Notice: A função _load_textdomain_just_in_time foi chamada <strong>incorretamente</strong>. O carregamento da tradução para o domínio <code>planet4-master-theme-backend</code> foi ativado muito cedo. Isso geralmente é um indicador de que algum código no plugin ou tema está sendo executado muito cedo. As traduções devem ser carregadas na ação <code>init</code> ou mais tarde. Leia como <a href="https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/ ">Depurar o WordPress</a> para mais informações. (Esta mensagem foi adicionada na versão 6.7.0.) in /var/www/html/wp-includes/functions.php on line 6121`

- The notice occurs because translation functions such as __() are being executed before the text domain translations are loaded.

- The issue is caused by the [MasterSite](https://github.com/greenpeace/planet4-master-theme/blob/main/src/Loader.php#L80) and [MediaReplacer](https://github.com/greenpeace/planet4-master-theme/blob/main/src/Loader.php#L109) classes calling the __() function too early in the lifecycle.

---

Ref: 
https://greenpeace-planet4.atlassian.net/browse/PLANET-8081
### Testing

- In the local development environment, install the pt_BR language and activate it
```
npx wp-env run cli wp language core install pt_BR
npx wp-env run cli wp language core activate pt_BR
```
- Check for the above mentioned notice
- With this PR applied, the notice should be fixed and no longer appear

### DIFF
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -98,15 +98,17 @@ class MasterSite extends \Timber\Site
         add_filter('wp_dropdown_users_args', [$this, 'filter_authors'], 10, 1);
         add_filter('http_request_timeout', fn () => 10);
 
-        register_nav_menus(
-            [
-                'navigation-bar-menu' => __('Navigation Bar Menu', 'planet4-master-theme-backend'),
-                'donate-menu' => __('Donate Button', 'planet4-master-theme-backend'),
-                'footer-primary-menu' => __('Footer Primary Menu', 'planet4-master-theme-backend'),
-                'footer-secondary-menu' => __('Footer Secondary Menu', 'planet4-master-theme-backend'),
-                'footer-social-menu' => __('Footer Social Menu', 'planet4-master-theme-backend'),
-            ]
-        );
+        add_action('after_setup_theme', function (): void {
+            register_nav_menus(
+                [
+                    'navigation-bar-menu' => __('Navigation Bar Menu', 'planet4-master-theme-backend'),
+                    'donate-menu' => __('Donate Button', 'planet4-master-theme-backend'),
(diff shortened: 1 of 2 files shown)


mardelnet (commented):
Hey @sagarsdeshmukh, can you please add to the testing instructions the steps to install the pt_BR language and activate it?

mardelnet (commented):
@sagarsdeshmukh, I cannot replicate the error you're trying to fix.
On my local instance, using the `main` branch and the `pt_BR` language, the website behaves normally. 

However, I still can see this translation issue produced by the WP Stateless plugin:
It seems that the issue with the plugin was solved in [version 4.2.1](https://wordpress.org/plugins/wp-stateless/#developers)
cc @comzeradd 

```
Notice: Function _load_textdomain_just_in_time was called incorrectly. Translation loading for the stateless-media domain was triggered too early. This is usually an indicator for some code in the plugin or theme running too early. Translations should be loaded at the init action or later. Please see [Debugging in WordPress](https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/) for more information. (This message was added in version 6.7.0.) in /var/www/html/wp-includes/functions.php on line 6131 Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/wp-includes/functions.php:6131) in /var/www/html/wp-includes/functions.php on line 7182 Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/wp-includes/functions.php:6131) in /var/www/html/wp-admin/admin-header.php on line 14
```

PR #2907, opened 2026-03-02
PLANET-8116 Implement new Timeline block year navigation

mleray:
### Summary

Ref: [PLANET-8116](https://greenpeace-planet4.atlassian.net/browse/PLANET-8116)

**Requirements**
- Follow the [mockups](https://www.figma.com/design/9NtbY8n3at8uOEJTsLrETb/P4-Design-System?node-id=9010-18014) on implementing the years sub-navigation at the top of the block.
- We should only show the years that have at least one item.
- If the amount of years is more that can fit on the screen it should be scrollable horizontally.
- We should use the nav component as the one in the Secondary navigation block.
- Clicking on a year should scroll on the respective year first item.

**Note**: SonarCloud is complaining about duplicated code, which there is a lot of with the Secondary Navigation block. I think it's fine for now but we should try in the future to refactor these 2 so that there isn't so much duplication.

### Testing

You can use this [page](https://www-dev.greenpeace.org/test-phobos/timelineee) I made for UAT.

[PLANET-8116]: https://greenpeace-planet4.atlassian.net/browse/PLANET-8116?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/src/blocks/Timeline/NewTimelineFrontend.js
+++ b/assets/src/blocks/Timeline/NewTimelineFrontend.js
@@ -1,3 +1,5 @@
+import {YearsNavigation} from './YearsNavigation';
+
 const {useState, useEffect} = wp.element;
 const {__, sprintf} = wp.i18n;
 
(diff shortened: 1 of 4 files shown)


Osong-Michael (commented):
On the figma designs the years navigation and the content below are on the same margin left and right, in your implementation the navigation is longer

Image

Osong-Michael (commented):
Also, if you add `scroll-margin-top` of about `100px` to every [div](https://github.com/greenpeace/planet4-master-theme/blob/main/assets/src/blocks/Timeline/NewTimelineFrontend.js#L185) with a year id, it ensures that when a user clicks and it scrolls to that year, the year is visible just below the navigation instead of it scrolling past

Osong-Michael (commented):
Code looks good to me, let me know when UAT passes then I will take one final look and approve the PR

sagarsdeshmukh (approved):
Great job @mleray 👍  Functionality wise the PR looks good to me.

PR #2948, opened 2026-04-10
PLANET-7975: Improve Actions List carousel accessibility

mardelnet:
### Summary

This PR fulfills the following requirements:
- Wrap the Previous/Next controls in a navigation landmark so they are easy to find.
- Add a label name "Actions List carousel controls".
- When activating an arrow, the focus moves to the next slide container. 

---

Ref: https://greenpeace-planet4.atlassian.net/browse/PLANET-7975

### Testing

1. Launch the website locally.
2. Add a new Actions List block with a carousel layout to any page.
3. Check that the carousel arrows are wrapped in a navigation landmark.
4. Check that the navigation landmark contains an ARIA label with the value "Actions List carousel controls".
5. Check that the focus is moved from the arrows to the first carousel item when an arrow is hit.
6. Now check a previously existing Actions List block. Points 3 and 4 should not be fulfilled.
7. Reset the database by running `npm run db:reset` 
8. Run the migration: `npx wp-env run cli wp p4-run-activator` 
9. Check again a previously existing Actions List block. 
10. Points 3, 4, and 5 should be fulfilled now.

### DIFF
--- a/assets/src/blocks/ActionsList/index.js
+++ b/assets/src/blocks/ActionsList/index.js
@@ -35,7 +35,7 @@ const carouselButtons = [
           'core/button',
           {
             className: 'carousel-control-prev',
-            text: __('Previous Carousel Slide', 'planet4-master-theme-backend'),
+            text: __('Previous carousel slide', 'planet4-master-theme-backend'),
             tagName: 'button',
           },
         ],
@@ -43,7 +43,7 @@ const carouselButtons = [
           'core/button',
           {
             className: 'carousel-control-next',
-            text: __('Next Carousel Slide', 'planet4-master-theme-backend'),
+            text: __('Next carousel slide', 'planet4-master-theme-backend'),
             tagName: 'button',
           },
         ],
--- a/assets/src/blocks/ActionsList/index.js
+++ b/assets/src/blocks/ActionsList/index.js
@@ -1,4 +1,4 @@
-import {LISTS_BREADCRUMBS, carouselButtons} from '../PostsList';
+import {LISTS_BREADCRUMBS} from '../PostsList';
 import {TAX_BREADCRUMB_BLOCK_NAME} from '../../block-editor/setupTaxonomyBreadcrumbBlock';
 
 export const ACTIONS_LIST_BLOCK_NAME = 'planet4-blocks/actions-list';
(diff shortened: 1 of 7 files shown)


planet-4:

:watch: 2026.04.29 07:46:45

assets/src/js/query_loop_carousel.js
```diff
@@ -149,22 +149,32 @@ export const setupQueryLoopCarousel = () => {
 
         const carouselButtons = layout.querySelectorAll(`${BUTTONS_CLASS} ${CONTROLS_CLASS}-next, ${BUTTONS_CLASS} ${CONTROLS_CLASS}-prev`);
 
-        // This resets the focus to the ghost element so users can tab over the new slide.
+        // Moves the focus to the next slide when the carousel arrows are hit.
```
mleray: This will impact the Posts List block as well, right? Was this tested?

src/Migrations/Utils/Functions.php
```diff
@@ -631,30 +631,33 @@ public static function create_block_query_no_results(array $inner_blocks, array
...
             isset($attrs['className']) ?
             'wp-block-group ' . $attrs['className'] :
             'wp-block-group';
 
+        $tagname = isset($attrs['tagName']) ? $attrs['tagName'] : 'div';
+
         // IMPORTANT: DO NOT MODIFY THIS FORMAT!
         $inner_html =
-        '<div class="' . $classname . '">
+        '<' . $tagname . ' class="' . $classname . '" aria-label="' . $aria_label . '">
```
mleray: Do we want to add the aria-label even if it's empty? I think in this case we would be better off not adding the attribute at all. I'm not sure how screen readers handle an empty aria-label 🤔

src/Migrations/M063ActionsListCarouselAccessibility.php
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    /**
+     * Removes a class from a piece of html.
+     *
+     * @param string $html - The HTML.
+     * @param string $classToRemove - The class to be removed.
+     * @return string.
+     */
+    private static function remove_class_from_html(string $html, string $classToRemove): string
```
mleray: This seems like a pretty generic function that could be useful in other cases, can we move it somewhere else?

mardelnet: Good idea. It was moved to the `src/Migrations/Utils/Functions.php` file.

src/Migrations/Utils/Functions.php
mardelnet: Nice catch. Fixed!

assets/src/js/query_loop_carousel.js
mardelnet: Excellent catch!
The existing logic was kept intact and wrapped in a conditional to continue affecting the Posts List blocks.
The new logic was wrapped in another conditional to affect only the Actions List blocks.

assets/src/blocks/ActionsList/index.js
```diff
@@ -16,6 +16,42 @@ const ACT_PAGE = window.p4_vars.options.take_action_page || -1;
...
+      {
+        lock: {move: true},
+        layout: {type: 'flex', justifyContent: 'space-between', orientation: 'horizontal', flexWrap: 'nowrap'},
+      },
+      [
+        [
+          'core/button',
+          {
+            className: 'carousel-control-prev',
+            text: __('Previous Carousel Slide', 'planet4-master-theme-backend'),
```
mleray: Should we remove the capital first letters in `Carousel` and `Slide`? To be consistent with our other copy (for example `Actions List carousel controls` just above)

mardelnet: Done!

mleray (approved):
Looks good to me 👏

PR #2950, opened 2026-04-12
Remove duplicate URL in pa11y accessibility config

rushikeshmore:
### Summary

The `urls` array in `pa11y.json` contains the same URL (`https://www-dev.greenpeace.org/`) twice. This causes pa11y-ci to test the identical page redundantly. This PR removes the duplicate entry.

---

Ref: N/A (no ticket — discovered during codebase review)

### Testing

1. Verify `pa11y.json` contains a single URL in the `urls` array
2. No functional change — pa11y-ci will test the same page, just once instead of twice

### DIFF
--- a/pa11y.json
+++ b/pa11y.json
@@ -16,7 +16,7 @@
     "includeNotices": true
   },
   "urls": [
-    "https://www-dev.greenpeace.org/",
-    "https://www-dev.greenpeace.org/"
+    "http://www.planet4.test/",
+    "http://www.planet4.test/?s="
   ]
 }


pa11y.json
```diff
@@ -16,7 +16,6 @@
     "includeNotices": true
   },
   "urls": [
-    "https://www-dev.greenpeace.org/",
```
sagarsdeshmukh, line 19: Hi @rushikeshmore ,  Thank you for your contribution 🎉
Instead of removing this entry, could you please update it similar to the [.pa11yci](https://github.com/greenpeace/planet4-master-theme/blob/main/.pa11yci#L16-L17) file?
For example:
```
"http://www.planet4.test/",
"http://www.planet4.test/?s="
```

rushikeshmore: Good call, thanks for pointing that out! Updated the URLs to match `.pa11yci` -- now using `http://www.planet4.test/` and `http://www.planet4.test/?s=` instead of just removing the duplicate.

comzeradd (approved):

PR #2955, opened 2026-04-16
PLANET-8155: Set Custom as the default query type for the Query Loop

mardelnet:
### Summary

The default Query Type of the native Query Loop block does not render the posts as expected (but it does in the editor). Besides this not being a bug but the normal behavior, it can be misleading for editors. Taking that into consideration, this PR sets "Custom" as the default Query Type of the native Query Loop block.

Screenshot from 2026-04-16 10-10-40

---

Ref: 
- https://greenpeace-planet4.atlassian.net/browse/PLANET-8155
- https://greenpeace-gpi.slack.com/archives/C0160MX64AG/p1776244669542409

### Testing

1. Run the website locally.
2. Switch to the `main` branch and add a new native Query Loop block.
3. Check the default Query Type in the block settings. It should be "Default".
4. Switch to this PR branch and add a new native Query Loop block.
3. Check the default Query Type in the block settings. It should be "Custom".

### DIFF
--- a/src/MasterSite.php
+++ b/src/MasterSite.php
@@ -77,7 +77,6 @@ class MasterSite extends \Timber\Site
             // Disable WP Block Patterns.
             remove_theme_support('core-block-patterns');
         }
-
         add_post_type_support('page', 'excerpt'); // Added excerpt option to pages.
 
         add_filter('timber/context', [$this, 'add_to_context']);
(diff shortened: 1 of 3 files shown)


planet-4:

:watch: 2026.04.20 10:54:48

Osong-Michael (approved):
👍 works as expected

mardelnet:
@comzeradd, do we have a way to identify if NROs are using the Query Loop block, and check whether the frontend is rendered correctly or not?

comzeradd:
> do we have a way to identify if NROs are using the Query Loop block, and check whether the frontend is rendered correctly or not?

We could slightly adjust our block report to get those numbers. Currently it also includes our block variations

PR #2959, opened 2026-04-22
Remove unused wp-scripts security placeholder devDependency

rushikeshmore:
### Summary

Removes `wp-scripts@^0.0.1-security` from `devDependencies`.

The `wp-scripts` package at version `0.0.1-security` is an [npm security placeholder](https://www.npmjs.com/package/wp-scripts) published by the npm security team. It has no `bin`, no dependencies, and contains only a README. The actual `wp-scripts` CLI used by `npm run build`, `start`, and `analyze` is provided by `@wordpress/scripts`, which is already a devDependency.

Before removal: `node_modules/.bin/wp-scripts -> ../@wordpress/scripts/bin/wp-scripts.js`
After removal: same symlink, still functional.

---

Ref: none

### Testing

1. Checkout the branch.
2. `nvm use` (Node 20).
3. `npm ci --legacy-peer-deps` — installs without the placeholder package.
4. `ls -la node_modules/.bin/wp-scripts` — still links to `@wordpress/scripts/bin/wp-scripts.js`.
5. `npm run build` — compiles successfully (84 pre-existing sass/size warnings, no errors).
6. `npm run lint` — 0 errors, 44 pre-existing React hooks warnings unchanged.

### Notes

The `package-lock.json` diff is larger than the one-line removal because npm re-classifies many transitive packages from `devOptional` to `dev` once the placeholder leaves the dependency graph. This is a metadata change only and has no runtime impact.

PR #2960, opened 2026-04-22
Drop unused react-scripts CRA dependency

rushikeshmore:
### Summary

Removes the `react-scripts` dependency. Create React App is [officially deprecated](https://react.dev/blog/2025/02/14/sunsetting-create-react-app), and `react-scripts` is not imported anywhere in this repo.

The one thing it was silently providing was `regenerator-runtime`, which `assets/src/frontendIndex.js` depends on. A comment in that file explicitly noted the reliance on a transitive dep:

```js
// This ESLint error is disabled since 'regenerator-runtime/runtime' has already been added by another package.
// There is no need to explicitly include it in the list of dependencies in the package.json file.
```

This PR:

1. Drops `react-scripts` from `dependencies`.
2. Adds `regenerator-runtime@^0.14.1` as an explicit dependency so `frontendIndex.js` no longer relies on a transitive.
3. Removes the now-incorrect `eslint-disable-next-line import/no-extraneous-dependencies` comment in `frontendIndex.js`.

Net effect on `package-lock.json`: about 6,300 lines removed, because the CRA transitive tree (jest, workbox, eslint-webpack-plugin, etc.) is no longer pulled in.

---

Ref: none

### Testing

1. Checkout the branch.
2. `nvm use` (Node 20).
3. `npm ci --legacy-peer-deps`.
4. `npm run build` — compiles successfully (84 pre-existing sass/size warnings, same as baseline, no errors).
5. `npm run lint` — 0 errors, 44 pre-existing React hooks warnings unchanged.
6. `grep -R "react-scripts" --exclude-dir=node_modules --exclude=package-lock.json .` — no matches, confirming no code references.

### DIFF
--- a/assets/src/frontendIndex.js
+++ b/assets/src/frontendIndex.js
@@ -1,6 +1,3 @@
-// This ESLint error is disabled since 'regenerator-runtime/runtime' has already been added by another package.
-// There is no need to explicitly include it in the list of dependencies in the package.json file.
-// eslint-disable-next-line import/no-extraneous-dependencies
 import 'regenerator-runtime/runtime';
 
 import {setupBlockFrontend} from './blocks/components/BlockFrontend/setupBlockFrontend';
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,7 @@
   "packages": {
     "": {
       "name": "planet4-master-theme",
-      "license": "GPL-3.0-or-later",
+      "license": "GPL-3.0",
       "dependencies": {
         "@popperjs/core": "^2.11.8",
         "@wordpress/edit-post": "^7.28.1",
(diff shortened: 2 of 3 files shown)


rushikeshmore:
Friendly check-in — same shape as #2959, dropping the unused `react-scripts` devDependency. CI green. cc @sagarsdeshmukh @comzeradd

GP-Dan-Tovbein (approved):
Tested locall and works fine. Great job!

comzeradd:
@rushikeshmore can you rebase your branch?

rushikeshmore:
Rebased on main, force-pushed. Lockfile regenerated cleanly (no more react-scripts entries).

PR #2961, opened 2026-04-22
Discover PHPUnit tests using the Test.php suffix convention

rushikeshmore:
### Summary

The PHPUnit suite was configured to discover only files matching the `test-*.php` prefix:

```xml
<directory prefix="test-" suffix=".php">./tests/</directory>
```

Two existing test files do not match that pattern and are silently skipped by CI today:

- `tests/SqlParameters.php` (class `SqlParameters extends \PHPUnit\Framework\TestCase`, one test method)
- `tests/Post/ReadingTimeCalculatorTest.php` (class `ReadingTimeCalculatorTest extends TestCase`, full test class)

Both are pure PHPUnit tests with no WordPress bootstrap dependency, and both cover production code (`P4\MasterTheme\SqlParameters` and `P4\MasterTheme\Post\ReadingTimeCalculator`).

This PR:

1. Adds a second `<directory>` entry to the unit testsuite so files ending in `Test.php` are also discovered.
2. Renames `tests/SqlParameters.php` to `tests/SqlParametersTest.php` so it matches the new suffix rule, and renames the class from `SqlParameters` to `SqlParametersTest`. This also stops the test class from shadowing the production class `P4\MasterTheme\SqlParameters`.

The existing `test-*.php` files continue to be discovered; they sit next to the new rule, not under it.

---

Ref: none

### Testing

1. Checkout the branch.
2. From the repo root, inspect `phpunit.xml.dist` — the `<testsuite name="unit">` block now contains two `<directory>` children.
3. In a working PHP test environment (WP test suite installed), run `vendor/bin/phpunit`. The output should report two more tests than before: `SqlParametersTest::testOrder` and the full `ReadingTimeCalculatorTest` class.
4. `php -l tests/SqlParametersTest.php` — no syntax errors.
5. `xmllint --noout phpunit.xml.dist` — XML remains valid.

### Notes

Local phpcs could not be run because the vendored `wp-coding-standards/wpcs@2.x` is incompatible with recent PHP versions on my machine; CI runs phpcs in a pinned Docker image and should pass as usual. No code-style change is introduced by the rename.


sagarsdeshmukh:
Hi @rushikeshmore, Thanks for catching the silent-skip bug, definitely worth fixing.

Could we keep the existing `test-*.php` convention for consistency with the rest of `tests/` instead of introducing a second discovery rule? i.e. rename `tests/SqlParameters.php` => `tests/test-sql-parameters.php` and `tests/Post/ReadingTimeCalculatorTest.php` => `tests/Post/test-reading-time-calculator.php`, and drop the `phpunit.xml.dist` change. Open to keeping both conventions if there's an explicit reason.

rushikeshmore:
@sagarsdeshmukh good call on consistency. Force-pushed:

- Dropped the `phpunit.xml.dist` change
- Renamed `tests/SqlParameters.php` -> `tests/test-sql-parameters.php` (class kept as `SqlParametersTest` to avoid shadowing the production `P4\MasterTheme\SqlParameters`)
- Renamed `tests/Post/ReadingTimeCalculatorTest.php` -> `tests/Post/test-reading-time-calculator.php`

Both files now match the existing `test-*.php` convention and are discovered by the unit testsuite without any config change.

sagarsdeshmukh (approved):
Nice job @rushikeshmore 👏 
The PR looks good to me 👍

PR #2303, opened 2024-06-04
PLANET-7527 Move blocks report into master theme

sagarsdeshmukh:
Ref. https://jira.greenpeace.org/browse/PLANET-7527

> Requirements
> 
>     Migrate Blocks Report code to master theme.
>     That should also include classic blocks report functionality.
>     Verify that the API endpoints also work with the plugin de-activated.

The following folder name is changed during the migration to the master theme:
[classes/search](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/tree/main/classes/search) => [src/BlockReportSearch ](https://github.com/greenpeace/planet4-master-theme/tree/PLANET-7527-move-block-report/src/BlockReportSearch)

**Testing:**
- You can test this [PR](https://github.com/greenpeace/planet4-master-theme/pull/2303) along with blocks repo [PR](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/1215) on Local environment or on [test-venus](https://www-dev.greenpeace.org/test-venus/wp-admin/?page=plugin_blocks_report) test instance
- ~~De-activate the Blocks plugin, and check all [blocks reports](https://www-dev.greenpeace.org/test-sinope/wp-admin/?page=plugin_blocks_report), It should work as expected~~ The blocks report functionality should work as as expected
- The Plugins [blocks report API](https://www-dev.greenpeace.org/test-venus/wp-json/plugin_blocks/v3/plugin_blocks_report/) should work as expected.

PR #2963, opened 2026-04-22
Correct misleading docblocks in Like SQL class

rushikeshmore:
### Summary

The `Like` class in `src/BlockReportSearch/Block/Sql/Like.php` was added alongside its sibling `Regex` in [PR #2303](https://github.com/greenpeace/planet4-master-theme/pull/2303). Both classes share the same structure, and the docblocks on `Like.php` were copied from `Regex.php` but never updated for the new class. They describe the file as implementing a regex, when in fact the class builds SQL `LIKE` patterns.

Current (incorrect) docblocks:
- File-level: `* Block search regex`
- Class-level: `* Regular expression used in SQL query`

The class uses `%` wildcards and is consumed by a `LIKE` query (compare the two classes — `Regex.php` uses `.*`, `[^/]*`, regex anchors; `Like.php` uses `%` wildcards and produces a LIKE pattern).

This PR updates both docblocks to describe the `LIKE` pattern behaviour the class actually implements, so maintainers looking at the file in isolation aren't misled.

No behaviour change — comment-only edit.

Ref: commit [1de71054](https://github.com/greenpeace/planet4-master-theme/commit/1de71054) PLANET-7527 Move blocks report into master theme

### Testing

1. `git diff` — only the two docblock lines change.
2. `composer sniffs` — no new phpcs errors.
3. No code paths touched, no functional change.

### DIFF
--- a/src/BlockReportSearch/Block/Sql/Like.php
+++ b/src/BlockReportSearch/Block/Sql/Like.php
@@ -1,7 +1,7 @@
 <?php
 
 /**
- * Block search regex
+ * Block search LIKE pattern
  */
 
 namespace P4\MasterTheme\BlockReportSearch\Block\Sql;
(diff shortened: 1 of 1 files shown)


sagarsdeshmukh (approved):
Thanks @rushikeshmore 👍 
PR looks good to me.

PR #2965, opened 2026-04-24
PLANET-8045: Apply new new layout of validation summary

GP-Dan-Tovbein:
### Summary

This PRs applies the [new layout](https://www.figma.com/design/9NtbY8n3at8uOEJTsLrETb/P4-Design-System?node-id=370-24165&p=f&t=3Lh49kT2jTubh1Pu-0). Also set the Validation summary by default.

Since we need to modify the icon displayed on box, and Gravity Forms Theme does not provide that, we need to include Material Icons. Check it out the [Slack](https://greenpeace-gpi.slack.com/archives/G015K63081W/p1776957297524719) conversation.

---

Ref: https://greenpeace-planet4.atlassian.net/browse/PLANET-8045

### Testing

1. Create a new form (ideally First Name, Last Name, Email and any other.
2. Set all of them as `required`.
3. Check that Validation Summary is `enabled`.
4. Save.
5. Include the created form to a new page
6. Navigate through that page.
7. Click on submit and validate the layout on [Figma](https://www.figma.com/design/9NtbY8n3at8uOEJTsLrETb/P4-Design-System?node-id=370-24165&p=f&t=3Lh49kT2jTubh1Pu-0)


src/GravityFormsExtensions.php
```diff
@@ -147,6 +150,25 @@ public function dequeue_gf_scripts(): void
...
+    public function include_material_icons_font(): void
+    {
+        // Check if Gravity Forms is active
+        if (!class_exists('GFForms')) {
+            return;
+        }
+
+        wp_enqueue_style(
+            'material-symbols-outlined',
+            'https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined:opsz,wght,FILL,GRAD@24,400,0,0',
```
comzeradd: Left a comment in the slack discussion, as I think it's better to go for a single svg icon for now, since we don't use other icons from this Font Library yet.

But if we wanted to add it, wouldn't make more sense to add it in the same way we load our rest of the fonts?

GP-Dan-Tovbein: That's a good idea. 
@comzeradd could you share me the bucket url we use for that since it looks I have no permission on that? I suspect it's on [planet4-production](https://console.cloud.google.com/storage/browser?project=planet4-production&folder=&organizationId=), am I right?

comzeradd: I think for the scope of this ticket we can go for a single svg icon, instead of the font library. We can add the Material icon font once the design team has reviewed our existing icons and give us the green light to replace them.

planet-4:

:watch: 2026.05.13 18:47:09

assets/src/scss/layout/_gravity-forms.scss
```diff
@@ -8,14 +8,67 @@
...
     box-shadow: none;
     width: fit-content;
 
     > h2 {
-      color: var(--p4-action-yellow-600);
+      color: var(--red-800) !important;
+      display: flex;
+      flex-direction: row;
+      align-items: center;
+      font-size: var(--font-size-xxs--font-family-primary) !important;
```
Osong-Michael, line 23: The font weight on the test instance for the error does not really match the designs on Figma in my opinion, needs to be bolder. _Last name_ text should be bolder.


Image

Osong-Michael (commented):
Also, I don't know if these changes affect the behaviour, but when I create a new GF to test on this instance, it creates the form as inactive so it does not show up in the editor, the user needs to go back and activate the form before they can use it, is this expected behaviour?

assets/src/scss/layout/_gravity-forms.scss
GP-Dan-Tovbein: @Osong-Michael which browser are you using? and wich text should be bolder?

On my local FF and Chrome looks like the same as it's on Figma

**Chrome**
Screenshot 2026-05-13 at 7 24 40 AM

**Firefox**
Screenshot 2026-05-13 at 7 24 50 AM

**Figma**
Screenshot 2026-05-13 at 7 27 06 AM

GP-Dan-Tovbein:
> Also, I don't know if these changes affect the behaviour, but when I create a new GF to test on this instance, it creates the form as inactive so it does not show up in the editor, the user needs to go back and activate the form before they can use it, is this expected behaviour?

Good catch! It was happens with my code. The problem is that I previously add a new field to its array and then instead of update the current one, created a new one. That's why I'm getting the form first, and then update it

GP-Dan-Tovbein:
@Osong-Michael FYI, the PR it's now passing the UAT

Osong-Michael (approved):

PR #2969, opened 2026-04-29
PLANET-8047: Load JS scripts only where they are needed

mardelnet:
### Summary

Originally, this PR/ticket was meant to cover importing JS scripts only on pages where the affected elements were included, as it was done [here](https://github.com/greenpeace/planet4-master-theme/pull/2814).

However, a deeper look at the task revealed that the targeted JS scripts could be converted into PHP scripts, moving the logic to the back and improving the websites' performance.  

In summary, this PR converts the following JS scripts into PHP: `setupPDFIcon()`, `setupExternalLinks()`, `removeNoPostText()`, and `removeRelatedPostsSection()`. 

This implementation also reduces the JS bundle size a little bit:

|  Before | Now  |
|---|---|
|Screenshot from 2026-04-24 13-47-09|Screenshot from 2026-04-24 13-45-39|

---

Ref: https://greenpeace-planet4.atlassian.net/browse/PLANET-8047

### Testing

1. Check that the "Related Posts" section is removed from posts when it contains no related posts.
2. Check that the "No Results" inner block from a Post List or an Actions List block is removed when the title and the description of the block are empty.
3. Check that the class `pdf-link` is added to the links connected to PDF files.
4. Check that the class `external-link` is added to the external links.


planet-4:

:watch: 2026.05.04 14:58:18

src/HtmlPostProcessor.php
```diff
@@ -0,0 +1,207 @@
+<?php
+
+namespace P4\MasterTheme;
+
+use DOMDocument;
+use DOMXPath;
+
+/**
+ * Class OutputBufferManager
```
mleray: This is the wrong class name

src/HtmlPostProcessor.php
```diff
@@ -0,0 +1,207 @@
...
+            }
+
+            $classes = $link->getAttribute('class');
+            $link->setAttribute('class', trim($classes . ' pdf-link'));
+            $link->setAttribute('title', __('This link will open a PDF file', 'planet4-master-theme'));
+        }
+    }
+
+    /**
+     * Adds the class "pdf-link" to the links connected to PDF files.
```
mleray: This doc is incorrect

src/HtmlPostProcessor.php
mardelnet: Fixed!

src/HtmlPostProcessor.php
mardelnet: Fixed!

mleray:
For the Actions/Posts List no results, when it is shown the copy in the frontend includes `(This default text can be edited)` which should be removed I think? It's not related to your changes though so we can definitely fix this in a follow-up. 

Screenshot 2026-05-04 at 16 40 55

mleray (approved):
I've added a comment but as I said it can be addressed in a follow-up, everything seems to be working as expected 👏

PR #2973, opened 2026-04-30
PLANET-8154 GP Media: Improve image alt text accessibility

Osong-Michael:
### Summary
Every time an image is being imported through Greenpeace Media we use 3 main metadata sources to compose what we use in the frontend. 

We currently do this:
   - alt-text: title + credit
   - caption (shown under the image): description + credit
   - title tag (shown on hover): title + credit

We should adjust those based on accessibility best practices while taking into account how screen readers interact with those metadata.

---

**Ref:  https://greenpeace-planet4.atlassian.net/browse/PLANET-8154**

### Testing

First, import and Image from the GP Media Library, then create a new page and add an Image block to the page with the imported image.

- Remove credit from the hover title and use only the title. ✅ 
- Use description for the alt-text instead of the title.✅ 
- Keep caption intact (description & credit).✅ 
- Ensure caption is not being picked up by screen readers (eg. by using aria-hidden="true"). This is needed to avoid duplication, as screen readers will pick up the alt-text anyway.✅

### DIFF
--- a/src/ImageHandler.php
+++ b/src/ImageHandler.php
@@ -45,6 +45,29 @@ class ImageHandler
         add_filter('wp_image_editors', [$this, 'force_image_compression']);
         add_filter('wp_handle_upload_prefilter', [$this, 'image_type_validation']);
         add_filter('jpeg_quality', fn () => 60);
+        add_filter('render_block_core/image', [$this, 'add_figcaption_exception'], 10, 1);
+    }
+
+    /**
+     * Adds an aria-hidden attribute to all figcaption elements in image blocks.
+     *
+     * This ensures that screen readers do not read duplicate content when the image
+     * alt text and caption contain the same information, improving accessibility.
+     *
+     * @param string $block_content The HTML content of the rendered block.
+     *
+     * @return string Modified block HTML with updated figcaption attributes.
+     */
+    public function add_figcaption_exception(string $block_content): string
+    {
+        // Add aria-hidden to figcaption
+        $block_content = preg_replace(
+            '/<figcaption(?![^>]*aria-hidden)/',
+            '<figcaption aria-hidden="true"',
+            $block_content
+        );
+
+        return $block_content;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


planet-4:

:watch: 2026.05.19 08:47:29

src/ImageHandler.php
```diff
@@ -45,8 +45,35 @@ public function __construct()
...
+        // Add aria-hidden to figcaption
+        $block_content = preg_replace(
+            '/<figcaption(?![^>]*aria-hidden)/',
+            '<figcaption aria-hidden="true"',
+            $block_content
+        );
+
+        return $block_content;
+    }
+    // phpcs:enable SlevomatCodingStandard.Functions.UnusedParameter,
```
sagarsdeshmukh: This comment has a trailing comma(,)

src/ImageHandler.php
```diff
@@ -45,8 +45,35 @@ public function __construct()
         add_filter('wp_image_editors', [$this, 'force_image_compression']);
         add_filter('wp_handle_upload_prefilter', [$this, 'image_type_validation']);
         add_filter('jpeg_quality', fn () => 60);
+        add_filter('render_block_core/image', [$this, 'add_figcaption_exception'], 10, 2);
```
sagarsdeshmukh: Since we’re using only one of the three [function parameters](https://developer.wordpress.org/reference/hooks/render_block_this-name/), we can pass 1 as the last argument instead of 2. This way, we won’t need to use a phpcs:disable comment.

src/ImageHandler.php
```diff
@@ -104,26 +127,35 @@ public function register_meta_fields(): void
 
     /**
      * Override the Gutenberg core/image block render method output,
-     * to add credit field in its caption text & image alt text as title.
+     * to add credit field in its caption text, alt and title attributes.
```
sagarsdeshmukh: As per the ticket requirements -

> Remove credit from the hover title and use only the title.
> Remove credit from alt-text.

We are removing the credit from image alt and title attributes. right?

src/ImageHandler.php
```diff
@@ -221,4 +253,57 @@ public static function resize_image(
...
+     */
+    private function update_image_title_with_gp_media(string $title, string $content): string
+    {
+        $content = preg_replace_callback(
+            '/<img[^>]+>/i',
+            function ($matches) use ($title) {
+                $img = $matches[0];
+
+                // If title already exists, don't overwrite it
+                if (strpos($img, 'title=') !== false) {
```
sagarsdeshmukh: The `title` collision check is too loose. `strpos($img, 'title=') !== false` will match `data-title=`, `aria-titled=`, etc. 
Here we can use a word-boundary check - e.g. `preg_match('/\stitle\s*=/i', $img)`.

src/ImageHandler.php
```diff
@@ -221,4 +253,57 @@ public static function resize_image(
...
+                    return $img;
+                }
+
+                return str_replace(
+                    '<img',
+                    '<img title="' . esc_attr($title) . '"',
```
sagarsdeshmukh, line 292: Same as above here as well, the `title=` will match `data-title=`, `aria-titled=` etc.

src/ImageHandler.php
```diff
@@ -104,26 +127,35 @@ public function register_meta_fields(): void
...
             return $content;
         }
 
         $credit = $img_post_meta[self::CREDIT_META_FIELD][0] ?? '';
-        $alt_text = $img_post_meta['_wp_attachment_image_alt'][0] ?? '';
 
-        if ($alt_text) {
-            $content = str_replace(' alt=', ' title="' . esc_attr($alt_text) . '" alt=', $content);
+        // Replace alt text with image description
+        if ($image_description) {
```
sagarsdeshmukh: A description of "   " would pass the truthy check then become empty. In my opinion we should trim first, then check truthy.

src/ImageHandler.php
```diff
@@ -45,6 +45,29 @@ public function __construct()
...
+     * alt text and caption contain the same information, improving accessibility.
+     *
+     * @param string $block_content The HTML content of the rendered block.
+     *
+     * @return string Modified block HTML with updated figcaption attributes.
+     */
+    public function add_figcaption_exception(string $block_content): string
+    {
+        // Add aria-hidden to figcaption
+        $block_content = preg_replace(
```
sagarsdeshmukh: The WP 6.2+ offers a `WP_HTML_Tag_Processor` [core class](https://developer.wordpress.org/reference/classes/wp_html_tag_processor/) used to modify attributes in an HTML document for tags matching a query, which looks more safe and clean.

sagarsdeshmukh:
We use two filters(In `src/ImageHandler.php` file) for same blocks: `add_figcaption_exception` runs on `render_block_core/image`, and `p4_core_image_block_render` is the block's render callback. Both manipulate caption-related HTML. if we consolidating logic into the existing render callback would avoid running two passes per image.

src/ImageHandler.php
Osong-Michael: Yes but we still need it in the captions

sagarsdeshmukh: Yes, that’s right. However, the function comment says “add credit field in its caption text, alt, and title attributes.” Since we’re only keeping the credit text in the caption, I suggest updating the comment accordingly. Does that make sense?

sagarsdeshmukh (approved):
Tested on local, PR looks good to me.
Nice job 👍

PR #2974, opened 2026-04-30
PLANET-7974  Add live region and clear slide context to Actions List

mleray:
### Summary

This is for screen readers and therefore better accessibility. This commit also updates the indicators' click behaviour to be the same as the arrows

Ref: [PLANET-7974](https://greenpeace-planet4.atlassian.net/browse/PLANET-7974)

**Requirements**

1. Wrap the entire Action List block in a `div` with the following attributes:
  - role="region"
  - aria-label="Actions list" 
  - aria-roledescription="carousel"

2. Add a visually hidden live status element that updates to “Slide X of Y” when the user navigates between these sets. 

### Testing

On local you can go to the [Take Action page](http://www.planet4.test/take-action/) for example, since it has an Actions List block already, you just need to change it to a carousel. Or you can use the [venus Take Action page](https://www-dev.greenpeace.org/test-venus/take-action/) that is also used for UAT. Then you can test the following:
1. When clicking on the indicators (the dots) the focus behaviour is now the same as with the arrows, meaning that the focus goes to the first Action in the new slide.
2. Using a screen reader, make sure that `Slide X of Y` is announced when changing slides.

[PLANET-7974]: https://greenpeace-planet4.atlassian.net/browse/PLANET-7974?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/src/js/query_loop_carousel.js
+++ b/assets/src/js/query_loop_carousel.js
@@ -141,7 +141,7 @@ export const setupQueryLoopCarousel = () => {
           carouselItem.classList.add('carousel-item', 'carousel-li', `carousel-slide-${totalCarouselItems}`);
           list.append(carouselItem);
 
-          itemWrapper = document.createElement('div');
+          itemWrapper = document.createElement('ul');
           itemWrapper.classList.add('carousel-item-wrapper');
 
           carouselItem.append(itemWrapper);
--- a/assets/src/js/query_loop_carousel.js
+++ b/assets/src/js/query_loop_carousel.js
@@ -1,7 +1,7 @@
 import {v4 as uuid} from 'uuid';
 
 // Constants
-const {__} = wp.i18n;
+const {__, sprintf} = wp.i18n;
 const ARROW_DIRECTIONS = ['prev', 'next'];
 const LAYOUTS = {
   carousel: 'carousel',
@@ -57,6 +57,18 @@ export const setupQueryLoopCarousel = () => {
       const isPostsList = layout.className.includes('posts-list');
       const itemsPerSlide = isPostsList ? 4 : 3;
 
+      // Add some accessibility attributes
+      layout.setAttribute('role', 'region');
+      layout.setAttribute('aria-label', __('Actions List', 'planet4-master-theme'));
+      layout.setAttribute('aria-roledescription', 'carousel');
+
+      // Add an aria-live div to announce slide changes.
+      const announcement = document.createElement('div');
+      announcement.setAttribute('aria-live', 'polite');
+      announcement.classList.add('visually-hidden');
+      announcement.id = `announce-${uniqueId}`;
+      layout.appendChild(announcement);
+
       // Adapt it as bootstrap carousel
       const carousel = document.createElement('div');
       carousel.setAttribute('id', uniqueId);
@@ -69,7 +81,7 @@ export const setupQueryLoopCarousel = () => {
 
       const backToList = document.createElement('a');
       backToList.href = `#${uniqueId}`;
-      backToList.textContent = __('Back to Actions List', 'planet4-master-theme-backend');
+      backToList.textContent = __('Back to Actions List', 'planet4-master-theme');
       backToList.classList.add('carousel-skip-link');
       backToList.setAttribute('role', 'link');
 
@@ -175,12 +187,24 @@ export const setupQueryLoopCarousel = () => {
         }
 
         if (layout.className.includes(BLOCK_CLASSNAMES.actionsList)) {
-          // Moves the focus to the next slide when the carousel arrows are hit.
-          carouselButtons.forEach(button => {
+          // Moves the focus to the next slide when the carousel arrows or indicators are hit.
+          // Also update the aria-live text so that the screen reader announces the slide change.
+          const indicatorButtons = layout.querySelectorAll('.carousel-indicators li');
+          [...carouselButtons, ...indicatorButtons].forEach(button => {
             button.addEventListener('click', () => {
 
               const observer = new MutationObserver(() => {
                 const currentSlide = layout.querySelector('.carousel-item.active');
+                const currentSlideClasses = currentSlide.classList.value.split(' ');
+                const currentSlideNumber = currentSlideClasses.find(c => c.indexOf('carousel-slide-') !== -1).replace('carousel-slide-', '');
+                const slides = layout.querySelectorAll('.carousel-item').length;
+
+                announcement.innerText = sprintf(
+                /* translators: 1: current slide number, 2: total amount of slides */
+                  __('Slide %1$d from %2$d', 'planet4-master-theme'),
+                  Number(currentSlideNumber) + 1,
+                  slides
+                );
 
                 if (!currentSlide) {return;}
 
--- a/assets/src/js/query_loop_carousel.js
+++ b/assets/src/js/query_loop_carousel.js
@@ -124,15 +124,18 @@ export const setupQueryLoopCarousel = () => {
 
       posts.forEach((post, index) => {
         if (index % itemsPerSlide === 0) {
-          // Reset link that is only accessible via JS
-          slideReset = document.createElement('a');
-          slideReset.classList.add('carousel-ghost-link');
-          slideReset.setAttribute('aria-hidden', 'true');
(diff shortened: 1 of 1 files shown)


planet-4:

:watch: 2026.05.15 07:42:51

assets/src/js/query_loop_carousel.js
```diff
@@ -57,6 +57,18 @@ export const setupQueryLoopCarousel = () => {
       const isPostsList = layout.className.includes('posts-list');
       const itemsPerSlide = isPostsList ? 4 : 3;
 
+      // Add some accessibility attributes
+      layout.setAttribute('role', 'region');
+      layout.setAttribute('aria-label', __('Actions List', 'planet4-master-theme'));
+      layout.setAttribute('aria-roledescription', 'carousel');
```
mardelnet: We are adding these attributes to the Posts List block, too.
If we keep them, we should at least rename the `aria-label` attribute

assets/src/js/query_loop_carousel.js
```diff
@@ -69,7 +81,7 @@ export const setupQueryLoopCarousel = () => {
 
       const backToList = document.createElement('a');
       backToList.href = `#${uniqueId}`;
-      backToList.textContent = __('Back to Actions List', 'planet4-master-theme-backend');
+      backToList.textContent = __('Back to Actions List', 'planet4-master-theme');
```
mardelnet: This also affects the Posts List block. 
Should we rename it for that case?

assets/src/js/query_loop_carousel.js
```diff
@@ -57,6 +57,18 @@ export const setupQueryLoopCarousel = () => {
...
+      layout.setAttribute('role', 'region');
+      layout.setAttribute('aria-label', __('Actions List', 'planet4-master-theme'));
+      layout.setAttribute('aria-roledescription', 'carousel');
+
+      // Add an aria-live div to announce slide changes.
+      const announcement = document.createElement('div');
+      announcement.setAttribute('aria-live', 'polite');
+      announcement.classList.add('visually-hidden');
+      announcement.id = `announce-${uniqueId}`;
+      layout.appendChild(announcement);
```
mardelnet: This also affects the Posts List block. 
Is that expected?

assets/src/js/query_loop_carousel.js
```diff
@@ -112,21 +124,24 @@ export const setupQueryLoopCarousel = () => {
...
-          slideReset.classList.add('carousel-ghost-link');
-          slideReset.setAttribute('aria-hidden', 'true');
-          slideReset.setAttribute('tabindex', '-1');
-          slideReset.style.position = 'absolute';
-          slideReset.style.opacity = '0';
-          slideReset.style.pointerEvents = 'none';
-          list.appendChild(slideReset);
+          // Add a reset link that is only accessible via JS.
+          // Only for the Posts List block, not Actions List.
+          if (layout.className.includes(BLOCK_CLASSNAMES.postsList)) {
```
mardelnet: Have you changed this as a consequence of [these edits](https://github.com/greenpeace/planet4-master-theme/pull/2948/changes)?

mardelnet (commented):
@mleray, is a migration script necessary for the already existing carousels?

assets/src/js/query_loop_carousel.js
```diff
@@ -175,12 +194,24 @@ export const setupQueryLoopCarousel = () => {
...
+          [...carouselButtons, ...indicatorButtons].forEach(button => {
             button.addEventListener('click', () => {
 
               const observer = new MutationObserver(() => {
                 const currentSlide = layout.querySelector('.carousel-item.active');
+                const match = currentSlide.className.match(/carousel-slide-(\d+)/);
+                const slideIndex = match ? parseInt(match[1], 10) : null;
+                const slides = layout.querySelectorAll('.carousel-item').length;
+
+                announcement.innerText = sprintf(
```
mardelnet: This text is not added until a slider is activated. 
Should we add it on DOM ready instead?

mleray: No, it's meant to be updated only when the slide changes so that the screen reader announces it

assets/src/js/query_loop_carousel.js
mleray: Yes, the reset link is no longer needed for the Actions List after these changes

assets/src/js/query_loop_carousel.js
mleray: For now the changes are only for the Actions List, I'll update the code

assets/src/js/query_loop_carousel.js
mleray: For now the changes are only for the Actions List, I'll update the code

assets/src/js/query_loop_carousel.js
mleray: For now the changes are only for the Actions List, I'll update the code

mleray:
> @mleray, is a migration script necessary for the already existing carousels?

@mardelnet since the changes are added in JS, it shouldn't be necessary!

assets/src/js/query_loop_carousel.js
```diff
@@ -112,21 +131,24 @@ export const setupQueryLoopCarousel = () => {
...
-          slideReset.classList.add('carousel-ghost-link');
-          slideReset.setAttribute('aria-hidden', 'true');
-          slideReset.setAttribute('tabindex', '-1');
-          slideReset.style.position = 'absolute';
-          slideReset.style.opacity = '0';
-          slideReset.style.pointerEvents = 'none';
-          list.appendChild(slideReset);
+          // Add a reset link that is only accessible via JS.
+          // Only for the Posts List block, not Actions List.
+          if (layout.className.includes(BLOCK_CLASSNAMES.postsList)) {
```
mardelnet:
```suggestion
          if (if (isPostsList)) {
```

assets/src/js/query_loop_carousel.js
```diff
@@ -45,16 +45,19 @@ export const setupQueryLoopCarousel = () => {
       return;
     }
 
+    // This is for the carousel layout, we need to setup the arrows and indicators.
+    // Or hide them if there are not enough items to scroll.
     if (layout.className.includes(LAYOUTS.carousel)) {
```
mardelnet, line 50: I wonder if this condition that wraps all the logic can be turned into a much simpler guard condition (an early return).

assets/src/js/query_loop_carousel.js
```diff
@@ -67,13 +70,12 @@ export const setupQueryLoopCarousel = () => {
...
       const posts = list.querySelectorAll('.wp-block-post');
 
-      const backToList = document.createElement('a');
-      backToList.href = `#${uniqueId}`;
-      backToList.textContent = __('Back to Actions List', 'planet4-master-theme-backend');
-      backToList.classList.add('carousel-skip-link');
-      backToList.setAttribute('role', 'link');
-
-      layout.append(backToList);
+      if (isActionsList) {
```
mardelnet, line 73: Can this condition be combined with the other one on line 105?

mardelnet (approved):
Well done!
I have some optional suggestions.
Feel free to merge the PR without working on them if they're not necessary.

assets/src/js/query_loop_carousel.js
mleray: In this case it's not applicable, because if it's not a carousel there is an `else` statement where we remove the arrows!

assets/src/js/query_loop_carousel.js
mleray: I didn't do this because the one on line 105 should be executed only when there is more than one slide, it's within the `if (posts.length > itemsPerSlide)` condition. The `aria-live` and back to list link should be added only if there are arrows and indicators too!

assets/src/js/query_loop_carousel.js
mleray: We could revert the statement and show the `removeArrows` section first, if you think that's more readable

assets/src/js/query_loop_carousel.js
mleray: Updated!

PR #2975, opened 2026-05-01
PLANET-8084 Refactor Cookies block to utilize block.json

mleray:
### Summary

Ref: [PLANET-8084](https://greenpeace-planet4.atlassian.net/browse/PLANET-8084)

### Testing

The Cookies block should still behave as expected ([example page](https://www-dev.greenpeace.org/test-deimos/cookies-test/)). If testing on local, you will need to run `npm run build`.

[PLANET-8084]: https://greenpeace-planet4.atlassian.net/browse/PLANET-8084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


planet-4:

:watch: 2026.05.18 08:00:41

assets/src/blocks/Cookies/edit.js
```diff
@@ -23,6 +17,34 @@ export const CookiesEditor = ({setAttributes, attributes, isSelected}) => {
     all_cookies_description,
   } = attributes;
 
+  const COOKIES_DEFAULT_COPY = window.p4_vars?.options?.cookies_default_copy || {};
+
+  // Planet4 settings(Planet 4 > Cookies > Enable Analytical Cookies).
+  const ENABLE_ANALYTICAL_COOKIES = window.p4_vars?.options?.enable_analytical_cookies;
+
+  const CookiesFieldResetButton = ({fieldName, toAttribute, currentValue}) => {
```
sagarsdeshmukh: The previous implementation had CookiesFieldResetButton as a separate module/file. Is there a specific reason or advantage to adding it here?

mleray: Since it's only used here, I just thought this made more sense 🤷 And it simplifies the folder structure

sagarsdeshmukh: I’m still learning React, but based on my understanding, I tried to evaluate which approach would be better in the long run:
1)
```
const CookiesFieldResetButton = () => {}
```
A new function/component is created on every render of CookiesEditor.
&
2)
```
import {CookiesFieldResetButton}
```
The component is imported once and reused.
IMO, option 2 would be better for long-term maintainability. What do you think?

mleray: @sagarsdeshmukh I see your point! I've moved the component out of the file and into the `block-editor` folder, let me know what you think 👀

assets/src/block-editor/CookiesFieldResetButton/index.js
```diff
@@ -1,9 +1,8 @@
 const {__} = wp.i18n;
 const {Tooltip} = wp.components;
 
-const COOKIES_DEFAULT_COPY = window.p4_vars.options.cookies_default_copy || {};
-
 export const CookiesFieldResetButton = ({fieldName, toAttribute, currentValue}) => {
+  const COOKIES_DEFAULT_COPY = window.p4_vars.options.cookies_default_copy || {};
```
sagarsdeshmukh: In `edit.js`, we use defensive optional chaining,
`window.p4_vars?.options?.cookies_default_copy`
but here we do not. For consistency, should we use the same approach in both places?

sagarsdeshmukh:
`CookiesFieldResetButton` now lives in `assets/src/block-editor/CookiesFieldResetButton/index.js`. If this component is only used by the Cookies block, keeping it in a top-level shared directory might be unnecessary and reduces the “block-as-a-self-contained-unit” benefit that `block.json` aims to provide.

Should we consider keeping it inside `assets/src/blocks/Cookies/` unless there’s a concrete plan to reuse it across multiple blocks?

assets/src/block-editor/CookiesFieldResetButton/index.js
mleray: Good catch, updated!

mleray:
> Should we consider keeping it inside `assets/src/blocks/Cookies/` unless there’s a concrete plan to reuse it across multiple blocks?

Yes I agree with you @sagarsdeshmukh, updated!

sagarsdeshmukh (approved):
Nice job, @mleray 👏
I tested the PR locally and on the test instance. It looks good to me 👍

PR #2977, opened 2026-05-04
Disable username/password login when enforce_sso is on

comzeradd:
### Summary

Instead of relying just to redirection we can make sure that plain is completely disabled.

### Testing

On local environemnt:
1. Make sure enforce_sso and ga_login redirection is off:
```
npx wp-env run cli wp option patch delete planet4_features enforce_sso
npx wp-env run cli wp option patch delete galogin ga_auto_login
```
2. You should see the normal login screen
3. You should be able to login using credentials
4. Enable the sso flag:
```
npx wp-env run cli wp option patch insert planet4_features enforce_sso on
```
5. You should see the normal login screen
6. Trying to login with credentials you should be getting an error
7. Enable redirection:
```
npx wp-env run cli wp option patch update galogin ga_auto_login 1
```
8. You should be redirected straight to Google

### DIFF
--- a/src/LoginHandler.php
+++ b/src/LoginHandler.php
@@ -36,29 +36,13 @@ class LoginHandler
         add_action('login_footer', function (): void {
             $html = ob_get_clean();
 
-            $features = get_option('planet4_features', []);
-            $enforce_sso = !empty($features['enforce_sso']);
-
             // Clean up the HTML by removing the "Remember Me" checkbox
             $html = preg_replace('/<p[^>]*class=["\']forgetmenot["\'][^>]*>.*?<\/p>/is', '', $html);
 
-            if ($enforce_sso) {
-                if (isset($_GET['loggedout']) && $_GET['loggedout'] === 'true') {
-                    wp_redirect(home_url());
-                    exit;
-                }
-
-                $gal_instance = google_apps_login();
-                if (!method_exists($gal_instance, 'ga_start_auth_get_url')) {
-                    return;
-                }
-
-                $ga_url = $gal_instance->ga_start_auth_get_url();
-
-                if (!empty($ga_url)) {
-                    wp_redirect(esc_url_raw($ga_url));
-                    exit;
-                }
+            // Redirect to the homepage when the user has just logged out.
+            if (isset($_GET['loggedout']) && $_GET['loggedout'] === 'true') {
+                wp_redirect(home_url());
+                exit;
             }
 
             echo $html;
@@ -80,6 +64,7 @@ class LoginHandler
         add_filter('authenticate', [$this, 'enforce_google_signon'], 4, 3);
         add_filter('authenticate', [$this, 'check_google_login_error'], 30, 1);
         add_filter('authenticate', [$this, 'custom_block_login_if_rate_limited'], 30, 3);
+        add_filter('authenticate', [$this, 'disable_credentials_login'], 100, 1);
         add_filter('login_headerurl', [$this, 'add_login_logo_url']);
         add_filter('login_headertext', [$this, 'add_login_logo_url_title']);
         add_action('login_enqueue_scripts', [$this, 'add_login_stylesheet']);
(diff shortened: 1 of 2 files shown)


planet-4:

:watch: 2026.05.05 09:03:48

src/LoginHandler.php
```diff
@@ -373,4 +358,41 @@ private function password_policy_check(string $password): string|bool
...
+        $features = get_option('planet4_features', []);
+        $enforce_sso = !empty($features['enforce_sso']);
+
+        if (!$enforce_sso) {
+            return $user;
+        }
+
+        if ($user instanceof WP_User) {
+            return $user;
+        }
```
mardelnet: In my local environment, I am still able to log in using credentials when the SSO is active. 
That's fixed when this condition is removed.

src/LoginHandler.php
```diff
@@ -80,6 +64,7 @@ public function __construct()
         add_filter('authenticate', [$this, 'enforce_google_signon'], 4, 3);
         add_filter('authenticate', [$this, 'check_google_login_error'], 30, 1);
```
mardelnet, line 65: With the changes added by this PR, are these functions (`enforce_google_signon` and `check_google_login_error`) still needed?

src/LoginHandler.php
```diff
@@ -373,4 +358,37 @@ private function password_policy_check(string $password): string|bool
...
     }
+
+    /**
+     * Disables login with username and password when enforce_sso feature is enabled.
+     *
+     * @param WP_User|WP_Error $user The current user logging in.
+     * @param string           $username The username of the user.
+     * @param string           $password The password of the user.
+     */
+    public function disable_credentials_login(
```
mardelnet, line 367: Should we check if the Google Apps Login plugin is enabled, and add an early return if not?

mardelnet (commented):
Should we hide this element from the admin panel to prevent editors from unchecking it?

Image

src/LoginHandler.php
comzeradd: Probably not. It depends on the level of security we want to enforce on test instances where we don't enforce Google login. 

After we fix the redirection, we can spend some time refactoring this class, as it seems to me it has a lot of unnecessary complexity.

mardelnet: Agreed!

src/LoginHandler.php
comzeradd: I'm not sure what the workflow would be in that case 🤔 If an admin accidentally deactivates the plugin they won't be able to login anyway because we enforce Google login for accounts with `greenpeace.org` email.

mardelnet: I think there's a way to prevent plugins from being deactivated by using the filters `plugin_action_links` and `pre_update_option_active_plugins`. We could consider using them in a different PR/ticket.

comzeradd:
> Should we hide this element from the admin panel to prevent editors from unchecking it?

I added a css rule to hide it

mardelnet (commented):
Step 7 of the testing instructions (Enable redirection: `npx wp-env run cli wp option patch update galogin ga_auto_login 1`) is not working locally since no data exists for the key `ga_auto_login`, as it was deleted on Step 1. However, the workflow works as expected if the redirection is manually activated.

mardelnet (approved):
The changes look good to me, @comzeradd

src/LoginHandler.php
```diff
@@ -80,6 +64,7 @@ public function __construct()
         add_filter('authenticate', [$this, 'enforce_google_signon'], 4, 3);
         add_filter('authenticate', [$this, 'check_google_login_error'], 30, 1);
         add_filter('authenticate', [$this, 'custom_block_login_if_rate_limited'], 30, 3);
+        add_filter('authenticate', [$this, 'disable_credentials_login'], 100, 1);
```
sagarsdeshmukh, line 67: The `authenticate` hook can return [WP_User](https://developer.wordpress.org/reference/classes/wp_user/) object or [WP_Error](https://developer.wordpress.org/reference/classes/wp_error/) or null otherwise.
hence to be safe we should update function return type with `WP_User|WP_Error|null`
Ref. https://developer.wordpress.org/reference/hooks/authenticate/

comzeradd: Yes, good point. I amended the function definition

sagarsdeshmukh:
> Step 7 of the testing instructions (Enable redirection: `npx wp-env run cli wp option patch update galogin ga_auto_login 1`) is not working locally since no data exists for the key `ga_auto_login`, as it was deleted on Step 1. However, the workflow works as expected if the redirection is manually activated.

For step 7 `npx wp-env run cli wp option patch insert galogin ga_auto_login 1` command works for me.

sagarsdeshmukh (approved):
I tested the PR on local, and looks good to me 👍

PR #2978, opened 2026-05-06
PLANET-8025 Cache nav menu queries to reduce DB load

sagarsdeshmukh:
Versioned cache around wp_get_nav_menu_items() for footer menus + wp_get_object_terms() dedupe.

### Summary

Adds a versioned cache around wp_get_nav_menu_items() for the footer menus, plus dedupes wp_get_object_terms() within a single request, the duplicate-query pattern flagged by the PR #2784 experiment.

Measured on [planet4.test](http://www.planet4.test/):
```
  queries:  114 -> 101  (-13, -11%)
  DB time:  0.0281s -> 0.0138s  (-51%)
  page:     0.28s -> 0.18s  (-36%)
  memory:   16.1MB -> 15.2MB
```
---

Ref: [PLANET-8025](https://greenpeace-planet4.atlassian.net/browse/PLANET-8025)

### 1. Verify query reduction
1. Install Query Monitor plugin on local and load any front-end page (e.g. a story page) as an admin.
2. Note the **Queries by Caller** counts for `WP_Query->get_posts()`, `WP_Term_Query->get_terms()`, `update_meta_cache()`, and `_prime_post_caches()`.
3. Reload the same page once (warm cache).
4. Expected on warm cache: `WP_Query->get_posts()` -3, `WP_Term_Query->get_terms()` -3, `update_meta_cache()` -5, `_prime_post_caches()` -2. Total query count should drop ~10-13.

### 2. Verify HTML output is unchanged
1. Disable NavMenuCache (comment out `NavMenuCache::class` in `src/Loader.php`), reload, save HTML: `curl -s <url> > /tmp/before.html`
2. Re-enable, reload twice (cold then warm), save HTML: `curl -s <url> > /tmp/after.html`
3. `diff /tmp/before.html /tmp/after.html` should show only nonce/timestamp differences. The header nav, footer-primary, footer-secondary, and footer-social menus must render identically.

### 3. Verify cache invalidation on menu edits
1. `wp option get p4_nav_menu_cache_version` — note the value.
2. Admin > Appearance > Menus, edit any menu (drag an item, change a label), Save.
3. `wp option get p4_nav_menu_cache_version` — value should be incremented.
4. Front-end reload: change reflects immediately.

### 4. Verify WPML plugin instance
1. With WPML active, switch language and reload.
2. Front-end menu reflects the correct language.
3. Each language has its own cache entry (cache key varies on `ICL_LANGUAGE_CODE`).

### 5. Verify admin/editor unaffected
1. Admin > Appearance > Menus loads and saves normally.
2. Block editor menu picker works.
3. WP REST `/wp/v2/menus` and `/wp/v2/menu-items` return live data (cache is bypassed for these endpoints).

### 6. Visual test
- Header navigation renders with all items, correct labels, correct links.
- Footer Primary, Footer Secondary render with correct labels and `target` attributes.
- Footer Social Media icons render with correct `href`s.
- Mobile/burger menu opens and shows the same items.

### 7. Rollback test
- Comment out `NavMenuCache::class` in `src/Loader.php` and reload. site renders exactly as before this PR. The wrapper falls through to `wp_get_nav_menu_items()` on cache miss, so commenting out the loader entry is a complete kill switch.

### 8. warm-cache check
- Attach two Query Monitor screenshots: one with `NavMenuCache::class` commented out (baseline), one with it enabled and a warm reload (optimized). Caption the totals (queries / DB time / page time).

[PLANET-8025]: https://greenpeace-planet4.atlassian.net/browse/PLANET-8025?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


planet-4:

:watch: 2026.06.29 07:15:49

mardelnet:
Hey @sagarsdeshmukh:
1. Can you fix the conflict with the `src/Loader.php` file?
2. Can you please add testing instructions?
3. How did you measure the results in memory, DB time, etc?

sagarsdeshmukh:
> Hey @sagarsdeshmukh:
> 
> 1. Can you fix the conflict with the `src/Loader.php` file?
> 2. Can you please add testing instructions?
> 3. How did you measure the results in memory, DB time, etc?

The PR has been updated with testing instructions, and the code conflicts have also been resolved. I used the Query Monitor plugin for the stats comparison.

src/NavMenuCache.php
```diff
@@ -0,0 +1,377 @@
...
+     */
+    public function __construct()
+    {
+        // Bypass everything in admin/customize/preview to keep menu editing fresh.
+        if (is_admin() || (defined('REST_REQUEST') && REST_REQUEST && $this->is_menu_rest_request())) {
+            $this->register_invalidation_hooks();
+            return;
+        }
+
+        // Layer B is intentionally NOT a wp_get_nav_menu_items filter.
```
mardelnet: Where's Layer A and Layer C?

src/NavMenuCache.php
```diff
@@ -0,0 +1,377 @@
...
+     */
+    private static function cache_set(string $key, $value): bool
+    {
+        if (self::using_persistent_cache()) {
+            $found = false;
+            wp_cache_get($key, self::CACHE_GROUP, false, $found);
+            if ($found) {
+                return true;
+            }
+            return (bool) wp_cache_set($key, $value, self::CACHE_GROUP, self::TTL);
```
mardelnet: Maybe all of this can be simplified by using [wp_cache_add](https://developer.wordpress.org/reference/functions/wp_cache_add/)?

mardelnet (commented):
@sagarsdeshmukh:

First: amazing job! I've been testing your changes and, so far, everything is working alright.

Second: I wonder if there's a way to measure the difference in queries, DB time, page, memory, and any other valuable variable using test instances instead of the local instance, so that we can have a more realistic idea of the advantages of your code changes.

Third and last: the `NavMenuCache` class is pretty complex and may be overkill if performance isn't significantly improved. Please, don't get me wrong, I am not trying to discard your job (which again, is great), I'm just saying that (if the performance does not improve so much) we can end up trading a small improvement with a big maintenance liability.

cc @comzeradd

src/NavMenuCache.php
```diff
@@ -0,0 +1,377 @@
...
+     * The numeric suffix is bumped when the cache *format* changes (e.g.
+     * stdClass-vs-WP_Post). Bumping it invalidates every previously-stored
+     * entry without needing a manual flush.
+     */
+    private const CACHE_GROUP = 'p4_nav_menu_v4';
+
+    /**
+     * Option name that stores the integer cache version.
+     */
+    private const VERSION_OPTION = 'p4_nav_menu_cache_version';
```
comzeradd: I noticed that in other places we use dash (`-`) for naming those cache keys. Not sure what's the most WP way, but just mentioning this for consistency.

src/NavMenuCache.php
```diff
@@ -0,0 +1,377 @@
...
+ */
+class NavMenuCache
+{
+    /**
+     * Object cache group used when an external object cache is present.
+     * The numeric suffix is bumped when the cache *format* changes (e.g.
+     * stdClass-vs-WP_Post). Bumping it invalidates every previously-stored
+     * entry without needing a manual flush.
+     */
+    private const CACHE_GROUP = 'p4_nav_menu_v4';
```
comzeradd: In other places we use `p4-cache` as a prefix for cache groups

src/NavMenuCache.php
```diff
@@ -0,0 +1,377 @@
...
+    /**
+     * Option name that stores the integer cache version.
+     */
+    private const VERSION_OPTION = 'p4_nav_menu_cache_version';
+
+    /**
+     * Default TTL for cached entries (12 hours). Invalidation is version-based,
+     * so this is a safety ceiling, not the primary lifetime control.
+     */
+    private const TTL = 12 * HOUR_IN_SECONDS;
```
comzeradd: Do you think it's too much to cache for 24h? As usually that's what we used in other use cases.

src/NavMenuCache.php
```diff
@@ -0,0 +1,377 @@
...
+     *
+     * Without this guard, set_transient() writes wp_options on every request
+     * even when the value is unchanged — turning the cache into a query
+     * amplifier. With the guard, transient writes only happen on cache miss.
+     *
+     * @param mixed $value
+     */
+    private static function cache_set(string $key, $value): bool
+    {
+        if (self::using_persistent_cache()) {
```
comzeradd: Maybe we can simplify this class my removing those checks completely? I don't think we have applied those in other places where we cache keys. 

We do use Redis, so persistent cache is always there, but even in the unlikely scenario that the connection with Redis is lost, WP will save those keys in memory when using `wp_cahe_` functions, which is probably fine temporarily. While my understanding is that the `_transient` functions save keys in the database.

src/NavMenuCache.php
sagarsdeshmukh: 24h is fine.

src/NavMenuCache.php
sagarsdeshmukh: Fixed. Removed the Layer A/B/C/D references from the comments since they were temporary development scaffolding and became confusing after optimization changes. The comments now describe the actual implementation more clearly.

src/NavMenuCache.php
sagarsdeshmukh: Done.

src/NavMenuCache.php
sagarsdeshmukh: I updated the code to use same dash (-) for naming cache keys.

src/NavMenuCache.php
sagarsdeshmukh: Added `p4-cache` as a prefix for cache groups.

src/NavMenuCache.php
sagarsdeshmukh: Done. Removed `using_persistent_cache()` entirely. `cache_get()` and `cache_set()` now use only `wp_cache_get()` / `wp_cache_add()`.

sagarsdeshmukh:
> @sagarsdeshmukh:
> 
> First: amazing job! I've been testing your changes and, so far, everything is working alright.
> 
> Second: I wonder if there's a way to measure the difference in queries, DB time, page, memory, and any other valuable variable using test instances instead of the local instance, so that we can have a more realistic idea of the advantages of your code changes.
> 
> Third and last: the `NavMenuCache` class is pretty complex and may be overkill if performance isn't significantly improved. Please, don't get me wrong, I am not trying to discard your job (which again, is great), I'm just saying that (if the performance does not improve so much) we can end up trading a small improvement with a big maintenance liability.
> 
> cc @comzeradd

I simplified the `NavMenuCache` class from 377 lines down to 306 lines. Could you please take another look now?

mardelnet (commented):
@sagarsdeshmukh, on my local instance, the Query Monitor plugin states that 134 database queries are called with this PR (same as the `main` branch). Would you mind confirming this?

Image

On the other hand, I tried a much simpler approach using Timber's built-in cache, and I got a significant reduction in the query calls:

Image

You can test the latest by replacing the menu's logic in the `MasterSite` class with this:

```php
$footer_social_menu = \Timber\Helper::transient(
    'footer_social_menu',
    function () {
        if (has_nav_menu('footer-social-menu')) {
            return Timber::get_menu('footer-social-menu');
        }
        return wp_get_nav_menu_items('Footer Social');
    },
    DAY_IN_SECONDS
);
$context['footer_social_menu'] = $footer_social_menu;

$footer_primary_menu = \Timber\Helper::transient(
    'footer_primary_menu',
    function () {
        if (has_nav_menu('footer-primary-menu')) {
            return Timber::get_menu('footer-primary-menu');
        }
        return wp_get_nav_menu_items('Footer Primary');
    },
    DAY_IN_SECONDS
);
$context['footer_primary_menu'] = $footer_primary_menu;

$footer_secondary_menu = \Timber\Helper::transient(
    'footer_secondary_menu',
    function () {
        if (has_nav_menu('footer-secondary-menu')) {
            return Timber::get_menu('footer-secondary-menu');
        }
        return wp_get_nav_menu_items('Footer Secondary');
    },
    DAY_IN_SECONDS
);
$context['footer_secondary_menu'] = $footer_secondary_menu;
```

sagarsdeshmukh:
> @sagarsdeshmukh, on my local instance, the Query Monitor plugin states that 134 database queries are called with this PR (same as the `main` branch). Would you mind confirming this?
> 
> Image
> On the other hand, I tried a much simpler approach using Timber's built-in cache, and I got a significant reduction in the query calls:
> 
> Image
> You can test the latest by replacing the menu's logic in the `MasterSite` class with this:
> 
> ```
> $footer_social_menu = \Timber\Helper::transient(
>     'footer_social_menu',
>     function () {
>         if (has_nav_menu('footer-social-menu')) {
>             return Timber::get_menu('footer-social-menu');
>         }
>         return wp_get_nav_menu_items('Footer Social');
>     },
>     DAY_IN_SECONDS
> );
> $context['footer_social_menu'] = $footer_social_menu;
> 
> $footer_primary_menu = \Timber\Helper::transient(
>     'footer_primary_menu',
>     function () {
>         if (has_nav_menu('footer-primary-menu')) {
>             return Timber::get_menu('footer-primary-menu');
>         }
>         return wp_get_nav_menu_items('Footer Primary');
>     },
>     DAY_IN_SECONDS
> );
> $context['footer_primary_menu'] = $footer_primary_menu;
> 
> $footer_secondary_menu = \Timber\Helper::transient(
>     'footer_secondary_menu',
>     function () {
>         if (has_nav_menu('footer-secondary-menu')) {
>             return Timber::get_menu('footer-secondary-menu');
>         }
>         return wp_get_nav_menu_items('Footer Secondary');
>     },
>     DAY_IN_SECONDS
> );
> $context['footer_secondary_menu'] = $footer_secondary_menu;
> ```

Thanks for your feedback @mardelnet ,
The local env (`planet4.test`) doesn't have a persistent object cache, so the previous `wp_cache_*` calls were falling back to in-memory storage that resets every request. I adopted your `\Timber\Helper::transient()` suggestion.
The PR is now ready for another look 🙏

src/MasterSite.php
```diff
@@ -110,6 +110,17 @@ protected function hooks(): void
...
+        // Bust footer-menu transients on menu edits.
+        $invalidate_footer_menus = static function (): void {
+            delete_transient('p4_footer_social_menu');
+            delete_transient('p4_footer_primary_menu');
+            delete_transient('p4_footer_secondary_menu');
+        };
+        add_action('wp_update_nav_menu', $invalidate_footer_menus);
+        add_action('wp_update_nav_menu_item', $invalidate_footer_menus);
+        add_action('wp_delete_nav_menu', $invalidate_footer_menus);
+        add_action('customize_save_after', $invalidate_footer_menus);
```
mardelnet: Do you think this and the `resolve_footer_menu_items` function can be moved to a separate class to shrink the `MasterSite` and keep it cleaner?

mardelnet (commented):
@sagarsdeshmukh, should we also include the `navigation-bar-menu` and the `donate-menu`?

mardelnet (commented):
Amazing job, @sagarsdeshmukh 
The logic looks solid, and the database queries are significantly reduced.

As I mentioned [here](https://github.com/greenpeace/planet4-master-theme/pull/2978#pullrequestreview-4531184400), I think the remaining menus can be included as well. What do you think?

As a final optional suggestion, we might move all the logic in `MasterSite` related to navigation menus into a separate class: `register_nav_menus`, `invalidate_footer_menus`, `enable_transparent_nav`, etc...

sagarsdeshmukh:
> @sagarsdeshmukh, should we also include the `navigation-bar-menu` and the `donate-menu`?

I'd like to leave these out of this PR. They have constraints the footer menus don't.
Per-request "current" highlighting the navbar templates do:
```
<!-- templates/desktop-menu.twig:5 and navigation-bar.twig:57 -->
<li class="nav-item {{ item.class }} {{ item == item.current ? 'active' : '' }}">
```
`item.current` is set by Timber based on the page being rendered. If we cached the items, the `active` class would freeze to whichever page first warmed the cache.

sagarsdeshmukh:
> Amazing job, @sagarsdeshmukh The logic looks solid, and the database queries are significantly reduced.
> 
> As I mentioned [here](https://github.com/greenpeace/planet4-master-theme/pull/2978#pullrequestreview-4531184400), I think the remaining menus can be included as well. What do you think?
> 
> As a final optional suggestion, we might move all the logic in `MasterSite` related to navigation menus into a separate class: `register_nav_menus`, `invalidate_footer_menus`, `enable_transparent_nav`, etc...

Thanks Pedro 🙏 

I updated the PR by extracting the nav menu logic into a separate `NavMenus` class. I also added a [comment](https://github.com/greenpeace/planet4-master-theme/pull/2978#issuecomment-4786869549) regarding the inclusion of the `navigation-bar-menu` and `donate-menu`.

The PR is ready for another look.

mardelnet (approved):
Great job @sagarsdeshmukh 

On my local, I see that the theme's query calls were reduced from 64 to 48, and the duration from about 0.025 to 0.015 seconds.

Do you think there's also room for improvement for the other navigation menus? If so, should we open a ticket for them?

sagarsdeshmukh:
Thanks Pedro 🙏 

Really appreciate the thorough review.

Yes, I think there's room, but the navbar and donate menus need a different caching shape than the footers because of -
1. Per-request `item.current` highlighting (the `active` class on  the current page's menu item)
2. Templates that depend on Timber\MenuItem accessors (`item.link`, `item.children`)

I'll open a follow-up [ticket](https://greenpeace-planet4.atlassian.net/browse/PLANET-8209) for navigation-bar-menu and donate-menu caching.

PR #2980, opened 2026-05-11
PLANET-7940: UX fixes and adjustments in the Actions List block

mardelnet:
### Summary

This PR introduces the following changes to the Actions List block:

- In the editor:
  - For the Grid layout, display a maximum of 6 elements instead of 5.
  - For the Carousel layout, fix items looking very thin when there are fewer than 3.
  - Update the block description: “Display your Actions using dynamic cards that can be filtered by Category, Tag, or Action type, and shown in either a grid or carousel view.“
  - Update the Manual Override feature of the block: "CAUTION: Adding actions individually will override the automatic functionality of this block."

- In the front:
  - Remove the flicker from the “Load More” button when fewer than 6 elements are displayed using the Grid layout.

---

Ref: https://greenpeace-planet4.atlassian.net/browse/PLANET-7940

### Testing

1. Check that the changes detailed above are correctly introduced.

### DIFF
--- a/assets/src/js/actions_list_load_more.js
+++ b/assets/src/js/actions_list_load_more.js
@@ -22,8 +22,9 @@ export const setupActionsListLoadMore = () => {
     }
 
     const posts = [...block.querySelectorAll('.wp-block-post')];
-    if (!posts || posts.length <= postsPerRow * 2) {
-      loadMoreButtonContainer.classList.add('d-none');
+    if (posts && posts.length > postsPerRow * 2) {
+      loadMoreButtonContainer.classList.add('d-flex');
+    } else {
       return;
     }
 
--- a/assets/src/js/actions_list_load_more.js
+++ b/assets/src/js/actions_list_load_more.js
@@ -9,6 +9,10 @@ export const setupActionsListLoadMore = () => {
   if (window.innerWidth >= 768 && window.innerWidth < 992) {
     postsPerRow = 2;
   }
+  // For small screens we only show 1 post per row.
+  else if (window.innerWidth < 768) {
+    postsPerRow = 1;
+  }
 
   if (!gridBlocks.length) {
     return;
@@ -24,8 +28,10 @@ export const setupActionsListLoadMore = () => {
     const posts = [...block.querySelectorAll('.wp-block-post')];
     if (posts && posts.length > postsPerRow * 2) {
       loadMoreButtonContainer.classList.add('d-flex');
+      loadMoreButtonContainer.classList.remove('d-none');
     } else {
-      return;
+      loadMoreButtonContainer.classList.remove('d-flex');
+      loadMoreButtonContainer.classList.add('d-none');
     }
 
     const loadMoreButton = loadMoreButtonContainer.querySelector('.wp-element-button');
--- a/assets/src/js/actions_list_load_more.js
+++ b/assets/src/js/actions_list_load_more.js
@@ -26,6 +26,7 @@ export const setupActionsListLoadMore = () => {
     }
 
     const posts = [...block.querySelectorAll('.wp-block-post')];
+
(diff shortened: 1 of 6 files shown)


planet-4:

:watch: 2026.05.28 10:46:24

assets/src/js/app.js
```diff
@@ -26,11 +29,4 @@ document.addEventListener('DOMContentLoaded', () => {
       setupQueryLoopCarousel();
     });
   }
-
-  if(document.querySelectorAll('.actions-list.is-custom-layout-grid').length) {
-    import('./actions_list_load_more').then(({setupActionsListLoadMore}) => {
-      setupActionsListLoadMore();
-      window.addEventListener('resize', setupActionsListLoadMore);
```
mardelnet, line 33: @mleray, the resize listener is not working as expected within the conditional import, so I had to remove it from it.

assets/src/js/actions_list_load_more.js
```diff
@@ -9,6 +9,10 @@ export const setupActionsListLoadMore = () => {
   if (window.innerWidth >= 768 && window.innerWidth < 992) {
     postsPerRow = 2;
   }
+  // For small screens we only show 1 post per row.
+  else if (window.innerWidth < 768) {
```
mleray: Can we put this `else if` on the same line as the previous closing accolade? And maybe move the comments? e.g. 
```
if (window.innerWidth >= 768 && window.innerWidth < 992) {
    // For medium screens we only show 2 posts per row.
    postsPerRow = 2;
  } else if (window.innerWidth < 768) { 
    // For small screens we only show 1 post per row.
    postsPerRow = 1;
  }
```
I personally prefer this version

assets/src/js/actions_list_load_more.js
```diff
@@ -9,6 +9,10 @@ export const setupActionsListLoadMore = () => {
...
+    postsPerRow = 1;
+  }
 
   if (!gridBlocks.length) {
```
mleray: Should we move this check before the `if` condition to set the posts per row? We should break as soon as possible

assets/src/scss/blocks/ActionsList/ActionsListStyle.scss
```diff
@@ -7,6 +7,7 @@
 
   .load-more-actions-container {
     margin-top: $sp-4;
+    display: none;
```
mleray, line 10: Should we add `d-none` to the element directly in the Actions List definition ([here](https://github.com/greenpeace/planet4-master-theme/blob/70abbc91db9b0e2688e9d5f92d6a0b1b0351d8ef/assets/src/blocks/ActionsList/index.js#L115))?
It won't be applied to existing blocks though, only new ones, so we can keep this here as well unless we want to do a migration

mleray: Yeah nevermind, it's not worth a migration... and if we need to keep this code anyway there's no point in bothering with the class

mleray (dismissed):
Tested on the instance, looks good 👏 

I've added a couple of comments but they are suggestions, not blockers so I'll let you be the judge 😄

assets/src/js/actions_list_load_more.js
mardelnet: Done!

assets/src/js/actions_list_load_more.js
mardelnet: Done!

assets/src/scss/blocks/ActionsList/ActionsListStyle.scss
mardelnet: I'll disregard this if you don't mind.

mleray (approved):
👍

PR #2985, opened 2026-05-18
PLANET-8163 Added filter for HTML post processor for NROs

Osong-Michael:
### Summary

This fix is in reference to the issue post in this [conversation](https://greenpeace.slack.com/archives/C0151L0KKNX/p1777990444705289)

---

Ref: https://greenpeace-planet4.atlassian.net/browse/PLANET-8163

### Testing

I added the fix to the Denmark dev instance, you can test it [here](https://www-dev.greenpeace.org/denmark/vaer-med/ikkemintilladelse/).

[Nordics Child theme code](https://github.com/greenpeace/planet4-child-theme-nordic/blob/test/processor-filter/functions.php#L461)

### DIFF
--- a/src/HtmlPostProcessor.php
+++ b/src/HtmlPostProcessor.php
@@ -28,9 +28,16 @@ class HtmlPostProcessor
     public function manage_output_buffer(string $buffer): string
     {
         // Filter to disable the output buffer processing, if needed.
+        $context = [
+            'template' => get_page_template_slug(),
+            'post_id' => get_the_ID(),
+            'post_type' => get_post_type(),
+        ];
+
         $should_process = apply_filters(
             'planet4_master_theme_process_buffer',
             true,
+            $context,
             $buffer
         );
 
--- a/src/HtmlPostProcessor.php
+++ b/src/HtmlPostProcessor.php
@@ -27,6 +27,17 @@ class HtmlPostProcessor
      */
     public function manage_output_buffer(string $buffer): string
     {
+        // Filter to disable the output buffer processing, if needed.
+        $should_process = apply_filters(
+            'planet4_master_theme_process_buffer',
+            true,
+            $buffer
+        );
+
+        if (!$should_process) {
+            return $buffer;
+        }
+
         // Create a new DOM document to parse and manipulate the HTML.
         $dom = new DOMDocument();
 


planet-4:

:watch: 2026.06.05 10:45:43

src/HtmlPostProcessor.php
```diff
@@ -27,6 +27,17 @@ public function __construct()
      */
     public function manage_output_buffer(string $buffer): string
     {
+        // Filter to disable the output buffer processing, if needed.
+        $should_process = apply_filters(
+            'planet4_master_theme_process_buffer',
```
mardelnet, line 38: Can we add this filter to the list of Planet 4 hooks on the support page: https://support.greenpeace.org/planet4/tech/hooks?

src/HtmlPostProcessor.php
```diff
@@ -27,6 +27,17 @@ public function __construct()
...
+            true,
+            $buffer
```
mardelnet: Can we pass the page context as a parameter instead of the full buffer, so devs don't have to do string-searching on a large HTML blob? Maybe something like this?

```php 
$should_process = apply_filters(
        'planet4_master_theme_process_buffer',
        true,
        [
            'template' => get_page_template_slug(),
            'post_id'  => get_the_ID(),
            'post_type' => get_post_type(),
        ]
    );
```

mardelnet (commented):
Nice job, @Osong-Michael 
I left a few comments.
Let's also wait for Svilena's feedback before merging this.

src/HtmlPostProcessor.php
Osong-Michael: Yes, but I think the code needs to be approved and merged first then Quentin will update the page, I am not sure of the process

mardelnet: We can simply open a PR in the Docs repository [as Sagar did](https://github.com/greenpeace/planet4-docs/pull/9) a few weeks ago once these changes merged.

Osong-Michael: All done, here is the [PR](https://github.com/greenpeace/planet4-docs/pull/10).

elncetoo:
Hi :)) I tested a couple of things and here is what I found:
- The post type approach didn't work for our use case because the form is an ACF block embedded on a regular page, not a dedicated post type; 
- What worked was exposing `$buffer` as a 3rd argument in the `planet4_master_theme_process_buffer` filter, which allows us to reliably skip processing based on the rendered HTML content `(str_contains($buffer, 'leads-form'))` from the child theme.

Would this approach works on your end too?

sagarsdeshmukh (approved):
PR looks good to me 👍

comzeradd:
I guess we can close this now, since #3003 was implemented and merged

PR #2986, opened 2026-05-19
Hide Avatars plugin license notice

comzeradd:
### Summary

image

This may be confusing for NRO admins, so better hide it. At least in production.

There is no specific class to target, so instead targetting the href content.

---

### Testing

1. Even in this branch, you should be able to see the notice on almost all Admin screens.
2. Change `WP_APP_ENV` to `production` in `.wp-env.override.json`.
3. Restart the environment.
4. The notice should no longer be visible.

### DIFF
--- a/admin/css/notices.css
+++ b/admin/css/notices.css
@@ -11,6 +11,7 @@
 #gf_dashboard_message,
 [data-ep-notice],
 .otgs-installer-notice-plugin-recommendation,
-.welcome-panel-content {
+.welcome-panel-content,
+div.error:has(a[href*="gpavatars_list_options"]) {
   display: none;
 }


planet-4:

:watch: 2026.05.19 07:52:06

admin/css/notices.css
```diff
@@ -11,6 +11,7 @@
 #gf_dashboard_message,
 [data-ep-notice],
 .otgs-installer-notice-plugin-recommendation,
-.welcome-panel-content {
+.welcome-panel-content,
+div.error:has(a[href*="gpavatars_list_options"]) {
```
sagarsdeshmukh, line 15: It would be nice to add a small comment about this change here. I’m hoping we can remove this patch after 13 Jun 2026; otherwise, we might miss other notices from Google Profile Avatars that use the gpavatars_list_options in href.

sagarsdeshmukh:
I pinned the `avatars-notice` branch on the [Titan test instance](https://github.com/greenpeace/planet4-test-titan/blob/main/composer-local.json#L6) to check the fix, but it looks like the notice is still visible. 🤔

comzeradd:
> I pinned the `avatars-notice` branch on the [Titan test instance](https://github.com/greenpeace/planet4-test-titan/blob/main/composer-local.json#L6) to check the fix, but it looks like the notice is still visible. 🤔

We only enqueue this css files on production environments (see [here](https://github.com/greenpeace/planet4-master-theme/blob/v1.368.0/src/Settings.php#L855-L863)). That's why to test it locally you need to change `WP_APP_ENV`.

admin/css/notices.css
comzeradd: Not sure if we want to remove it (and re-add it next year), since any important notification will still be visible on dev and local environments.

sagarsdeshmukh (approved):
👍

PR #2990, opened 2026-05-25
PLANET-8165: Serve Hubspot pages under P4 domain and path

mardelnet:
### Summary

This PR introduces a mechanism to intercept requests to a P4 website path, fetch content from Hubspot, and render it. Additionally, the following changes were included:
- The code in the `admin/js/options.js` file was arranged within functions to make it clearer.
- The code in the `src/Settings.php` file was adjusted to enqueue admin CSS and JS only in admin pages. 

---

Ref: https://greenpeace-planet4.atlassian.net/browse/PLANET-8165

### Testing

1. Log in to the website admin area.
2. Go to Planet 4 > Hubspot (`/wp-admin/admin.php?page=planet4_settings_hubspot`)
3. Check the "Reverse Proxy" box.
4. Complete the text fields using the data from "Reverse Proxy Data Examples" (check below)
5. Use the data from column Hubspot Domain to complete the field Hubspot Domain.
2. Use the data from column Hubspot Path to complete the field Hubspot Path, or leave it empty.
3. Use the data from column Hubspot Slug to create the page URL as explained in step 9.
4. Choose any string for the field P4 Path (for example, `p4-path`).
8. Save the changes.
9. Visit the webpage with the URL structure `[website-domain]/[P4 path]` or `[website-domain]/[p4-path]/[hubspot-slug]`, for example, `http://www.planet4.test/p4-path` or `http://www.planet4.test/p4-path/no-more-oil-wars`.
10. Check that Hubspot's content is rendered.
11. Check that the content is not rendered if the "Reverse Proxy" box is unchecked.
12. Check that the text fields in Planet 4 > Hubspot are hidden if the "Reverse Proxy" box is unchecked.
13. Check that the Save button in Planet 4 > Hubspot is disabled if at least one of the text fields is empty.

#### Reverse Proxy Data Examples:

| Hubspot Domain                  | Hubspot Path | Hubspot Slug                               |
|---------------------------------|--------------|--------------------------------------------|
| act.greenpeace.ca               | en-ca        | -                                          |
| act.greenpeace.ca               | en-ca        | no-more-oil-wars                           |
| pages.greenpeaceafrica.org      | -            | -                                          |
| pages.greenpeaceafrica.org      | -            | plastic-audit-supermarket-edition          |
| pages.greenpeaceafrica.org      | -            | stop-the-industrial-plunder-of-west-africa |
| acthubspot.greenpeace.org.au    | -            | -                                          |
| acthubspot.greenpeace.org.au    | -            | stop-big-gas                               |
| donatehubspot.greenpeace.org.au | -            | -                                          |
| donatehubspot.greenpeace.org.au | -            | stop-big-gas                               |

### DIFF
--- a/admin/js/options.js
+++ b/admin/js/options.js
@@ -1,4 +1,5 @@
 
+const {__} = wp.i18n;
 
 function removeSpamButtonsBox () {
   const spamButtons = document.querySelectorAll('.checkforspam');
@@ -70,7 +71,7 @@ function toggleHubspotReverseProxySaveButton() {
     alertMessage = document.createElement('span');
     alertMessage.id = 'hubspot_reverse_alert';
     alertMessage.classList.add('hidden');
-    alertMessage.innerHTML = 'WARNING: The "Hubspot Domain" and the "P4 Path" are mandatory!';
+    alertMessage.innerHTML = __('WARNING: The "Hubspot Domain" and the "P4 Path" are mandatory!', 'planet4-master-theme-backend');
     alertMessage.style.marginInlineStart = '10px';
     saveButton.after(alertMessage);
   }
--- a/admin/js/options.js
+++ b/admin/js/options.js
@@ -1,11 +1,15 @@
-const spamButtons = document.querySelectorAll('.checkforspam');
 
-spamButtons.forEach(box => {
-  box.remove();
-});
+
+function removeSpamButtonsBox () {
+  const spamButtons = document.querySelectorAll('.checkforspam');
+
+  spamButtons.forEach(box => {
+    box.remove();
+  });
+}
 
 // Show/hide old special pages when enabling or disabling the new IA.
-document.addEventListener('DOMContentLoaded', () => {
+function toggleNewIaPages() {
   const newIASetting = document.querySelector('#new_ia');
   const analyticalCookiesCheckbox = document.querySelector('#enable_analytical_cookies');
 
@@ -45,4 +49,73 @@ document.addEventListener('DOMContentLoaded', () => {
       descriptionTextField.classList.toggle('hidden', !checked);
     });
   }
+}
+
+// Control the elements of the Hubspot Reverse Proxy form.
+function toggleHubspotReverseProxySaveButton() {
+  const reverseProxy = document.querySelector('#hubspot_reverse_proxy');
+  const saveButton = document.querySelector('#option_metabox input[type=submit]');
+  const domain = document.querySelector('.cmb2-id-hubspot-reverse-proxy-domain');
+  const p4Path = document.querySelector('.cmb2-id-hubspot-reverse-proxy-p4-path');
+  const hubspotPath = document.querySelector('.cmb2-id-hubspot-reverse-proxy-hubspot-path');
+
+  if (!reverseProxy || !saveButton || !domain || !p4Path || !hubspotPath) {
+    return;
+  }
+
+  // Create and insert an alert message:
+  let alertMessage = document.querySelector('#hubspot_reverse_alert');
+
+  if (!alertMessage) {
+    alertMessage = document.createElement('span');
+    alertMessage.id = 'hubspot_reverse_alert';
+    alertMessage.classList.add('hidden');
+    alertMessage.innerHTML = 'WARNING: All the fields must be complete to save the changes!';
+    alertMessage.style.marginInlineStart = '10px';
+    saveButton.insertAdjacentElement('afterend', alertMessage);
+  }
+
+  // Enable/disable the save button, and show/hide the alert message:
+  function toggleSaveButton() {
+    const proxyEnabled = reverseProxy.checked;
+    const fieldsEmpty =
+        domain.querySelector('input').value.trim() === '' ||
+        p4Path.querySelector('input').value.trim() === '' ||
+        hubspotPath.querySelector('input').value.trim() === '';
+
+    const shouldDisable = proxyEnabled && fieldsEmpty;
+    saveButton.disabled = shouldDisable;
+    alertMessage.classList.toggle('hidden', !shouldDisable);
+  }
+
+  // Show/hide the form fields:
+  function toggleTextFields() {
+    if (reverseProxy.checked) {
+      domain.classList.remove('hidden');
+      p4Path.classList.remove('hidden');
+      hubspotPath.classList.remove('hidden');
+    } else {
+      domain.classList.add('hidden');
+      p4Path.classList.add('hidden');
+      hubspotPath.classList.add('hidden');
+    }
+  }
+
+  toggleSaveButton();
+  toggleTextFields();
+
+  reverseProxy.addEventListener('change', () => {
+    toggleTextFields();
+    toggleSaveButton();
+  });
+
+  domain.querySelector('input').addEventListener('input', toggleSaveButton);
+  p4Path.querySelector('input').addEventListener('input', toggleSaveButton);
+  hubspotPath.querySelector('input').addEventListener('input', toggleSaveButton);
+}
+
+document.addEventListener('DOMContentLoaded', () => {
+  removeSpamButtonsBox();
+  toggleNewIaPages();
+  toggleHubspotReverseProxySaveButton();
 });
--- a/admin/js/options.js
+++ b/admin/js/options.js
@@ -70,7 +70,7 @@ function toggleHubspotReverseProxySaveButton() {
     alertMessage = document.createElement('span');
     alertMessage.id = 'hubspot_reverse_alert';
     alertMessage.classList.add('hidden');
-    alertMessage.innerHTML = 'WARNING: All the fields must be complete to save the changes!';
+    alertMessage.innerHTML = 'WARNING: The "Hubspot Domain" and the "P4 Path" are mandatory!';
     alertMessage.style.marginInlineStart = '10px';
     saveButton.after(alertMessage);
   }
@@ -78,10 +78,7 @@ function toggleHubspotReverseProxySaveButton() {
   // Enable/disable the save button, and show/hide the alert message:
   function toggleSaveButton() {
     const proxyEnabled = reverseProxy.checked;
-    const fieldsEmpty =
-        domain.querySelector('input').value.trim() === '' ||
-        p4Path.querySelector('input').value.trim() === '' ||
-        hubspotPath.querySelector('input').value.trim() === '';
+    const fieldsEmpty = domain.querySelector('input').value.trim() === '' || p4Path.querySelector('input').value.trim() === '';
 
     const shouldDisable = proxyEnabled && fieldsEmpty;
     saveButton.disabled = shouldDisable;
--- a/admin/js/options.js
+++ b/admin/js/options.js
@@ -72,7 +72,7 @@ function toggleHubspotReverseProxySaveButton() {
     alertMessage.classList.add('hidden');
     alertMessage.innerHTML = 'WARNING: All the fields must be complete to save the changes!';
     alertMessage.style.marginInlineStart = '10px';
-    saveButton.insertAdjacentElement('afterend', alertMessage);
+    saveButton.after(alertMessage);
   }
 
   // Enable/disable the save button, and show/hide the alert message:
--- a/src/HubspotReverseProxy.php
+++ b/src/HubspotReverseProxy.php
@@ -124,7 +124,7 @@ class HubspotReverseProxy
         // Forward only HubSpot-relevant cookies, never P4 session cookies.
         // https://knowledge.hubspot.com/privacy-and-consent/hubspot-cookie-security-and-privacy
         $allowed_cookies = [
-            '__cfduid', '__cf_bm', '__cfuvid', ' __hs_opt_out', '_hs_cookie_cat_pref',
+            '__cfduid', '__cf_bm', '__cfuvid', '__hs_opt_out', '_hs_cookie_cat_pref',
             'laboratory-anonymous-id', 'hs_login_email', 'hsgn', 'ak_bmsc', 'Hubspotapi',
             'Hubspotapi-csrf', 'Hubspotapi-lax', 'Hubspotapi-prefs', 'Hubspotapi-strict',
             'hs_langswitcher_choice', 'Hubspotutk', 'gbu9uvfhph6a0mdatwbzomssrlboczvs',
--- a/src/HubspotReverseProxy.php
+++ b/src/HubspotReverseProxy.php
@@ -64,7 +64,7 @@ class HubspotReverseProxy
         $hubspot_domain = rtrim($options['hubspot_reverse_proxy_domain'] ?? '', '/');
         $hubspot_path = trim($options['hubspot_reverse_proxy_hubspot_path'] ?? '', '/');
 
-        if (! $p4_path || ! $hubspot_domain || ! $hubspot_path) {
+        if (! $p4_path || ! $hubspot_domain) {
             return null;
         }
 
--- a/src/HubspotReverseProxy.php
+++ b/src/HubspotReverseProxy.php
@@ -22,7 +22,7 @@ class HubspotReverseProxy
     /**
      * Handles the reverse proxy logic on every WordPress request.
      *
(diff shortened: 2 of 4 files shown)


planet-4:

:watch: 2026.06.19 08:32:46

comzeradd:
I added the CA example in the test instance settings, to test out a multilingual site as well.

So this page: https://act.greenpeace.ca/en-ca/no-more-oil-wars
It's accessible also here: https://www-dev.greenpeace.org/test-leda/hubspot-test/no-more-oil-wars

The page renders fine, but I don't see the form. Looking at web console the one error that potentially could relate to this:

```
Refused to execute script from '<URL>' because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.
```

mardelnet:
> I added the CA example in the test instance settings, to test out a multilingual site as well.
> 
> So this page: https://act.greenpeace.ca/en-ca/no-more-oil-wars It's accessible also here: https://www-dev.greenpeace.org/test-leda/hubspot-test/no-more-oil-wars
> 
> The page renders fine, but I don't see the form. Looking at web console the one error that potentially could relate to this:
> 
> ```
> Refused to execute script from '<URL>' because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.
> ```

Thanks @comzeradd, now it's working: 
https://www-dev.greenpeace.org/test-leda/hubspot-test/no-more-oil-wars

src/HubspotReverseProxy.php
```diff
@@ -0,0 +1,182 @@
...
+
+        if (empty($options['hubspot_reverse_proxy'])) {
+            return null;
+        }
+
+        $p4_path = trim($options['hubspot_reverse_proxy_p4_path'] ?? '', '/');
+        $hubspot_domain = rtrim($options['hubspot_reverse_proxy_domain'] ?? '', '/');
+        $hubspot_path = trim($options['hubspot_reverse_proxy_hubspot_path'] ?? '', '/');
+
+        if (! $p4_path || ! $hubspot_domain) {
```
comzeradd: Cosmetic comment, but you have some inconsistencies on spacing through out this class. In this case you have a space after the `!` operator while in lines 31 and 37 you don't.

In general, we don't add spaces I think anymore in those cases. In contradiction to what WP does.

src/HubspotReverseProxy.php
```diff
@@ -0,0 +1,182 @@
...
+
+        $forward_headers = [];
+
+        foreach ([ 'Accept', 'Accept-Language', 'Accept-Encoding', 'Cookie' ] as $h) {
+            $key = 'HTTP_' . strtoupper(str_replace('-', '_', $h));
+            if (empty($_SERVER[ $key ])) {
+                continue;
+            }
+
+            $forward_headers[ $h ] = $_SERVER[ $key ];
```
comzeradd: Same here for keys. Not sure if there is a way to make `phpcs` catch those inconsistencies. We probably have more like this already in the code base.

comzeradd (commented):
Looks good and working as expected. I deployed it also on GPCA dev site, to test it out a bit further with a multilingual setup.

I left only some minor cosmetic comments.

admin/js/options.js
```diff
@@ -45,4 +49,70 @@ document.addEventListener('DOMContentLoaded', () => {
...
+    toggleSaveButton();
+  });
+
+  domain.querySelector('input').addEventListener('input', toggleSaveButton);
+  p4Path.querySelector('input').addEventListener('input', toggleSaveButton);
+  hubspotPath.querySelector('input').addEventListener('input', toggleSaveButton);
+}
+
+document.addEventListener('DOMContentLoaded', () => {
+  removeSpamButtonsBox();
```
sagarsdeshmukh, line 116: Thanks for nicely refactored this files functionality. 👍
It improves readability.

src/HubspotReverseProxy.php
```diff
@@ -0,0 +1,182 @@
...
+    {
+        $query_string = $_SERVER['QUERY_STRING'] ?? '';
+
+        if ($query_string) {
+            $target_url .= '?' . $query_string;
+        }
+
+        $forward_headers = [];
+
+        foreach ([ 'Accept', 'Accept-Language', 'Accept-Encoding', 'Cookie' ] as $h) {
```
sagarsdeshmukh: Here the cookie forwarding leaks all user session cookies to HubSpot. can we forward whitelisted / HubSpot-relevant cookies only?

src/HubspotReverseProxy.php
```diff
@@ -0,0 +1,182 @@
...
+
+    /**
+     * Rewrites root-relative URLs in proxied HTML to point to the HubSpot domain.
+     * The regex matches src="..." and href="..." attributes where the value starts with /
+     *
+     * @param string $html The HTML content to modify.
+     * @param string $hubspot_domain The HubSpot domain (with scheme, no trailing slash).
+     * @return string The modified HTML with rewritten URLs.
+     */
+    private function rewrite_relative_urls(string $html, string $hubspot_domain): string
```
comzeradd, line 201: One issue I spotted with GPCA example is that some urls start with `//` instead of `https://` which probably creates a problem here.

See for instance this page: https://www-dev.greenpeace.org/canada/en/act/no-more-oil-wars

It has its own language selector.
The link for French, originally looks like this:
`//act.greenpeace.ca/fr-ca/halte-aux-guerres-petrolieres`
through the proxy it looks like this:
`https://act.greenpeace.ca//act.greenpeace.ca/fr-ca/halte-aux-guerres-petrolieres`

It may be an edge case, but not sure if they can change the url or is it Hubspot adding them this way.

src/HubspotReverseProxy.php
```diff
@@ -0,0 +1,182 @@
...
+        if (!$content) {
+            return;
+        }
+
+        if (str_contains($content['type'], 'text/html')) {
+            $content['body'] = $this->inject_canonical($content['body'], $target_data['target_url']);
+            $content['body'] = $this->rewrite_relative_urls($content['body'], $target_data['hubspot_domain']);
+        }
+
+        echo $content['body']; // phpcs:ignore WordPress.Security.EscapeOutput
```
sagarsdeshmukh, line 58: Output of raw remote HTML (`echo $content['body']`).
This is fundamentally what a reverse proxy does, but it does mean any XSS on the HubSpot side runs under the P4 origin and can read P4 cookies/localStorage. Worth at minimum documenting the trust assumption.

comzeradd:
Do you think it's easy to only proxy the urls under the P4 Path but not the root of the path?

For example, if the P4 Path is set to `/act` then proxy `/act/no-more-oil-wars` but not `/act/`. It's probably a common scenario that NROs use `/act` for their "Act / Take Action" page and then link Hubspot pages from there.

src/HubspotReverseProxy.php
```diff
@@ -0,0 +1,182 @@
...
+     * The regex matches src="..." and href="..." attributes where the value starts with /
+     *
+     * @param string $html The HTML content to modify.
+     * @param string $hubspot_domain The HubSpot domain (with scheme, no trailing slash).
+     * @return string The modified HTML with rewritten URLs.
+     */
+    private function rewrite_relative_urls(string $html, string $hubspot_domain): string
+    {
+        return preg_replace_callback(
+            '/(\b(?:src|href|action)=")(\\/[^"]*")/',
```
sagarsdeshmukh: The regex only matches double-quoted attributes that start with `/`. It misses single-quoted attributes, `srcset`, inline `style="background:url(/...)"`, `data-*` attributes used by JS. 
It works now but will keep breaking as HubSpot tweaks its markup. Worth at minimum extending to single quotes and `srcset`

src/HubspotReverseProxy.php
```diff
@@ -0,0 +1,182 @@
...
+        foreach ([ 'Accept', 'Accept-Language', 'Accept-Encoding', 'Cookie' ] as $h) {
+            $key = 'HTTP_' . strtoupper(str_replace('-', '_', $h));
+            if (empty($_SERVER[ $key ])) {
+                continue;
+            }
+
+            $forward_headers[ $h ] = $_SERVER[ $key ];
+        }
+
+        $response = wp_remote_get($target_url, [
```
sagarsdeshmukh, line 158: Is it feasible to cache the HubSpot target URL response? Right now, every P4 request synchronously fetches data from HubSpot.

src/Settings.php
```diff
@@ -604,21 +604,48 @@ public function set_sub_pages(): void
...
+                    [
+                        'name' => __('Hubspot Path', 'planet4-master-theme-backend'),
+                        'id' => 'hubspot_reverse_proxy_hubspot_path',
                         'type' => 'text',
+                        'attributes' => ['type' => 'text'],
+                        'classes' => 'hidden',
+                        'description' => __('Path slug your Hubspot landing pages use. Usually empty or language specific.', 'planet4-master-theme-backend'),
+                    ],
+                    [
+                        'name' => __('P4 Path', 'planet4-master-theme-backend'),
```
sagarsdeshmukh, line 621: Just a concern regarding potential P4 path collisions: if an existing page slug gets added here, it could cause issues. A small reserved-slugs check in the save filter could help catch this. Alternatively, we could add some helper text to warn users and avoid such situations.

(You can ignore this if it’s not relevant. it just came to mind while reviewing 🙂 )

sagarsdeshmukh:
I tested the PR on local and test instance and its working as expected. nice work @mardelnet  💯

src/Settings.php
mardelnet: Thanks, @sagarsdeshmukh 
I do think your concern is relevant.
I also wonder if it can be fixed by applying [this adjustment](https://github.com/greenpeace/planet4-master-theme/pull/2990#issuecomment-4562704195).
What do you think?

src/HubspotReverseProxy.php
mardelnet: I like this, but it's a bit out of this PR scope.
What do you think @comzeradd? Do I create a new ticket so we can work on it in the future?

src/HubspotReverseProxy.php
mardelnet: Should I include all the cookies listed [here](https://knowledge.hubspot.com/privacy-and-consent/hubspot-cookie-security-and-privacy)? cc @comzeradd

src/HubspotReverseProxy.php
mardelnet: Good catch! Done!

mardelnet:
> Do you think it's easy to only proxy the urls under the P4 Path but not the root of the path?
> 
> For example, if the P4 Path is set to `/act` then proxy `/act/no-more-oil-wars` but not `/act/`. It's probably a common scenario that NROs use `/act` for their "Act / Take Action" page and then link Hubspot pages from there.

Done, @comzeradd

src/HubspotReverseProxy.php
mardelnet: I added a comment with that information (check lines 25 to 29)

src/HubspotReverseProxy.php
mardelnet: Fixed!

src/HubspotReverseProxy.php
mardelnet: Fixed!

src/HubspotReverseProxy.php
mardelnet: Fixed!
Let me know if you want me to create a ticket and make `phpcs` catch those inconsistencies.

comzeradd: Yes, that would be a nice improvement 👍🏻

src/HubspotReverseProxy.php
comzeradd: As far as I can see, our url is already cached by Cloudflare. Adding an extra layer of caching with WP/Redis will create the extra challenge of how we invalidate it 🤔 

But yes, let's create a ticket to investigate further. We may also need to think how we can invalidate CF cache for these endpoints.

src/HubspotReverseProxy.php
comzeradd: Yes, we can try that 👍🏻

src/HubspotReverseProxy.php
mardelnet: Done! https://greenpeace-planet4.atlassian.net/browse/PLANET-8183

src/HubspotReverseProxy.php
mardelnet: Done!

src/HubspotReverseProxy.php
mardelnet: Done: https://greenpeace-planet4.atlassian.net/browse/PLANET-8184

src/HubspotReverseProxy.php
```diff
@@ -0,0 +1,233 @@
...
+            return null;
+        }
+
+        if (!preg_match('#^https?://#', $hubspot_domain)) {
+            $hubspot_domain = 'https://' . $hubspot_domain;
+        }
+
+        $request_path = trim($wp->request, '/');
+
+        // Only proxy paths that have content beyond the base segment, but ever the root itself.
```
sagarsdeshmukh: @mardelnet -  could you please confirm do you mean? -

> Only proxy paths that have content beyond the base segment, but **never** the root itself.

src/HubspotReverseProxy.php
```diff
@@ -0,0 +1,233 @@
...
+                continue;
+            }
+
+            $forward_headers[$h] = $_SERVER[$key];
+        }
+
+        // Forward only HubSpot-relevant cookies, never P4 session cookies.
+        // https://knowledge.hubspot.com/privacy-and-consent/hubspot-cookie-security-and-privacy
+        $allowed_cookies = [
+            '__cfduid', '__cf_bm', '__cfuvid', ' __hs_opt_out', '_hs_cookie_cat_pref',
```
sagarsdeshmukh: Just a quick question, why do we have a leading space here?
`' __hs_opt_out'`
Is the space intentional, or could it be removed?

src/Loader.php
```diff
@@ -96,6 +96,7 @@ private function load_services(array $services): void
             Sentry::class,
             PostFunctionsHandler::class,
             HtmlPostProcessor::class,
+            HubspotReverseProxy::class,
```
sagarsdeshmukh, line 99: We [check](https://github.com/greenpeace/planet4-master-theme/blob/PLANET-8165_serve-hubspot-pages-under-p4-domain-and-path/src/Settings.php#L606) whether the GF HubSpot Add-On is active before showing the HubSpot Reverse Proxy settings page.

Should we add the same check here as well?

At the moment, `HubspotReverseProxy::class` is still being added to `default_services` in `Loader.php` unconditionally. As a result, the `parse_request` hook runs on every front-end request, even when the GF HubSpot Add-On is not installed or activated.

admin/js/options.js
```diff
@@ -45,4 +49,70 @@ document.addEventListener('DOMContentLoaded', () => {
...
+  }
+
+  // Create and insert an alert message:
+  let alertMessage = document.querySelector('#hubspot_reverse_alert');
+
+  if (!alertMessage) {
+    alertMessage = document.createElement('span');
+    alertMessage.id = 'hubspot_reverse_alert';
+    alertMessage.classList.add('hidden');
+    alertMessage.innerHTML = 'WARNING: The "Hubspot Domain" and the "P4 Path" are mandatory!';
```
sagarsdeshmukh: Can we use the `__()` function here to make this string translatable?

mardelnet: Good idea, done!

src/Loader.php
mardelnet: Great catch! I added the check to the class constructor.

src/HubspotReverseProxy.php
mardelnet: Great catch! Fixed!

src/HubspotReverseProxy.php
mardelnet: Fixed! The correct word is "never"

sagarsdeshmukh (approved):
Nice job @mardelnet 👍

PR #2991, opened 2026-05-26
Make sure Carousel Header focus changes only with tabbing

mleray:
### Summary

Before this fix the focus went to the heading also when clicking on the arrows, which looked weird. It needs to happen only when the user is tabbing.

Ref: [Slack thread](https://greenpeace.slack.com/archives/C0151L0KKNX/p1779748245904929)

### Testing

Make sure that the focus goes to the heading when interacting with the arrows/indicators while tabbing, but **not** when clicking. Also, clicking on the heading should no longer show the focus styles, they should only be applied when tabbing.

PR #2993, opened 2026-05-28
Make sure focus is updated only when tabbing in Actions List

mleray:
### Summary

When clicking on the arrows, the focus should not move. It should only happen when tabbing. This is a similar issue to what we recently solved for the Carousel Header [here](https://github.com/greenpeace/planet4-master-theme/pull/2991).

### Testing

Add an Actions List block to a page, with the carousel layout. Make sure that the focus only moves to the first item when tabbing on the arrows, but not when clicking on them.

Screenshot 2026-05-28 at 17 39 41

### DIFF
--- a/assets/src/js/query_loop_carousel.js
+++ b/assets/src/js/query_loop_carousel.js
@@ -203,8 +203,7 @@ export const setupQueryLoopCarousel = () => {
           // Also update the aria-live text so that the screen reader announces the slide change.
           const indicatorButtons = layout.querySelectorAll('.carousel-indicators li');
           [...carouselButtons, ...indicatorButtons].forEach(button => {
-            button.addEventListener('click', () => {
-
+            button.addEventListener('click', e => {
               const observer = new MutationObserver(() => {
                 const currentSlide = layout.querySelector('.carousel-item.active');
                 const match = currentSlide.className.match(/carousel-slide-(\d+)/);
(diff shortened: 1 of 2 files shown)


assets/src/js/query_loop_carousel.js
```diff
@@ -222,13 +221,17 @@ export const setupQueryLoopCarousel = () => {
...
-                const focusTarget = currentSlide.querySelector('a');
-
-                if (focusTarget) {
-                  focusTarget.focus();
-                } else {
-                  currentSlide.setAttribute('tabindex', '-1');
-                  currentSlide.focus();
+                // This means that it's a tabbing event and not a click,
+                // therefore we need to move the focus to the first item.
+                if (e.target === layout.ownerDocument.activeElement) {
```
mleray, line 226: Should I refactor the Carousel Header implementation to use the same check, instead of a separate `onKeydown` function? 🤔

mardelnet: Can you share the code to be refactored?

mleray: [This](https://github.com/greenpeace/planet4-master-theme/blob/main/assets/src/blocks/CarouselHeader/CarouselControls.js#L53-L59) could be moved to the `onClick` function defined [above](https://github.com/greenpeace/planet4-master-theme/blob/main/assets/src/blocks/CarouselHeader/CarouselControls.js#L46-L51), by using the same `e.target === activeElement` check

mardelnet: Sounds good to me, maybe here or in another PR.

planet-4:

:watch: 2026.05.29 13:56:43

assets/src/js/query_loop_carousel.js
mleray: Updated! I've had to disable the `@wordpress/no-global-active-element` ESLint rule, I couldn't find online a proper explanation as to why it's bad 🤷‍♀️

mardelnet (approved):
It works like a charm!

PR #2998, opened 2026-06-04
Write an e2e test for the Actions List block manual override

Osong-Michael:
### Summary
Expanding Actions List block test scenarios to include manual override.

---

**Ref: https://greenpeace-planet4.atlassian.net/browse/PLANET-8054**

### Testing

 - Login to the Admin Panel.

 - Create two new Actions.

 - Create two new Pages that are children pages of the designated “Take Action” page.

 - Create a new Page.

 - Add a Actions List block:

    - Pick the Grid layout

   - Use the “Manual Override” to select the two Actions and the two Pages created in previous steps.

   - Change block title to "Campaigns"

 - Publish the page.

 - Navigate to the frontend view of that Page and verify that all of the above properties are working as expected.

   - Block title

   - There are 4 actions displayed

   - Assess the 4 actions are indeed the ones created (eg. by title)

### DIFF
--- a/tests/e2e/blocks/actions-list.spec.js
+++ b/tests/e2e/blocks/actions-list.spec.js
@@ -33,8 +33,8 @@ test.describe('Test Actions List block', () => {
     await admin.visitAdminPage('admin.php', 'page=planet4_settings_analytics');
     const takeActionPageId = await page.locator('#take_action_page').inputValue();
 
-    // Skip Test if the new IA is not enabled or if the "Take Action" page ID is not available.
-    test.skip(!isNewIA || !takeActionPageId, 'The new IA must be enabled to run this test.');
+    // Skip Test if the new IA is not enabled or if the "Take Action" page ID is not available or gotten from the Analytics settings.
+    test.skip(!isNewIA || !takeActionPageId, 'The new IA must be enabled or the "Take Action" page ID must be available to run this test.');
 
     // Create 2 actions to be selected via the Manual Override.
     const regularActionTitles = [
@@ -64,13 +64,13 @@ test.describe('Test Actions List block', () => {
       });
     }
 
-    // Combine all action titles if needed downstream.
+    // Combine all action titles.
     const actionTitles = [...regularActionTitles, ...childActionTitles];
 
     // Create a page to hold the Actions List block.
     await createPostWithFeaturedImage({page, admin, editor}, {title: 'Test Actions List, Manual Override', postType: 'page'});
 
-    // Add a Actions List block using the Manual Override to select the 2 actions created above.
+    // Add a Actions List block using the Manual Override to select the actions created above.
     await addActionsListBlockWithManualOverride(page, actionTitles);
 
     // Publish page.
--- a/tests/e2e/blocks/actions-list.spec.js
+++ b/tests/e2e/blocks/actions-list.spec.js
@@ -1,9 +1,11 @@
-import {test, expect} from '../tools/lib/test-utils.js';
+import {test} from '../tools/lib/test-utils.js';
 import {publishPostAndVisit, createPostWithFeaturedImage} from '../tools/lib/post.js';
-import {searchAndInsertBlock} from '../tools/lib/editor';
-
-const TEST_TITLE = 'Campaigns';
-const TEST_CATEGORY = 'Energy';
+import {
+  addActionsListBlock,
+  checkActionsListBlock,
+  addActionsListBlockWithManualOverride,
+  checkActionsListBlockWithManualOverride,
+} from '../tools/lib/actions-list.js';
 
 test.useAdminLoggedIn();
 
@@ -14,36 +16,65 @@ test.describe('Test Actions List block', () => {
     await createPostWithFeaturedImage({page, admin, editor}, {title: 'Test Actions List Grid Layout', postType: 'page'});
 
     // Add Actions List block.
-    await searchAndInsertBlock({page}, 'Actions List');
-
-    // Set Actions per page to 2.
-    await page.getByRole('spinbutton', {name: 'Items per page'}).fill('2');
-
-    // Filter by "Energy" category.
-    const editorSettings = page.getByRole('region', {name: 'Editor settings'});
-    await editorSettings.getByRole('button', {name: 'Filters options'}).click();
-    await page.getByLabel('Show Taxonomies').click();
-    await editorSettings.getByLabel('Categories').fill(TEST_CATEGORY);
-    await editorSettings.locator(
-      '.components-form-token-field__suggestion', {hasText: TEST_CATEGORY}
-    ).click();
-    await expect(editorSettings.locator(
-      '.components-form-token-field__token-text', {hasText: TEST_CATEGORY})
-    ).toBeVisible();
-
-    // Change the title.
-    await page.getByRole('document', {name: 'Block: Heading'}).fill(TEST_TITLE);
+    await addActionsListBlock(page);
 
     // Publish page.
     await publishPostAndVisit({page, editor});
 
-    // Test that the block is displayed as expected in the frontend.
-    const block = page.locator('.p4-query-loop');
-    await expect(block).toHaveClass(/is-custom-layout-grid/);
-    await expect(block.locator('h2.wp-block-heading')).toHaveText(TEST_TITLE);
-    await expect(block.locator('.wp-block-post')).toHaveCount(2);
-    for (const category of await block.locator('.wp-block-post-terms').all()) {
-      await expect(category).toHaveText(TEST_CATEGORY);
+    // Check the Actions List block.
+    await checkActionsListBlock(page);
+  });
+
+  test('Test the Manual Override', async ({page, admin, editor, requestUtils}) => {
+    // Fetch the "Take Action" parent page ID
+    const takeActionPages = await requestUtils.rest({
+      path: '/wp/v2/pages',
+      method: 'GET',
+      params: {slug: 'take-action'},
+    });
+    const takeActionPageId = takeActionPages[0].id;
+
+    // Create 2 actions to be selected via the Manual Override.
+    const regularActionTitles = [
+      'Test Actions List Manual Override Action 1',
+      'Test Actions List Manual Override Action 2',
+    ];
+
+    for (const title of regularActionTitles) {
+      await requestUtils.rest({
+        path: '/wp/v2/p4_action',
+        method: 'POST',
+        data: {title, status: 'publish'},
+      });
     }
+
+    // Create 2 actions as children of the "Take Action" page.
+    const childActionTitles = [
+      'Test Actions List Manual Override Child Action 1',
+      'Test Actions List Manual Override Child Action 2',
+    ];
+
+    for (const title of childActionTitles) {
+      await requestUtils.rest({
+        path: '/wp/v2/p4_action',
+        method: 'POST',
+        data: {title, status: 'publish', parent: takeActionPageId},
+      });
+    }
+
+    // Combine all action titles if needed downstream.
+    const actionTitles = [...regularActionTitles, ...childActionTitles];
+
+    // Create a page to hold the Actions List block.
+    await createPostWithFeaturedImage({page, admin, editor}, {title: 'Test Actions List, Manual Override', postType: 'page'});
+
+    // Add a Actions List block using the Manual Override to select the 2 actions created above.
+    await addActionsListBlockWithManualOverride(page, actionTitles);
+
+    // Publish page.
+    await publishPostAndVisit({page, editor});
+
+    // Check that the block displays correctly in the frontend.
+    await checkActionsListBlockWithManualOverride(page, actionTitles);
   });
 });
--- a/tests/e2e/blocks/actions-list.spec.js
+++ b/tests/e2e/blocks/actions-list.spec.js
@@ -6,6 +6,7 @@ import {
   addActionsListBlockWithManualOverride,
   checkActionsListBlockWithManualOverride,
 } from '../tools/lib/actions-list.js';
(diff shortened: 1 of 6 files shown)


planet-4:

:watch: 2026.06.30 09:34:54

tests/e2e/blocks/actions-list.spec.js
```diff
@@ -14,36 +16,65 @@ test.describe('Test Actions List block', () => {
...
+  });
+
+  test('Test the Manual Override', async ({page, admin, editor, requestUtils}) => {
+    // Fetch the "Take Action" parent page ID
+    const takeActionPages = await requestUtils.rest({
+      path: '/wp/v2/pages',
+      method: 'GET',
+      params: {slug: 'take-action'},
+    });
+    const takeActionPageId = takeActionPages[0].id;
```
mleray: What happens if for some reason the Take Action page is not set? For example if a test instance uses the old IA?

Osong-Michael: Hmmm, didn't really think of that since the assumption is that all test instances are on the new IA.

I remember running into some issues locally when I once ran our test suits with an NRO's DB, it was chaos

mleray: Sometimes we need to revert to the old IA to fix a bug or something, would be annoying if the test fails in that case

Osong-Michael: Ok, so what do you suggest, since this is a requirement of the ticket?

mleray: Let's ask @comzeradd 
We could either skip the test in this case, or we implement the old IA version as well which uses Pages (the p4_action type would not exist)

mleray: I see that we already have one test ([special pages](https://github.com/greenpeace/planet4-master-theme/blob/main/tests/e2e/special-pages.spec.js#L9)) where we check whether or not the new IA is enabled

Osong-Michael: Hey @comzeradd need your input on this, regarding this test do we skip if new IA is not enabled or do we use the old implementation?

comzeradd: I wouldn't spend much time on covering the "Old IA" use case, as we are planning on phasing it out. Unless it's a fairly easy conditional we can add, I'd focus mostly on the assumption that "New IA" is enabled

mleray: There is still a possibility that the Take Action page is not set though, even if the new IA is enabled, should we implement this use case?

tests/e2e/blocks/actions-list.spec.js
```diff
@@ -14,36 +16,72 @@ test.describe('Test Actions List block', () => {
...
-    await expect(block.locator('h2.wp-block-heading')).toHaveText(TEST_TITLE);
-    await expect(block.locator('.wp-block-post')).toHaveCount(2);
-    for (const category of await block.locator('.wp-block-post-terms').all()) {
-      await expect(category).toHaveText(TEST_CATEGORY);
+    // Check the Actions List block.
+    await checkActionsListBlock(page);
+  });
+
+  test('Test the Manual Override', async ({page, admin, editor, requestUtils}) => {
+    await admin.visitAdminPage('admin.php', 'page=planet4_settings_navigation');
```
mleray: Maybe we can create a reusable function for this check? Since we already use it at least twice

tests/e2e/blocks/actions-list.spec.js
```diff
@@ -14,36 +17,70 @@ test.describe('Test Actions List block', () => {
...
+    const childActionTitles = [
+      'Test Actions List Manual Override Child Action 1',
+      'Test Actions List Manual Override Child Action 2',
+    ];
+
+    for (const title of childActionTitles) {
+      await requestUtils.rest({
+        path: '/wp/v2/p4_action',
+        method: 'POST',
+        data: {title, status: 'publish', parent: takeActionPageId},
```
mleray, line 63: There is still a possibility that the Take Action page is not set, even if the new IA is enabled, should we implement this use case?

Osong-Michael: In this test or another test?

tests/e2e/blocks/actions-list.spec.js
comzeradd: So the test expects to find a  "Take Action" page but looking for a page with the slug `take-action`. That may work in most test instances, but it doesn't seem very reliable. Can't we pick the "Take Action" page from the settings?

tests/e2e/blocks/actions-list.spec.js
mleray: In this test! What happens here if `takeActionPageId` is undefined?

tests/e2e/blocks/actions-list.spec.js
mleray: Right, actually on local with the default db I have the "Act" page selected there, not "Take Action"

tests/e2e/blocks/actions-list.spec.js
```diff
@@ -14,36 +17,66 @@ test.describe('Test Actions List block', () => {
...
+
+  test('Test the Manual Override', async ({page, admin, editor, requestUtils}) => {
+    const isNewIA = await isNewIAEnabled(admin, page);
+
+    // Get the "Take Action" page ID from the Analytics settings.
+    await admin.visitAdminPage('admin.php', 'page=planet4_settings_analytics');
+    const takeActionPageId = await page.locator('#take_action_page').inputValue();
+
+    // Skip Test if the new IA is not enabled or if the "Take Action" page ID is not available.
+    test.skip(!isNewIA || !takeActionPageId, 'The new IA must be enabled to run this test.');
```
mleray: Can we update this explanation message to also include the fact that a Take Action page must be selected in the settings?

tests/e2e/blocks/actions-list.spec.js
```diff
@@ -14,36 +17,66 @@ test.describe('Test Actions List block', () => {
...
+
+    for (const title of childActionTitles) {
+      await requestUtils.rest({
+        path: '/wp/v2/p4_action',
+        method: 'POST',
+        data: {title, status: 'publish', parent: takeActionPageId},
+      });
+    }
+
+    // Combine all action titles if needed downstream.
```
mleray: Can we remove "if needed downstream"? We know we need it later 😄

tests/e2e/blocks/actions-list.spec.js
```diff
@@ -14,36 +17,66 @@ test.describe('Test Actions List block', () => {
...
+    const actionTitles = [...regularActionTitles, ...childActionTitles];
+
+    // Create a page to hold the Actions List block.
+    await createPostWithFeaturedImage({page, admin, editor}, {title: 'Test Actions List, Manual Override', postType: 'page'});
+
+    // Add a Actions List block using the Manual Override to select the 2 actions created above.
```
mleray: We select more than 2 Actions right? Maybe we can just remove the number from this comment

tests/e2e/tools/lib/query-loop-utils.js
```diff
@@ -0,0 +1,78 @@
+import {searchAndInsertBlock} from './editor.js';
+import {expect} from './test-utils.js';
+
+export async function addListBlock(page, blockName, itemsPerPage, config = {}) {
```
mleray, line 12: Can we add docs to the functions in this file?

tests/e2e/tools/lib/posts-list.js
```diff
@@ -1,95 +1,42 @@
...
-  await editorSettings.locator(
-    '.components-form-token-field__suggestion', {hasText: TEST_CATEGORY}
-  ).click();
-  await expect(editorSettings.locator(
-    '.components-form-token-field__token-text', {hasText: TEST_CATEGORY})
-  ).toBeVisible();
-
-  // Change the title.
-  await page.getByRole('document', {name: 'Block: Heading'}).fill(TEST_TITLE);
+export async function addPostsListBlock(page, layout) {
```
mleray, line 15: Can we add docs to the functions in this file?

tests/e2e/tools/lib/actions-list.js
```diff
@@ -0,0 +1,32 @@
+import {addListBlock, addListBlockWithManualOverride, checkListBlock} from './query-loop-utils.js';
+
+const BLOCK_NAME = 'Actions List';
+const TEST_TITLE = 'Campaigns';
+const TEST_CATEGORY = 'Energy';
+const MANUAL_OVERRIDE_TITLE = 'Actions';
+
+export async function addActionsListBlock(page, layout) {
```
mleray, line 14: Can we add docs to the functions in this file?

tests/e2e/tools/lib/check-new-ia.js
```diff
@@ -0,0 +1,8 @@
+async function isNewIAEnabled (admin, page) {
```
mleray, line 7: Can we add docs?

tests/e2e/blocks/actions-list.spec.js
mleray: Hey @Osong-Michael you might have missed this comment 👀

tests/e2e/tools/lib/query-loop-utils.js
```diff
@@ -0,0 +1,100 @@
...
+    await editorSettings.locator(
+      '.components-form-token-field__suggestion', {hasText: config.category}
+    ).click();
+    await expect(editorSettings.locator(
+      '.components-form-token-field__token-text', {hasText: config.category})
+    ).toBeVisible();
+  }
+
+  if (config.title) {
+    await page.getByRole('document', {name: 'Block: Heading'}).fill(config.title);
```
mleray, line 35: With WordPress 7.0 this will need to be updated to `await editor.canvas.getByRole('document', {name: 'Block: Heading 2'})`
Hopefully it works with WordPress 6.9 too 🤞

tests/e2e/tools/lib/query-loop-utils.js
```diff
@@ -0,0 +1,100 @@
...
+    await editorSettings.locator('.components-form-token-field__input').fill(title);
+    await editorSettings.locator(
+      '.components-form-token-field__suggestion', {hasText: title}
+    ).first().click();
+    await expect(editorSettings.locator(
+      '.components-form-token-field__token-text', {hasText: title})
+    ).toBeVisible();
+  }
+
+  await page.getByRole('document', {name: 'Block: Heading'}).fill(overrideTitle);
```
mleray, line 68: Same comment here, with WordPress 7.0 we need to use `await editor.canvas.getByRole('document', {name: 'Block: Heading 2'})` instead

tests/e2e/blocks/actions-list.spec.js
Osong-Michael: No I updated the code to say the ID should be gotten from the Analytics settings page

mleray: Ah okay, but I wasn't talking about the comment (although it's good to have it be accurate as well) I was talking about the explanation message, currently `The new IA must be enabled to run this test.`

tests/e2e/tools/lib/query-loop-utils.js
mleray: If it doesn't work with WordPress 6.9 we can merge it as is and I will update it in my upgrade PR

tests/e2e/tools/lib/query-loop-utils.js
Osong-Michael: If that is the case we will need a different ticket to update the code for WP 7 then because we have a lot of test files that still use this selector

tests/e2e/blocks/actions-list.spec.js
Osong-Michael: Ok gotcha

tests/e2e/tools/lib/query-loop-utils.js
Osong-Michael: Since you are working on the upgrade won't it be better you just update it there along side any other code required. I can also test it on my other PR too.

Osong-Michael: Hey @mleray can you let me know if it works well on your end I tested on my and get the error below

  ```
1) [chromium] › tests/e2e/blocks/actions-list.spec.js:12:7 › Test Actions List block › Test the Grid layout 

    TimeoutError: locator.fill: Timeout 40000ms exceeded.
    Call log:
      - waiting for locator('[name="editor-canvas"]').contentFrame().getByRole('document', { name: 'Block: Heading' })


      33 |
      34 |     // Change the title.
    > 35 |     await editor.canvas.getByRole('document', {name: 'Block: Heading'}).fill(TEST_TITLE);
         |                                                                         ^
      36 |
      37 |     // Publish page.
      38 |     await publishPostAndVisit({page, editor});
```

tests/e2e/tools/lib/query-loop-utils.js
mleray: I'm making the changes in [my PR](https://github.com/greenpeace/planet4-master-theme/pull/3018), luckily it's not for all uses of this selector but "just" for the WYSIWYG blocks

tests/e2e/tools/lib/query-loop-utils.js
mleray: I'm making the changes in [my PR](https://github.com/greenpeace/planet4-master-theme/pull/3018), luckily it's not for all uses of this selector but "just" for the WYSIWYG blocks
As I said we can indeed merge this as is and I will update it afterwards

mleray (approved):
👍

PR #3003, opened 2026-06-08
Replace the HTML parser in the `HtmlPostProcessor` class

mardelnet:
### Summary

The `HtmlPostProcessor` parses the raw HTML with `DOMDocument`, which misinterprets some Vue's attributes and identifies them as invalid HTML. This is causing issues in some websites relying on Vue.

Taking that into consideration:
- To remove the "no post text" from the query block, this PR encapsulates the HTML that is parsed via `DOMDocument` only within the Query block, so the rest of the DOM remains untouched.
- To remove the "Related Posts" section when a post has no related posts, this PR refactors the `single.php`.
- To add CSS classes to PDF links and external links, this PR replaces `DOMDocument` with a combination of `WP_HTML_Tag_Processor` and CSS.

---

Ref: https://greenpeace-gpi.slack.com/archives/C0160MX64AG/p1780904946715889

### Testing

1. Check that the "Related Posts" section is removed from posts when it contains no related posts.
2. Check that the "No Results" inner block from a Post List or an Actions List block is removed (in the front) when the title and the description of the block are empty.
3. Check that the class `pdf-link` is added to the links connected to PDF files (pay attention to the exceptions!).
4. Check that the class external-link is added to the external links  (pay attention to the exceptions!).
5. Check that Vue elements remain intact. This can be tested by simulating a Vue template, as it was explained [here](https://greenpeace-gpi.slack.com/archives/C0160MX64AG/p1780907423459559?thread_ts=1780904946.715889&cid=C0160MX64AG).

### DIFF
--- a/src/Blocks/QueryLoopExtension.php
+++ b/src/Blocks/QueryLoopExtension.php
@@ -309,6 +309,9 @@ class QueryLoopExtension
      * Removes the "No Results" inner block from a Post List or an Actions List block
      * when the title and the description of the block are empty.
      *
+     * Removes the "See all Posts" link from a Post List or an Actions List block
+     * when the block has no posts.
+     *
      * @param string $block_content The HTML generated for the block.
      * @param array  $block         The block.
      *
@@ -330,22 +333,34 @@ class QueryLoopExtension
         libxml_clear_errors();
         $xpath = new DOMXPath($dom);
 
+        $posts_list = $xpath->query("//*[contains(@class, 'wp-block-post-template')]");
         $post_title = $xpath->query("//*[contains(@class, 'wp-block-heading')]");
         $post_description = $xpath->query(
             "//*[contains(@class, 'p4-query-loop')]//p[not(ancestor::*[contains(@class, 'wp-block-query-no-results')])]"
         );
 
+        $list_is_empty = $posts_list->length === 0 || trim($posts_list->item(0)->textContent) === '';
         $title_is_empty = $post_title->length === 0 || trim($post_title->item(0)->textContent) === '';
         $description_is_empty = $post_description->length === 0 || trim($post_description->item(0)->textContent) === '';
 
-        if (!$title_is_empty || !$description_is_empty) {
+        if (!$list_is_empty) {
             return $block_content;
         }
 
-        // Strip only the no-results block, return the rest intact
-        $no_results = $xpath->query("//*[contains(@class, 'wp-block-query-no-results')]");
-        foreach ($no_results as $node) {
-            $node->parentNode->removeChild($node);
+        // Strip the "see all posts" link, keep the rest intact.
+        if ($list_is_empty) {
+            $see_all_link = $xpath->query("//*[contains(@class, 'see-all-link')]");
+            foreach ($see_all_link as $node) {
+                $node->parentNode->removeChild($node);
+            }
+        }
+
+        // Strip the no-results block, keep the rest intact.
+        if ($title_is_empty && $description_is_empty) {
+            $no_results = $xpath->query("//*[contains(@class, 'wp-block-query-no-results')]");
+            foreach ($no_results as $node) {
+                $node->parentNode->removeChild($node);
+            }
         }
 
         return $dom->saveHTML();
--- a/src/Blocks/QueryLoopExtension.php
+++ b/src/Blocks/QueryLoopExtension.php
@@ -5,6 +5,8 @@ declare(strict_types=1);
 namespace P4\MasterTheme\Blocks;
 
 use WP_Block;
+use DOMDocument;
+use DOMXPath;
 
 /**
  * Enhances the query loop block with custom filtering logic for specific blocks.
@@ -36,6 +38,7 @@ class QueryLoopExtension
         add_filter('render_block_data', [self::class, 'track_posts_list_context'], 10, 1);
         add_filter('render_block', [self::class, 'add_data_attributes_to_posts_list_block'], 10, 2);
         add_filter('render_block', [self::class, 'add_data_attributes_to_actions_list_block'], 10, 2);
+        add_filter('render_block_core/query', [self::class, 'remove_no_post_text'], 10, 2);
     }
 
     /**
@@ -283,7 +286,6 @@ class QueryLoopExtension
      *
      * @return string The updated block content.
      */
-
     public static function add_data_attributes_to_actions_list_block(string $block_content, array $block): string
     {
         if (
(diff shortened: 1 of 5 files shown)


planet-4:

:watch: 2026.06.18 08:25:28

src/HtmlPostProcessor.php
```diff
@@ -2,206 +2,111 @@
...
-        $xpath = new DOMXPath($dom);
+        while ($processor->next_tag('a')) {
+            $href = $processor->get_attribute('href') ?? '';
+            $classes = $processor->get_attribute('class') ?? '';
 
-        $this->remove_related_section_no_posts($xpath);
-        $this->remove_no_post_text($xpath);
-        $this->setup_pdf_icon($xpath);
-        $this->setup_external_links($xpath);
+            if (empty($href)) {
```
mleray: Could we add here the check to also continue if the link is neither PDF or external?

src/HtmlPostProcessor.php
```diff
@@ -2,206 +2,111 @@
...
-
-            // Skip headings, image links, and empty links
-            if (in_array(strtoupper($parent_tag), ['H1', 'H2', 'H3', 'H4', 'H5', 'H6']) || $has_image || $text === '') {
-                continue;
-            }
-
-            $classes = $link->getAttribute('class');
-            $link->setAttribute('class', trim($classes . ' pdf-link'));
-            $link->setAttribute('title', __('This link will open a PDF file', 'planet4-master-theme'));
+        if (
```
mleray, line 91: Maybe here we could directly return the condition? Something like 
```
return !str_contains($href, $this->home_host) &&
            !str_contains($href, '.pdf') &&
            !str_starts_with($href, '/') && ...
```
That would also maybe get rid of this SonarCloud failure we get: `Reduce the number of conditional operators (6) used in the expression (maximum allowed 3)`

mleray: Ah, nevermind, I just saw there is an additional check afterwards so this suggestion is not possible 😬

src/HtmlPostProcessor.php
mardelnet: Done!

src/HtmlPostProcessor.php
```diff
@@ -2,206 +2,113 @@
...
-        foreach ($sections as $section) {
-            $has_post_template = $xpath->query(".//*[contains(@class, 'wp-block-post-template')]", $section);
+            if ($is_pdf) {
+                $processor->add_class('pdf-link');
+                $processor->set_attribute('title', __('This link will open a PDF file', 'planet4-master-theme'));
+                continue;
+            }
 
-            if ($has_post_template->length !== 0) {
+            if (!$is_external_link) {
```
mleray: This should no longer be necessary right? Since we now check for `!$is_pdf && !$is_external_link)` earlier

mleray (commented):
On [this post](https://www-dev.greenpeace.org/test-sinope/press/1113/duis-posuere-6/) I still see the `Related Posts` even though there are no posts

Screenshot 2026-06-16 at 14 02 39

src/HtmlPostProcessor.php
mleray: Probably just an `else` statement should be enough

mardelnet: Fixed!

mardelnet:
> On [this post](https://www-dev.greenpeace.org/test-sinope/press/1113/duis-posuere-6/) I still see the `Related Posts` even though there are no posts

Great catch, @mleray 
I think it's better now:
- https://www-dev.greenpeace.org/test-sinope/press/1113/duis-posuere-6/
- https://www-dev.greenpeace.org/test-sinope/press/1111/duis-posuere-5/
- https://www-dev.greenpeace.org/test-sinope/press/1109/duis-posuere-4/
- https://www-dev.greenpeace.org/test-sinope/press/1107/duis-posuere-3/
- https://www-dev.greenpeace.org/test-sinope/press/1105/duis-posuere-2/
- https://www-dev.greenpeace.org/test-sinope/press/263/duis-posuere/

mleray (commented):
On [this page](https://www-dev.greenpeace.org/test-sinope/about-us-2/) I've added a Posts List without posts and I still see the `See all posts` in the frontend

Screenshot 2026-06-17 at 10 32 06

mardelnet:
> On [this page](https://www-dev.greenpeace.org/test-sinope/about-us-2/) I've added a Posts List without posts and I still see the `See all posts` in the frontend

Nice catch, @mleray 
It seems that the `main` branch behaves the same way, so the issue was not incorporated by this PR. If so, I can create another ticket to tackle it.
Would you mind checking it?

mleray:
> Nice catch, @mleray It seems that the `main` branch behaves the same way, so the issue was not incorporated by this PR. If so, I can create another ticket to tackle it. Would you mind checking it?

@mardelnet you're right! Yes let's open a follow-up ticket.

src/Blocks/QueryLoopExtension.php
```diff
@@ -302,4 +304,50 @@ public static function add_data_attributes_to_actions_list_block(string $block_c
...
+
+        $title_is_empty = $post_title->length === 0 || trim($post_title->item(0)->textContent) === '';
+        $description_is_empty = $post_description->length === 0 || trim($post_description->item(0)->textContent) === '';
+
+        if (!$title_is_empty || !$description_is_empty) {
+            return $block_content;
+        }
+
+        // Strip only the no-results block, return the rest intact
+        $no_results = $xpath->query("//*[contains(@class, 'wp-block-query-no-results')]");
```
mleray: I know we talked about opening a separate ticket for the `See all posts` link, but it seems to me it would be relatively easy to add it here with a similar implementation? Since it has the `see-all-link` class. But if I'm mistaken and it's more complicated than that, feel free to ignore and open the follow-up ticket as planned.

mardelnet: Just in case, let's wait for confirmation [here](https://greenpeace-gpi.slack.com/archives/G015K63081W/p1781686307111929).

mleray (dismissed):
Looks good 👏 
Only the issue with the `See all posts` link remains, but as I said in my latest comment I'll leave it up to you to handle separately or not.

src/Blocks/QueryLoopExtension.php
mleray: Ah, I see, seems quite obvious to me but sure let's wait for their response!

mardelnet: @mleray, would you mind rechecking this PR?
I applied the removal of the "See all posts" link when required:
https://github.com/greenpeace/planet4-master-theme/pull/3003/changes/ebf0be2df1154814bb367091b94131ab6cb6730c

mleray (approved):
Looks good 👏

PR #3004, opened 2026-06-09
PLANET-8191 Fix publish validation and sidebar errors on wp_template,…

sagarsdeshmukh:
… and wp_template_part post types

### Summary
- Updated the validation logic for custom template post_type(custom templates are stored as wp_template & wp_template_part post_type) & allow it without validation checks
- Added `registeredSidebars = []` to keep track of the currently registered sidebars.
- On every post type change, all previously registered sidebars are unregistered using `wp.plugins.unregisterPlugin()` before any new sidebars are registered.
- For `wp_template` and `wp_template_part`, `sidebarsForPostType()` now returns `null`. As a result, existing sidebars are unregistered and no new sidebars are registered, preventing the editor crash.

---

Ref: 
https://greenpeace-planet4.atlassian.net/browse/PLANET-8191

### Testing

1. Add a new page/post click on template from page sidebar , add new template/edit existing template
2. the add/edit template works without validation error

### DIFF
--- a/assets/src/block-editor/setupBlockEditorValidations.js
+++ b/assets/src/block-editor/setupBlockEditorValidations.js
@@ -22,11 +22,11 @@ export const setupBlockEditorValidations = () => {
  */
 const getValidationState = select => {
   const {getEditedPostAttribute, getCurrentPostType} = select('core/editor');
-  // Reusable blocks / synced patterns are stored as the 'wp_block' post type.
-  // They're design artifacts, not editorial content, so the post-level
-  // requirements (title, featured image, topic link, valid forms) don't apply. Returning a
-  // fully-valid state stops these checks from blocking pattern create/edit.
-  if (getCurrentPostType() === 'wp_block') {
+  // Reusable blocks / synced patterns ('wp_block'), custom templates ('wp_template'),
+  // and template parts ('wp_template_part') are design artifacts, not editorial content,
+  // so the post-level requirements (title, featured image, topic link, valid forms) don't apply.
+  // Returning a fully-valid state stops these checks from blocking their create/edit.
+  if (['wp_block', 'wp_template', 'wp_template_part'].includes(getCurrentPostType())) {
     return {postTitle: true, featuredImage: true, topicLink: true, isValid: true, validForms: true};
   }
   const {getBlocks} = select('core/block-editor');
--- a/assets/src/block-editor/setupBlockEditorValidations.js
+++ b/assets/src/block-editor/setupBlockEditorValidations.js
@@ -25,20 +25,19 @@ const ALLOWED_POST_TYPES = ['post', 'page', 'p4_action', 'campaign'];
 
 const getValidationState = select => {
   const {getEditedPostAttribute, getCurrentPostType} = select('core/editor');
-  // Skip validation for post types not in the whitelist (e.g. synced patterns,
-  // templates, template parts) they are design artifacts, not editorial content.
-  if (!ALLOWED_POST_TYPES.includes(getCurrentPostType())) {
-    return {postTitle: true, featuredImage: true, topicLink: true, isValid: true, validForms: true};
-  }
   const {getBlocks} = select('core/block-editor');
+  // skip=true for post types not in the whitelist (e.g. synced patterns, templates,
+  // template parts) They are design artifacts, not editorial content, so all checks
+  // short-circuit to valid. New checks added below automatically respect this flag.
+  const skip = !ALLOWED_POST_TYPES.includes(getCurrentPostType());
   const allBlocks = getBlocks();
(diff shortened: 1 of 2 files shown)


planet-4:

:watch: 2026.06.15 06:17:36

assets/src/block-editor/setupBlockEditorValidations.js
```diff
@@ -22,11 +22,11 @@ export const setupBlockEditorValidations = () => {
...
-  // Reusable blocks / synced patterns are stored as the 'wp_block' post type.
-  // They're design artifacts, not editorial content, so the post-level
-  // requirements (title, featured image, topic link) don't apply. Returning a
-  // fully-valid state stops these checks from blocking pattern create/edit.
-  if (getCurrentPostType() === 'wp_block') {
+  // Reusable blocks / synced patterns ('wp_block'), custom templates ('wp_template'),
+  // and template parts ('wp_template_part') are design artifacts, not editorial content,
+  // so the post-level requirements (title, featured image, topic link) don't apply.
+  // Returning a fully-valid state stops these checks from blocking their create/edit.
+  if (['wp_block', 'wp_template', 'wp_template_part'].includes(getCurrentPostType())) {
```
mardelnet: Hey @sagarsdeshmukh!
Checking [this PR](https://github.com/greenpeace/planet4-master-theme/pull/3005/), I came to the idea that instead of returning an object with all its values set as `true` as we're doing on line 29, we might better skip the whole validation for any post type but the ones we want: post, pages, and actions.
Otherwise, we might be adding the validation incorrectly to post types created by external plugins or child themes.
What do you think?
cc @comzeradd

sagarsdeshmukh: Yes, I agree with the idea of introducing a whitelist of post types for which this validation should apply.
The whitelisted post types are post, pages, actions & Campaigns. right?

mardelnet: I think so, yes.

sagarsdeshmukh: PR updated 👍

assets/src/block-editor/setupBlockEditorValidations.js
```diff
@@ -20,13 +20,14 @@ export const setupBlockEditorValidations = () => {
  * @param {Function} select - The WordPress data `select` function used to access store selectors.
  * @return {Object} - An object containing individual validation flags and a combined `isValid` flag.
  */
+// Post types for which publish validation (title, featured image, topic link) is enforced.
+const ALLOWED_POST_TYPES = ['post', 'page', 'p4_action', 'campaign'];
+
 const getValidationState = select => {
```
mardelnet, line 26: Can we use something like this, so there is no need to update the early return (line 29 to 32) every time a new check is added (as we had to do [here](https://github.com/greenpeace/planet4-master-theme/pull/3005/changes))?

```js
const getValidationState = select => {
  const {getEditedPostAttribute, getCurrentPostType} = select('core/editor');
  const {getBlocks} = select('core/block-editor');
  const skip = !ALLOWED_POST_TYPES.includes(getCurrentPostType());
  const postTitle = skip || Boolean(getEditedPostAttribute('title'));
  const featuredImage = skip || Boolean(getEditedPostAttribute('featured_media'));
  const topicLink = skip || checkTopicLinks(getBlocks());
  return {postTitle, featuredImage, topicLink, isValid: postTitle && featuredImage && topicLink};
};
```

assets/src/block-editor/setupCustomSidebar.js
```diff
@@ -53,11 +54,18 @@ export const setupCustomSidebar = () => {
       }
 
       currentPostType = newPostType;
+
+      // Unregister sidebars from the previous post type so they don't render
+      // (and crash) when their expected meta fields are absent (e.g. wp_template).
+      registeredSidebars.forEach(sidebar => wp.plugins.unregisterPlugin(sidebar.getId()));
```
mardelnet, line 64: Is this used to remove the custom sidebars from patterns and other post types, or is it also used for something else? If so, can you extend the comment to explain that, because otherwise the change seems hard to understand for me.

assets/src/block-editor/setupBlockEditorValidations.js
sagarsdeshmukh: PR updated!

assets/src/block-editor/setupCustomSidebar.js
sagarsdeshmukh: I updated the comment message to explain why sidebar cleanup is needed on post type change. please have a look.

mleray (commented):
SonarCloud has a potentially interesting suggestion here:

Image

mardelnet (approved):
Nice job!

PR #3006, opened 2026-06-16
Update tests to ensure consistency when running locally and in CI

Osong-Michael:
### Summary
To ensure consistency in test suit

---

### Testing

You can run all tests locally with this branch and it should all pass 🤞

### DIFF
--- a/tests/e2e/blocks/actions-list.spec.js
+++ b/tests/e2e/blocks/actions-list.spec.js
@@ -41,10 +41,10 @@ test.describe('Test Actions List block', () => {
       const suffix = Date.now();
 
       // Create 2 actions to be selected via the Manual Override.
-      const regularActionTitles = [
-        `Test Actions List Manual Override Action 1 ${suffix}`,
-        `Test Actions List Manual Override Action 2 ${suffix}`,
-      ];
+      const regularActionTitles = Array.from(
+        {length: 2},
+        (_, i) => `Test Actions List Manual Override Action ${i + 1} ${suffix}`
+      );
 
       for (const title of regularActionTitles) {
         const created = await requestUtils.rest({
@@ -56,10 +56,10 @@ test.describe('Test Actions List block', () => {
       }
 
       // Create 2 actions as children of the "Take Action" page.
-      const childActionTitles = [
-        `Test Actions List Manual Override Child Action 1 ${suffix}`,
-        `Test Actions List Manual Override Child Action 2 ${suffix}`,
-      ];
+      const childActionTitles = Array.from(
+        {length: 2},
+        (_, i) => `Test Actions List Manual Override Child Action ${i + 1} ${suffix}`
+      );
 
       for (const title of childActionTitles) {
         const created = await requestUtils.rest({
--- a/tests/e2e/blocks/actions-list.spec.js
+++ b/tests/e2e/blocks/actions-list.spec.js
@@ -1,4 +1,4 @@
-import {test} from '../tools/lib/test-utils.js';
+import {test, expect} from '../tools/lib/test-utils.js';
 import {publishPostAndVisit, createPostWithFeaturedImage} from '../tools/lib/post.js';
 import {
   addActionsListBlock,
@@ -12,18 +12,15 @@ import {isNewIAEnabled} from '../tools/lib/check-new-ia.js';
 test.useAdminLoggedIn();
 
 test.describe('Test Actions List block', () => {
+  test.describe.configure({mode: 'serial'});
   // This is the default layout, so we don't need to select it manually.
   test('Test the Grid layout', async ({page, admin, editor}) => {
     test.slow();
-    await createPostWithFeaturedImage({page, admin, editor}, {title: 'Test Actions List Grid Layout', postType: 'page'});
+    const uniqueTitle = `Test Actions List Grid Layout ${Date.now()}`;
 
-    // Add Actions List block.
+    await createPostWithFeaturedImage({page, admin, editor}, {title: uniqueTitle, postType: 'page'});
     await addActionsListBlock({page, editor});
-
-    // Publish page.
     await publishPostAndVisit({page, editor});
-
-    // Check the Actions List block.
     await checkActionsListBlock(page);
   });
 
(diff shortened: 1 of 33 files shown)


planet-4:

:watch: 2026.08.18 18:47:49

tests/e2e/blocks/carousel-header.spec.js
```diff
@@ -5,69 +5,98 @@ import {searchAndInsertBlock} from '../tools/lib/editor.js';
...
+};
+
+/**
+ * Set the Carousel Header block's attributes directly via wp.data,
+ * bypassing the block editor UI entirely.
+ *
+ * @param {{Page}} page       Playwright page object
+ * @param {Object} attributes Block attributes to apply (carousel_autoplay, slides, currentImageIndex)
+ */
+const setCarouselAttributes = async (page, attributes) => {
```
mleray, line 63: I understand that this is probably solving a lot of problems, but we will now not know if the Carousel Header editor UI is broken 😕

tests/e2e/blocks/secondary-navigation.spec.js
```diff
@@ -32,25 +31,15 @@ test('Test Secondary Navigation block', async ({page, admin, editor}) => {
...
-  // Add content (headings and paragraphs).
-  for (let index = 0; index < HEADINGS.length; index++) {
-    await addHeadingOrParagraph({page, editor}, 'Heading', 'h2', i, HEADINGS[index]);
-    // For the paragraphs, we need to use i + 1 because there is a paragraph in the Page Header.
-    await addHeadingOrParagraph({page, editor}, 'Paragraph', 'p', i + 1, PARAGRAPH_CONTENT);
-
-    i++;
-  };
+  // // Add content (headings and paragraphs).
+  await addHeadingOrParagraph({editor, page}, content);
```
mleray: Maybe a nitpick, but should we rename this function? Since it now adds all the content at once?

tests/e2e/blocks/spreadsheet.spec.js
```diff
@@ -50,18 +50,16 @@ test('Test Spreadsheet block', async ({page, admin, editor}) => {
   await page.locator('button[aria-label="green"]').click();
   await expect(editorTable).toHaveClass(/is-color-green/);
 
+
```
mleray: Unnecessary added space

tests/e2e/blocks/topic-link.spec.js
```diff
@@ -10,15 +10,24 @@ test('Test Topic Link block', async ({page, admin, editor, baseURL}) => {
...
+  await searchAndInsertBlock({editor, page}, 'planet4-blocks/topic-link');
 
   // Select the Nature category
   await page.getByRole('combobox', {name: 'Select Category:'}).selectOption({label: CATEGORY});
 
   // Add a background image.
   await page.getByRole('button', {name: 'Select Background Image'}).click();
   const imageModal = page.getByRole('dialog', {name: 'Select or Upload Media'});
-  await imageModal.locator('[data-id="357"]').click();
+  await expect(imageModal.locator('[data-id]').first()).toBeVisible();
```
mleray, line 21: Can we create a function for this? It's used it at least twice

tests/e2e/tools/lib/editor.js
```diff
@@ -1,6 +1,74 @@
...
+      .getBlockVariations('core/query')
+      .find(item => item.name === name);
+
+    if (!variation) {
+      throw new Error(
+        `Block variation "${name}" not found`
+      );
+    }
+
+    const createBlocksFromTemplate = (blocks = []) =>
```
mleray, line 60: Can we add JS docs here?

tests/e2e/tools/lib/editor.js
```diff
@@ -21,91 +89,82 @@ async function openComponentPanel({page, editor}, panelTitle) {
...
-      );
-    }
-    return blocksList.getByRole('option', {name: blockName});
-  };
-
-  const blockOption = getBlockOption();
-
-  await expect(blockOption).toBeVisible();
-  await blockOption.click();
+  await editor.insertBlock({
```
mleray: This can be one line

tests/e2e/tools/lib/gravity-forms.js
```diff
@@ -114,14 +114,28 @@ const changeConfirmationType = async ({page, admin}, formId, label) => {
     'admin.php',
     `page=gf_edit_forms&view=settings&subview=confirmation&id=${formId}`
   );
-  await page.locator('#the-list > tr:first-child').hover();
-  await page.getByRole('link', {name: 'Edit', exact: true}).click();
+  const confrimationRow = await page.locator('#the-list > tr').first();
```
mleray: Small typo here, `confrimationRow` => `confirmationRow`

tests/e2e/blocks/carousel-header.spec.js
Osong-Michael: True, the reason I did this with this block in particular is because, during testing, I realised the content for the slide was not getting updated synchronously as the image, meaning even though we add the title, description, button and image, only the image will get added then the other attributes were not, instead Playwright for reason unknown to me, would batch the events then trigger them while the second slide was getting updated which meant crashed locators since the UI was getting re-rendered a lot. Actually got this work around from the WP Gutenberg tests, any interaction that causes a lot of re-renders, like adding blocks to pages, they used the API to prevent that since as it makes the test less flaky.

tests/e2e/tools/lib/media-library.js
```diff
@@ -34,7 +34,11 @@ async function deleteImageFromLibrary(page, imageId) {
 
   // Open image link via the sidebar.
   await expect(sidebar).toBeVisible();
-  await sidebar.locator('.sidebar-action').click();
+  await expect(sidebar.locator('.sidebar-action')).not.toHaveText('Import to Library', {timeout: 60000});
+
+  const sidebarAction = sidebar.locator('.sidebar-action');
```
mleray, line 39: Can we declare this earlier? Then we can use it also in line 37

tests/e2e/blocks/secondary-navigation.spec.js
Osong-Michael: This function still adds only the Headings and Paragraphs to the page for the Secondary navigation test

tests/e2e/tools/lib/post.js
```diff
@@ -11,34 +36,35 @@ import {expect} from './test-utils';
...
 /**
- * Updates a post and waits for the confirmation snackbar.
+ * Updates a post and waits both for the underlying save request to
+ * succeed and for the confirmation snackbar to appear.
  *
  * @param {Object} params      - Parameters for updating the post.
  * @param {Object} params.page - The page object representing the browser page.
  *
- * @return {Promise<void>} - A promise that resolves when the snackbar confirming the update is visible.
+ * @return {Promise<void>} - A promise that resolves when the update is confirmed.
```
mleray: Looks like this actually no longer returns anything, no?

tests/e2e/country-selector.spec.js
```diff
@@ -2,7 +2,25 @@ import {test, expect} from './tools/lib/test-utils.js';
...
+
+  await toggleButton.click();
+
+  const countriesList = page.locator('.countries-list');
+
+  // WebKit fix for click() event
+  try {
+    await expect(countriesList).toBeVisible({timeout: 5000});
+  } catch {
+    await toggleButton.dispatchEvent('click');
```
mleray, line 23: Can't we just use `dispatchEvent` for all browsers?

tests/e2e/footer.spec.js
```diff
@@ -11,19 +11,28 @@ test('check footer menu', async ({page}) => {
...
+    await menuSelect.selectOption({label: footerMenu});
 
-    const arrayOfLocators = page.locator('ul#menu-to-edit li');
-    const elementsCount = await arrayOfLocators.count();
+    await Promise.all([
+      page.waitForNavigation({waitUntil: 'domcontentloaded'}),
+      page.locator('span.submit-btn input.button').click(),
+    ]);
+
+    await expect(menuSelect).toHaveValue(await menuSelect.locator(`option:has-text("${footerMenu}")`).getAttribute('value'));
```
mleray, line 23: This line is confusing me a bit, can we maybe clarify it with some variables?

tests/e2e/blocks/topic-link.spec.js
Osong-Michael: It is used twice, but the context is different, and that matters with tests a lot as I found out.

tests/e2e/navigation-bar.spec.js
```diff
@@ -5,37 +5,58 @@ test.useAdminLoggedIn();
 test('Test navigation bar menu', async ({page, admin, requestUtils}) => {
   const testId = Math.floor(Math.random() * 10000); // NOSONAR
```
mleray: Maybe we can use `Date.now()` here as well for consistency with the other tests

tests/e2e/tools/lib/media-library.js
Osong-Michael: As I mentioned earlier context matters in testing, storing locators and re-using them after an event has had a few issues in my testing so the reason why I query the locator again so that incase the page reloads we don't have an issue

tests/e2e/blocks/actions-list.spec.js
```diff
@@ -32,65 +29,76 @@ test.describe('Test Actions List block', () => {
...
-        method: 'POST',
-        data: {title, status: 'publish', parent: takeActionPageId},
-      });
+    const createdActionIds = [];
+
+    try {
+      const suffix = Date.now();
+
+      // Create 2 actions to be selected via the Manual Override.
+      const regularActionTitles = [
```
GP-Dan-Tovbein: Just an idea, this can be also more dynamic
```suggestion
      const regularActionTitles = Array.from(
         { length: 2 },
         (_, i) => `Test Actions List Manual Override Action ${i + 1} ${suffix}`
       );
```

tests/e2e/blocks/actions-list.spec.js
```diff
@@ -32,65 +29,76 @@ test.describe('Test Actions List block', () => {
...
+        const created = await requestUtils.rest({
+          path: '/wp/v2/p4_action',
+          method: 'POST',
+          data: {title, status: 'publish'},
+        });
+        createdActionIds.push(created.id);
+      }
+
+      // Create 2 actions as children of the "Take Action" page.
+      const childActionTitles = [
```
GP-Dan-Tovbein: Same here


```suggestion
      const childActionTitles = Array.from(
  { length: 2 },
  (_, i) => `Test Actions List Manual Override Child Action ${i + 1} ${suffix}`
);
```

tests/e2e/country-selector.spec.js
Osong-Michael: This is a workaround for webkit only, I kept the normal `click` for the other browsers in case something happened in the future with that selector, we can use them as a control

tests/e2e/blocks/carousel-header.spec.js
mleray: Alright, I'm fine to keep it as is for now, since we need to prioritise fixing the tests asap. But I think in the future it would be nice to look again into testing the WYSIWYG! Pinging @comzeradd

tests/e2e/blocks/secondary-navigation.spec.js
mleray: Right, but it used to add one at a time hence `addHeadingOrParagraph`, now something like `addHeadingsAndParagraphs` or even simply `addContent` would make more sense

tests/e2e/blocks/topic-link.spec.js
mleray: Okay, I'll trust you on that one!

tests/e2e/country-selector.spec.js
mleray: Yes, but if `dispatchEvent` works for all browsers, it would be simpler to use that only no?

Osong-Michael: Yea it would, but it is like what you mentioned with the Carousel Block, testing UI elements, `dispatchEvent` will fire even if the element had an overlay an was not interactable, but `click()` ensures the element is visible and interactable before firing the event. I could still update it if you think it makes no difference.

mleray: No your reasoning makes sense, thank you!

tests/e2e/blocks/gallery.spec.js
```diff
@@ -26,10 +26,21 @@ test('Test Gallery basic functionalities', async ({page, admin, editor}) => {
...
+    .evaluateAll(items =>
+      items.map(item => item.dataset.id)
+    );
+
+  // Pick 3 available media items for the gallery block.
+  const imageId1 = mediaIds[50];
+  const imageId2 = mediaIds[51];
+  const imageId3 = mediaIds[52];
+
+  await imageModal.locator(`[data-id="${imageId1}"]`).click();
```
GP-Dan-Tovbein, line 41: Do you it's better if we include these 3 promises into one? Like `Promise.all([..., ..])` ?

tests/e2e/blocks/posts-list.spec.js
```diff
@@ -52,32 +53,52 @@ test.describe('Test Posts List block', () => {
...
-        method: 'POST',
-        data: {title, status: 'publish'},
-      });
+    const createdPostIds = [];
+
+    try {
+      const suffix = Date.now();
+
+      // Create 4 posts to be selected via the Manual Override.
+      const postTitles = [
```
GP-Dan-Tovbein: Same as https://github.com/greenpeace/planet4-master-theme/pull/3006/changes#r3802782466

tests/e2e/blocks/secondary-navigation.spec.js
```diff
@@ -61,9 +50,9 @@ test('Test Secondary Navigation block', async ({page, admin, editor}) => {
   await expect(secondaryNavigationBlock).toBeVisible();
 
   // The links in the block match all h2 elements present in the page content.
-  for (let index = 0; index < HEADINGS.length; index++) {
-    await expect(secondaryNavigationBlock.locator(NAV_LINK_CLASS).nth(index)).toHaveText(HEADINGS[index]);
-  };
+  for (let index = 0; index < content.length; index++) {
```
GP-Dan-Tovbein, line 53: What do you think to use `for (const item of array)` instead? I think it's cleaner and safer

tests/e2e/blocks/gallery.spec.js
Osong-Michael: These are not 3 separate calls, we are querying all the images from the media library DOM when the tab loads, saving that in an array and using that to assign images for the Gallery Block

tests/e2e/blocks/secondary-navigation.spec.js
Osong-Michael: We need the `index` for the Playwright locator we use.

tests/e2e/blocks/carousel-header.spec.js
comzeradd: Yes, ideally we do want to test the editor behavior as well. But for now, we can make an exception if we think that the CH block is often flaky in tests.

mleray (approved):
👏 👏 🥳

PR #2678, opened 2025-07-07
PLANET-7869: Return back `post__not_in` to main query

GP-Dan-Tovbein:
### Summary
We might need to rollback this functionality to avoid displaying sticky posts through main query. For somehow it's been removed removed [here](https://github.com/greenpeace/planet4-master-theme/commit/d1fa83eab88813ff4e32ea1c850954a5158bcb12#diff-01c9525afc2c87cfcb03124117c9d0ebb067029f6c955a607fa0fd4274aca556L392).

Ref: https://jira.greenpeace.org/browse/PLANET-7869

### Testing

1. Set at least 4 post as `sticky`
2. Then navigate to the News & Stories page
3. Expected bahaviour:
  3.a Featured posts must be displayed on top
  3.b Sticky posts that are displayed on top, shouldn't be displayed on bottom

Screen Shot 2025-07-07 at 11 17 07

PR #3011, opened 2026-06-19
PLANET-8024: optimize news and stories db queries

GP-Dan-Tovbein:
### Summary
This PR improves the DB Query performance of the News & Stories page `/news-stories`.

After the analysis I made, I detected one significant change that improves the performance. According to the `WP_Query` official documentation [see](https://developer.wordpress.org/reference/classes/wp_query/), it suggest to set `no_found_rows` as `true` which improves the performance. 

> Whether to skip counting the total rows found. Enabling can improve performance. Default false.

Here is also the [doc](https://github.com/WordPress/wordpress-develop/blob/7.0/src/wp-includes/class-wp-query.php#L718) from their Github repo.

I confirm that it help to reduce the performance since if not it includes `SELECT SQL_CALC_FOUND_ROWS`.

#### Before
```sql
SELECT SQL_CALC_FOUND_ROWS wp_posts.ID
FROM wp_posts
...
LIMIT 0, 4
```

#### After
```sql
SELECT wp_posts.ID
FROM wp_posts
...
LIMIT 0, 4
```
It reduces 1 DB query. 

👀 I've also left some inline comments.

---

Ref: [PLANET-8024](https://greenpeace-planet4.atlassian.net/browse/PLANET-8024)

### Testing
I'd suggest to validate the same page with the `main` branch too.

1. Run on local and make sure to enable at least 4 sticky posts to make the `news-stories` page as much as completed.
2. Open the `Query Monitor` plugin
3. Validate that it's being reduced one query (~117~ 116).

[PLANET-8024]: https://greenpeace-planet4.atlassian.net/browse/PLANET-8024?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


planet-4:

:watch: 2026.07.02 08:44:14

src/ListingPage.php
```diff
@@ -160,11 +160,12 @@ private function set_featured_posts(): void
...
-            'post_type' => 'post',
-            'post_status' => 'publish',
-            'fields' => 'ids',
-            'posts_per_page' => -1,
+            'post__not_in'        => $featured_post_ids,
+            'post_type'           => 'post',
+            'post_status'         => 'publish',
+            'fields'              => 'ids',
+            'posts_per_page'      => -1,
+            'no_found_rows'       => true,
```
GP-Dan-Tovbein:
```
@type bool   $no_found_rows   Whether to skip counting the total rows found. Enabling can improve performance. Default false.
```

mardelnet (commented):
@GP-Dan-Tovbein, question: what is the `featured_post_ids` variable used for in the `templates/featured-posts.twig` template?

GP-Dan-Tovbein:
> `featured_post_ids`

What we initially implemented is to exclude `sticky` posts from main list, because those should be shown ONLY  through the featured section. It was a complex implementation it's not possible to get them within `get_posts()`.

See this #2678 for more reference

mardelnet:
> > `featured_post_ids`
> 
> What we initially implemented is to exclude `sticky` posts from main list, because those should be shown ONLY through the featured section. It was a complex implementation it's not possible to get them within `get_posts()`.
> 
> See this #2678 for more reference

I think we don't need the query in the `set_featured_posts()` function, nor the `wp:query` block in the `templates/featured-posts.twig` file. We can get the featured posts data directly from the query in the `get_sticky_posts()` function, add it to the Timber context, and finally render it in the corresponding Twig file. I did some quick tests and the number of queries can be reduced by 10.

GP-Dan-Tovbein:
> > > `featured_post_ids`
> > 
> > 
> > What we initially implemented is to exclude `sticky` posts from main list, because those should be shown ONLY through the featured section. It was a complex implementation it's not possible to get them within `get_posts()`.
> > See this #2678 for more reference
> 
> I think we don't need the query in the `set_featured_posts()` function, nor the `wp:query` block in the `templates/featured-posts.twig` file. We can get the featured posts data directly from the query in the `get_sticky_posts()` function, add it to the Timber context, and finally render it in the corresponding Twig file. I did some quick tests and the number of queries can be reduced by 10.

Are you sure those sticky posts are not being rendered through the posts list, too?

mardelnet:
> > > > `featured_post_ids`
> > > 
> > > 
> > > What we initially implemented is to exclude `sticky` posts from main list, because those should be shown ONLY through the featured section. It was a complex implementation it's not possible to get them within `get_posts()`.
> > > See this #2678 for more reference
> > 
> > 
> > I think we don't need the query in the `set_featured_posts()` function, nor the `wp:query` block in the `templates/featured-posts.twig` file. We can get the featured posts data directly from the query in the `get_sticky_posts()` function, add it to the Timber context, and finally render it in the corresponding Twig file. I did some quick tests and the number of queries can be reduced by 10.
> 
> Are you sure those sticky posts are not being rendered through the posts list, too?

Yes, I am. The `pre_get_posts` filter in the `functions.php` file takes care of it.

GP-Dan-Tovbein:
> > > > > `featured_post_ids`
> > > > 
> > > > 
> > > > What we initially implemented is to exclude `sticky` posts from main list, because those should be shown ONLY through the featured section. It was a complex implementation it's not possible to get them within `get_posts()`.
> > > > See this #2678 for more reference
> > > 
> > > 
> > > I think we don't need the query in the `set_featured_posts()` function, nor the `wp:query` block in the `templates/featured-posts.twig` file. We can get the featured posts data directly from the query in the `get_sticky_posts()` function, add it to the Timber context, and finally render it in the corresponding Twig file. I did some quick tests and the number of queries can be reduced by 10.
> > 
> > 
> > Are you sure those sticky posts are not being rendered through the posts list, too?
> 
> Yes, I am. The `pre_get_posts` filter in the `functions.php` file takes care of it.

Let check it, then 👁️ . 

What do you mean about "`wp:query` block in the `templates/featured-posts.twig` file", the whole block or just the `eclude`?

mardelnet:
> > > > > > `featured_post_ids`
> > > > > 
> > > > > 
> > > > > What we initially implemented is to exclude `sticky` posts from main list, because those should be shown ONLY through the featured section. It was a complex implementation it's not possible to get them within `get_posts()`.
> > > > > See this #2678 for more reference
> > > > 
> > > > 
> > > > I think we don't need the query in the `set_featured_posts()` function, nor the `wp:query` block in the `templates/featured-posts.twig` file. We can get the featured posts data directly from the query in the `get_sticky_posts()` function, add it to the Timber context, and finally render it in the corresponding Twig file. I did some quick tests and the number of queries can be reduced by 10.
> > > 
> > > 
> > > Are you sure those sticky posts are not being rendered through the posts list, too?
> > 
> > 
> > Yes, I am. The `pre_get_posts` filter in the `functions.php` file takes care of it.
> 
> Let check it, then 👁️ .
> 
> What do you mean about "`wp:query` block in the `templates/featured-posts.twig` file", the whole block or just the `eclude`?

The whole block. We can pass the posts data directly to the context, there's no need to use the Query block.

GP-Dan-Tovbein:
> > > > > > > `featured_post_ids`
> > > > > > 
> > > > > > 
> > > > > > What we initially implemented is to exclude `sticky` posts from main list, because those should be shown ONLY through the featured section. It was a complex implementation it's not possible to get them within `get_posts()`.
> > > > > > See this #2678 for more reference
> > > > > 
> > > > > 
> > > > > I think we don't need the query in the `set_featured_posts()` function, nor the `wp:query` block in the `templates/featured-posts.twig` file. We can get the featured posts data directly from the query in the `get_sticky_posts()` function, add it to the Timber context, and finally render it in the corresponding Twig file. I did some quick tests and the number of queries can be reduced by 10.
> > > > 
> > > > 
> > > > Are you sure those sticky posts are not being rendered through the posts list, too?
> > > 
> > > 
> > > Yes, I am. The `pre_get_posts` filter in the `functions.php` file takes care of it.
> > 
> > 
> > Let check it, then 👁️ .
> > What do you mean about "`wp:query` block in the `templates/featured-posts.twig` file", the whole block or just the `eclude`?
> 
> The whole block. We can pass the posts data directly to the context, there's no need to use the Query block.

Could you share the code to test it locally, please? I'm not sure if I get your point

mardelnet:
> > > > > > > > `featured_post_ids`
> > > > > > > 
> > > > > > > 
> > > > > > > What we initially implemented is to exclude `sticky` posts from main list, because those should be shown ONLY through the featured section. It was a complex implementation it's not possible to get them within `get_posts()`.
> > > > > > > See this #2678 for more reference
> > > > > > 
> > > > > > 
> > > > > > I think we don't need the query in the `set_featured_posts()` function, nor the `wp:query` block in the `templates/featured-posts.twig` file. We can get the featured posts data directly from the query in the `get_sticky_posts()` function, add it to the Timber context, and finally render it in the corresponding Twig file. I did some quick tests and the number of queries can be reduced by 10.
> > > > > 
> > > > > 
> > > > > Are you sure those sticky posts are not being rendered through the posts list, too?
> > > > 
> > > > 
> > > > Yes, I am. The `pre_get_posts` filter in the `functions.php` file takes care of it.
> > > 
> > > 
> > > Let check it, then 👁️ .
> > > What do you mean about "`wp:query` block in the `templates/featured-posts.twig` file", the whole block or just the `eclude`?
> > 
> > 
> > The whole block. We can pass the posts data directly to the context, there's no need to use the Query block.
> 
> Could you share the code to test it locally, please? I'm not sure if I get your point

Sure: https://github.com/greenpeace/planet4-master-theme/compare/main...PLANET-8024_optimize-news-stories-db-queries-v2

GP-Dan-Tovbein:
> > > > > > > > > `featured_post_ids`
> > > > > > > > 
> > > > > > > > 
> > > > > > > > What we initially implemented is to exclude `sticky` posts from main list, because those should be shown ONLY through the featured section. It was a complex implementation it's not possible to get them within `get_posts()`.
> > > > > > > > See this #2678 for more reference
> > > > > > > 
> > > > > > > 
> > > > > > > I think we don't need the query in the `set_featured_posts()` function, nor the `wp:query` block in the `templates/featured-posts.twig` file. We can get the featured posts data directly from the query in the `get_sticky_posts()` function, add it to the Timber context, and finally render it in the corresponding Twig file. I did some quick tests and the number of queries can be reduced by 10.
> > > > > > 
> > > > > > 
> > > > > > Are you sure those sticky posts are not being rendered through the posts list, too?
> > > > > 
> > > > > 
> > > > > Yes, I am. The `pre_get_posts` filter in the `functions.php` file takes care of it.
> > > > 
> > > > 
> > > > Let check it, then 👁️ .
> > > > What do you mean about "`wp:query` block in the `templates/featured-posts.twig` file", the whole block or just the `eclude`?
> > > 
> > > 
> > > The whole block. We can pass the posts data directly to the context, there's no need to use the Query block.
> > 
> > 
> > Could you share the code to test it locally, please? I'm not sure if I get your point
> 
> Sure: [main...PLANET-8024_optimize-news-stories-db-queries-v2](https://github.com/greenpeace/planet4-master-theme/compare/main...PLANET-8024_optimize-news-stories-db-queries-v2)

I've just seen the code and it seems to be rendered the expected posts, however, I'm a bit doubt about if refactoring `templates/featured-posts.twig` without using the wp query it's the best approach. I mean we need to refactor the whole template using custom html. Do you think if it's still better ?

mardelnet:
> > > > > > > > > > `featured_post_ids`
> > > > > > > > > 
> > > > > > > > > 
> > > > > > > > > What we initially implemented is to exclude `sticky` posts from main list, because those should be shown ONLY through the featured section. It was a complex implementation it's not possible to get them within `get_posts()`.
> > > > > > > > > See this #2678 for more reference
> > > > > > > > 
> > > > > > > > 
> > > > > > > > I think we don't need the query in the `set_featured_posts()` function, nor the `wp:query` block in the `templates/featured-posts.twig` file. We can get the featured posts data directly from the query in the `get_sticky_posts()` function, add it to the Timber context, and finally render it in the corresponding Twig file. I did some quick tests and the number of queries can be reduced by 10.
> > > > > > > 
> > > > > > > 
> > > > > > > Are you sure those sticky posts are not being rendered through the posts list, too?
> > > > > > 
> > > > > > 
> > > > > > Yes, I am. The `pre_get_posts` filter in the `functions.php` file takes care of it.
> > > > > 
> > > > > 
> > > > > Let check it, then 👁️ .
> > > > > What do you mean about "`wp:query` block in the `templates/featured-posts.twig` file", the whole block or just the `eclude`?
> > > > 
> > > > 
> > > > The whole block. We can pass the posts data directly to the context, there's no need to use the Query block.
> > > 
> > > 
> > > Could you share the code to test it locally, please? I'm not sure if I get your point
> > 
> > 
> > Sure: [main...PLANET-8024_optimize-news-stories-db-queries-v2](https://github.com/greenpeace/planet4-master-theme/compare/main...PLANET-8024_optimize-news-stories-db-queries-v2)
> 
> I just seen the code and it seems to be rendered the expected posts, however, I'm a bit doubt about if refactoring `templates/featured-posts.twig` without using the wp query it's the best approach. I mean we need to refactor the whole template using custom html. Do you think if it's still better ?

I do. I think it's worth it if we can save 8 or 10 query calls. 
That will also make the code more maintainable: right now we have one WP Query to get the sticky posts, plus another WP Query to get all posts except the sticky ones, plus one Query block to render the sticky posts we already have from the first query.
What do you think?

GP-Dan-Tovbein:
@mardelnet OK, so we have2 different scenarios are available.

### Case 1 (apply your suggested changes)
- Reducing 9 queries BUT:
- Don't use the query block through the featured-posts template
- Apply a custom HTML
- Implement new styles to fit the current style
- Include more code into the `functions.php` which the idea was even to reduce it its complexity 
- More harder to maintain if we're not following wp native implementation


### Case 2 
Keep as it is (current PR) but without reducing 9 queries.

Am I correct?

Copying @comzeradd to see his point of view 👁️

comzeradd:
I like @mardelnet’s suggestion. Sure it ads a few more lines of code in some places but it’s definitely easier to understand compared to the current implementation. And we get lower db queries as a bonus.

GP-Dan-Tovbein:
> I like @mardelnet’s suggestion. Sure it ads a few more lines of code in some places but it’s definitely easier to understand compared to the current implementation. And we get lower db queries as a bonus.

Thanks both! Let's do that then.

src/ListingPage.php
```diff
@@ -159,19 +159,23 @@ private function set_featured_posts(): void
...
-            'post_status' => 'publish',
-            'fields' => 'ids',
-            'posts_per_page' => -1,
-        ]);
-
-        $excluded_post_ids = $featured_query->posts;
-
-        $this->context['featured_post_ids'] = $featured_post_ids;
+        $this->context['featured_post_ids'] = array_map(
+            fn($post) => [
```
GP-Dan-Tovbein: This is a new way of define anonymous functions in php 7. It helps to not create unnecessary variables

templates/featured-posts.twig
```diff
@@ -1,38 +1,50 @@
...
+                {% for post in featured_post_ids %}
+                    <li class="wp-block-post post-{{ post.ID }} post type-post status-publish format-standard has-post-thumbnail sticky hentry category-energy tag-renewables p4-page-type-press">
+                        <div class="query-list-item-image query-list-item-image-max-width is-layout-flow wp-block-post-template-is-layout-flow">
+                            <a href="{{ post.permalink }}" target="_self">
+                                <img width="2500" height="1281" src="{{ post.thumbnail }}" class="attachment- size- wp-post-image" alt="" sizes="(min-width: 1600px) 389px, (min-width: 1200px) 335px, (min-width: 1000px) 281px, (min-width: 780px) 209px, (min-width: 580px) 516px, calc(100vw - 24px)" decoding="async" fetchpriority="high" srcset="{{ post.thumbnail }} 2500w, {{ post.thumbnail }}-300x154.jpg 300w, {{ post.thumbnail }}-768x394.jpg 768w, {{ post.thumbnail }}-1024x525.jpg 1024w, {{ post.thumbnail }}-2048x1049.jpg 2048w, {{ post.thumbnail }}-510x261.jpg 510w">
+                            </a>
                         </div>
-                    </div>
+                        <div class="query-list-item-body">
```
GP-Dan-Tovbein: I've used the template provided by `wp:query`. This will help us to use "native" code and easy for maintain. In addition, we won't include extra CSS.

functions.php
```diff
@@ -283,7 +283,10 @@ function ($query): void {
             return;
         }
 
-        $query->set('post__not_in', ListingPage::get_sticky_posts());
+        $query->set('post__not_in', array_map(
```
GP-Dan-Tovbein: Same as https://github.com/greenpeace/planet4-master-theme/pull/3011/changes#diff-581a5e708dfbcacebf4404805030e096e2f2b48e0d96f2a036b0b66c1cb734d3R163

src/ListingPage.php
mardelnet: Can we update the variable name, since it now contains posts instead of IDs?

src/ListingPage.php
```diff
@@ -159,19 +159,23 @@ private function set_featured_posts(): void
...
-            'post_type' => 'post',
-            'post_status' => 'publish',
-            'fields' => 'ids',
-            'posts_per_page' => -1,
-        ]);
-
-        $excluded_post_ids = $featured_query->posts;
-
-        $this->context['featured_post_ids'] = $featured_post_ids;
+        $this->context['featured_post_ids'] = array_map(
```
mardelnet: Can we update the array key, since it now contains posts instead of IDs?

src/ListingPage.php
```diff
@@ -159,19 +159,23 @@ private function set_featured_posts(): void
...
+            fn($post) => [
+                'ID' => $post->ID,
+                'title' => $post->post_title,
+                'excerpt' => $post->post_excerpt,
+                'permalink' => get_permalink($post->ID),
+                'thumbnail' => has_post_thumbnail($post->ID) ? get_the_post_thumbnail($post->ID, 'medium') : null,
+                'terms' => wp_get_post_terms($post->ID, 'category', ['fields' => 'all']),
+                'tags' => wp_get_post_terms($post->ID, 'post_tag', ['fields' => 'all']),
```
mardelnet: These small adjustments reduce the number of queries by 8 units:

```suggestion
                'thumbnail' => get_the_post_thumbnail($post, 'medium') ?: null,
                'terms' => get_the_terms($post, 'category') ?: [],
                'tags' => get_the_terms($post, 'post_tag') ?: [],
```

src/ListingPage.php
mardelnet, line 77: The `set_news_page_link` function is calling `get_option('page_for_posts')`, which was already called previously on line 60. I think the whole function can be removed and something like this added as a replacement within the condition `if ($news_page_id === $current_page_id)` on line 63:
```php
$this->context['news_page_link'] = get_permalink($news_page_id);
```

functions.php
```diff
@@ -283,7 +283,10 @@ function ($query): void {
...
+            fn($post) => $post->ID,
+            ListingPage::get_sticky_posts()
```
mardelnet: Do you think it's possible to cache this call so the function `get_sticky_posts` is not called twice (here and by `ListingPage::set_featured_posts`)?

GP-Dan-Tovbein:
@mardelnet after [latest suggested changes](https://github.com/greenpeace/planet4-master-theme/pull/3011/changes/ec2f36168790fa3772921644c944796170dbd71c) I pushed now I'm getting `122` queries running on local whereas I got `119` and `104` against the `main` branch. Do you also get the same result?

#### Previous changes
Screenshot 2026-06-29 at 7 30 59 AM


#### Latest suggested changes
Screenshot 2026-06-29 at 7 22 17 AM

#### `main` branch
Screenshot 2026-06-29 at 7 27 24 AM

mardelnet:
> @mardelnet after [latest suggested changes](https://github.com/greenpeace/planet4-master-theme/pull/3011/changes/ec2f36168790fa3772921644c944796170dbd71c) I pushed now I'm getting `122` queries running on local whereas I got `119` and `104` against the `main` branch. Do you also get the same result?

@GP-Dan-Tovbein, in my local environment, after rebasing this PR I see:
- `main`: 196 queries
- This PR branch: 202 queries
- This PR branch with [these changes](https://github.com/greenpeace/planet4-master-theme/pull/3011#discussion_r3490154743): 194 queries

GP-Dan-Tovbein:
> > @mardelnet after [latest suggested changes](https://github.com/greenpeace/planet4-master-theme/pull/3011/changes/ec2f36168790fa3772921644c944796170dbd71c) I pushed now I'm getting `122` queries running on local whereas I got `119` and `104` against the `main` branch. Do you also get the same result?
> 
> @GP-Dan-Tovbein, in my local environment, after rebasing this PR I see:
> 
>     * `main`: 196 queries
> 
>     * This PR branch: 202 queries
> 
>     * This PR branch with [these changes](https://github.com/greenpeace/planet4-master-theme/pull/3011#discussion_r3490154743): 194 queries

That's weird. You get a completely results 🤔 
How do you test it? Because you get too many queries more than me.

`main`
you: 196
me: 102

94 DB queries less

current branch
me: 122
you: 202

80 DB queries less

suggested and applied changes
me: 119
you: 194 

75 DB queries less

Am I doing something wrong?

mardelnet:
> > > @mardelnet after [latest suggested changes](https://github.com/greenpeace/planet4-master-theme/pull/3011/changes/ec2f36168790fa3772921644c944796170dbd71c) I pushed now I'm getting `122` queries running on local whereas I got `119` and `104` against the `main` branch. Do you also get the same result?
> > 
> > 
> > @GP-Dan-Tovbein, in my local environment, after rebasing this PR I see:
> 
> That's weird. You get a completely results 🤔 How do you test it? Because you get too many queries more than me.

I am simply running the local instance with the Query Monitor plugin.
Perhaps we can install the plugin in the Proteus test instance so we can see the same output.

mardelnet:
@GP-Dan-Tovbein, in my local environment, I now see this (similar to you):

 - `main` branch: 111 queries
- This PR branch: 132 queries
- This PR branch + suggested changes: 124 queries

I was using the `WP Dummy Content Generator` plugin to create additional posts and had changed the limit of posts from 10 to 100. The combination of both was causing the difference we saw in the number of queries.

GP-Dan-Tovbein:
> `main` you: 196 me: 102
> 
> 94 DB queries less

Thanks Pedro! That makes more sense

src/ListingPage.php
GP-Dan-Tovbein: Done

src/ListingPage.php
GP-Dan-Tovbein: Done

src/ListingPage.php
GP-Dan-Tovbein: Done

functions.php
GP-Dan-Tovbein: Done

src/ListingPage.php
```diff
@@ -22,6 +22,9 @@ class ListingPage
     /** @static int */
     public static int $STICKY_POSTS_TO_SHOW = 4;
 
+    /** @var array|null $STICKY_POSTS_CACHE used to run sticky posts twice or more. */
```
mardelnet:
```suggestion
    /** @var array|null $STICKY_POSTS_CACHE used to avoid getting the sticky posts more than once. */
```

src/ListingPage.php
```diff
@@ -38,6 +41,8 @@ public function __construct(array $templates = [ 'index.twig' ], array $context
         $this->templates = $templates;
         $this->context = $context;
 
+        $this->news_page_id = (int) get_option('page_for_posts');
```
mardelnet, line 47: This property is undefined.

templates/featured-posts.twig
mardelnet, line 1: The spacing has to be slightly adjusted:

| Main Branch                                                                                                                         | This PR                                                                                                                             |
|-------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------|
| Image | Image |

src/ListingPage.php
GP-Dan-Tovbein: This happens because I'm not assigning before the constructor.

templates/featured-posts.twig
GP-Dan-Tovbein: This happens because I didn't close the `` thumbnail and then a `<br>` it's been rendered.

mardelnet: I still see this problem: https://www-dev.greenpeace.org/test-proteus/news-stories/

As an alternative, I could finally find a way to filter the sticky posts while keeping the WP Query block. You can check it [here](https://github.com/greenpeace/planet4-master-theme/compare/PLANET-8024_optimize-news-stories-db-queries...PLANET-8024_optimize-news-stories-db-queries-v2). I leave it to you whether to switch to that implementation or keep the current one, as you have already worked a lot on this one.

GP-Dan-Tovbein: @mardelnet what's about the DB queries? 
Because that's was the main reason to don't use `wp:query` in templates.

mardelnet: According to my testing, the number of queries per branch is:

- main: 118
- This PR: 112
- The alternative PR: 116

So this PR is better at reducing query calls, but it replaces the block with custom HTML.
So for me, whichever option you choose will be fine, we cannot get the best of both worlds in this case.

GP-Dan-Tovbein: BTW, I see the layout issue fixed on my local

Screenshot 2026-07-01 at 8 08 35 AM

mardelnet: The separation between elements is still too big compared with the main branch.
Check the first image here: https://github.com/greenpeace/planet4-master-theme/pull/3011#discussion_r3497072565

GP-Dan-Tovbein: > According to my testing, the number of queries per branch is:
> 
>     * main: 118
> 
>     * This PR: 112
> 
>     * The alternative PR: 116
> 
> 
> So this PR is better at reducing query calls, but it replaces the block with custom HTML. So for me, whichever option you choose will be fine, we cannot get the best of both worlds in this case.

Ok! That was my initial concern to re-implement the whole template.. initially was about to reduce 9 queries but now the result it's completely different.

GP-Dan-Tovbein: @mardelnet FYI, I've just updated the PR

mardelnet (approved):
Amazing job, @GP-Dan-Tovbein!

PR #3029, opened 2026-07-02
PLANET-8207 Support multiple Timeline blocks

mleray:
### Summary

Ref: [PLANET-8207](https://greenpeace-planet4.atlassian.net/browse/PLANET-8207)

### Testing

In the frontend, make sure that the years navigation and scrolling behaviour work as expected.

On local:
1. Enable the new design for the block via `Planet 4 > Features`.
2. Add at least 2 Timeline blocks (you can use [this sheet](https://docs.google.com/spreadsheets/d/1tYlLd_Fx0T_7ZEaf2y9dLfRnr5HzEOW_s0wELp5-j4s/) for example) to a page and save.

Alternatively, you can use [this page](https://www-dev.greenpeace.org/test-mars/timelines-2/) from the test instance.

[PLANET-8207]: https://greenpeace-planet4.atlassian.net/browse/PLANET-8207?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/src/blocks/Timeline/NewTimelineEditorScript.js
+++ b/assets/src/blocks/Timeline/NewTimelineEditorScript.js
@@ -5,9 +5,15 @@ import {getUniqueId} from '../../functions/getUniqueId';
 
 const {InspectorControls, RichText} = wp.blockEditor;
 const {PanelBody, Tooltip} = wp.components;
+import {select} from '@wordpress/data';
 const {debounce} = wp.compose;
 const {useCallback, useState, useEffect} = wp.element;
 const {__} = wp.i18n;
+
+const isTimelineIdReserved = timelineId => {
+  const timelines = select('core/editor').getBlocks().filter(block => block.name === 'planet4-blocks/timeline');
+  return timelines.some(timeline => timeline.attributes.timeline_id === timelineId);
+};
 
 const renderEdit = (
   sheetURL,
@@ -83,11 +89,9 @@ export const NewTimelineEditor = ({isSelected, attributes, setAttributes}) => {
 
   // Set a timeline id in the attributes, if it doesn't exist yet.
   useEffect(() => {
-    if (attributes.timeline_id) {
-      return;
+    if (!attributes.timeline_id || isTimelineIdReserved(attributes.timeline_id)) {
+      setAttributes({timeline_id: getUniqueId('tl')});
     }
-
-    setAttributes({timeline_id: getUniqueId('tl')});
   }, []);
 
   return (
--- a/assets/src/blocks/Timeline/NewTimelineEditorScript.js
+++ b/assets/src/blocks/Timeline/NewTimelineEditorScript.js
@@ -10,9 +10,28 @@ const {debounce} = wp.compose;
 const {useCallback, useState, useEffect} = wp.element;
 const {__} = wp.i18n;
 
+/**
+ * Get all blocks and inner blocks from a page.
+ *
+ * @param {Array} blocks - an array of blocks.
+ *
+ * @return {Array} All blocks from the page.
+ */
+const flattenBlocks = blocks => blocks.flatMap(block => [block, ...flattenBlocks(block.innerBlocks || [])]);
+
+/**
+ * Check if the given timeline id already exists on this page.
+ * This can happen if we duplicate a block for example.
+ *
+ * @param {string} timelineId - the Timeline block id.
+ *
+ * @return {boolean} Whether the timeline id is already given or not.
+ */
 const isTimelineIdReserved = timelineId => {
-  const timelines = select('core/editor').getBlocks().filter(block => block.name === 'planet4-blocks/timeline');
-  return timelines.some(timeline => timeline.attributes.timeline_id === timelineId);
+  const allBlocks = flattenBlocks(select('core/editor').getBlocks());
+  return allBlocks.some(
+    block => block.name === 'planet4-blocks/timeline' && block.attributes.timeline_id === timelineId
+  );
 };
 
 const renderEdit = (
--- a/assets/src/blocks/Timeline/NewTimelineEditorScript.js
+++ b/assets/src/blocks/Timeline/NewTimelineEditorScript.js
@@ -1,11 +1,12 @@
 import {URLInput} from '../../block-editor/URLInput/URLInput';
 import {NewTimelineFrontend} from './NewTimelineFrontend';
 import {URLDescriptionHelp} from './URLDescriptionHelp';
+import {getUniqueId} from '../../functions/getUniqueId';
 
 const {InspectorControls, RichText} = wp.blockEditor;
 const {PanelBody, Tooltip} = wp.components;
 const {debounce} = wp.compose;
-const {useCallback, useState} = wp.element;
+const {useCallback, useState, useEffect} = wp.element;
 const {__} = wp.i18n;
 
 const renderEdit = (
@@ -79,6 +80,15 @@ export const NewTimelineEditor = ({isSelected, attributes, setAttributes}) => {
   // Using a state to prevent the input losing the cursor position, a React issue reported multiple times
   const [sheetURL, setSheetURL] = useState(attributes.google_sheets_url);
   const debounceSheetURLUpdate = useCallback(debounce(toAttribute('google_sheets_url'), 300), []);
+
+  // Set a timeline id in the attributes, if it doesn't exist yet.
+  useEffect(() => {
+    if (attributes.timeline_id) {
+      return;
+    }
+
+    setAttributes({timeline_id: getUniqueId('tl')});
+  }, []);
 
   return (
     <>
--- a/assets/src/blocks/Timeline/NewTimelineFrontend.js
+++ b/assets/src/blocks/Timeline/NewTimelineFrontend.js
@@ -9,6 +9,11 @@ const getMonthName = monthNumber => {
     .format(new Date(2000, monthNumber - 1));
 };
 
+const uniqueId = () => {
+  const r = Math.floor(Math.random() * 10000);
+  return `timeline-${r}`;
+};
+
 const getLocalizedDate = (day, month) => {
   const locale = document.documentElement.lang || 'en';
 
@@ -50,6 +55,7 @@ export const NewTimelineFrontend = ({attributes}) => {
   const [loading, setLoading] = useState(false);
   const [sheetData, setSheetData] = useState(null);
   const [processedSheetData, setProcessedSheetData] = useState(null);
+  const [timelineId, setTimelineId] = useState('');
 
 
   const TimelineEvent = ({event}) => {
@@ -217,7 +223,9 @@ export const NewTimelineFrontend = ({attributes}) => {
     }
   }
 
-  if (loading || !processedSheetData) {
+  useEffect(() => setTimelineId(uniqueId()), []);
+
+  if (loading || !processedSheetData || !timelineId) {
     return null;
   }
 
@@ -231,7 +239,7 @@ export const NewTimelineFrontend = ({attributes}) => {
   );
 
   return (
-    <section className={`block timeline-block new-timeline-block ${className ?? ''} alignfull`} aria-label={summaryText}>
+    <section id={timelineId} className={`block timeline-block new-timeline-block ${className ?? ''} alignfull`} aria-label={summaryText}>
       <div className="container">
         {!!timeline_title && !isEditing &&
           <h2 className="page-section-header text-center">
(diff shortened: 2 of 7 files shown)


planet-4:

:watch: 2026.07.23 12:43:26

mardelnet (commented):
@mleray 
I added multiple Timeline blocks without sheet URLs on any.
Then, when the URL was added to the first block, all the blocks rendered the same content. Would you mind checking this?

assets/src/blocks/Timeline/NewTimelineFrontend.js
```diff
@@ -217,7 +223,9 @@ export const NewTimelineFrontend = ({attributes}) => {
     }
   }
 
-  if (loading || !processedSheetData) {
+  useEffect(() => setTimelineId(uniqueId()), []);
```
mardelnet: Do we need a separate function for this?

assets/src/blocks/Timeline/NewTimelineFrontend.js
```diff
@@ -9,6 +9,11 @@ const getMonthName = monthNumber => {
     .format(new Date(2000, monthNumber - 1));
 };
 
+const uniqueId = () => {
+  const r = Math.floor(Math.random() * 10000);
+  return `timeline-${r}`;
```
mardelnet: This is a super picky and bizarre edge case, but thanks to the [birthday paradox](https://en.wikipedia.org/wiki/Birthday_problem), we would hit a 50% chance of generating the same ID if "only" 120 timelines are added to the same page.

```suggestion
  return `timeline-${Date.now()}-${r}`;
```

mleray:
> @mleray I added multiple Timeline blocks without sheet URLs on any. Then, when the URL was added to the first block, all the blocks rendered the same content. Would you mind checking this?

@mardelnet I don't think this is what is happening, if you see [this page](https://www-dev.greenpeace.org/test-mars/timelines-2-0/) I created, in the frontend only one block is rendering because it's the only one that has a URL. In the editor however it looks confusing because the second Timeline block appears kind of on top of the previous one (hope that makes sense 😬) I'm guessing that we are setting some sort of max height for the block but it isn't well applied, maybe we need to hide the overflow or something like that?

Screenshot 2026-07-10 at 12 22 21

assets/src/blocks/Timeline/NewTimelineFrontend.js
mleray: But wouldn't `Date.now()` be the same for both blocks? If they are rendered at the same time?

assets/src/blocks/Timeline/NewTimelineFrontend.js
mleray: For what? Generating the id?

mardelnet: Yes, can't we just do something like this?
```js
useEffect(() => {
    const r = Math.floor(Math.random() * 10000);
    setTimelineId(`timeline-${r}`());
  }, []);
```

assets/src/blocks/Timeline/NewTimelineFrontend.js
mardelnet: You're probably right. You can forget this.

mardelnet:
> > @mleray I added multiple Timeline blocks without sheet URLs on any. Then, when the URL was added to the first block, all the blocks rendered the same content. Would you mind checking this?
> 
> @mardelnet I don't think this is what is happening, if you see [this page](https://www-dev.greenpeace.org/test-mars/timelines-2-0/) I created, in the frontend only one block is rendering because it's the only one that has a URL. In the editor however it looks confusing because the second Timeline block appears kind of on top of the previous one (hope that makes sense 😬) I'm guessing that we are setting some sort of max height for the block but it isn't well applied, maybe we need to hide the overflow or something like that?

If that's the case, we need a solution for the editor.
I wonder if rendering the block in the editor is actually necessary since there are no internal components that can be edited, right? Maybe we can simply use a placeholder.

mleray:
> If that's the case, we need a solution for the editor. I wonder if rendering the block in the editor is actually necessary since there are no internal components that can be edited, right? Maybe we can simply use a placeholder.

@mardelnet we do need to edit the title and description, for the rest we could indeed not show everything. But I think it's nice to get at least an idea of what it'll look like in the frontend, I'll look into making sure there is a respected max height.

assets/src/blocks/Timeline/NewTimelineFrontend.js
```diff
@@ -9,6 +9,11 @@ const getMonthName = monthNumber => {
     .format(new Date(2000, monthNumber - 1));
 };
 
+const uniqueId = () => {
```
mardelnet: @mleray 
I think that generating an ID every time the block is rendered won't work.
Let's say editors want to share a link to a specific part of the timeline. It won't be possible: the link shared will contain an ID that will hardly match the one of the recipient.
We should consider another solution, maybe based on the post ID and something else. What do you think?

mleray: That's a very good point. Using the post id is a good idea but I'm not sure what else to add in case there are multiple blocks, as you said a random number will change every time and simply using a number (1 2 3 etc) would not be reliable in case editors change the order of the blocks 🤔

mardelnet: What about the block's `clientId`?
https://github.com/WordPress/gutenberg/issues/25195

mleray: I've found this hook [useId](https://react.dev/reference/react/useId) that seems to do exactly what we need, but I've tried it on local and it's not working properly, maybe it's a React version issue...

mardelnet: Maybe we can add a new attribute to this block (ID), auto-generate a value only the first time the block is added, and populate the attribute with it.

assets/src/scss/blocks/Timeline/TimelineEditorStyle.scss
```diff
@@ -8,6 +8,9 @@ div[data-type="planet4-blocks/timeline"] {
 }
 
 .new-timeline-block {
+  max-height: 800px;
```
mardelnet: It seems that using a fixed value creates an empty space.
Switching to `100%` seems to fix it:

```suggestion
  max-height: 100%;
```

Image

assets/src/blocks/Timeline/NewTimelineFrontend.js
mleray: That might work! Would require a migration for existing blocks 🤔

mardelnet: Is the new Timeline already being used?

mleray: No, but it's the same block as the old one, just re-designed, os if we add a new attribute we will need to update existing blocks

assets/src/blocks/Timeline/NewTimelineFrontend.js
mleray: I've updated the PR, please have a look 🙏

assets/src/blocks/Timeline/NewTimelineFrontend.js
mleray: I've updated the PR, and also think I found a way to update existing blocks without an actual migration script, by using the `deprecated` functions... let me know what you think!

mardelnet (commented):
@mleray 
Nice job adding the unique ID!
I have some issues on my local, though.
Can you try to reproduce them?

<hr>

1. Switch to the `main` branch.
2. Create a page and add a few Timeline blocks linked to a sheet.
3. Switch to this PR branch.
4. Check the page editor. The blocks should be broken and must be recovered.
5. Check the page front. The blocks should be empty.

<hr>

1. Switch to this PR branch.
2. Go to the front of a page that contains a Timeline block.
3. Click on a year and copy the URL that contains the unique ID.
4. Open a new browser tab and go to the copied URL.
5. Check that you are not scrolled to the corresponding year.

mleray:
> @mleray Nice job adding the unique ID! I have some issues on my local, though. Can you try to reproduce them?

@mardelnet I will check the first scenario, however for the second one it is a known bug with the new implementation and there is already a [ticket](https://greenpeace-planet4.atlassian.net/browse/PLANET-8206) for it!

mleray:
@mardelnet I'm not sure that we need to worry about the issue you raised, since it seems we will have to manually update existing blocks anyway? And this one in particular is about updating blocks that already use the new design, which shouldn't happen

mardelnet:
> @mardelnet I'm not sure that we need to worry about the issue you raised, since it seems we will have to manually update existing blocks anyway? And this one in particular is about updating blocks that already use the new design, which shouldn't happen

Are you talking about this, @mleray?

- Switch to the main branch.
- Create a page and add a few Timeline blocks linked to a sheet.
- Switch to this PR branch.
- Check the page editor. The blocks should be broken and must be recovered.
- Check the page front. The blocks should be empty.

mleray:
> > @mardelnet I'm not sure that we need to worry about the issue you raised, since it seems we will have to manually update existing blocks anyway? And this one in particular is about updating blocks that already use the new design, which shouldn't happen
> 
> Are you talking about this, @mleray?
> 
> * Switch to the main branch.
> * Create a page and add a few Timeline blocks linked to a sheet.
> * Switch to this PR branch.
> * Check the page editor. The blocks should be broken and must be recovered.
> * Check the page front. The blocks should be empty.

@mardelnet yes

mardelnet (commented):
@mleray:
Could you please check [this page](https://www-dev.greenpeace.org/test-mars/timeline-test)?
If you scroll down slowly, at a certain point, you'll be automatically scrolled back to the previous year.

mleray:
> @mleray: Could you please check [this page](https://www-dev.greenpeace.org/test-mars/timeline-test)? If you scroll down slowly, at a certain point, you'll be automatically scrolled back to the previous year.

@mardelnet I think I know what the problem is, when you duplicate the block you keep the id and therefore when scrolling you always go back to the first block with that id 😬 that's a problem for sure, we need to find a way to detect duplication and not copy the id

assets/src/blocks/Timeline/NewTimelineEditorScript.js
```diff
@@ -1,13 +1,20 @@
...
+import {select} from '@wordpress/data';
 const {debounce} = wp.compose;
-const {useCallback, useState} = wp.element;
+const {useCallback, useState, useEffect} = wp.element;
 const {__} = wp.i18n;
 
+const isTimelineIdReserved = timelineId => {
+  const timelines = select('core/editor').getBlocks().filter(block => block.name === 'planet4-blocks/timeline');
+  return timelines.some(timeline => timeline.attributes.timeline_id === timelineId);
+};
```
mardelnet, line 35: If the Timeline is nested, we will need a recursive function (haven't tested it 100% yet).
Also, can you add JS Docs, please?

```suggestion
const flattenBlocks = blocks =>
  blocks.flatMap(block => [block, ...flattenBlocks(block.innerBlocks || [])]);

const isTimelineIdReserved = timelineId => {
  const allBlocks = flattenBlocks(select('core/editor').getBlocks());
  return allBlocks.some(
    block => block.name === 'planet4-blocks/timeline' && block.attributes.timeline_id === timelineId
  );
};
```

mardelnet (commented):
I am testing the Timeline with these spreadsheets ([1](https://docs.google.com/spreadsheets/d/e/2PACX-1vT3tXZ14L09LosXjsQT2beRGpgjgXUJilvgbmVfY2oBzKY0aV2bc2WLwUkeOcRj_-by9T4iMWAgxOkg/pubhtml) and [2](https://docs.google.com/spreadsheets/d/e/2PACX-1vRqvZPeiB2Wf1MLRUItQBFZC_cwTc0YK4gPlcUUe3W6tNkdvYuXG9DUVa2ZVioOjsKaVijSEmkQJ65-/pubhtml)), and none of them work.
The issue was probably not introduced by this PR, but the block fails silently.
Should we open a PR to at least give some feedback to the user to better understand what's wrong?

assets/src/blocks/Timeline/NewTimelineEditorScript.js
mleray: Updated!

mardelnet (approved):
Well done!

PR #3031, opened 2026-07-08
Adjust registration emails and forms when enforcing SSO

comzeradd:
## Summary

Small adjustments to registration forms and email, based on the recent changes with SSO

- Hide password field on new user registration form.
- Hide password field on user edit form.
- Adjust new user notification email to remove password references.
- Add website url on new user notification email sent to admin.

### Testing

Testing emails locally is not easy unless you tap on `wp_mail()`. But I temporarily switched `enforce_sso` to the test instance assigned.

Login to the test instance:
1. There should be no password field when you edit your profile.
2. There should be no password field when creating a new user.
3. Create a new user, use your email address with a plus sign alias (eg. `myusername+testing@`) so you can get the registration email. Leave the option for email notification on. Check the email you got.

### DIFF
--- a/src/RegistrationHandler.php
+++ b/src/RegistrationHandler.php
@@ -21,6 +21,8 @@ class RegistrationHandler
     private function hooks(): void
     {
         add_filter('wp_new_user_notification_email_admin', [$this, 'add_site_url_to_admin_notification_email'], 10, 3);
+        add_action('user_profile_update_errors', [$this, 'validate_password_policy'], 10, 3);
+        add_action('validate_password_reset', [$this, 'validate_password_reset'], 10, 2);
 
         $features = get_option('planet4_features', []);
         $enforce_sso = !empty($features['enforce_sso']);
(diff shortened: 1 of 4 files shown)


planet-4:

:watch: 2026.07.10 10:25:51

mardelnet (commented):
@comzeradd, with this PR, are these changes still necessary: https://github.com/greenpeace/planet4-master-theme/pull/2736/changes?

comzeradd:
> @comzeradd, with this PR, are these changes still necessary: https://github.com/greenpeace/planet4-master-theme/pull/2736/changes?

I'm leaning towards keeping it, just to protect the few instances where SSO is not enforced.

src/RegistrationHandler.php
```diff
@@ -0,0 +1,96 @@
...
+     * phpcs:disable SlevomatCodingStandard.Functions.UnusedParameter
+     */
+    public function hide_password_reset_on_new_user_notification_email(
+        array $email,
+        \WP_User $user,
+        string $blogname
+    ): array {
+        $email['message'] = "Hi {$user->user_login},\n\n" .
+            "Your account is ready.\n" .
+            "Please sign in using your account at:\n" .
```
mardelnet: Should it be translatable?

src/RegistrationHandler.php
```diff
@@ -0,0 +1,96 @@
...
+        // Hide password fields from user profile edit screen.
+        add_filter('show_password_fields', '__return_false');
+    }
+
+    /**
+    * Add site URL to admin notification email.
+    *
+    * @param array $email User email data.
+    * @param WP_User $user User object.
+    * @param string $blogname Site name name.
```
mardelnet:
```suggestion
    * @param string $blogname Site name.
```

mardelnet:
> > @comzeradd, with this PR, are these changes still necessary: https://github.com/greenpeace/planet4-master-theme/pull/2736/changes?
> 
> I'm leaning towards keeping it, just to protect the few instances where SSO is not enforced.

Cool! Optional suggestions:

Perhaps we could move the logic from the MasterSite (`validate_password_policy` and `validate_password_reset`) to the class you've created, since they are all part of the same solution. 

We could also use the `assets/src/js/hide_weak_pw_checkbox.js` file to hide the password field instead of the `hide_password_field_on_registration_form` function, so we don't have to inject CSS.

src/RegistrationHandler.php
```diff
@@ -0,0 +1,96 @@
...
+     *
+     * @return array Modified email data.
+     * phpcs:disable SlevomatCodingStandard.Functions.UnusedParameter
+     */
+    public function hide_password_reset_on_new_user_notification_email(
+        array $email,
+        \WP_User $user,
+        string $blogname
+    ): array {
+        $email['message'] = "Hi {$user->user_login},\n\n" .
```
mardelnet: Is it okay to completely replace the message instead of augmenting it?

src/RegistrationHandler.php
mardelnet: This message sounds a bit confusing to me, since it mentions the "account" twice, but it's referring to two different accounts: the WordPress one, and the GP email one, right? Maybe it could be tweaked a tiny bit?

src/RegistrationHandler.php
comzeradd: It's a very short message anyway. I thought it's easier to just replace it.

src/RegistrationHandler.php
comzeradd: I made a small adjustment to the message. Added also the new user email to make it more clear.

image

comzeradd:
> Perhaps we could move the logic from the MasterSite (`validate_password_policy` and `validate_password_reset`) to the class you've created, since they are all part of the same solution.

Yes, good point. I moved them in a separate commit so it's easier to review.

> We could also use the `assets/src/js/hide_weak_pw_checkbox.js` file to hide the password field instead of the `hide_password_field_on_registration_form` function, so we don't have to inject CSS.

I'm not sure if that's feasibe since we only hide the password field when sso is enforced. While `hide_weak_pw_checkbox.js` is always there. Implementing this with JS also has a small visual hiccup because you can see the paasword field momentarily.

To make that part more clean I changed it to enqueue a css file instead of injecting css code.

mardelnet (approved):
Great job!

PR #3034, opened 2026-07-08
PLANET-8216: fix query url conflict in action pages

GP-Dan-Tovbein:
### Summary

The implementation is about to display the proper action page based on `/:action_type/slug`. Currently, when you browse to any `Action Page` the core doesn't differentiate taxonomies, and everything it's a `p4_action`.

For example, if you create two actions:
- `/:action_type(petitions)/:slug(oceans)` and `/:action_type(joinus)/:slug(oceans)`  the core will prioritise to display the first created since by the core everything it's a `p4_action`.

What we need to fix is
- To override the query and look for an action page based on `action_type` and `post_name` (slug), and then display the proper action.
- To allow editors to use same slugs, like `oceans`.

---

Ref: https://greenpeace-planet4.atlassian.net/browse/PLANET-8216

### Testing

1. Create 2 actions
2. Set up with different `action_types`
3. Use the same `slug` (like `oceans`)
First check
1. You can use the same slug
2. When you browse them the will be displayed properly
- `/petitions/oceans` should display its related content
- `/joinus/oceans` should display its related content

### DIFF
--- a/src/CustomPostType/ActionPage.php
+++ b/src/CustomPostType/ActionPage.php
@@ -74,6 +74,32 @@ class ActionPage
         add_action('save_post_' . self::POST_TYPE, [ $this, 'save_taxonomy_action_type_on_quick_edit' ], 10, 2);
         // Update action type on add/edit of action.
         add_action('rest_after_insert_' . self::POST_TYPE, [ $this, 'save_taxonomy_action_type_on_edit' ], 10, 1);
+
+        // This hook allows to filtrer properly action pages by slug and name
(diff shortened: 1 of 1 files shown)


comzeradd:
On the `main` branch I can't create two Actions with the same slug, even they are assigned a different Action Type. So I can't reproduce the bug.

I'm not sure what other side effects it may have if we actually allow doing that.


https://github.com/user-attachments/assets/c3554888-6817-40ca-a408-356cabcc20a6

PR #3036, opened 2026-07-09
PLANET-8189: implement automation for importing external landing pages with redirection

mardelnet:
### Summary

This PR implements a mechanism to automate the import of external landing pages, turn them into Actions, and add redirects to the source URL.

---

Ref: https://greenpeace-planet4.atlassian.net/browse/PLANET-8189

### Testing

1. Go to the website admin area.
2. Look at the sidebar and check that a link labeled "Import Action" exists under "Actions".
3. Go to "Actions" and check that a button labeled "Import Action" exists next to "Add Action".
4. Go to "Import Action".
5. Add an external non-secure URL (http instead of https) and submit the form. An error message should be shown.
6. Add a non-valid URL and submit the form. An error message should be shown.
7. Add a valid and secure external URL and submit the form. A success message should be shown.
8. Add the same URL used in the previous step and submit the form. An error message should be shown.
9. Go to "Actions" and check that a new action exists with data from the external URL: title, description, featured image, and excerpt.
10. Go to "Tools" > "Redirections" and check that a new group called "Actions" exists.
11. Check that a new redirection from the recently created action to the source URL exists.
12. Move the recently created action to the trash and check that the linked redirection is suspended.
13. Completely delete the recently created action and check that the linked redirection is removed.
14. Verify that the Action is shown in all relevant places: Actions List block, Listing pages, Search page.

### DIFF
--- a/admin/js/action_importer.js
+++ b/admin/js/action_importer.js
@@ -8,6 +8,6 @@ document.addEventListener('DOMContentLoaded', () => {
     actionImportButton.href = 'edit.php?post_type=p4_action&page=import-action';
     actionImportButton.className = 'add-new-h2';
     actionImportButton.textContent = 'Import Action';
-    pageTitleAction.insertAdjacentElement('afterend', actionImportButton);
+    pageTitleAction.after(actionImportButton);
   }
 });
--- a/admin/js/action_importer.js
+++ b/admin/js/action_importer.js
@@ -1,0 +1,13 @@
+// Adds a link to the Import Action page.
+
+document.addEventListener('DOMContentLoaded', () => {
+  const pageTitleAction = document.querySelector('.wp-admin.post-type-p4_action .page-title-action');
+
+  if(pageTitleAction) {
+    const actionImportButton = document.createElement('a');
+    actionImportButton.href = 'edit.php?post_type=p4_action&page=import-action';
+    actionImportButton.className = 'add-new-h2';
+    actionImportButton.textContent = 'Import Action';
+    pageTitleAction.insertAdjacentElement('afterend', actionImportButton);
+  }
+});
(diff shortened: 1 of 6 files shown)


planet-4:

:watch: 2026.07.10 08:39:01

mardelnet:
@comzeradd:
I have applied some of the adjustments suggested by Sentry and eliminated all of the reported errors.
However, there are still several warnings. Most of them complain about the number of returns per function. I think they can be ignored, as some functions require many early returns as guard conditions. 
Other warnings report that the class is too large and should be split. Please let me know if I should work on the latter.

comzeradd:
Functionality-wise it works great. Just reviewing the code still. 

I noticed that when an Action lacks taxonomies, we show just a dot on the listing page breadcrumbs. I know that this issue was not introduced in the change, but since we assign no tag or category by default on imported Actions, we may end up with this UI more often than now. Ideally, if there is no category or tag assigned, we should hide the dot.

image

mardelnet:
> I noticed that when an Action lacks taxonomies, we show just a dot on the listing page breadcrumbs. I know that this issue was not introduced in the change, but since we assign no tag or category by default on imported Actions, we may end up with this UI more often than now. Ideally, if there is no category or tag assigned, we should hide the dot.

Good catch, @comzeradd
I've created [this function](https://github.com/greenpeace/planet4-master-theme/blob/ad3e6b522e6cee75d4696a0b741f1c8be8acdca8/src/BlockRenderOverrides.php#L46) to remove the taxonomies when empty.
I didn't know where to make that function live in our codebase, so I've created the class `BlockRenderOverrides` and also moved there other similar "orphaned" functions that lived in the `MasterSite` class and the `functions.php` file.
Let me know if you don't agree with these changes.

comzeradd (approved):
Great work 🧉 

From my tests the functionality works perfect. We can add some additional features in follow up tickets.

PR #3043, opened 2026-07-19
Move eslint to devDependencies

rushikeshmore:
### Summary

Moves `eslint` (pinned 8.56.0) from `dependencies` to `devDependencies`. It is a lint tool consumed only by the `lint:js` script and lint-staged, and no source file imports it at runtime. All the other lint packages (`@wordpress/eslint-plugin`, `eslint-plugin-react`, `eslint-plugin-react-hooks`) already live in `devDependencies`, so this aligns the dependency groups with actual usage, in the same spirit as #2959 and #2960.

Regenerating `package-lock.json` (npm 10) also synced two stale leftovers in the lockfile: the root `license` field now matches package.json (`GPL-3.0-or-later`, from #2962), and the orphaned `wp-scripts` placeholder entries from before #2959 are gone.

---

Ref: -

### Testing

1. `npm ci && npm run build` - completes identically to `main` (93 warnings, 0 errors, webpack 5.103.0)
2. `npm run lint` - passes; eslint 8.56.0 resolves from devDependencies exactly as before
3. `git grep -E "require\('eslint'|from 'eslint'"` - no runtime usage anywhere


rushikeshmore:
Fixed the commitlint failure and rebased onto current main.

**commitlint:** the subject was fine, the body was tripping `body-case: sentence-case` because it opened with a lowercase "eslint". Reworded that first word to "ESLint". Verified with the same command CI runs, on the versions the repo pins:

```
$ node -v && npm -v
v20.20.2
10.8.2
$ npx commitlint -V --from main --to HEAD
⧗   input: Move eslint to devDependencies
✔   found 0 problems, 0 warnings
```

Also rebased onto `767d3bc7` so the merge base is current. No commit since I filed has touched `package.json` or `package-lock.json`, so the lockfile applies unchanged. Re-verified after the rebase: `npm ci` clean, `npm run build` 0 errors, `npm run lint` 0 errors, and `lint:js` still resolves eslint correctly from devDependencies, which is the thing this PR needed to not break.

**request-instance:** I do not think this one is related to the diff. The Jul 29 pipeline was triggered through the API rather than by a PR event:

```
$ curl -s https://circleci.com/api/v2/pipeline/3170d2ec-a862-443e-8e91-9385b342597b
"circleci": { "event_type": "create pipeline run api", ... }
```

so there is no `pull_request` context for `<< pipeline.event.context.github.pr_url >>` (config line 140) to resolve to. `check-pr-exists.sh` passes the empty-ish value through, and `book-test-instance.py` then crashes on an unguarded regex match:

```
File "/usr/local/bin/book-test-instance.py", line 166, in <module>
    pr_endpoint, _ = get_repo_endpoints(pr_url=pr_url)
File "/usr/lib/python3.10/p4/github.py", line 43, in get_repo_endpoints
    repository = matches.group(1) or None
AttributeError: 'NoneType' object has no attribute 'group'
```

`comment-pr` already handles exactly this case and skips instead of failing:

```yaml
- run:
    name: Check if there is a PR
    command: |
      if [ ! -f /tmp/workspace/pr ]; then
        circleci-agent step halt
      fi
```

`request-instance` does not have that guard, so it hard-fails. Re-running the workflow from a PR-triggered pipeline should get it green without any change on my side. If it is worth filing separately, adding the same `step halt` guard to `request-instance` would stop this from hitting future fork PRs.

Everything substantive is already green on this SHA: `create-zip`, `frontend-tests`, `php83-tests`, and both SonarCloud checks.

rushikeshmore:
One practical note so the re-run does not mislead: the copy of this branch on the main repo (`refs/heads/move-eslint-to-devdependencies`) is still at the pre-fix commit `599d7fd7`, while the PR head is now `a442303c`. Re-running the existing pipeline as-is would lint the old message and fail again for the same reason.

Whatever produced that mirrored branch would need to point at `a442303c` for CI to see the corrected commit. Happy to do it differently if you would rather I opened a fresh PR from a branch on the main repo instead.

GP-Dan-Tovbein (dismissed):
Looks good to me !

We should also consider to test it out newest versions of eslint since they are recently released v10 and we're using v8.

See https://www.npmjs.com/package/eslint

mleray:
There are now conflicts that need to be resolved before merging 🙏

rushikeshmore:
Rebased onto current main (2f586ee4), conflicts resolved. Sorry for the delay.

The conflict was package-lock.json against the pa11y removal in 28fbfbbf. Took the lockfile from main and regenerated it with the pinned npm (node 20.20.2 / npm 10.8.2), so the diff is now just the eslint move in package.json plus the matching dev flag flips in the lockfile. The stale license and wp-scripts leftovers the original commit mentioned were already cleaned up by that regeneration on main, so I dropped that paragraph from the commit body.

Checked against main at the same commit: build 92 warnings / 0 errors on both, lint:js 47 warnings / 0 errors on both, and lint:js still resolves eslint from node_modules after npm ci.

Same note as before on CI: the mirrored branch on this repo is still at 599d7fd7, the PR head is now eb4d8d47.

Note: the push dismissed the earlier approval, so this needs a re-approve.

PR #3047, opened 2026-07-22
PLANET-8181: Dynamically render content in Listing Pages

mardelnet:
### Summary

This PR makes the content rendering for the Listing Pages dynamic.
This includes the posts list, the toggle layout button, the paginator, and, for the News & Stories page, the filters too.

NOTE: The test `CategoryPageTest::test_category_page_results` was temporarily disabled as it was failing.
I'll work on fixing it, maybe in a separate PR.

---

Ref: https://greenpeace-planet4.atlassian.net/browse/PLANET-8181

### Testing

1. Visit all different kinds of Listing Pages (News & Stories, taxonomies, etc.).
2. Check they still look and work as usual.
3. Check the toggle layout button works as usual and keeps the selected layout on page reload.  
4. Check the paginator works as usual and reloads the posts list dynamically, without a full page reload.
5. In the News & Stories page, check that the filters work as usual and reload the posts list dynamically, without a full page reload.
6. In the News & Stories page, check that the filters work when URL parameters are included.
7. In the News & Stories page, check that the featured section works as usual.

#### Some References

- News & Stories page: http://www.planet4.test/news-stories/
- News & Stories page + URL parameters: http://www.planet4.test/news-stories/?post-type=press&category=energy
- Tag page: http://www.planet4.test/tag/renewables/
- Post Type page: http://www.planet4.test/press/
- Category page: http://www.planet4.test/category/nature/
- All Actions page: http://www.planet4.test/petitions/


planet-4:

:watch: 2026.07.28 08:01:09

assets/src/js/Components/ListingPage/ListingPagePosts.js
```diff
@@ -0,0 +1,457 @@
...
+
+  /**
+   * Fetches the available post types, categories, and tags, and marks
+   * taxonomies as loaded once the request is completed.
+   *
+   * @return {Promise<void>}
+   */
+  const getTaxonomies = useCallback(async () => {
+    try {
+      const baseUrl = document.body.dataset.nro;
```
Osong-Michael: I see we use this in a few places, could we create a constant for this as the value never changes

assets/src/js/Components/ListingPage/ListingPagePosts.js
```diff
@@ -0,0 +1,457 @@
...
+        Sentry.captureMessage('ListingPagePosts: missing document.body.dataset.nro in getPosts');
+      }
+
+      const {data, totalPages: pages} = await fetchJson(
+        `${baseUrl}/wp-json/wp/v2/${addQueryArgs(endpoint, args)}`
+      );
+
+      // Ignore this response if a newer request has been fired.
+      if (requestId !== requestIdRef.current) {
+        if (typeof Sentry !== 'undefined') {
```
Osong-Michael: We use this Sentry declaration a lot, is it possible to create a helper function, makes the code less busy?

assets/src/js/Components/ListingPage/ListingPagePosts.js
```diff
@@ -0,0 +1,457 @@
...
+   * Re-fetches posts whenever `getPosts` changes identity.
+   */
+  useEffect(() => {
+    getPosts();
+  }, [getPosts]);
+
+  /**
+   * Restores a previously saved layout preference from `localStorage` on mount.
+   */
+  useEffect(() => {
```
Osong-Michael: So I see we initialise the layout variable then use this Effect to update it, this cause two renders I think 🤔.

We could instead initialise directly from localStorage, test the code below and adjust as needed.


```suggestion
  const [layout, setLayout] = useState(() => {
    try {
        const stored = localStorage.getItem(LAYOUTS.STORAGE_NAME);

        return stored === LAYOUTS.GRID
            ? LAYOUTS.GRID
            : LAYOUTS.LIST;
    } catch {
        return LAYOUTS.LIST;
    }
});
```

assets/src/js/Components/ListingPage/ListingPagePosts.js
```diff
@@ -0,0 +1,457 @@
...
+ * @param {Object}      props                         Component props.
+ * @param {HTMLElement} [props.filtersContainer]      DOM node to portal the filter controls into, if present.
+ * @param {HTMLElement} [props.layoutToggleContainer] DOM node to portal the layout toggle button into, if present.
+ *
+ * @return {JSX.Element} The rendered listing page posts section.
+ */
+const ListingPagePosts = ({filtersContainer, layoutToggleContainer}) => {
+  const [posts, setPosts] = useState([]);
+  const [isLoadingPosts, setIsLoadingPosts] = useState(true);
+  const [postTypes, setPostTypes] = useState([]);
```
Osong-Michael: Since we already doing the same with the filters can we do the same here...?


```suggestion
  const [taxonomies, setTaxonomies] = useState({
    postTypes: [],
    categories: [],
    tags: [],
});
```

assets/src/js/Components/ListingPage/ListingPagePosts.js
```diff
@@ -0,0 +1,457 @@
...
+  const [categories, setCategories] = useState([]);
+  const [tags, setTags] = useState([]);
+  const [taxonomiesLoaded, setTaxonomiesLoaded] = useState(false);
+  const [layout, setLayout] = useState(LAYOUTS.LIST);
+  const [page, setPage] = useState(1);
+  const [totalPages, setTotalPages] = useState(1);
+  const [filters, setFilters] = useState({
```
Osong-Michael: On lines 345 and 371, you have two useEffects that are responsible for synchronizing the filters with the URL, since these are essentially two halves of the same feature, i think we can extract them as a separate custom hook then use it directly on the filters state, makes this component much cleaner.


```suggestion
  const [filters, setFilters] = useFilterUrlSync({
    postTypes,
    categories,
    tags,
    loaded: taxonomiesLoaded,
  });
```

assets/src/js/Components/ListingPage/ListingPagePosts.js
```diff
@@ -0,0 +1,457 @@
...
+    } catch (e) {
+      if (typeof Sentry !== 'undefined') {
+        // eslint-disable-next-line no-undef
+        Sentry.captureException(e);
+      }
+    }
+  }, []);
+
+  /**
+   * Reads filters from the URL once taxonomies have loaded.
```
Osong-Michael: This useEffect and the one below can be combined into a new custom hook that can be imported like below


```suggestion
   import {useEffect, useRef, useState} from '@wordpress/element';

const URL_PARAMS = {
  postType: 'post-type',
  category: 'category',
  tag: 'tag',
};

function getIdBySlug(list, slug) {
  return list.find(item => item.slug === slug)?.id ?? '';
}

function getSlugById(list, id) {
  return list.find(item => item.id === id)?.slug ?? '';
}

export default function useFilterUrlSync({
  postTypes,
  categories,
  tags,
  loaded,
}) {
  const [filters, setFilters] = useState({
    postType: '',
    category: '',
    tag: '',
  });

  const hasSyncedFromUrl = useRef(false);

  /**
   * Read filters from URL once.
   */
  useEffect(() => {
    if (!loaded || hasSyncedFromUrl.current) {
      return;
    }

    const params = new URLSearchParams(window.location.search);

    setFilters({
      postType: getIdBySlug(
        postTypes,
        params.get(URL_PARAMS.postType)
      ),
      category: getIdBySlug(
        categories,
        params.get(URL_PARAMS.category)
      ),
      tag: getIdBySlug(
        tags,
        params.get(URL_PARAMS.tag)
      ),
    });

    hasSyncedFromUrl.current = true;
  }, [loaded, postTypes, categories, tags]);

  /**
   * Keep URL in sync.
   */
  useEffect(() => {
    if (!loaded) {
      return;
    }

    const params = new URLSearchParams(window.location.search);

    const values = {
      [URL_PARAMS.postType]:
				getSlugById(postTypes, filters.postType),
      [URL_PARAMS.category]:
				getSlugById(categories, filters.category),
      [URL_PARAMS.tag]:
				getSlugById(tags, filters.tag),
    };

    Object.entries(values).forEach(([key, value]) => {
      if (value) {
        params.set(key, value);
      } else {
        params.delete(key);
      }
    });

    const query = params.toString();

    window.history.replaceState(
      null,
      '',
      `${window.location.pathname}${
        query ? `?${query}` : ''
      }${window.location.hash}`
    );
  }, [filters, loaded, postTypes, categories, tags]);

  return [filters, setFilters];
}

```

Osong-Michael (commented):
Your code is really nice, I left just a few suggestions

assets/src/js/Components/ListingPage/ListingPagePosts.js
mardelnet: Good idea! Done!

assets/src/js/Components/ListingPage/ListingPagePosts.js
mardelnet: Good idea! Done!

assets/src/js/Components/ListingPage/ListingPagePosts.js
mardelnet: Good idea! Done!

assets/src/js/Components/ListingPage/ListingPagePosts.js
mardelnet: Good idea! Done!

assets/src/js/Components/ListingPage/ListingPagePosts.js
mardelnet: Good idea! Done!

assets/src/js/Components/ListingPage/ListingPagePosts.js
mardelnet: Done!

Osong-Michael (approved):
Great job 👍 

One question: if the tests in `test-category-page.php` is not needed, isn't it better to delete it rather than comment it out? or does it need fixing?

mardelnet:
> One question: if the tests in `test-category-page.php` is not needed, isn't it better to delete it rather than comment it out? or does it need fixing?

I'll create a ticket so we can fix it, @Osong-Michael

mardelnet:
> > One question: if the tests in `test-category-page.php` is not needed, isn't it better to delete it rather than comment it out? or does it need fixing?
> 
> I'll create a ticket so we can fix it, @Osong-Michael

Done! https://greenpeace-planet4.atlassian.net/browse/PLANET-8244

PR #3056, opened 2026-07-31
PLANET-8206 Scroll to correct year on load in Timeline block

mleray:
### Summary

Ref: [PLANET-8206](https://greenpeace-planet4.atlassian.net/browse/PLANET-8206)

### Testing

You can use [this page](https://www-dev.greenpeace.org/test-phoebe/timeline-test/) I made for example. Try scrolling to a year or use the navigation, then reload the page and make sure that the block loads to the correct year.

[PLANET-8206]: https://greenpeace-planet4.atlassian.net/browse/PLANET-8206?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/src/blocks/Timeline/YearsNavigation.js
+++ b/assets/src/blocks/Timeline/YearsNavigation.js
@@ -12,16 +12,12 @@ export const YearsNavigation = ({years, isEditing, timelineId}) => {
 
   const isRTL = document.dir === 'rtl';
 
+  /**
+   * Handle the behaviour when clicking on a year.
+   *
+   * @param {string} year - The year that was just clicked.
+   */
   const handleClick = year => {
-    const isFirefox = navigator.userAgent.toLowerCase().includes('firefox');
-
-    if (isFirefox) {
-      const target = document.getElementById(`${timelineId}-${year}`);
-      if (target) {
-        target.scrollIntoView({behavior: 'smooth', block: 'start'});
-      }
-    }
-
     isClicking.current = true;
     setActiveYear(`${timelineId}-${year}`);
 
@@ -34,6 +30,11 @@ export const YearsNavigation = ({years, isEditing, timelineId}) => {
     addEventListener('scrollend', () => isClicking.current = false, {once: true});
   };
 
+  /**
+   * Handle the behaviour when clicking the navigation arrows.
+   *
+   * @param {string} direction - The direction that was clicked, either left or right.
+   */
   const scrollNav = direction => {
     isManualScroll.current = true;
 
@@ -160,11 +161,30 @@ export const YearsNavigation = ({years, isEditing, timelineId}) => {
     return () => observer.disconnect();
   }, [isRTL]);
 
+  // Scroll to the year in the URL on load, if present. Only in the frontend.
+  useEffect(() => {
+    if (isEditing) {
+      return;
+    }
+
+    const loadItem = window.location.hash.substring(1);
+    if (!loadItem || !loadItem.startsWith(timelineId)) {
+      return;
+    }
+
+    const target = document.getElementById(loadItem);
+    if (!target) {
+      return;
+    }
+
+    target.scrollIntoView({behavior: 'smooth', block: 'start'});
+  }, []);
+
   return (
     <nav
       className={`years-navigation d-flex justify-content-center ${isEditing ? 'no-scroll' : ''}`}
       aria-label={sprintf(
-      /* translators: 1: amount of years in the timeline */
+        /* translators: 1: amount of years in the timeline */
         __('Timeline, list with %1$d items', 'planet4-blocks'),
         years.length
       )}
--- a/assets/src/blocks/Timeline/YearsNavigation.js
+++ b/assets/src/blocks/Timeline/YearsNavigation.js
@@ -177,7 +177,8 @@ export const YearsNavigation = ({years, isEditing, timelineId}) => {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


planet-4:

:watch: 2026.09.11 13:42:59

assets/src/blocks/Timeline/YearsNavigation.js
```diff
@@ -12,16 +12,12 @@ export const YearsNavigation = ({years, isEditing, timelineId}) => {
 
   const isRTL = document.dir === 'rtl';
 
+  /**
+   * Handle the behaviour when clicking on a year.
+   *
+   * @param {string} year - The year that was just clicked.
+   */
   const handleClick = year => {
-    const isFirefox = navigator.userAgent.toLowerCase().includes('firefox');
```
mleray, line 16: I'm not sure why this was added in the first place, it seems to work fine in Firefox without 🤷‍♀️

sagarsdeshmukh (dismissed):
The PR looks good to me. 👍
I’d like to suggest a UAT for this PR.

mleray:
> The PR looks good to me. 👍 I’d like to suggest a UAT for this PR.

@sagarsdeshmukh I've tried many things and I cannot get it to scroll to the exact right position, like it is when we click 😕 If you have any ideas they are very welcome!

comzeradd:
> 've tried many things and I cannot get it to scroll to the exact right position, like it is when we click 😕 If you have any ideas they are very welcome!

When reloading, I cann see the issue. The page is scrolling slightly further than the selected year. And the year is not visible. If I slowly start scrolling over I can see that the year changes in the url anchor to the previous year before its elements are even visible. Maybe that is what creates the problem? Every year section expanding a bit further than it should.

https://github.com/user-attachments/assets/39efcbcf-a6ef-49ea-a67d-db69a7c71217

mleray:
> When reloading, I cann see the issue. The page is scrolling slightly further than the selected year. And the year is not visible. If I slowly start scrolling over I can see that the year changes in the url anchor to the previous year before its elements are even visible. Maybe that is what creates the problem? Every year section expanding a bit further than it should.

@comzeradd I'm just confused because when clicking on the links it works just fine, I would expect the same behaviour here. All solutions I've found online point to `scroll-margin-top` which [we already use](https://github.com/greenpeace/planet4-master-theme/blob/main/assets/src/scss/blocks/Timeline/TimelineStyle.scss#L47) 😕 The only thing I can think of is to use `scrollTo` and hardcode some pixels to get to the exact position, but that would be tricky (depending on screen size etc) and might also need to be changed frequently if for example we change some margins.

comzeradd:
Can you try to use `calc(120px - 4rem);` in [this line](https://github.com/greenpeace/planet4-master-theme/blob/v1.380.0/assets/src/scss/blocks/Timeline/TimelineStyle.scss#L47)?

I think we need to account for `scroll-padding-top` that we have in [body.scss](https://github.com/greenpeace/planet4-master-theme/blob/v1.380.0/assets/src/scss/base/_body.scss#L5)

assets/src/blocks/Timeline/YearsNavigation.js
```diff
@@ -160,11 +161,30 @@ export const YearsNavigation = ({years, isEditing, timelineId}) => {
...
+    if (!loadItem || !loadItem.startsWith(timelineId)) {
+      return;
+    }
+
+    const target = document.getElementById(loadItem);
+    if (!target) {
+      return;
+    }
+
+    target.scrollIntoView({behavior: 'smooth'});
```
comzeradd: From my tests, it seems that the problem is that the block is quite long and takes time to load. On first load the scroll position is indeed off. But if you click on the address bar and hit enter it works fine.

I did some quick research on how people solve this and all I found is the use of [requestAnimationFrame](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame) function to request from the browser to execute the animation at a later repaint.

Here is an example I tried and it seems to work. 

```javascript
let settleTimer;
let secondFrame;

const scrollToTarget = () => {
  target.scrollIntoView({behavior: 'smooth', block: 'start'});
};

const firstFrame = requestAnimationFrame(() => {
  secondFrame = requestAnimationFrame(scrollToTarget);
  settleTimer = setTimeout(scrollToTarget, 500);
});

return () => {
  cancelAnimationFrame(firstFrame);
  cancelAnimationFrame(secondFrame);
  clearTimeout(settleTimer);
};
```

It practically waits for two render cycles before scrolling. I don't like the `500` timeout, but that's practically an additional scrolling to account for images lazy loading. I couldn't make it work without it.

cc @mleray @sagarsdeshmukh

assets/src/blocks/Timeline/YearsNavigation.js
```diff
@@ -160,11 +161,31 @@ export const YearsNavigation = ({years, isEditing, timelineId}) => {
...
+      return;
+    }
+
+    const target = document.getElementById(loadItem);
+    if (!target) {
+      return;
+    }
+
+    const scrollToTarget = () => target.scrollIntoView({behavior: 'smooth'});
+    setTimeout(scrollToTarget, 1000);
```
comzeradd, line 181: It's a better workaround. Better delay the scrolling a bit than scroll to the wrong place.

comzeradd (approved):
👍🏻

PR #3059, opened 2026-07-31
PLANET-8023_optimize-query-db-single-posts

GP-Dan-Tovbein:
### Summary
While working on Query DB perfomance ticket I've noticed that the official WP stack exchange request to `wp_get_object_terms` with `get_the_terms`.

> get_the_terms relies on the object cache. This gives a scaling boost and a speed boost. This means WP won't fetch the data a second time and reuses the results from the first time. Why grab a post from the database a second time? Store it in a variable to optimise performance!
> Unless you have a permanent object cache installed, the object cache only exists for that request though. If you have a permanent object cache set up this boost increases speed dramatically by storing this data in external software instead of variables, e.g. Redis/Memcached.

Ref: https://wordpress.stackexchange.com/questions/313114/when-to-not-to-use-wp-get-post-terms-vs-get-the-terms

---

Ref: https://greenpeace-planet4.atlassian.net/browse/PLANET-8023

### Testing

1. It should work the same

### DIFF
--- a/src/Post.php
+++ b/src/Post.php
@@ -48,6 +48,7 @@ class Post extends \Timber\Post
         $post = parent::build($wp_post);
         $post->set_page_types();
         $post->set_author();
+        $post->post_meta = get_post_meta($post->ID);
 
         return $post;
     }
@@ -240,6 +241,15 @@ class Post extends \Timber\Post
         }
     }
 
+    public function get_meta(string $key, int|null $index = null): string
+    {
+        if (null !== $index) {
+            return $this->post_meta[$key][$index] ?? '';
+        }
+
+        return $this->post_meta[$key] ?? '';
+    }
+
     /**
      * Get post/page custom planet4 type.
      * ACTION, DOCUMENT, PAGE, POST
@@ -256,7 +266,7 @@ class Post extends \Timber\Post
      */
     public function get_og_title(): string
     {
-        return get_post_meta($this->id, 'p4_og_title', true);
+        return $this->get_meta('p4_og_title', 0);
     }
 
     /**
@@ -265,7 +275,7 @@ class Post extends \Timber\Post
      */
     public function get_og_description(): string
     {
-        $og_desc = get_post_meta($this->id, 'p4_og_description', true);
+        $og_desc = $this->get_meta('p4_og_description', 0);
 
         if ('' === $og_desc) {
             return $this->post_excerpt;
@@ -280,12 +290,12 @@ class Post extends \Timber\Post
      */
     public function get_og_image(): array
     {
-        $meta = get_post_meta($this->id);
         $image_id = null;
         $image_metas = ['p4_og_image_id', '_thumbnail_id', 'background_image_id'];
         foreach ($image_metas as $image_meta) {
-            if (!empty($meta[$image_meta][0])) {
-                $image_id = $meta[$image_meta][0];
+            $image_meta_id = $this->get_meta($image_meta, 0);
+            if (!empty($image_meta_id)) {
+                $image_id = $image_meta_id;
                 break;
             }
         }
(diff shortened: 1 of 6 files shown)


planet-4:

:watch: 2026.08.25 11:22:06

sagarsdeshmukh:
Hi @GP-Dan-Tovbein , Could you please check the phpunit test errors at [here](https://app.circleci.com/pipelines/gh/greenpeace/planet4-master-theme/15710/workflows/5fbc7dea-3df3-48ff-b1fc-804c4134b336/jobs/141044/tests).

src/Post.php
```diff
@@ -351,18 +352,18 @@ public function social_share_platforms(): array
...
+        return (bool) $this->post_meta['p4_author_override'][0] ?? false;
     }
 
     /**
      * Sets the User author of this Post.
      */
     public function set_author(): void
     {
-        $author_override = get_post_meta($this->id, 'p4_author_override', true);
+        $author_override = $this->get_author_override();
```
sagarsdeshmukh: The author name on posts seems to have disappeared. Could you please double-check this logic?

For example, on **test-iocaste**:
https://www-dev.greenpeace.org/test-iocaste/press/1113/duis-posuere-6/

Working example on **test-titan**:
https://www-dev.greenpeace.org/test-titan/press/1113/duis-posuere-6/

src/Post.php
```diff
@@ -48,6 +48,7 @@ public static function build(\WP_Post $wp_post): self
         $post = parent::build($wp_post);
         $post->set_page_types();
         $post->set_author();
+        $post->post_meta = get_post_meta($post->ID);
```
sagarsdeshmukh: It looks like set_author() is called before post_meta is populated, so it's reading unset metadata.
Moving the post_meta assignment before calling set_author() should fix the issue.

```
$post->set_author();                        // reads $this->post_meta ...
$post->post_meta = get_post_meta($post->ID); // ... which is set only here
```

sagarsdeshmukh: Line #51 can also trigger a "Creation of dynamic property is deprecated" warning, since creating dynamic properties has been deprecated in PHP 8.2 and later.
https://php.watch/versions/8.2/dynamic-properties-deprecated

GP-Dan-Tovbein: Here is the proposal solution. 
https://www.php.net/manual/en/function.array-key-exists.php

The problem is that `p4_author_override` key it's only set through the array when the value it's not empty.

src/Post.php
```diff
@@ -347,24 +348,23 @@ public function social_share_platforms(): array
...
     {
-        $author_override = get_post_meta($this->id, 'p4_author_override', true);
+        $author_override = $this->get_author_override();
         $real_author = Timber::get_user((int) $this->post_author);
 
-        if ('' !== $author_override) {
+        if (true === $author_override) {
             $fake_user = $real_author;
-            $fake_user->display_name = $author_override;
+            $fake_user->display_name = $this->post_meta['p4_author_override'][0];
```
GP-Dan-Tovbein, line 366: The `post_meta` it's already cached by wp, so it's fine to check it again

src/Post.php
sagarsdeshmukh: The post_meta is still an undeclared property. The class declares issues_nav_data, content_type, page_types, author, data_layer but not post_meta. 
Assigning $post->post_meta = … creates a dynamic property, deprecated in PHP 8.2+

src/Post.php
```diff
@@ -347,24 +348,22 @@ public function social_share_platforms(): array
 
     /**
      * Get post's author override status.
-     *
      */
     public function get_author_override(): bool
     {
-        return !empty(get_post_meta($this->id, 'p4_author_override', true));
+        return array_key_exists('p4_author_override', $this->post_meta);
```
sagarsdeshmukh, line 354: The array_key_exists() function returns true whenever the meta key exists, regardless of its value. This means get_author_override() will also return true when the meta value is an empty string.

sagarsdeshmukh (approved):
Nice work @GP-Dan-Tovbein , PR looks good to me 👍

PR #3070, opened 2026-08-17
Update ToC to show all Timeline blocks in the frontend

Osong-Michael:
### Summary

Updated Timeline block observer to pick up all h2 if multiple timeline blocks present 

---

### Testing

Create a new page and add as many _Timeline_ blocks as you want, then add the _Table of Content_ Block, the ToC block should pick up all the H2s of the timeline block both in the editor and in the frontend


comzeradd (approved):
👍🏻

PR #3072, opened 2026-08-19
Fix media-replacer test

Osong-Michael:
### Summary

On test instances, the media replacer test gets two success messages when a file is replaced, one for the replacer and the other for cache, this fix ensures the right message is picked up so the tests don't fail. 

Also the data returned by the spreadsheet block in my testing is not consistent, it some times returns `data: {header, rows} `and other times `{header, rows}`, so this fix normalizes the data to `{header, rows}` so it is more consistent.

---

### Testing
All tests should pass now with no issues.

### DIFF
--- a/assets/src/blocks/Spreadsheet/spreadsheet.js
+++ b/assets/src/blocks/Spreadsheet/spreadsheet.js
@@ -15,6 +15,33 @@ const placeholderData = {
   ],
 };
 
+/**
+ * Normalizes the spreadsheet REST response into a consistent { header, rows }
+ * shape, since the endpoint sometimes returns it flat and sometimes nested
+ * under a `data` key.
+ *
+ * @param {Object} rawData - Raw response from get-spreadsheet-data.
+ * @return {{header: Array, rows: Array}|null} Normalized data, or null if unusable.
+ */
+function normalizeSpreadsheetData(rawData) {
+  if (!rawData) {
+    return null;
+  }
+
+  const source = Array.isArray(rawData?.data?.header) || Array.isArray(rawData?.data?.rows) ?
+    rawData.data :
+    rawData;
+
+  if (!Array.isArray(source.header) || !Array.isArray(source.rows)) {
+    return null;
+  }
+
+  return {
+    header: source.header,
+    rows: source.rows,
+  };
+}
+
 export const SpreadsheetFrontend = ({
   url,
   color,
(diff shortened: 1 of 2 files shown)


planet-4:

:watch: 2026.08.19 10:03:22

assets/src/blocks/Spreadsheet/spreadsheet.js
```diff
@@ -15,6 +15,33 @@ const placeholderData = {
...
 
+/**
+ * Normalizes the spreadsheet REST response into a consistent { header, rows }
+ * shape, since the endpoint sometimes returns it flat and sometimes nested
+ * under a `data` key.
+ *
+ * @param {Object} rawData - Raw response from get-spreadsheet-data.
+ * @return {{header: Array, rows: Array}|null} Normalized data, or null if unusable.
+ */
+function normalizeSpreadsheetData(rawData) {
```
mleray, line 26: Isn't that something we should do in the block directly? Rather than in the tests?

tests/e2e/media-replacer.spec.js
```diff
@@ -140,7 +140,9 @@ test('Replace Media file (PDF) in WordPress', async ({page, admin, requestUtils}
     // --- Replace the media file ---
     await replaceMediaFile(page, newFile);
 
-    const successNotice = page.locator('.notice-success');
+    const successNotice = page.locator('.notice-success').filter({
+      hasText: 'These files were successfully replaced:',
+    });
     await expect(successNotice).toContainText('These files were successfully replaced:', {timeout: 20000});
```
mleray, line 146: Do we need to check this if we already filter on the text? Shouldn't we just make sure it's visible?

assets/src/blocks/Spreadsheet/spreadsheet.js
Osong-Michael: This is the block code 😄

mleray: Ah my bad 🤦‍♀️

tests/e2e/media-replacer.spec.js
Osong-Michael: I mean, the assertion does two things, it checks if the notice is available and also makes sure we get the correct string back.

mleray: Ah, right

mleray (approved):
👍

PR #3080, opened 2026-08-27
PLANET-8264 Add sidebar options for customizing meta tags

sagarsdeshmukh:
### Summary

Editors should be able to override the meta tags (title and description) independently. Currently, the Open Graph/Social fields define the meta title and the Excerpt defines the meta description.

The meta description serves a different purpose from Open Graph standards. It is intended for search engines, is typically around 120–160 characters, and should be written to encourage users to click on the page in search results, not as a summary of the article.

### Requirements

- Add two new options in the sidebar, inside the Search Engines Optimizations section.
  - One input text field. Title: Meta Title. Help Text: Leave empty to use the Open Graph Title (if set) or the Post Title.
  - One textarea field: Title: Meta Description. Help Text: Leave empty to use Open Graph Description (if set) or the Post Excerpt.
- Add an additional help text above both fields: Check optimal length at [SERP Snippet Generator](https://app.sistrix.com/en/serp-snippet-generator).
- Both fields should be above the Canonical URL option (see screenshot).
- Precedence requirements for meta tags:
  - Meta Title tag: Meta Title > Open Graph Title > Post Title
  - Meta Description tag: Meta Description > Open Graph Description > Post Excerpt

---

Ref. https://greenpeace-planet4.atlassian.net/browse/PLANET-8264

### Testing
1. Check out this branch on local and rebuild the editor assets(`npm run build`)
2. Edit published Post or Page, open the "Search Engine Optimizations" panel in the sidebar
3. Confirm the panel now shows, in this order - 
    - A help line: "Check optimal length at SERP Snippet Generator
    - Meta Title
    - Meta Description
    - Canonical link

4. Repeat on a Page, an Action and  Campaign, the two new fields should appear on all four post types
5. Enter values in Meta Title and Meta Description, Update, reload the editor, Confirm both values persisted
6. Check Meta Title precedence: Meta Title > OG Title > Post Title
7. Check Meta Description precedence: Meta Description > OG Description > Post Excerpt

### DIFF
--- a/assets/src/block-editor/Sidebar/SearchEngineOptimizationsSidebar.js
+++ b/assets/src/block-editor/Sidebar/SearchEngineOptimizationsSidebar.js
@@ -3,20 +3,46 @@ import {URLInput} from '../URLInput/URLInput';
 const {PluginDocumentSettingPanel} = wp.editor;
 const {useDispatch, useSelect} = wp.data;
 const {__} = wp.i18n;
+const {TextControl, TextareaControl, ExternalLink} = wp.components;
+const {createInterpolateElement} = wp.element;
 
 const CANONICAL_URL = 'p4_seo_canonical_url';
+const META_TITLE = 'p4_seo_meta_title';
+const META_DESCRIPTION = 'p4_seo_meta_description';
 
 export const SearchEngineOptimizationsSidebar = {
   getId: () => 'planet4-seo-sidebar',
   render: () => {
     const meta = useSelect(select => select('core/editor').getEditedPostAttribute('meta'), []);
-    const {editPost} = useDispatch('core/editor', [meta[CANONICAL_URL]]);
+    const {editPost} = useDispatch('core/editor');
 
     return (
       <PluginDocumentSettingPanel
         name="planet4-search-engine-optimizations"
         title={__('Search Engine Optimizations', 'planet4-master-theme-backend')}
       >
+        <p className="components-base-control__help">
+          {createInterpolateElement(
+            __('Check optimal length at <a>SERP Snippet Generator</a>.', 'planet4-master-theme-backend'),
+            {
+              a: (
+                <ExternalLink href="https://app.sistrix.com/en/serp-snippet-generator" />
+              ),
+            }
+          )}
+        </p>
+        <TextControl
+          label={__('Meta Title', 'planet4-master-theme-backend')}
+          value={meta[META_TITLE] || ''}
+          onChange={value => editPost({meta: {[META_TITLE]: value}})}
+          help={__('Leave empty to use the Open Graph Title (if set) or the Post Title.', 'planet4-master-theme-backend')}
+        />
+        <TextareaControl
+          label={__('Meta Description', 'planet4-master-theme-backend')}
+          value={meta[META_DESCRIPTION] || ''}
+          onChange={value => editPost({meta: {[META_DESCRIPTION]: value}})}
+          help={__('Leave empty to use Open Graph Description (if set) or the Post Excerpt.', 'planet4-master-theme-backend')}
+        />
         <URLInput
           label={__('Canonical link', 'planet4-master-theme-backend')}
           value={meta[CANONICAL_URL]}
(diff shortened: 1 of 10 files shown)


planet-4:

:watch: 2026.09.03 09:48:36

assets/src/block-editor/Sidebar/SearchEngineOptimizationsSidebar.js
```diff
@@ -3,20 +3,46 @@ import {URLInput} from '../URLInput/URLInput';
...
     const meta = useSelect(select => select('core/editor').getEditedPostAttribute('meta'), []);
-    const {editPost} = useDispatch('core/editor', [meta[CANONICAL_URL]]);
+    const {editPost} = useDispatch('core/editor');
 
     return (
       <PluginDocumentSettingPanel
         name="planet4-search-engine-optimizations"
         title={__('Search Engine Optimizations', 'planet4-master-theme-backend')}
       >
+        <p className="components-base-control__help">
```
mardelnet: Can we add some additional spacing between the help text and the title?

Image

templates/blocks/meta_fields.twig
```diff
@@ -30,7 +34,13 @@
...
     {% endif %}
 
-    <meta name="title" content="{{ title }}"/>
+    {% if seo_meta_title %}
+        {% set meta_title = seo_meta_title|decode_entities~' - '~site.name %}
+    {% else %}
+        {% set meta_title = site.name %}
+    {% endif %}
+
+    <meta name="title" content="{{ meta_title }}"/>
```
mardelnet, line 47: We're adding a new meta tag (`meta name=title`) but we are not replacing the actual `title` tag, which is still using the value of the Open Graph field.

mardelnet: The Open Graph value or the page title is not included here when the Meta Title field is empty.

assets/src/block-editor/Sidebar/SearchEngineOptimizationsSidebar.js
mardelnet, line 1: It seems that our custom fields take precedence over other fields with a similar purpose. For example, I tested this using the Yoast SEO plugin (which I think it is used by some NROs). Maybe this is something they should be aware of.

templates/blocks/meta_fields.twig
sagarsdeshmukh: You’re right @mardelnet , I forgot to commit this change. Could you please check it now?

assets/src/block-editor/Sidebar/SearchEngineOptimizationsSidebar.js
sagarsdeshmukh: In the case where Yoast SEO or a similar plugin is active, should we add any functional condition in this PR to avoid conflicts, or is simply informing the NRO sufficient? 🤔

cc @comzeradd

comzeradd: What kind of conflict do we see? Are the Yoast SEO custom fields taking precedence over our sidebar options? If that's the case, I think it's fine to just include it as notice when the feature is released (in comms and in the handbook). I prefer we don't add conditional logic for plugins we don't support.

mardelnet: The opposite, @comzeradd, our sidebar options take precedence over Yoast.
I think that informing the NRO will be sufficient, though.

templates/blocks/meta_fields.twig
mardelnet: @sagarsdeshmukh, the `<title>` tag is still unaffected by the new Title fields.
Only the `<meta name="title">` tag changes.
As far as I know, the "title" meta tag is no longer used by modern search engines, and it only works as a fallback for the title tag. The latter is the one with real SEO impact.

comzeradd: We could adjust [title.twig](https://github.com/greenpeace/planet4-master-theme/blob/v1.379.0/templates/blocks/title.twig) to follow the same precedence rules: Meta Title > Open Graph Title > Post Title.

sagarsdeshmukh: The PR has been updated as discussed.

mardelnet (approved):
Nice job @sagarsdeshmukh 
Maybe, in a different PR, we could extend this also to populate the SEO and Meta fields when an Action is imported (cc @comzeradd )

PR #3092, opened 2026-09-09
Make sure there is no homepage title before adding hidden H1

mleray:
### Summary

This is to make sure that we don't have two H1 on the page, which would be bad for accessibility and also SEO. This doesn't take Heading blocks into account, but it's an extra check.

### DIFF
--- a/templates/page.twig
+++ b/templates/page.twig
@@ -1,8 +1,9 @@
 {% extends "base.twig" %}
 
 {% block content %}
-    <main class="page-content container {% if hide_page_title or not header_title %}no-page-title{% endif %}" id="content">
-        {% if fn('is_front_page') %}
+    {% set no_page_title = hide_page_title or not header_title %}
+    <main class="page-content container {% if no_page_title %}no-page-title{% endif %}" id="content">
+        {% if fn('is_front_page') and no_page_title %}
             <h1 class="visually-hidden" tabindex="0">
                 {{ __('%s homepage', 'planet4-master-theme')|format(site.name) }}
             </h1>


planet-4:

:watch: 2026.09.09 12:47:58

mardelnet (commented):
Hey @mleray 
I am a bit confused: In which case does the homepage display the page title? I am running the `main` branch locally. I created a new empty page, turned it into the homepage, and the title is never shown, regardless of whether the "Hide page title" box is checked or not.

mleray:
> Hey @mleray I am a bit confused: In which case does the homepage display the page title? I am running the `main` branch locally. I created a new empty page, turned it into the homepage, and the title is never shown, regardless of whether the "Hide page title" box is checked or not.

@mardelnet you're right, I've just tested it myself and see the same happening 🤔 From what I see [here](https://github.com/greenpeace/planet4-master-theme/blob/2f586ee442df7faa5346dd23ff022899d2e92f16/src/Context.php#L22) we take the `p4_title` from the metadata as header for the homepage specifically

mardelnet:
> > Hey @mleray I am a bit confused: In which case does the homepage display the page title? I am running the `main` branch locally. I created a new empty page, turned it into the homepage, and the title is never shown, regardless of whether the "Hide page title" box is checked or not.
> 
> @mardelnet you're right, I've just tested it myself and see the same happening 🤔 From what I see [here](https://github.com/greenpeace/planet4-master-theme/blob/2f586ee442df7faa5346dd23ff022899d2e92f16/src/Context.php#L22) we take the `p4_title` from the metadata as header for the homepage specifically

And is this expected? Is it a bug? How do we proceed, @mleray?

mleray:
> And is this expected? Is it a bug? How do we proceed, @mleray?

I don't know @mardelnet! Maybe @comzeradd can tell us, and if indeed there is a bug there we can open a follow-up ticket. But in any case adding this extra check doesn't hurt, right?

mardelnet:
> > And is this expected? Is it a bug? How do we proceed, @mleray?
> 
> I don't know @mardelnet! Maybe @comzeradd can tell us, and if indeed there is a bug there we can open a follow-up ticket. But in any case adding this extra check doesn't hurt, right?

But what's the benefit of the changes introduced in this PR?
Aren't they producing the exact same output as the `main` branch?

mleray:
> But what's the benefit of the changes introduced in this PR? Aren't they producing the exact same output as the `main` branch?

@mardelnet the benefit is an extra check to make sure that we don't add the hidden H1 in case there is a page title, even though we haven't been able to add one it might still be possible. Especially if it's a bug that we end up solving!

mardelnet (approved):
Good job!

Maybe we can discuss with the designers what we can do to prevent editors from adding more than one H1 element per page.

mleray:
> Maybe we can discuss with the designers what we can do to prevent editors from adding more than one H1 element per page.

@mardelnet it is apparently possible since WordPress 6.7 to [disable heading levels](https://developer.wordpress.org/news/snippets/how-to-disable-heading-levels-in-the-editor/)! I think it would make sense for us to disable H1 everywhere, but indeed let's discuss with the design team

comzeradd:
I don't think it's a bug, but maybe it was added in the template before we even had the "Hide Page title" option in the sidebar. So we can't just remove it because I'm sure there will be homepages that don't have that toggle enabled and then some homepage titles will pop up right above the Carousel Header 😁 

But let's discuss options with the design team about `h1` and then we could do a simple migration specifically for the homepage to enable "Hide Page title" and remove the check from the template.

mleray:
> I don't think it's a bug, but maybe it was added in the template before we even had the "Hide Page title" option in the sidebar. So we can't just remove it because I'm sure there will be homepages that don't have that toggle enabled and then some homepage titles will pop up right above the Carousel Header 😁
> 
> But let's discuss options with the design team about `h1` and then we could do a simple migration specifically for the homepage to enable "Hide Page title" and remove the check from the template.

@comzeradd which check are you referring to in the template? If there's a title? We definitely want to keep that because it applies to all pages for the "Hide page title". Our question was that in the homepage case it seems that even when there is a title and "Hide page title" is disabled, the title is not shown. It seems that for the homepage specifically, we take the [`p4_title` meta data value](https://github.com/greenpeace/planet4-master-theme/blob/2f586ee442df7faa5346dd23ff022899d2e92f16/src/Context.php#L22), where is that even defined? Sorry if it's confusing

comzeradd:
> > I don't think it's a bug, but maybe it was added in the template before we even had the "Hide Page title" option in the sidebar. So we can't just remove it because I'm sure there will be homepages that don't have that toggle enabled and then some homepage titles will pop up right above the Carousel Header 😁
> > But let's discuss options with the design team about `h1` and then we could do a simple migration specifically for the homepage to enable "Hide Page title" and remove the check from the template.
> 
> @comzeradd which check are you referring to in the template? If there's a title? We definitely want to keep that because it applies to all pages for the "Hide page title". Our question was that in the homepage case it seems that even when there is a title and "Hide page title" is disabled, the title is not shown. It seems that for the homepage specifically, we take the [`p4_title` meta data value](https://github.com/greenpeace/planet4-master-theme/blob/2f586ee442df7faa5346dd23ff022899d2e92f16/src/Context.php#L22), where is that even defined? Sorry if it's confusing

I mean this last thing you linked. We could simplify the code so even the homepage just respects the "Hide page title" option.

`p4_title` seems to be a leftover from when we had custom page title and subtitle in the sidebar. In the same section as the "Hide page title". But we didn't cleaned up the [custom fields](https://github.com/greenpeace/planet4-master-theme/blob/v1.381.0/src/PageMeta.php#L14-L16) when we did, that's why apparently it still works.

We can probably tackle that as part of [PLANET-8215](https://greenpeace-planet4.atlassian.net/browse/PLANET-8215)

[PLANET-8215]: https://greenpeace-planet4.atlassian.net/browse/PLANET-8215?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3094, opened 2026-09-15
PLANET-8296: Fix flickering in secondary navigation block

mardelnet:
### Summary

This PR fixes the flickering in the Secondary Navigation block when H2 elements are added asynchronously, for instance, by the Timeline block. This problem can be tested using [this link](https://www.greenpeace.org/techvision/deep-dives/beautiful-basics/) or with this block code:  [blocks.txt](https://github.com/user-attachments/files/32229653/blocks.txt)
 

The changes implemented by this PR include:
- The data obtained by the `getHeadingsFromDom` function is now memoized. It only recomputes when `levels` change or the DOM mutates, giving a stable reference the rest of the time.
- A new `useEffect` was added to watch for DOM structural changes (e.g. Timeline's H2 being mounted) and trigger a headings recompute.
- The arrows are removed when only one H2 element is found.
- `headings` was added to the effect that toggles the arrows, so it re-runs whenever the nav's item list changes and not only once on mount.

---

Ref: https://greenpeace-planet4.atlassian.net/browse/PLANET-8296

### Testing

1. Check that no flickering exists when a page uses both the Secondary Navigation block and the Timeline block.
2. Check that the mobile version of the Secondary Navigation block works correctly.
3. Check that the Secondary Navigation block works correctly with other blocks.


planet-4:

:watch: 2026.09.15 07:31:49

mleray (commented):
I'm wondering if we could add a similar fix for the Timeline block scroll on load? 🤔 For now we've added a [timeout](https://github.com/greenpeace/planet4-master-theme/pull/3056/changes#diff-5bdb7b54dd99620ae48204e8670d32b46ade3b294e4fe2aa9b67ece17a09df67R181), but this might be better, because we basically had the same issue that we needed to wait for the block to fully render

mleray (approved):
Looks good to me 👏 

Something I've noticed (but can be fixed in a follow-up) is that it's not always scrolling to the correct anchor on load. For example [this page](https://www-dev.greenpeace.org/test-pandora/about-us-2/#this-is-a-heading-3) I used for testing.

mardelnet:
> Looks good to me 👏
> 
> Something I've noticed (but can be fixed in a follow-up) is that it's not always scrolling to the correct anchor on load. For example [this page](https://www-dev.greenpeace.org/test-pandora/about-us-2/#this-is-a-heading-3) I used for testing.

Done, @mleray! https://greenpeace-planet4.atlassian.net/browse/PLANET-8306

PR #3100, opened 2026-09-15
PLANET-8243: Announces langs and countries with locale pronunciations

GP-Dan-Tovbein:
### Summary

This implementation announces countries in its proper locale pronunciation instead of everything in English. 
For example: Belgium Françoise it is announced different than Belgium Dutch.

---

Ref: https://greenpeace-planet4.atlassian.net/browse/PLANET-8243

### Testing

1. Open the screen reader
2. Navigate through to any GP site
3. Focus the countries selector and start tabbing
4. Each country should be announced in its proper language


planet-4:

:watch: 2026.09.23 07:43:28

templates/footer_country_selector.twig
```diff
@@ -38,7 +38,10 @@
                 {% set main_lang, lang_count = country.lang|first, country.lang|length %}
                 <li role="listitem">
                     <a href="{{ main_lang.url }}"
-                        hreflang="{{ main_lang.locale[0] }}"
+                        {% if main_lang.locale[0] is defined %}
```
mleray, line 41: So if `main_lang.locale[0]` is not defined, we don't add anything?

GP-Dan-Tovbein: Exactly, because it checks whenever it has a multiple languages added.

templates/footer_country_selector.twig
```diff
@@ -48,10 +51,14 @@
...
                             <a href="{{ lang.url }}"
-                                hreflang="{{ lang.locale[0] }}"
+                                {% if lang.locale[0] is defined %}
+                                    lang="{{ lang.locale[0] }}"
+                                    hreflang="{{ lang.locale[0] }}"
+                                {% endif %}
                                 data-ga-category="Country Selector"
                                 data-ga-action="{{ country.name ~ ' | ' ~ lang.name }}"
                                 data-ga-label="n/a"
+                                aria-label="{{ lang.name ~ ', ' ~ country.name }}"
```
mleray, line 61: Could this be different in RTL languages for example?

GP-Dan-Tovbein: I'm sure it won't affect that, the browser is on change on that. It's basically the same of https://github.com/greenpeace/planet4-master-theme/pull/3100/changes/BASE..b9e983aa983dd53f8cee306c9fd31782b95f475f#diff-26a6ff6c33f468b55994daf5fd0006c20cef6073c1f06feec8200a30fff33c8aR59

mleray: It is different, the GA action is for us for analytics so it doesn't really matter, the aria-label is for accessibility. For RTL languages that means we would have the country name first and then the language, no? Probably not a big deal but worth thinking about

PR #3101, opened 2026-09-16
PLANET-8234: Avoid Search Field to jump to a second line in the Navigation Menu

mardelnet:
### Summary

This PR prevents the Navigation Menu from breaking into a second line when the number of items/length of characters in the Menu Items is too large. With the changes applied in this PR, the regular menu is converted into the hamburger menu when we detect that the menu layout can be broken.

---

Ref: https://greenpeace-planet4.atlassian.net/browse/PLANET-8234

### Testing

1. Edit the Navigation Bar Menu in ways that would normally break it into two lines and check that the desktop menu is replaced by the hamburger menu.
2. Check that the navigation menu behaves normally on window resizing.


planet-4:

:watch: 2026.09.17 09:12:01

mleray (commented):
On the test instance I still see this happening at 1024px width:

Image

mardelnet:
> On the test instance I still see this happening at 1024px width:

Nice catch, @mleray!
I think it's fixed now. Would you mind checking it again?

assets/src/js/header/setupDynamicHambMenu.js
```diff
@@ -0,0 +1,49 @@
+// Set up displaying the hamburger menu when the desktop menu is too large.
+
+const dynamicHambMenu = () => {
```
mleray: Nit-picky suggestion I know, but can we call it `dynamicHamburgerMenu` instead? It's not that much longer, but clearer imo.

assets/src/js/header/setupDynamicHamburgerMenu.js
```diff
@@ -0,0 +1,49 @@
+// Set up displaying the hamburger menu when the desktop menu is too large.
+
+const dynamicHambMenu = () => {
+  const header = document.querySelector('#header');
+  const mainMobileNav = document.querySelector('#nav-main');
+
+  if (!header || !mainMobileNav) {
```
mleray, line 11: Do we also need to check the screen size before continuing? 🤔 If it's less than 992px we also don't need to do anything since it's always the hamburger menu anyways.

assets/src/js/header/setupDynamicHambMenu.js
```diff
@@ -0,0 +1,49 @@
...
+  const searchForm = header.querySelector('#search_form');
+
+  if (!logo || !mainDesktopNav || !searchForm || !mainMobileNavBtn) {
+    return;
+  }
+
+  const headerHeight = header.getBoundingClientRect().height;
+  const logoHeight = logo.getBoundingClientRect().height;
+
+  if (headerHeight > logoHeight) {
```
mleray: Could it be better to have early return here?

assets/src/js/header/setupDynamicHambMenu.js
```diff
@@ -0,0 +1,49 @@
...
+    mainDesktopNav.classList.remove('d-lg-flex');
+    mainMobileNav.classList.remove('d-lg-none');
+    mainMobileNavBtn.style.display = 'initial';
+    logo.style.marginLeft = '0';
```
mleray: How does this work with RTL sites?

mleray (commented):
Do we need a FLAG on this? Or documentation? To warn the community about the change

mleray:
There's now no more search in 1024px screen 😬 

Screenshot 2026-09-17 at 09 58 24

mardelnet:
@mleray, I've already addressed most of your suggested changes.
But before you continue with the code review, let me include @comzeradd in this thread.
In a previous conversation I had with him, we noticed that these changes might be overkill and hard to maintain, so maybe there's no reason to keep working on them.
Also, I'd like to know your opinion (and Nikos', too) on that. Do you think these changes are worth it?

mleray:
@mardelnet I tend to agree that it's overkill. We have limitation warnings set in the backend, editors know when they are making choices that might break the design.

mardelnet:
> @mleray, I've already addressed most of your suggested changes. But before you continue with the code review, let me include @comzeradd in this thread. In a previous conversation I had with him, we noticed that these changes might be overkill and hard to maintain, so maybe there's no reason to keep working on them. Also, I'd like to know your opinion (and Nikos', too) on that. Do you think these changes are worth it?

@comzeradd can we have your opinion on this, please?

comzeradd:
I tend to agree with your reservations. Even if we make it work in this PR, it practically makes the navigation code more prone to regression bugs every time we make a change on the navigation bar.

It seems like we are trying to resolve a content management issue with code. It does seem like an overkill to me as well.

